PR #5, opened 2023-06-05
Models for Organization and UserOrganizationMembership

pxwxnvermx:
This PR includes models and migrations for Organization, UserOrganizationMembership and Opportunity


pxwxnvermx:
Linter failed due to some default imports that were created by `./manage.py createapp` command.

commcare_connect/users/models.py
```diff
@@ -33,3 +44,28 @@ def get_absolute_url(self) -> str:
 
         """
         return reverse("users:detail", kwargs={"pk": self.id})
+
+
+class Organization(BaseModel):
+    name = models.CharField(max_length=255, unique=True)
+    slug = models.SlugField(max_length=255)
```
snopoke, line 51: I think the slug should be unique, not the name.

commcare_connect/users/models.py
```diff
@@ -33,3 +44,28 @@ def get_absolute_url(self) -> str:
...
+        related_name='users',
+        related_query_name='user'
+    )
+    user = models.ForeignKey(
+        User,
+        on_delete=models.DO_NOTHING,
+        related_name='organizations',
+        related_query_name='organization'
+    )
+    role = models.CharField(max_length=255)
```
snopoke, line 71: This should have choices of `admin` and `member`

commcare_connect/users/models.py
```diff
@@ -33,3 +44,28 @@ def get_absolute_url(self) -> str:
...
         """
         return reverse("users:detail", kwargs={"pk": self.id})
+
+
+class Organization(BaseModel):
+    name = models.CharField(max_length=255, unique=True)
+    slug = models.SlugField(max_length=255)
+
+    def save(self, *args, **kwargs):
+        self.slug = slugify(self.name)
```
snopoke, line 54: We should probably handle conflicts by appending a suffix to make sure it's unique - something like this: https://code.djangoproject.com/wiki/SlugifyUniquely

snopoke:
> Linter failed due to some default imports that were created by `./manage.py createapp` command.

You can use pre-commit locally as well: `pre-commit install`

snopoke (commented):
Can you add Django admin configs for these models too

snopoke (approved):

snopoke (approved):

PR #16, opened 2023-06-21
Add API for getting opportunity details for CID Users

pxwxnvermx:
image

image


snopoke (approved):

shubham1g5:
Not sure what's the right channel to ask these questions so adding them here - 

1. Do we have any details on how we are going to map an opportunity to a user ? Is Opportunity a organisation level , a demographical level or a user level property or a combination of all of these ? 

2. Should we track how many users can accept a opportunity and validate it everytime someone claims an opportunity ?

PR #35, opened 2023-07-25
performance improvement to fetching app data form CC

snopoke:
* Fetch data in parallel (per domain)
* Cache data to avoid re-fetching when saving the form

In future we may need a 'refresh' button to allow the user to update the app data in case they have made recent changes in CCHQ.


pxwxnvermx (approved):
@snopoke Thanks a lot for these improvements. 🥇

PR #39, opened 2023-07-28
add oauth token introspection

calellowitz:
This sets commcare-connect to use connect id as a fallback server for checking OAuth tokens, and add oauth to the allowed authentication methods.


calellowitz:
Sorry, thought I had requested review on this a while ago. Not sure how I missed that

snopoke (approved):

commcare_connect/users/models.py
```diff
@@ -13,12 +14,24 @@ class User(AbstractUser):
...
+        _("username"),
+        max_length=150,
+        unique=True,
+        help_text=_("Required. 150 characters or fewer. Letters, digits and @/./+/-/_ only."),
+        validators=[username_validator],
+        error_messages={
+            "unique": _("A user with that username already exists."),
+        },
+        null=True,
+    )
```
snopoke, line 34: Is this just copied from `AbstractUser`? Does removing `username = None` not have the same effect?

snopoke (approved):

commcare_connect/users/models.py
calellowitz: It is copied but it adds `null=True` since email enrolled users don't have a username.

calellowitz:
@snopoke I think I got to all your comments but let me know if that is not true

commcare_connect/users/tests/test_views.py
```diff
@@ -106,3 +108,19 @@ def test_not_authenticated(self, user: User, rf: RequestFactory):
...
+
+    def test_view(self, mobile_user: User, rf: RequestFactory):
+        request = rf.post("/fake-url/", data={"commcare_username": "abc", "connect_username": mobile_user.username})
+        request.user = mobile_user
+        with mock.patch(
+            "oauth2_provider.views.mixins.ClientProtectedResourceMixin.authenticate_client"
+        ) as authenticate_client:
+            authenticate_client.return_value = True
+            response = create_user_link_view(request)
+            print(response)
```
snopoke, line 123: nit: print

snopoke (approved):

PR #41, opened 2023-07-31
Opportunity Access model and migrations

pxwxnvermx:


commcare_connect/opportunity/models.py
```diff
@@ -70,3 +70,18 @@ class DeliverForm(models.Model):
 
     def __str__(self):
         return self.name
+
+
+class OpportunityAccess(models.Model):
+    user = models.OneToOneField(
```
calellowitz, line 76: I think this should be a key to the actual User model, not `ConnectUser`

commcare_connect/users/models.py
```diff
@@ -69,3 +69,15 @@ class Role(models.TextChoices):
         related_name="memberships",
     )
     role = models.CharField(max_length=20, choices=Role.choices, default=Role.MEMBER)
+
+
+class ConnectUser(BaseModel):
```
calellowitz, line 74: As I mentioned on the ticket, I don't think these fields are quite right and we should hold off on adding this data until the details of ConnectID/Connect authentication are finished. As mentioned in the other comment, I don't think this is necessary for this PR either.

commcare_connect/opportunity/models.py
```diff
@@ -70,3 +70,18 @@ class DeliverForm(models.Model):
 
     def __str__(self):
         return self.name
+
+
+class OpportunityAccess(models.Model):
```
calellowitz, line 75: What happened to the other fields on this model? I know we mentioned not creating ones that referenced uncreated models, but some of the fields in the spec were specfic to this one and are still missing. Is that intentional?

calellowitz (commented):
A couple questions on this. It seems like it went a slightly different direction than we discussed on the ticket.

commcare_connect/users/models.py
pxwxnvermx: Removed ConnectUser. [45f8e61](https://github.com/dimagi/commcare-connect/pull/41/commits/45f8e61ef0d11a964d253808d2ed907ef223b36b)

commcare_connect/opportunity/models.py
```diff
@@ -70,3 +70,18 @@ class DeliverForm(models.Model):
...
+    user = models.OneToOneField(
+        ConnectUser,
+        on_delete=models.CASCADE,
+        related_name="claims",
```
calellowitz, line 79: I dont think you want to call these claims (default is probably fine, to be honest), since you can have access without yet having claimed the opportunity. In other words, some of these rows represent claims but some do not, so naming them like that is confusing.

commcare_connect/opportunity/models.py
```diff
@@ -70,3 +70,18 @@ class DeliverForm(models.Model):
...
+        related_query_name="claim",
+    )
+    opportunity = models.ForeignKey(
+        Opportunity,
+        on_delete=models.CASCADE,
+        related_name="claims",
```
calellowitz, line 85: Same comment about naming

commcare_connect/opportunity/models.py
calellowitz: `date_claimed` is the one I was thinking of. Agree that the others are not necessary yet.

commcare_connect/opportunity/models.py
calellowitz: Thanks

commcare_connect/opportunity/models.py
pxwxnvermx: Added. [4ad1eaf](https://github.com/dimagi/commcare-connect/pull/41/commits/4ad1eaf0cda305b5fbe0d1cde7ecfd476fcbbf52)

commcare_connect/opportunity/models.py
pxwxnvermx: Fixed. [4ad1eaf](https://github.com/dimagi/commcare-connect/pull/41/commits/4ad1eaf0cda305b5fbe0d1cde7ecfd476fcbbf52)

commcare_connect/opportunity/models.py
pxwxnvermx: Fixed. [4ad1eaf](https://github.com/dimagi/commcare-connect/pull/41/commits/4ad1eaf0cda305b5fbe0d1cde7ecfd476fcbbf52)

snopoke (approved):

PR #55, opened 2023-08-03
Save metadata from CC App to Database

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-93)

### DIFF
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -51,6 +51,12 @@ class Opportunity(BaseModel):
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs) -> None:
+        from commcare_connect.opportunity.tasks import create_learn_modules_assessments
+
+        super().save(*args, **kwargs)
+        create_learn_modules_assessments.delay(self.id)
+
 
 class DeliverForm(models.Model):
     app = models.ForeignKey(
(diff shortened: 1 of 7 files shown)


commcare_connect/opportunity/models.py
```diff
@@ -51,6 +51,12 @@ class Opportunity(BaseModel):
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs) -> None:
+        from commcare_connect.opportunity.tasks import create_learn_modules_assessments
+
+        super().save(*args, **kwargs)
+        create_learn_modules_assessments.delay(self.id)
```
snopoke, line 58: I don't think we should do this in the save function since it will get called during tests and other places.

Probably best to do it in the view.

snopoke (approved):
one comment but otherwise looks good

commcare_connect/opportunity/models.py
pxwxnvermx: fixed. [f8ff725](https://github.com/dimagi/commcare-connect/pull/55/commits/f8ff7253dfadcf67cd896f20deb9b20900ed5ce8)

snopoke (approved):

snopoke:
Just FYI - we don't have celery set up in AWS yet

snopoke:
(didn't mean to close)

PR #59, opened 2023-08-08
user links

calellowitz:
This implements the same user link model as in HQ for Connect. HQ will hit the newly created view (authing with oauth client credentials), when links are created there, replicating them to the connect server. This functionality relies on the token introspection branch that this PR is compared against, which automatically creates users in commcare connect with the same username as in connectid.

The PR is a draft because I am still working on the test cases but wanted to get it up so that the approach could be reviewed.


commcare_connect/users/views.py
```diff
@@ -44,3 +49,22 @@ def get_redirect_url(self):
 
 
 user_redirect_view = UserRedirectView.as_view()
+
+
+class CreateUserLinkView(ClientProtectedResourceMixin, APIView):
```
calellowitz, line 54: This mixin enforces OAuth client authentication, so it will only be accessible to systems like hq that have a client id and client secret.

commcare_connect/users/views.py
```diff
@@ -44,3 +49,22 @@ def get_redirect_url(self):
...
+    parser_classes = [parsers.JSONParser]
+
+    def post(self, request):
+        commcare_username = request.data.get('commcare_username')
+        connect_username = request.data.get('connect_username')
+        if not commcare_username or not connect_username:
+            return Response("commcare_username and connect_username required", status=status.HTTP_400_BAD_REQUEST)
+        try:
+            user = User.objects.get(username=connect_username)
```
calellowitz, line 63: This relies on the user having authenticated at least once with connect before this call, which I think is a fair assumption, but one I am willing to revisit.

commcare_connect/users/models.py
```diff
@@ -69,3 +69,11 @@ class Role(models.TextChoices):
...
     )
     role = models.CharField(max_length=20, choices=Role.choices, default=Role.MEMBER)
+
+
+class ConnectIDUserLink(models.Model):
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+    )
+    commcare_username = models.TextField()
```
calellowitz, line 79: This could be username or user id. I think both are in the forms so I went with username to match what we do in HQ, but could switch it if that is preferred.

commcare_connect/users/models.py
```diff
@@ -69,3 +69,11 @@ class Role(models.TextChoices):
         related_name="memberships",
     )
     role = models.CharField(max_length=20, choices=Role.choices, default=Role.MEMBER)
+
+
+class ConnectIDUserLink(models.Model):
```
calellowitz, line 74: An instance of this model will be created by connect at the same time it creates a user in hq for users where connect is creating their hq account.

commcare_connect/users/views.py
pxwxnvermx: `User` model uses the `email` field to identify users so I think this `username` field might not be populated because the `username` field type is set to None in the User model.

commcare_connect/users/views.py
calellowitz: One, sort of weird, alternative here is that HQ could forward the token from the user linking the account (effectively impersonating the connect user). It makes the auth a little less scary (client secret would no longer allow full access to this view), but is a bit gross having hq impersonate users by forwarding on their credentials.

commcare_connect/users/views.py
calellowitz: Sorry, I didn't realize that, thanks for flagging. I can wait for @snopoke to weigh in on how he imagined this working since he seems to have done the initial implementation there, but that seems unfortunate to me since the mobile users won't have an email, and instead have a ConnectID username/userid as their primary identifier, which I was envisioning using in the username here (following django-oauth-toolkit standard for token introspection).

@snopoke were you picturing an entirely separate user model for mobile users of connect?

commcare_connect/users/views.py
snopoke: +1 for having separate auth for HQ and not impersonating the user

commcare_connect/users/views.py
snopoke: This project was built using the Django cookiecutter which defaults to using email for auth (which is the pattern for most web apps). We can easily switch to supporting both:

1. Add back the `username` field to the user model
2. Change the `ACCOUNT_AUTHENTICATION_METHOD` setting to `username_email`

We may have to tweak a [few other settings](https://github.com/dimagi/commcare-connect/blob/ac92c81313c50311065e430cc3441187c1872b30/config/settings/base.py#L245-L255) as well.

snopoke (commented):
Thanks @calellowitz. This looks great.

commcare_connect/users/views.py
calellowitz: I started making changes on the token introspection PR since the user creation in that workflow requires username. I got most of the tests passing, but have one to go. Will pick it up again tomorrow, and do more proper localtesting, which I have yet to do.

calellowitz:
@snopoke I think this is good for final review

commcare_connect/users/models.py
```diff
@@ -46,3 +46,12 @@ def get_absolute_url(self) -> str:
 
         """
         return reverse("users:detail", kwargs={"pk": self.id})
+
+
+class ConnectIDUserLink(models.Model):
```
snopoke, line 51: I didn't notice before but I think there should be a unique constraint on `user + commcare_username` with error handling in the view (perhaps to just ignore the error if the link already exists or adjust the return code (201 created vs 200).

snopoke (approved):
Diff seems messed up but changes look good.

snopoke:
@calellowitz is this view used yet? I don't see any references to it in HQ

calellowitz:
Not yet, it only applies to users created on HQ and later linked to connect, which isn't a workflow we have yet.

PR #60, opened 2023-08-09
UI for Learn Progress Data

pxwxnvermx:
[Ticket Link](https://dimagi-dev.atlassian.net/browse/CCPC-95)

Learn Progress:
image

User Specific Learn Progress:
image

### DIFF
--- a/commcare_connect/templates/opportunity/learn_progress.html
+++ b/commcare_connect/templates/opportunity/learn_progress.html
@@ -4,35 +4,39 @@
 
 {% block content %}
   <div class="container">
-    <h2>Learn Progress for Users</h2>
-    <table class="table table-bordered">
-      <thead>
-      <tr>
-        <th>User Name</th>
-        <th>Learn Modules Completed</th>
-        <th></th>
-      </tr>
-      </thead>
-      <tbody class="align-middle">
-      {% for opportunity_access in page_obj %}
+    <div class="mt-2">
+      <h2>{{ opportunity.name }}</h2>
+
+      <h4 class="mt-4">Learn Progress for Users</h4>
+      <table class="table border mt-2 table-custom">
+        <thead class="table-secondary">
         <tr>
-          <td>{{ opportunity_access.user.name }}</td>
-          <td>
-            {{ opportunity_access.learn_progress }}%
-          </td>
-          <td>
-            <a
-              href="{% url 'opportunity:user_learn_progress' org_slug=org_slug opp_id=opportunity_access.opportunity.id pk=opportunity_access.id %}">
-              Details
-            </a>
-          </td>
+          <th>Name</th>
+          <th>Modules Completed</th>
+          <th></th>
         </tr>
-        {% empty %}
-        <tr>
-          <td colspan="3">No learn progress for users.</td>
-        </tr>
-      {% endfor %}
-      </tbody>
-    </table>
+        </thead>
+        <tbody class="align-middle">
+        {% for opportunity_access in page_obj %}
+          <tr>
+            <td>{{ opportunity_access.user.name }}</td>
+            <td>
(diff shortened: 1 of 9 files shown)


commcare_connect/users/signals.py
```diff
@@ -6,7 +6,9 @@
 
 @receiver(user_signed_up)
 def create_org_for_user(request, user, **kwargs):
-    if not user.members.exists():
```
pxwxnvermx, line 9: @snopoke This check was failing with AttributeError for members on the user object. I wrapped this in a try-except. I am not totally sure if this is a good solution for this.

commcare_connect/templates/opportunity/learn_progress.html
```diff
@@ -0,0 +1,38 @@
+{% extends "base.html" %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+{% block content %}
+  <div class="container">
```
snopoke, line 6: These pages should include some context about which opportunity the user is looking at. For now perhaps just a title with the opportunity name.

In future I'm thinking this info could be displayed on an Opportunity Details page which would show general info about the opp and probably have  the progress info shown below.

It would also have buttons for editing etc. Here's an example from another product:

image

commcare_connect/templates/opportunity/user_learn_progress.html
```diff
@@ -0,0 +1,29 @@
+{% extends "base.html" %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+{% block content %}
+  <div class="container">
```
snopoke, line 6: I'm not quite sure where this info belongs but fine to leave it on a separate page for now.

snopoke (commented):
Few small comments but looks good.

You should also update the nav so that the correct item is marked as active: https://github.com/dimagi/commcare-connect/blob/4258d6cd9dbf49814824d31dbb40f65b280f7d30/commcare_connect/templates/base.html#L51

commcare_connect/users/signals.py
pxwxnvermx: Thanks Fixed. [6c3f65f](https://github.com/dimagi/commcare-connect/pull/60/commits/6c3f65f5f29ad456da4622baefb3c72ae45347e6)

commcare_connect/templates/opportunity/learn_progress.html
```diff
@@ -0,0 +1,38 @@
...
+        <th>Learn Modules Completed</th>
+        <th></th>
+      </tr>
+      </thead>
+      <tbody class="align-middle">
+      {% for opportunity_access in page_obj %}
+        <tr>
+          <td>{{ opportunity_access.user.name }}</td>
+          <td>
+            {{ opportunity_access.learn_progress }}%
```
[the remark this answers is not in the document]
pxwxnvermx: I added all the fields from opportunity access to the table i.e. the `last_visit_date` and `visit_count`. [ca17a1f](https://github.com/dimagi/commcare-connect/pull/60/commits/ca17a1f7305842f249e7cdf71c76962b1af791e8)

commcare_connect/templates/opportunity/learn_progress.html
pxwxnvermx: I made some changes to clean up the Table and the UI to include the opportunity name. [92508b6](https://github.com/dimagi/commcare-connect/pull/60/commits/92508b60b355530905f76e56a9fcc6486c336be6)

pxwxnvermx: I was also thinking of renaming this page to the Opportunity Detail page and adding the learn progress table on this page.

snopoke (approved):

snopoke (approved):

PR #69, opened 2023-08-16
add api key model

calellowitz:
This adds an API Key model as discussed in [this ticket](https://dimagi-dev.atlassian.net/browse/CCPC-98). In draft mode because future commits will use this API key to create the user in HQ.

I chose to attach the API key to the opportunity for both convenience (this is the main place form entry point for data), and because API keys can be domain scoped so they may not apply to all opportunities that a user creates. This way a user can create multiple opportunities with different API keys (perhaps across different HQ environments), and the mobile can tell connect which API key to use by specifying the opportunity in the call, rather than needing to know the user or organization (something unlikely to be passed to the mobile).


commcare_connect/opportunity/models.py
```diff
@@ -167,3 +167,11 @@ class UserVisit(XFormBaseModel):
...
     )
     visit_date = models.DateTimeField()
+
+
+class HQApiKey(models.Model):
+    opportunity = models.ForeignKey(
+        Opportunity,
+        on_delete=models.CASCADE,
+    )
+    api_key = models.CharField(max_length=50)
```
snopoke, line 177: Should this be encrypted? I got most way through an `EncryptedField` implementation yesterday if you want to use it.

snopoke (commented):
Looks good

snopoke:
> I chose to attach the API key to the opportunity for both convenience (this is the main place form entry point for data), and because API keys can be domain scoped so they may not apply to all opportunities that a user creates. This way a user can create multiple opportunities with different API keys (perhaps across different HQ environments), and the mobile can tell connect which API key to use by specifying the opportunity in the call, rather than needing to know the user or organization (something unlikely to be passed to the mobile).

Interesting choice, I had not considered that as an option. One issue is if the user's auth fails when creating their first opportunity.

Also just wondering if we need this level of complexity or if we could assume that one Organization would only work with one HQ instance. 

Another approach would be to have users register different HQ instances with their org. Then when you create an opportunity they could select the HQ instance? 

It feels a bit like we're solving a problem that doesn't exist (yet).

commcare_connect/opportunity/models.py
```diff
@@ -22,6 +22,14 @@ def __str__(self):
         return self.name
 
 
+class HQApiKey(models.Model):
```
snopoke, line 25: Can we link this to an Organization as well. That would make it possible to lookup an API key in instances where the user is not present.

What's the thinking in attaching this to the user? Is it just a 'created_by' field?

commcare_connect/users/views.py
```diff
@@ -74,3 +76,29 @@ def post(self, request):
 
 
 create_user_link_view = CreateUserLinkView.as_view()
+
+
+@csrf_exempt
+@api_view(['POST'])
+@authentication_classes([OAuth2Authentication])
+def create_hq_user(request):
```
snopoke, line 84: What's the workflow here? I assume this is being called by mobile. Is this a v1 of a 'claim' endpoint? If so I think we should call it that since it will likely do more things than just create a user on HQ e.g. validate and update `OpportunityAccess` model

commcare_connect/opportunity/models.py
calellowitz: > Can we link this to an Organization as well

Yeah, it already is through the opportunity, but we can make that more explicit.

> What's the thinking in attaching this to the user? Is it just a 'created_by' field?

Since API keys are owned by people in HQ, it seemed right to link it to that person/account here (if the account is deleted we would probably want to delete api keys tied to that account). That is especially so for api keys attached to users that used hq credentials to sign up. They really own those keys. In a practical sense, to use the api key, you need a username, so this gives us a username (email) to use, again at least while users have signed up with hq credentials.

In the future we could have users specify an hq username when they add the api key, but for now assuming it matches the user creating it seemed fair.

commcare_connect/users/views.py
calellowitz: > I assume this is being called by mobile. Is this a v1 of a 'claim' endpoint? 

Yes to being called by mobile but I think this is separate from a "claim" endpoint, at least how the team has been using claim in other contexts. 

User progression through the opportunity at a high level will be:
1) Be given access
2) Complete learning
3) Claim job
4) Complete delivery

I was imagining this endpoint as being called before 2 and/or 4 depending on whether the user has a worker in the domain of the app. It might grow to encompass some of that logic (like a `get_or_create_hq_user_for_domain`), but I think it wants to remain separate from the claim logic (allocate some portion of the budget to this connect user).

calellowitz: I tried to write up my understanding of the claim workflows as currently discussed in (https://docs.google.com/document/d/1UiKPQrgJhA2UI9-sg0oriGsDYhniCvkwpnMzovvclJI/edit)[this doc]. That is based on conversations with Rama, and the workflow prototypes that we have based our design on.

I think there are a few key ways it differs from what you are describing that inform why I chose to implement it this way. However, if it ends up being simply a matter of naming, I am more than happy to rename this view.

> the point when a mobile user 'accepts' an opportunity they have been given access to. At this point we would need to create the HQ user so that they can get the learn app on their phone.

Users need to be able to start learning immediately, before "accepting/claiming" the job. In other words, their state on the connect backend should not need to change for them to download the learn app. 

I am envisioning creating a user as a part of the "download app" workflow from the mobile/user perspective. Mobile will tell connect it needs a user, and download the app from HQ. That could happen one or more times during the lifecycle of an opportunity, since if the deliver app is in a different domain, they will need to create the user again.

It could definitely be the case that doing this for the learn and deliver apps have separate URLs, rather than a shared one here, and what they share is a utility function that creates the user, but the reason I gave it its own endpoint was that I expected it to be used at multiple different points over the course of a user's interaction with the opportunity. User creation is not a one time thing at the point of claiming, but something that could be required each time they download an app.

All of that is to say, if you think this should be two URLs (`begin_learning` and `begin_delivery`) that take an opportunity and call this as a function, I am not opposed to that, but I do think user creation is a clear piece of reusable logic the phone will call multiple times, rather than a one time step along the user's progression.

snopoke: > Users need to be able to start learning immediately, before "accepting/claiming" the job. In other words, their state on the connect backend should not need to change for them to download the learn app.

I get that - but I'm sure we still want to record which user's have 'started learning'.

> before "accepting/claiming" the job

I think this is just terminology - all I mean is 'when the user starts learning'.

> I expected it to be used at multiple different points over the course of a user's interaction with the opportunity

There are only 2 points which are clearly defined workflow steps so I don't think there's a lot of value in have a standalone API for user creation. I also don't think the mobile app should be thinking in terms of 'user creation' but rather in terms of CCC workflow.

> All of that is to say, if you think this should be two URLs (begin_learning and begin_delivery) that take an opportunity and call this as a function, I am not opposed to that, but I do think user creation is a clear piece of reusable logic the phone will call multiple times, rather than a one time step along the user's progression.

I do think they should be separate. The user creation only happens in 2 places and I'd rather not think of it as 'user creation' otherwise we may not end up calling the API at all in instances where the user already has an HQ user in the domain (assuming mobile could determine that). If we did that we'd loose the information that the user had 'started learning'.

snopoke: Also just to say that we can change it later so fine to proceed as is for now.

snopoke (approved):

commcare_connect/users/helpers.py
```diff
@@ -14,3 +16,17 @@ def get_organization_for_request(request, view_kwargs):
 
     membership = request.user.memberships.select_related("organization").first()
     return membership.organization if membership else None
+
+
+def create_hq_user(user, domain, app, api_key):
```
snopoke, line 21: nit: `app` doesn't seem to be used

snopoke (approved):

commcare_connect/users/helpers.py
calellowitz: Ah yep. I was originally going to pull domain from there, then just passed that it in.

PR #72, opened 2023-08-16
Add CCHQ app fetch using all latest params

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-100)


snopoke (commented):
What happens if none of the requests succeed?

snopoke (approved):

pxwxnvermx:
> What happens if none of the requests succeed?

I have added code to raise an exception when no builds are available.

PR #75, opened 2023-08-18
Add learn modules to Opportunity Details UI

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-104)

image

### DIFF
--- a/commcare_connect/templates/opportunity/opportunity_detail.html
+++ b/commcare_connect/templates/opportunity/opportunity_detail.html
@@ -49,6 +49,22 @@
         </tr>
         </tbody>
       </table>
+      <h4>Learn App Modules</h4>
+      <div class="row mt-2">
+        {% for module in object.learn_app.learn_modules.all %}
+          <div class="col-sm">
+            <div class="card h-100">
+              <div class="card-body">
+                <h5 class="card-title">{{ module.name }}</h5>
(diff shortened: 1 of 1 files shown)


snopoke (commented):
What does this look like with more modules? Can you do a screenshot with 10 or 11?

pxwxnvermx:
@snopoke I replaced the cards with a table, with the cards the UI got cluttered with more than 10 modules.

snopoke (commented):
What do you think about making this one of the tabs? Also, when you first create an opportunity these models won't be populated. It would be nice to add a placeholder - though we may need an extra data element so that we know if the app has been processed or not. (probably leave for another PR).

snopoke (approved):

pxwxnvermx:
I think this would look better as a tab. But including this with the Learn Progress and Deliver Progress tabs does not make sense. We can make this a tab if in future we make changes to make an Opportunity have multiple deliver forms. Then the deliver forms and learn modules both can be individual tabs on the UI for Opportunity Details.

snopoke:
> But including this with the Learn Progress and Deliver Progress tabs does not make sense.

Why don't you think it makes sense? I think it would be fine. But I'm not going to hold up this PR.

PR #76, opened 2023-08-21
UI improvements for Opportunity List page

pxwxnvermx:
image

### DIFF
--- a/commcare_connect/templates/opportunity/opportunity_list.html
+++ b/commcare_connect/templates/opportunity/opportunity_list.html
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% load static %}
 
-{% block title %}Create new Opportunity{% endblock %}
+{% block title %}{{ org_slug }} - Opportunities{% endblock %}
 
 {% block content %}
   <div class="container">
--- a/commcare_connect/templates/opportunity/opportunity_list.html
+++ b/commcare_connect/templates/opportunity/opportunity_list.html
@@ -5,11 +5,15 @@
 
 {% block content %}
   <div class="container">
-    <div class="mb-2 mt-2">
+    <div class="my-2">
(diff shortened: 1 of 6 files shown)


commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -1,15 +1,19 @@
 {% extends "base.html" %}
 {% load static %}
 
-{% block title %}Create new Opportunity{% endblock %}
+{% block title %}{{ org_slug }} - Opportunities{% endblock %}
```
snopoke, line 4: how come this one has the `org_slug` but the others don't? If we want that to be consistent we should put it in the base template. We can use `request.org` since that's set by middleware.

snopoke (approved):
Small comment but otherwise 👍🏻

commcare_connect/templates/opportunity/opportunity_list.html
pxwxnvermx: This makes sense, I think adding the org name to the title makes it easier to understand which org the user is working on right now.

pxwxnvermx: Fixed. [8e9112c](https://github.com/dimagi/commcare-connect/pull/76/commits/8e9112cb8aa3d8d7517aa2c763d00c247956ebf2)

snopoke: Apologies for not being clearer, I think it's worth adding this to the base template [base.html](https://github.com/dimagi/commcare-connect/blob/1cf6da00aa100ec45fc3c781ae77d974a6d51ae3/commcare_connect/templates/base.html#L8). Something like:

```html
<title>
{% if request.org %}{{request.org.name}}: {% endif %}
{% block title %}CommCare Connect{% endblock title %}
</title>
```

PR #77, opened 2023-08-22
Refactor Opportunity Views to use request.org and remove unused context_data

pxwxnvermx:


commcare_connect/opportunity/views.py
```diff
@@ -21,33 +21,27 @@ class OpportunityList(OrganizationUserMixin, ListView):
...
     paginate_by = 10
 
-    def get_context_data(self, *args, object_list=None, **kwargs):
-        context_data = super().get_context_data(*args, **kwargs)
-        context_data["org_slug"] = self.kwargs["org_slug"]
-        return context_data
-
     def get_queryset(self):
-        return Opportunity.objects.filter(organization__slug=self.kwargs["org_slug"])
+        return Opportunity.objects.filter(organization__slug=self.request.org)
```
snopoke, line 25: `request.org` is actually the full organization so you either want `organization=self.request.org` or `organization_slug=self.request.org.slug`

pxwxnvermx: Thanks for pointing this out. Fixed https://github.com/dimagi/commcare-connect/pull/77/commits/dc154d1fbe67b6b26e0bc47a26f9803dea3cd7ae

snopoke: The same thing applies to the other changes where you have used `request.org` to refer to the `slug`. This only works now because the `__str__` method of `request.org` returns the slug. If that changes e.g. to use the name instead, then these changes will break. I think it would be safer to either stick with the view kwargs (nothing wrong with that) or use `request.org.slug`.

My reason for suggesting `request.org` in the other PR is that it is already available in all templates because of the request context processor so you don't need to remember to add something into the template context in each view.

snopoke (approved):

PR #78, opened 2023-08-23
User Visits Verification Status api endpoint

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-105)

image

### DIFF
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -45,4 +45,4 @@ class UserVisitVerificationStatusSerializer(serializers.Serializer):
     daily_visits = serializers.IntegerField()
     approved = serializers.IntegerField()
     rejected = serializers.IntegerField()
-    failed = serializers.IntegerField()
+    pending = serializers.IntegerField()
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from commcare_connect.opportunity.models import CommCareApp, Opportunity
+from commcare_connect.opportunity.models import CommCareApp, Opportunity, UserVisit
 
 
 class CommCareAppSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


commcare_connect/opportunity/api/views.py
```diff
@@ -30,3 +44,24 @@ def get(self, *args, **kwargs):
...
+    serializer_class = UserVisitVerificationStatusSerializer
+    permission_classes = [IsAuthenticated]
+
+    def get(self, *args, **kwargs):
+        user_visits = UserVisit.objects.filter(user=self.request.user, opportunity=kwargs.get("pk"))
+        total_visits = user_visits.count()
+        daily_visits = user_visits.filter(visit_date__date=datetime.today().date()).count()
+        approved = user_visits.filter(status=VisitValidationStatus.approved).count()
+        pending = user_visits.filter(status=VisitValidationStatus.pending).count()
+        rejected = user_visits.filter(status=VisitValidationStatus.rejected).count()
```
snopoke, line 59: you should be able to do these aggregations in a single query with conditional aggregations: https://docs.djangoproject.com/en/1.8/ref/models/conditional-expressions/#conditional-aggregation

snopoke: Would be good to put this is a tested function

commcare_connect/opportunity/api/views.py
```diff
@@ -30,3 +44,24 @@ def get(self, *args, **kwargs):
...
+        ret = {
+            "approved": approved,
+            "pending": pending,
+            "rejected": rejected,
+            "total_visits": total_visits,
+            "daily_visits": daily_visits,
+        }
+        return Response(UserVisitVerificationStatusSerializer(ret).data)
```
snopoke, line 67: For simple data like this where you've already constructed the dict there's no need for a serializer:
```suggestion
        return Response(ret)
```

calellowitz (changes requested):
I think this API should not be aggregate, but visit level. The phone can aggregate as needed, but will need to display form level information to the user

snopoke (approved):

commcare_connect/opportunity/api/serializers.py
```diff
@@ -38,3 +38,11 @@ class Meta:
...
     completed_modules = serializers.IntegerField()
     total_modules = serializers.IntegerField()
+
+
+class UserVisitSerializer(serializers.ModelSerializer):
+    deliver_form_name = serializers.CharField(source="deliver_form.name")
+
+    class Meta:
+        model = UserVisit
+        fields = ["id", "status", "visit_date", "deliver_form_name"]
```
calellowitz, line 48: should this also include  `xmlns` so forms can be matched if the name changes, and because its a bit nicer for machines than `deliver_form_name`.

Maybe @snopoke has thoughts?

calellowitz (approved):
Nothing blocking but one suggestion

commcare_connect/opportunity/api/serializers.py
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/78/commits/c1d56f49361c789fb99e7dd0c5b33bb487aa14b8

snopoke (approved):

PR #79, opened 2023-08-24
add users

calellowitz:
This adds a very rudimentary user upload workflow, allowing the user to submit a file with one phone number per line and have any numbers that match a ConnectID user added to the opportunity. It depends on the api introduced in https://github.com/dimagi/connect-id/pull/5

Opening as a draft pull request to allow review of the planned workflow, while I add tests. Notable future improvements include doing something with numbers that don't match (requirements for the POC are that all users will already be enrolled in connect id), and pulling more info about the users (name, etc.)


commcare_connect/opportunity/views.py
```diff
@@ -66,7 +66,10 @@ def get_success_url(self):
 
     def form_valid(self, form):
         form.instance.modified_by = self.request.user.email
-        return super().form_valid(form)
+        response = super().form_valid(form)
+        add_connect_users.delay(form.cleaned_data["users"], form.instance)
```
pxwxnvermx, line 70: Is there any reason for this task to be triggered when an Opportunity is edited? Is the UI for adding Phone Numbers a part of OpportunityEdit form?

commcare_connect/opportunity/forms.py
```diff
@@ -20,9 +20,12 @@ def __init__(self, *args, **kwargs):
             Row(Field("name")),
             Row(Field("description")),
             Row(Field("active")),
+            Row(Field("users")),
             Submit("submit", "Submit"),
         )
 
+        self.fields["users"] = forms.FileField()
```
snopoke, line 27: What kind of file are you expecting here? I imagine that the most likely type would be an Excel file.

If we want to support a plain text file we could just have a text box that you paste the numbers into.

commcare_connect/opportunity/tasks.py
```diff
@@ -17,3 +17,13 @@ def create_learn_modules_assessments(opportunity_id):
...
+
+
+@celery_app.task()
+def add_connect_users(data_file, opportunity):
+    numbers = [line.strip() for line in f]
+    result = request.get(f"{CONNECT_ID}/users/fetch_users", auth=(settings.CONNECTID_CLIENT_ID, settings.CONNECTID_CLIENT_SECRET), params={"phone_numbers": numbers})
+    data = result.json()
+    for user in data["found_users"]:
+        u = User.objects.get_or_create(username=user["username"])
+        OpportunityAccess.get_or_create(user=user, opportunity=oppportunity)
```
snopoke, line 29:
```suggestion
        OpportunityAccess.get_or_create(user=u, opportunity=oppportunity)
```

commcare_connect/opportunity/views.py
snopoke, line 70: I don't think you can pass a file to a celery task like this, also if you want the actual file you'll need to do `request.FILES["users"]`.

If you want to give celery access to the file you'll need to save it using Django storage backend:

```
name = uuid4().hex
storages["default"].save(name, request.FILES["users"])
```

commcare_connect/opportunity/forms.py
calellowitz: > If we want to support a plain text file we could just have a text box that you paste the numbers into.

That seems like a nice idea for now, so I switched to this.

calellowitz:
@snopoke This is ready for real review

commcare_connect/opportunity/forms.py
snopoke: Did you mean you switched to a text box? I still see a `FileField`.

calellowitz: Yep, I apparently stashed that change instead of committing it

snopoke (approved):

PR #98, opened 2023-09-04
fetch user bugs

calellowitz:
This removes an unnecessary column in `OpportunityAccess` that was non nullable (it will be replaced by foreign keys to the new claim model), and adds a simple test for the celery task that creates these.


pxwxnvermx (approved):
Awesome 🙌🏽

commcare_connect/opportunity/tests/test_tasks.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+
+from commcare_connect.opportunity.models import OpportunityAccess
+from commcare_connect.opportunity.tasks import add_connect_users
+from commcare_connect.opportunity.tests.factories import OpportunityFactory
+from commcare_connect.users.models import User
+
+
+class TestConnectUserCreation:
```
snopoke, line 11: 👏🏻

PR #101, opened 2023-09-05
Fixes for Linked users and Data Forwarding

pxwxnvermx:
Fixes in this PR
1. Added learn module jsonpath parser to match new forms being forwarded to CCC and save them as the correct completed module object.
2. Add `null=True and blank=True` to the email field on Users model. While adding more than 1 users to the Opportunity it was raising the unique constraint violation error as the new users were getting created with blank `email` field and more than 1 users cannot have the same email i.e blank in this case.


commcare_connect/users/models.py
```diff
@@ -20,7 +20,7 @@ class User(AbstractUser):
     name = models.CharField(_("Name of User"), blank=True, max_length=255)
     first_name = None  # type: ignore
     last_name = None  # type: ignore
-    email = models.EmailField(_("email address"), unique=True)
+    email = models.EmailField(_("email address"))
```
calellowitz, line 23: We can't make this field not unique, since it is the primary identifier of the user. What is the intent behind this change?

calellowitz (changes requested):
Can you add a description to this PR, out lining the intent behind this. This includes very major changes to how we manage users with no explanation or rationale.

commcare_connect/users/models.py
snopoke: I think you need to have `(null=True, blank=True, unique=True)`

snopoke (commented):
Would be great to have a test for this.

commcare_connect/users/models.py
calellowitz: @snopoke Does that create a partial index on non null values? That is what I was going to recommend.

calellowitz (commented):
Assuming Simon and I are on the same page about the DB stuff (which I think we are), this is fine by me. Leaving this as a "comment" rather than "approve" since I haven't seen the answer, but no longer blocking if he approves

calellowitz (approved):
Switching to approve because comment doesn't remove the "request changes" but still pending simon's approval

commcare_connect/users/models.py
snopoke: This doesn't create a partial index (though it's not strictly necessary for PostgreSQL since null values are not considered equal. In the long run a partial index would probably be better though.

@pxwxnvermx you can see how to do that here: https://docs.djangoproject.com/en/4.2/ref/models/constraints/#condition

(also remove the `unique=True` kwarg from the column)

pxwxnvermx: https://github.com/dimagi/commcare-connect/pull/101/commits/394c1794e8c37ef16679987dd16d4e133c2e4ac0

snopoke (approved):

PR #107, opened 2023-09-06
Pkv/fixes data forwarding

pxwxnvermx:
- Add jsonpath for assessment modules (missed in the last PR)
- Add `user.username` as the display instead of `user.name`, as `user.name` is not set for ConnectID users

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -70,18 +70,21 @@ def process_learn_modules(user, xform: XForm, app: CommCareApp, opportunity: Opp
     :param blocks: A list of learn module form blocks."""
     for module_data in blocks:
         module = get_or_create_learn_module(app, module_data)
-        CompletedModule.objects.update_or_create(
+        completed_module, created = CompletedModule.objects.get_or_create(
             user=user,
             module=module,
             opportunity=opportunity,
+            xform_id=xform.id,
             defaults={
                 "date": xform.received_on,
                 "duration": xform.metadata.duration,
-                "xform_id": xform.id,
                 "app_build_id": xform.build_id,
                 "app_build_version": xform.metadata.app_build_version,
             },
         )
+
+        if not created:
+            raise ProcessingError("Learn Module is already completed")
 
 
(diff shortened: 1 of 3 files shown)


commcare_connect/form_receiver/processor.py
```diff
@@ -70,15 +70,17 @@ def process_learn_modules(user, xform: XForm, app: CommCareApp, opportunity: Opp
     :param blocks: A list of learn module form blocks."""
     for module_data in blocks:
         module = get_or_create_learn_module(app, module_data)
-        CompletedModule.objects.create(
+        CompletedModule.objects.update_or_create(
```
snopoke, line 73: why is this update or create? I do think we should check that the form has not been processed but we should error if we receive two different forms with the same 'module'.

So I think this could be a `get_or_create` using the `xform_id` as one of the query params. If the model exists we verify that it's for the same module and ignore it if it is.

commcare_connect/opportunity/tables.py
```diff
@@ -15,7 +15,7 @@ class OpportunityAccessTable(tables.Table):
 
     class Meta:
         model = OpportunityAccess
-        fields = ("user.name", "learn_progress", "visit_count")
+        fields = ("user.username", "learn_progress", "visit_count")
```
calellowitz, line 18: this is fine for now, but long run we will want to fetch the name from connectid and use that rather than the username, since they are random strings and not user facing

commcare_connect/form_receiver/processor.py
pxwxnvermx: I was unsure on what should be done here, so i used the update or create, in case we recieved two entries it would avoid creating duplicate completed modules.
I will update the code to raise an error when it recieves a duplicate completed module. Will do the same for assessments also.

pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/107/commits/d679bb7c194c8a0ec1a7913f2a16abbff01bae8d

snopoke (approved):

PR #110, opened 2023-09-08
Claims model and apI

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-107)


commcare_connect/opportunity/api/views.py
```diff
@@ -43,3 +52,27 @@ class UserVisitViewSet(viewsets.GenericViewSet, viewsets.mixins.ListModelMixin):
...
+        claim, created = OpportunityClaim.objects.get_or_create(
+            opportunity_access=opportunity_access,
+            defaults={
+                "max_payments": opportunity.daily_max_visits_per_user,
+                "end_date": opportunity.end_date,
+            },
+        )
+
+        if not created:
+            return Response(status=400, data="Opportunity is already claimed")
```
snopoke, line 76: Maybe we just return a 200 here instead of an error since re-claiming isn't really an issue?

commcare_connect/opportunity/api/views.py
```diff
@@ -43,3 +52,27 @@ class UserVisitViewSet(viewsets.GenericViewSet, viewsets.mixins.ListModelMixin):
...
+
+class ClaimOpportunityView(APIView):
+    authentication_classes = [OAuth2Authentication]
+    permission_classes = [TokenHasReadWriteScope]
+
+    def post(self, *args, **kwargs):
+        opportunity_access = get_object_or_404(OpportunityAccess, user=self.request.user, opportunity=kwargs.get("pk"))
+        opportunity = opportunity_access.opportunity
+        if opportunity.learn_app.cc_domain != opportunity.deliver_app.cc_domain:
+            create_hq_user(self.request.user, opportunity.deliver_app.cc_domain, opportunity.api_key)
```
snopoke, line 65: Could put this after the claim object creation so that we only call it if there wasn't already a claim?

snopoke: Alternately I guess we could re-call it as long as it's idempotent.

snopoke (approved):

PR #114, opened 2023-09-13
remove username field

calellowitz:
`oauth2_provider` uses this attribute to determine where to save the username when creating users during token introspection. That has caused users it created to have their connectid username saved in the email column and username left blank. We already updated the [allauth settings](https://github.com/dimagi/commcare-connect/blob/main/config/settings/base.py#L253) to point to `username` so I don't believe this was doing much.


snopoke (approved):

snopoke (commented):
How does this impact 'admin' users?

calellowitz:
I don't think it should impact them at all (they are still authenticated and have those permissions), but I will confirm that in testing.

snopoke:
> I don't think it should impact them at all (they are still authenticated and have those permissions), but I will confirm that in testing.

The reason I ask is because of the changes to the tests you had to make which now use username instead of email but appear to be testing web user features.

calellowitz:
confirmed admin continues to work after i sign out and back in, at least with my user that was SSO created

PR #116, opened 2023-09-13
Claims: add budget existing user UI

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-109)

Form to Add budget to existing users:
image
image

This PR also includes a file for crispy-bootstrap5 modal template. This file can be removed once [this PR](https://github.com/django-crispy-forms/crispy-bootstrap5/pull/151) is merged for crispy-bootstrap5.


snopoke (commented):
This appears to mix up 'budget' with 'max_visits'. Since we don't currently have a 'budget' field per user I think we should be consistent with 'max_visits'.

It would be ideal if we could display the current 'max_visits' along with 'used_visits' for each user. I wonder if it shouldn't be shown as a table rather than a list of checkboxes? I think you could use an ['inline formset'](https://docs.djangoproject.com/en/4.2/topics/forms/modelforms/#inline-formsets).


The end date field could also use more explanation. Perhaps something like 'Extend Opportunity End Date'.

calellowitz (approved):
Besides simon's comment this looks good. One thing that changes is the budget calculation for the warning. The basics of the calculation are correct (users * budget per user), but budget per user will need to be (added visits * cost per visit) once you change the form to allow them to add visits/payments per user. That leaves a final calculation of (users selected * num added vists * payment per visit) to display in the warning

commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -0,0 +1,113 @@
...
+            </td>
+            <td>{{ form.instance.opportunity_access.visit_count }}</td>
+            <td>
+              <input type="date" name="{{ form.end_date.html_name }}" id="{{ form.end_date.auto_id }}"
+                     class="dateinput form-control{% if form.end_date.errors %} is-invalid{% endif %}"
+                     :value="isSelected('{{ form.instance.id }}') && end_date ? end_date : '{{ form.end_date.value|date:"Y-m-d" }}'"
+              />
+              {{ form.end_date|as_crispy_errors }}
+            </td>
+          </tr>
```
snopoke, line 63: How do these elements work (max visits and end date)? From the screenshot it seems like you could adjust them independently. But it also seems like the value is set from the 'global' controls at the bottom of the table.

But at least the JS budget calculation only takes the global max value.

I think we could possibly simplify this a bit if we just use the table to display the data and select the claims to update rather than using them to actually update the claims as well i.e. we could just do a single update query for all selected claims with the updated values.

commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -0,0 +1,113 @@
...
+{% block content %}
+  <div class="container">
+    <h1 class="display-5">Edit Opportunity</h1>
+    <h4>Adding Visits for Existing Users</h4>
+    <form method="post"
+          x-data="{
+            max_visits: 0,
+            end_date: '',
+            selected_users: [],
+            isSelected (id) { return this.selected_users.find(el => el === id)},
```
snopoke, line 16: adding in 'totalVisits' for using in the confirmation message:

```suggestion
            isSelected (id) { return this.selected_users.find(el => el === id)},
            totalVisits() { return this.selected_users.length * this.max_visits },
```

commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -0,0 +1,113 @@
...
+      <div class="modal fade" id="confirm-modal-add-visits" tabindex="-1" aria-labelledby="exampleModalLabel"
+           aria-hidden="true">
+        <div class="modal-dialog">
+          <div class="modal-content">
+            <div class="modal-header">
+              <h5 class="modal-title" id="exampleModalLabel">Confirm</h5>
+              <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+            </div>
+            <div class="modal-body">
+              Please confirm to add <b x-text="calculated_budget()"></b> budget to this Opportunity.
```
snopoke, line 102:
```suggestion
              You are about to allocate an additional <b x-text="totalVisits()"></b> which will result in an additional budget allocation of <b x-text="calculated_budget()"></b> to this Opportunity.
```

snopoke (commented):
These changes are looking great. Thanks for taking the suggestions on board. A few more comments about how the UI works.

Also, I think we could re-phrase the UI so that instead of setting a new 'max visits' we are adding N visits. This prevents us having to validate that the new max visits is > than existing max visits. It also works better this way if you have previously added some visits to some users and now want to add more.

commcare_connect/templates/opportunity/add_visits_existing_users.html
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/116/commits/611be9ab3e522249b619e82f13991ae0655eabf6

commcare_connect/templates/opportunity/add_visits_existing_users.html
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/116/commits/611be9ab3e522249b619e82f13991ae0655eabf6

commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -0,0 +1,113 @@
...
+          <tr>
+            <td colspan="4">
+              No Opportunity Claims yet.
+              <a href="{% url "opportunity:list" org_slug=request.org.slug %}">All Opportunities</a>
+            </td>
+          </tr>
+        {% endfor %}
+        </tbody>
+      </table>
+      <div class="row mb-2">
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/116/commits/f8d5e9c9740f23a58d35ede76dbca020f938918d

commcare_connect/opportunity/api/views.py
```diff
@@ -61,8 +60,7 @@ def get_queryset(self):
 
 
 class ClaimOpportunityView(APIView):
-    authentication_classes = [OAuth2Authentication]
-    permission_classes = [TokenHasReadWriteScope]
+    permission_classes = [IsAuthenticated]
```
snopoke, line 63: why did this need to change?

commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -0,0 +1,117 @@
...
+            <input type="number"
+                   class="numberinput form-control {% if form.additional_visits.errors %}is-invalid{% endif %}"
+                   name="additional_visits" id="id_additional_visits"
+                   x-model="additionalVisits"/>
+            {% for error in form.additional_visits.errors %}
+              <div class="invalid-feedback">
+                <strong>{{ error|escape }}</strong>
+              </div>
+            {% endfor %}
+          </div>
```
snopoke, line 68: Is there a reason to include the form markup directly instead of using crispy tags?

Crispy does have a field render tag: `{{ form.additional_visits|as_crispy_field }}`

https://django-crispy-forms.readthedocs.io/en/latest/api_templatetags.html#templatetags.crispy_forms_filters.as_crispy_field

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -39,6 +39,9 @@ <h1 class="display-4">Opportunities</h1>
             </a>
             <a class="btn btn-secondary btn-sm"
                href="{% url 'opportunity:edit' org_slug=request.org.slug pk=opportunity.id %}">Edit</a>
+            <a class="btn btn-secondary btn-sm"
+               href="{% url 'opportunity:add_budget_existing_users' org_slug=request.org.slug pk=opportunity.id %}">Add
+              Budget</a>
```
snopoke, line 44: We should also add a button to the opportunity details screen.

snopoke (approved):
Some questions and suggestions but nothing blocking.

commcare_connect/templates/opportunity/opportunity_list.html
pxwxnvermx: We already have the link for opportunity details above the `opportunity:edit` link.

commcare_connect/opportunity/api/views.py
pxwxnvermx: The authentication and permission here were setup for client_credentials workflow. Since we have ConnectID as a auth also we can directly use the IsAuthenticated permission as we do for other apis.

commcare_connect/templates/opportunity/add_visits_existing_users.html
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/116/commits/c06cebbdaa8351f00d43d9896b94026f7d93e2c3

commcare_connect/templates/opportunity/opportunity_list.html
snopoke: I meant you should add a button for 'add budget' to the opportunity details screen.

PR #120, opened 2023-09-19
Update opportunity api to include user progress and learn info 

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-115)

### DIFF
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -1,14 +1,34 @@
 from rest_framework import serializers
 
-from commcare_connect.opportunity.models import CommCareApp, CompletedModule, Opportunity, OpportunityClaim, UserVisit
+from commcare_connect.opportunity.models import (
+    CommCareApp,
+    CompletedModule,
+    LearnModule,
+    Opportunity,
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/api/serializers.py
```diff
@@ -34,12 +55,26 @@ class Meta:
...
+            return OpportunityClaimSerializer(claim.first()).data
+        return None
+
+    def get_learn_progress(self, obj):
+        opp_access = self._get_opp_access(obj)
+        total_modules = LearnModule.objects.filter(app=opp_access.opportunity.learn_app)
+        completed_modules = CompletedModule.objects.filter(opportunity=opp_access.opportunity)
+        return {"total_modules": total_modules.count(), "completed_modules": completed_modules.count()}
+
+    def _get_opp_access(self, obj):
```
snopoke, line 74: This is getting called twice now so it should be cached / memoized to prevent multiple DB queries.

pxwxnvermx: Can i use the `quickcache` decorator to cache this method?

snopoke (approved):

PR #127, opened 2023-09-21
Add deliver_progress and passing_score to opportunity api

pxwxnvermx:


commcare_connect/opportunity/api/serializers.py
```diff
@@ -24,7 +24,7 @@ class CommCareAppSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = CommCareApp
-        fields = ["cc_domain", "cc_app_id", "name", "description", "organization", "learn_modules"]
+        fields = ["cc_domain", "cc_app_id", "name", "description", "organization", "learn_modules", "passing_score"]
```
snopoke, line 27: I think we want to add the list of 'assessments' which has `score`, `passing_score` and `passed` fields.

pxwxnvermx: We are not sending the CompletedModules or Assessments in this API. The assessments submitted by user can be included in the learn_progress API.

snopoke (approved):

PR #130, opened 2023-09-25
add basic payment models forms

calellowitz:
This is a first pass at the payment models, a basic form to add payments, and a UI to view them. I am still working on getting the form to work perfectly but the intent should be clear.

I am looking for feedback on the approach and the data model as I finish these up.

Screenshot of the new payment list UI:
Screenshot from 2023-10-03 20-44-20


snopoke:
@calellowitz can you give more info about the workflow here and how the 2 models interact with eachother (PaymentUnit and Payment).

calellowitz:
Very valid question. This is supposed to lay the groundwork for these [user stories](https://docs.google.com/document/d/1z3Gm9N2o_0B8Z0ldscMi0h1E1k-77dVQyDydR6qnqHc/edit). It does not yet include a UI to configure `PaymentUnit` or the API to pull payments, but should include the rest (although I want to do a bit more work on the payment ui)

calellowitz:
I added a UI to import and export payment info (like the one for user visits), and pulled this out of draft. Ready for review on this work now.

calellowitz:
I also updated the delivery progress API to combine UserVisit and the new Payment info, so that the mobile would only require one api call for both and since they are in the same mobile UI screen. I renamed the existing user visit api, but kept it's functionality unchanged.

commcare_connect/opportunity/api/serializers.py
```diff
@@ -98,3 +99,21 @@ class UserVisitSerializer(serializers.ModelSerializer):
...
+        model = Payment
+        fields = ["amount", "date_paid"]
+
+
+class DeliveryProgressSerializer(serializers.Serializer):
+    deliveries = serializers.SerializerMethodField()
+    payments = serializers.SerializerMethodField()
+
+    def get_payments(self, obj):
+        return [PaymentSerializer(payment) for payment in obj.payment_set.all()]
```
pxwxnvermx, line 115: This serializer can be changed to use `many=True` that will automatically serialize all the objects in the queryset to a ListSerializer. 
```suggestion
        return PaymentSerializer(obj.payment_set.all(), many=True)
```

commcare_connect/opportunity/tasks.py
```diff
@@ -54,3 +54,13 @@ def generate_visit_export(opportunity_id: int, date_range: str, status: list[str
...
     storages["default"].save(export_tmp_name, ContentFile(content))
     return export_tmp_name
+
+
+@celery_app.task()
+def generate_payment_export(opportunity_id: int, export_format: str):
+    opportunity = Opportunity.objects.get(id=opportunity_id)
+    dataset = export_empty_payment_table(opportunity)
+    content = dataset.export("xlsx")
+    export_tmp_name = f"{now().isoformat()}_{opportunity.name}_payment_export.{export_format}"
```
pxwxnvermx, line 64: Will this lead to the file having different content-encoding (xlsx) and file format incase a user selects `csv`?

config/api_router.py
```diff
@@ -18,12 +19,13 @@
 
 router.register("users", UserViewSet)
 router.register("opportunity", OpportunityViewSet, basename="Opportunity")
-router.register("opportunity/(?P<opportunity_id>.+)/deliver_visit", UserVisitViewSet, basename="UserVisit")
+router.register("opportunity/(?P<opportunity_id>.+)/user_visit", UserVisitViewSet, basename="UserVisit")
```
pxwxnvermx, line 22: nit: Can we remove this API endpoint now since the data from this is also include in the deliver_progress endpoint?

commcare_connect/opportunity/api/serializers.py
```diff
@@ -98,3 +99,21 @@ class UserVisitSerializer(serializers.ModelSerializer):
...
+
+    def get_deliveries(self, obj):
+        deliveries = UserVisit.objects.filter(opportunity=obj.opportunity, user=obj.user)
+        return [UserVisitSerializer(user_visit) for user_visit in deliveries]
```
pxwxnvermx, line 119: Same as above.
```suggestion
        return UserVisitSerializer(deliveries, many=True)
```

config/api_router.py
calellowitz: I left it intentionally since it is a much more efficient API if you just want visit info, and there isn't really a cost, but happy to remove it if you feel strongly.

commcare_connect/opportunity/tasks.py
calellowitz: good catch, I thought I updated all those instances

commcare_connect/opportunity/api/serializers.py
calellowitz: nice, I was trying to figure out how to use `many` in the attribute level fields, but hadn't considered I could use it in the methods

config/api_router.py
pxwxnvermx: I am fine with keeping this API. Just wanted to get your thoughts about keeping this.

pxwxnvermx (approved):

PR #139, opened 2023-09-29
Use Deliver Units instead of 'forms'

snopoke:
Jira: https://dimagi-dev.atlassian.net/browse/CCPC-127

This updates the app to use delivery units instead of delivery forms:

* Creating an opportunity no longer requires you to select a delivery form, just a delivery app
* Form processing updated to process deliver unit XML blocks
* View and export updated to show deliver unit data


calellowitz (approved):
Is the intent here that these are entirely configured in HQ? That they are dynamically created in CCC based on received forms rather than matching hq forms to preconfigured units in CCC? I think that is potentially fine, but it might make some of the payment work trickier, since we will still need to associate these with payment configuration. Did you have a sense of how that would work? Otherwise, this looks great

snopoke:
> Is the intent here that these are entirely configured in HQ? That they are dynamically created in CCC based on received forms rather than matching hq forms to preconfigured units in CCC? I think that is potentially fine, but it might make some of the payment work trickier, since we will still need to associate these with payment configuration. Did you have a sense of how that would work? Otherwise, this looks great

That was my initial thought. We could easily parse them out of the app in the same way we do for learn modules.

One general issue with this approach is how to ensure that the data in Connect is kept in sync with the app e.g. if a new version is released that changes the learn / delivery blocks.

PR #143, opened 2023-10-03
use proper attribute

calellowitz:
`auto_now_added` is the intended attribute, setting the date on creation. `auto_created` does not do anything


pxwxnvermx (approved):
Thanks 🙌🏼

PR #146, opened 2023-10-04
Daily and Max visits limits on Deliver Visit processor

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-124)


commcare_connect/form_receiver/processor.py
```diff
@@ -145,6 +152,13 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         app_build_version=xform.metadata.app_build_version,
         form_json=xform.raw_form,
     )
+    if (
+        counts["daily"] >= opportunity.daily_max_visits_per_user
```
calellowitz, line 156: These (or at least the last two), should be based on the claim object for that user since they will vary across users (when budget is added and the end date is extended)

commcare_connect/opportunity/models.py
```diff
@@ -166,6 +166,7 @@ class VisitValidationStatus(models.TextChoices):
     pending = "pending", gettext("Pending")
     approved = "approved", gettext("Approved")
     rejected = "rejected", gettext("Rejected")
+    extra = "extra", "extra"
```
calellowitz, line 169: The second element should use `gettext` for translation like the others and follow the same capitalization convention

calellowitz (commented):
I believe we need to add a filter to the delivery progress API, to exclude forms in this state since they do not count for the user. We could have the mobile do it, but since we know it doesn't need them, it might be easier for us to do it on the web side. We also need to update all the reports in the opportunity detail UI to exclude forms in this state, from the count. I think we can leave them in the export for now, although we can confirm with Rama tomorrow.

commcare_connect/opportunity/models.py
```diff
@@ -166,6 +174,7 @@ class VisitValidationStatus(models.TextChoices):
     pending = "pending", gettext("Pending")
     approved = "approved", gettext("Approved")
     rejected = "rejected", gettext("Rejected")
+    extra = "extra", gettext("Extra")
```
snopoke, line 177: What about calling this 'Over Limit' or something? 'Extra' seems a bit vague to me.

pxwxnvermx: Yeah, that makes more sense. Renamed. https://github.com/dimagi/commcare-connect/pull/146/commits/d23c83ccfe5a61d050f95a6a440ec974836bb4b2

commcare_connect/form_receiver/processor.py
```diff
@@ -133,7 +138,13 @@ def process_deliver_form(user, xform: XForm, app: CommCareApp, opportunity: Oppo
 
 def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Opportunity, deliver_unit_block: dict):
     deliver_unit = get_or_create_deliver_unit(app, deliver_unit_block)
-    UserVisit.objects.create(
+    counts = UserVisit.objects.filter(opportunity=opportunity, user=user).aggregate(
+        daily=Count("pk", filter=Q(visit_date__date=datetime.date.today())), total=Sum("pk", default=0)
```
calellowitz, line 142: Apologies for missing this the first time, I think this needs to use the date of actual delivery (form date), rather than date we receive it, which could happen at a delay due to connectivity issues. I am not sure we save that now so we may need to update the model as well, but in short if you do one visit on three consecutive days, but submit all those forms at once on the third day, all should count since they were done on different days.

calellowitz (changes requested):
Sorry this feedback came in so late, but I think there is a fundamental issue in how we are counting this. The limit should apply to how many visits completed on a given day, not how many forms make it to connect on that day.

commcare_connect/form_receiver/processor.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/146/commits/92a105a5b74eaaa92e080bc05d358c8432489c5f

I have added the check to count visits on the actual visit_date of the form that we are currently processing. We save the form visit_date on the UserVisit model already.

calellowitz (approved):

PR #149, opened 2023-10-05
Add claimed_budget and utilised_budget to model and ui

pxwxnvermx:
[Ticket1](https://dimagi-dev.atlassian.net/browse/CCPC-118)
[Ticket2](https://dimagi-dev.atlassian.net/browse/CCPC-122)

image


commcare_connect/opportunity/models.py
```diff
@@ -67,13 +67,34 @@ def __str__(self):
...
         )["max_payments__sum"]
         if used_budget is None:
             used_budget = 0
-        return self.total_budget - used_budget
+        return used_budget
+
+    @property
+    def approved_visits(self):
+        approved_user_visits = (
+            UserVisit.objects.filter(opportunity=self, status=VisitValidationStatus.approved)
```
calellowitz, line 93: I think this is redundant. It counts only `approved` visits and excludes `over_limit`, but since a visit can't be both, we don't need the `exclude` statement.

calellowitz (approved):
Nothing blocking, but I think there is a redundant clause in one of the queries

commcare_connect/opportunity/models.py
pxwxnvermx: Removed. https://github.com/dimagi/commcare-connect/pull/149/commits/76ef45bcdcc0d69cfa5bfb567668b70905d92908

calellowitz (approved):

PR #151, opened 2023-10-05
set default protocol

calellowitz:
This is used by `allauth` in the `build_absolute_uri` and will cause it to use `https` instead of `http`. It looks like the `sites` framework expects the URL to lack a protocol, so I will remove that as well as part of this change.

Right now the constructed URLs begin `http://https://` with `https` coming from the URL in the `Site` object, and `http` coming from `allauth`

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -131,3 +131,5 @@ SPECTACULAR_SETTINGS["SERVERS"] = [  # noqa: F405
 ]
 # Your stuff...
 # ------------------------------------------------------------------------------
+
+DEFAULT_HTTP_PROTOCOL = "https"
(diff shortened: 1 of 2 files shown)


snopoke (approved):

snopoke (commented):
Just looking at the allauth docs, should this be `ACCOUNT_DEFAULT_HTTP_PROTOCOL`?

calellowitz:
ugh, probably. I pulled the name from the code https://github.com/pennersr/django-allauth/blob/main/allauth/utils.py#L275 and hadn't noticed that it wasn't using the normal `settings` but `account_settings`

PR #157, opened 2023-10-09
Add deliver units to opportunity detail UI, accordion for Learn Modules and Deliver units

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-137)

https://github.com/dimagi/commcare-connect/assets/31195143/bb9c166f-f96a-4c11-aa0a-38a3634767c3

Deliver Units Table:
image

### DIFF
--- a/commcare_connect/templates/opportunity/opportunity_detail.html
+++ b/commcare_connect/templates/opportunity/opportunity_detail.html
@@ -156,7 +156,7 @@
                 <thead class="table-light">
                 <tr>
                   <th>Name</th>
-                  <th>Slug</th>
+                  <th>ID</th>
                 </tr>
                 </thead>
                 <tbody>
--- a/commcare_connect/templates/opportunity/opportunity_detail.html
+++ b/commcare_connect/templates/opportunity/opportunity_detail.html
@@ -114,27 +114,64 @@
         </tr>
         </tbody>
       </table>
-      <h4>Learn App Modules</h4>
-
-      <table class="table border table-custom">
-        <thead class="table-light">
-        <tr>
-          <th>Name</th>
-          <th>Description</th>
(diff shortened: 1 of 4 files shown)


pxwxnvermx:
Yes, these would also be parsed from the app on Opportunity Creation.

commcare_connect/templates/opportunity/opportunity_detail.html
```diff
@@ -114,27 +114,64 @@ <h1 class="display-5 mb-0">{{ object.name }}</h1>
...
+              Deliver Units
+            </button>
+          </h2>
+          <div id="deliver-units" class="accordion-collapse collapse" aria-labelledby="headingTwo" data-bs-parent="#accordionExample">
+            <div class="accordion-body p-0 m-0">
+              <table class="table table-borderless table-custom">
+                <thead class="table-light">
+                <tr>
+                  <th>Name</th>
+                  <th>Slug</th>
```
calellowitz, line 159: Nit: I think this column should be called "ID". That is a much more understandable name to non developers, and matches what the field is called in the form builder.

calellowitz (approved):
One nit, but otherwise looks good. Confirming that column will match the ID field in new delivery units and only shows the mangled xmlns because of the data migration? If so, seems fine.

commcare_connect/templates/opportunity/opportunity_detail.html
pxwxnvermx: Renamed. https://github.com/dimagi/commcare-connect/pull/157/commits/6fc12416796f2f55a9fd97c43da8fc1a0f55ae5b

This column shows the ID for the deliver units. I have attached a screenshot with the correct table.

PR #165, opened 2023-10-12
Payment Unit model, Forms and UI

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-148)

- The form only shows those deliver units which are not already included in a payment unit.

## Payment Unit Form:
image

## Payment Unit Table:
image


commcare_connect/opportunity/tables.py
```diff
@@ -63,3 +63,18 @@ class Meta:
...
+    details = columns.LinkColumn(
+        "opportunity:edit_payment_unit",
+        verbose_name="",
+        text="Edit",
+        args=[utils.A("opportunity.organization.slug"), utils.A("opportunity.id"), utils.A("pk")],
+    )
+
+    class Meta:
+        model = PaymentUnit
+        fields = ("name", "amount")
```
calellowitz, line 78: This should probably also show the delivery units, so the users can see what they have configured.

calellowitz (changes requested):
Few small changes but otherwise looks great

commcare_connect/opportunity/models.py
```diff
@@ -162,6 +162,13 @@ def last_visit_date(self):
         return
 
 
+class PaymentUnit(models.Model):
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.PROTECT)
+    amount = models.PositiveIntegerField()
+    name = models.CharField()
```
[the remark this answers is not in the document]
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/165/commits/8155b59faaeb6e9fb413a0246b4ff32a69bd9846
I have added the same max_length that we have in other models for name charfield.

commcare_connect/opportunity/tables.py
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/165/commits/5a9b2594207aae7d452aa3c1f418cac1f5c1bad5

calellowitz (approved):

PR #176, opened 2023-10-20
tabbed table pagination

calellowitz:
https://dimagi-dev.atlassian.net/browse/CCPC-161
This should be a fix for the above ticket. It replaces the default pagination controls in the django tables template with custom ones using htmx that take into account our tabbed table setup, by reloading only the relevant table when a pagination control is selected.

This is mostly just copying the pagination block from the [template](https://github.com/jieter/django-tables2/blob/master/django_tables2/templates/django_tables2/bootstrap5.html#L57-L95) we were previously using, and replacing the `<a>` tags with htmx gets that replace the ancestor table element.

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -72,6 +72,7 @@ class UserPaymentsTable(tables.Table):
         fields = ("amount", "date_paid")
         orderable = False
         empty_text = "No payments made for this user"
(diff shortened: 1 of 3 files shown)


OrangeAndGreen (approved):

pxwxnvermx (approved):
Awesome. Thanks a lot

snopoke (approved):

PR #184, opened 2023-10-26
Opportunity APIs Tests

pxwxnvermx:
This PR adds tests for all the apis which are in the Opportunity django project. Also removed unused factories and fixtures.


calellowitz (approved):
Looks fine. Is there a reason you are checking the length of the keys instead of the content

pxwxnvermx:
Yeah, i am only creating a single object which is sent back through the API, so just checking the length of the list of objects in the response.

PR #189, opened 2023-11-01
budget updates

calellowitz:
The two commits in this PR address two issues
1) As seen in the screenshot below, it adds a new field of "max users" and automatically calculates the total budget so users do not need a calculator. It does not change anything on the backend, and we still only store the total budget.
2) Marks an opportunity as accepted when the user starts learning. This aligns with common sense understandings of what it means to accept an opportunity, and will be reflected in new language in the mobile UI indicating that starting learning consents to sharing the user info (in this case just name) with the project.

image


pxwxnvermx (approved):
Thanks 🙌🏽

PR #196, opened 2023-11-10
Add logging to create hq user

pxwxnvermx:
Create HQ User currently does not log anything when the api request return any status other than 201.
This PR adds logging to the function in case the api return an error.

### DIFF
--- a/commcare_connect/users/helpers.py
+++ b/commcare_connect/users/helpers.py
@@ -36,8 +36,6 @@ def create_hq_user(user, domain, api_key):
         },
         headers={"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"},
     )
-    if hq_request.status_code == 201:
-        return True
     try:
         hq_request.raise_for_status()
     except httpx.HTTPStatusError as e:
@@ -45,7 +43,7 @@ def create_hq_user(user, domain, api_key):
             f"{e.response.status_code} Error response {e.response.text} while creating user {user.username}"
         )
 
-    return False
+    return hq_request.status_code == 201
 
 
 def invite_user(user, opportunity_access):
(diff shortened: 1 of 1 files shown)


commcare_connect/users/helpers.py
```diff
@@ -37,6 +39,7 @@ def create_hq_user(user, domain, api_key):
     )
     if hq_request.status_code == 201:
         return True
+    logging.info(f"Create User Response: {hq_request.text}")
```
snopoke, line 42: Maybe this should be `logging.error` - alternatively what about raising an exception and capturing it in Sentry?

pxwxnvermx: I added httpx and then raised an exception on raise for status. Since we are only interested any errors which are not 2xx response codes. https://github.com/dimagi/commcare-connect/pull/196/commits/28e2b999ebd8abb6924121fd675cb427928b50d9

commcare_connect/users/helpers.py
```diff
@@ -37,6 +38,13 @@ def create_hq_user(user, domain, api_key):
...
     if hq_request.status_code == 201:
         return True
+    try:
+        hq_request.raise_for_status()
+    except httpx.HTTPStatusError as e:
+        raise CommCareHQAPIException(
+            f"{e.response.status_code} Error response {e.response.text} while creating user {user.username}"
+        )
+
     return False
```
snopoke, line 48: nit: can be refactored slightly:

```suggestion
    try:
        hq_request.raise_for_status()
    except httpx.HTTPStatusError as e:
        raise CommCareHQAPIException(
            f"{e.response.status_code} Error response {e.response.text} while creating user {user.username}"
        )

    return hq_request.status_code == 201
```

snopoke (approved):

PR #199, opened 2023-11-17
Fix additional visits save to opportunity total budget

pxwxnvermx:
This PR fixes two bugs in Opportuntiy Add budget to exisiting users workflow.
1. Additional visits assigned to users were not added to the total budget for the opportunity
2. When no end date was selected by users the end_date was unbound and led to an error.


calellowitz (approved):

calellowitz:
Tests would be good for this

PR #205, opened 2023-11-22
enforce unique apps

calellowitz:
Since this can cause all sorts of issues, I think we want to enforce unique app ids for now. We will have to do some prep work to delete duplicates in the current db, but since those are broken anyway, that should be fine. When we implement r reusable apps, we can remove this constraint, but for now its a giant trap that can accidentally ruin someone else's opportunity.


pxwxnvermx (approved):

pxwxnvermx:
We can also add the unique constraint to the learn_app and deliver_app columns on the Opportunity Model. That will ensure that no opportunties can reuse the apps from older opportunties.

pxwxnvermx:
Closing this PR. Please re-open if these changes are still needed.

PR #208, opened 2023-11-28
limit celery memory

calellowitz:
This limits the celery docker container to half the memory of the machine, to prevent the OOMs from large exports.

### DIFF
--- a/deploy/config/deploy.production.yml
+++ b/deploy/config/deploy.production.yml
@@ -11,6 +11,7 @@ servers:
     options:
       # create by ansible
       env-file: '/home/connect/www/commcare-connect/docker.env'
+      m: 1g
     cmd: /start_celery
     labels:
       traefik.enable: false


pxwxnvermx (approved):

deploy/config/deploy.production.yml
```diff
@@ -11,6 +11,7 @@ servers:
     options:
       # create by ansible
       env-file: '/home/connect/www/commcare-connect/docker.env'
+      m: 1g
```
pxwxnvermx, line 14: Is `m` an alias for `memory`?

calellowitz: yeah, https://docs.docker.com/config/containers/resource_constraints/#limit-a-containers-access-to-memory

pxwxnvermx: This might need to be `memory` because this is a different tool than docker.

pxwxnvermx: https://kamal-deploy.org/docs/configuration#using-accessories-for-database-cache-search-services

PR #210, opened 2023-11-28
end date is inclusive

calellowitz:
Title says it all. This affects both the opportunity list UI, and the form processor.


calellowitz:
This is a duplicate of https://github.com/dimagi/commcare-connect/pull/207

PR #213, opened 2023-11-30
Add additional columns to the User Status table

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-217)

image


calellowitz (changes requested):
The total line on the bottom should be more visually distinguished from the rest of the table, otherwise it is hard to tell that it is communicating different data. You can look in HQ reports to see how we do that in other places. At a minimum, I think you want a different background color, and bold font for the word "total"

pxwxnvermx:
I have changed the table footer color. I also tried a darker top border to distinguish between table footer and the table body, but that was not very visually appealing. Update the screenshot to reflect the changes.

config/settings/base.py
```diff
@@ -296,6 +295,9 @@
     "thead": {
         "class": "table-secondary",
     },
+    "tfoot": {
+        "class": "table-light fw-bold",
```
calellowitz, line 299: If it is possible to only bold the word "total" and not the numbers, that would be awesome, otherwise this is fine.

calellowitz (changes requested):

commcare_connect/opportunity/tables.py
```diff
@@ -74,12 +74,38 @@ class Meta:
         template_name = "django_tables2/bootstrap5.html"
 
 
+class AggregateColumn(columns.Column):
+    def render_footer(self, bound_column, table):
+        return sum(1 if bound_column.accessor.resolve(row) else 0 for row in table.data)
```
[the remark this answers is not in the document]
pxwxnvermx: This does total for the visible columns only. The calculation is based on the data returned by the database and only runs on the data in memory, without making any additional calls to the database.

commcare_connect/opportunity/models.py
```diff
@@ -205,6 +205,27 @@ def display_name(self):
         else:
             return "---"
 
+    @property
```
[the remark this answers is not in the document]
pxwxnvermx: Added an annotated query which calculates everything in a single query. The user status table view was making around 23 queries before to fetch 1 row of data. Now it is taking 8 queries.

commcare_connect/opportunity/tables.py
```diff
@@ -74,12 +74,38 @@ class Meta:
...
+
+
 class UserStatusTable(tables.Table):
-    display_name = columns.Column(verbose_name="Name")
+    display_name = columns.Column(verbose_name="Name", footer="Total")
+    accepted = BooleanAggregateColumn(verbose_name="Accepted")
+    claimed = AggregateColumn(verbose_name="Job Claimed", accessor="opportunityclaim.date_claimed")
+    started_learning = AggregateColumn(verbose_name="Started Learning", accessor="date_learn_started")
+    completed_learning = AggregateColumn(verbose_name="Completed Learning", accessor="date_learn_completed")
+    learn_passed = BooleanAggregateColumn(verbose_name="Learn Passed", accessor="has_passed_learn")
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/213/commits/3877811a8a90ad69e87b6208bb9790904aa00c37

calellowitz (commented):
How confident are you in the correctness of the query? Since these are so complex some tests may help, but once those are in, it could be good to go. I would deploy this branch to staging for performance testing, though.

commcare_connect/opportunity/helpers.py
```diff
@@ -0,0 +1,30 @@
...
+from commcare_connect.opportunity.models import Opportunity, OpportunityAccess
+
+
+def get_annotated_opportunity_access(opportunity: Opportunity):
+    access_objects = (
+        OpportunityAccess.objects.filter(opportunity=opportunity)
+        .select_related("user", "opportunityclaim")
+        .annotate(
+            last_visit_date_d=Max("user__uservisit__visit_date", filter=Q(user__uservisit__opportunity=opportunity)),
+            learn_progress_d=(Count("user__completed_modules") / Count("opportunity__learn_app__learn_modules")) * 100,
```
calellowitz, line 12: I think the completed module count needs an opportunity filter, since right now it is counting all completed modules for the user.

commcare_connect/opportunity/helpers.py
```diff
@@ -0,0 +1,30 @@
...
+            date_deliver_started=Min(
+                "user__uservisit__visit_date", filter=Q(user__uservisit__opportunity=opportunity)
+            ),
+            passed_assessment=Case(
```
calellowitz, line 16: I think this filters for whether they have taken the assessment, not whether they have passed.

commcare_connect/opportunity/helpers.py
calellowitz: Additionally when I run this function locally, I get a division by zero error and I suspect it is here (since I think this is the only place that does division. My user has done zero learning. It would be good to test this at each stage of the opportunity lifecycle to ensure it can handle all states of data.

calellowitz (changes requested):
I think I found some bugs in the query. Let me know if you have questions about those.

commcare_connect/opportunity/helpers.py
pxwxnvermx: > I think the completed module count needs an opportunity filter, since right now it is counting all completed modules for the user.
Thanks for pointing that out. I will add a fix for that and add a test which can check the data when opportunity is active with user having completed only some learning modules.

> Additionally when I run this function locally, I get a division by zero error and I suspect it is here (since I think this is the only place that does division. My user has done zero learning. It would be good to test this at each stage of the opportunity lifecycle to ensure it can handle all states of data.

I think this could be due to the Opportunity Learn App not having any learn modules as the division by zero error only comes when the divisor is 0 i.e learn_modules count in this case.

commcare_connect/opportunity/helpers.py
pxwxnvermx: Fixed. Added filter for passed assessments only. https://github.com/dimagi/commcare-connect/pull/213/commits/daff41020f739df242e773d68232b3958ff01042

commcare_connect/opportunity/helpers.py
pxwxnvermx: Added extra test cases which cover no learn data, partial learn data, learn complete and assessment passed, and all data.

calellowitz (approved):
There is an edge case around opportunities with no learn modules here (they haven't really completed the learning so it seems wrong to put that date), but I think this is fine to fix that in a follow up. It is a simple as an extra `AND` condition on that last bit.

commcare_connect/opportunity/tables.py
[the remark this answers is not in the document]
calellowitz: I think this should be for everyone (that is more standard in reporting UIs), but that can also be addressed in a follow up.

PR #216, opened 2023-12-06
sum assessments and skip total

calellowitz:
It was making a row for every value of `passed_assessment`, this aggregates all values into 1 by summing. If any assessment has value 1 (meaning it is for the right opportunity and is passed), they will get a true value, otherwise a failing value.


pxwxnvermx (approved):
Thanks.

PR #217, opened 2023-12-07
Fix user status table export aggregates queries

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-217)

This PR includes fixes for bugs related to the User Status Table.
- Exports through excel were failing due to date with timezone info, this bug is fixed now and the timestamp are converted to naive date time timestamps.
- The date for Completed learn was not showing up for users due to incorrect completed_modules_count being calculated as the `distinct=True` param was not specified for the Count function.
- Fixed the User Status Table not loading up due to None opportunity claims for users who have not yet claimed the opportunity, which also caused the total column to error.

### DIFF
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -22,6 +22,7 @@ def get_annotated_opportunity_access(opportunity: Opportunity):
             completed_modules_count=Count(
                 "user__completed_modules", filter=Q(user__completed_modules__opportunity=opportunity)
             ),
+            job_claimed=Case(When(Q(opportunityclaim__isnull=False), then="opportunityclaim__date_claimed")),
         )
         .annotate(
             date_learn_completed=Case(
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -30,7 +30,9 @@ def get_annotated_opportunity_access(opportunity: Opportunity):
             date_learn_completed=Case(
                 When(
                     Q(completed_modules_count=learn_modules_count),
-                    then=Max("user__completed_modules__date"),
+                    then=Max(
+                        "user__completed_modules__date", filter=Q(user__completed_modules__opportunity=opportunity)
+                    ),
                 )
             )
         )
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -20,7 +20,9 @@ def get_annotated_opportunity_access(opportunity: Opportunity):
                 )
             ),
             completed_modules_count=Count(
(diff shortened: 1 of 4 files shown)


commcare_connect/opportunity/helpers.py
```diff
@@ -20,14 +20,19 @@ def get_annotated_opportunity_access(opportunity: Opportunity):
                 )
             ),
             completed_modules_count=Count(
-                "user__completed_modules", filter=Q(user__completed_modules__opportunity=opportunity)
+                "user__completed_modules",
+                filter=Q(user__completed_modules__opportunity=opportunity),
+                distinct=True,
             ),
+            job_claimed=Case(When(Q(opportunityclaim__isnull=False), then="opportunityclaim__date_claimed")),
```
calellowitz, line 27: Why add this instead of what we had before?

commcare_connect/opportunity/helpers.py
```diff
@@ -20,14 +20,19 @@ def get_annotated_opportunity_access(opportunity: Opportunity):
                 )
             ),
             completed_modules_count=Count(
-                "user__completed_modules", filter=Q(user__completed_modules__opportunity=opportunity)
+                "user__completed_modules",
+                filter=Q(user__completed_modules__opportunity=opportunity),
+                distinct=True,
```
calellowitz, line 25: What bug does this fix?

commcare_connect/opportunity/tables.py
```diff
@@ -75,7 +75,8 @@ class Meta:
 
 
 class AggregateColumn(columns.Column):
-    pass
+    def render_footer(self, bound_column, table):
+        return sum(1 if bound_column.accessor.resolve(row) else 0 for row in table.data)
```
calellowitz, line 79: Does this still only total what is visible? I think we will want to fix that before adding the total row back.

calellowitz (commented):
It would be good to add a description that describes the changes and how they fix issues, otherwise it is quite hard to review.

commcare_connect/opportunity/helpers.py
pxwxnvermx: This would be the completed learn date not showing up. It was counting the rows incorrectly so i added the distinct to make the count avoid counting duplicates/ count the rows only once.

commcare_connect/opportunity/helpers.py
pxwxnvermx: The total column tries to resolve a related object i.e opportunityclaim in this case which could also be None leading to no related object found error. So i added this aggregation which tries to avoid using django to get the foreign key to opportunityclaim and does it in aggregation instead.

PR #220, opened 2023-12-11
Deliver Status Table

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-221)

- Deliver Status Table
image

- User Visits List
image


calellowitz:
I think this wants to replace the `Delivery Forms` tab and be called `Delivery Status` rather than a new tab, otherwise the UI looks good. I will review the code more closely soon.

commcare_connect/opportunity/tests/test_export.py
```diff
@@ -163,3 +173,31 @@ def test_export_user_status_table_data(opportunity: Opportunity):
     dataset = export_user_status_table(opportunity)
     prepared_test_dataset = _get_prepared_dataset_for_user_status_test(rows)
     assert prepared_test_dataset.export("csv") == dataset.export("csv")
+
+
+@pytest.mark.django_db
+def test_export_deliver_status_date(opportunity: Opportunity):
```
calellowitz, line 179: This test is fine as an end to end test of export, but it would be good to add specific tests with known outcomes just for the db query. These can focus on edge cases (zeros in certain fields, or all fields), as well very common but untested scenarios like multiple users, or users with visits in other opportunities. Testing just the query results will cut down on test time (no need to generate a csv or use tablib), and possible complicating factors (errors occurring in other parts of the code).

pxwxnvermx: I added some tests for the Query. Since the UserVisit factory randomly assigns the status it helps in testing if the query performs correctly for a lot of different cases. https://github.com/dimagi/commcare-connect/pull/220/commits/8f1ab1bae2edb630cb915e5ff10294ee2a7b42fe

PR #221, opened 2023-12-13
Fix Payment import failing on sending notification

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-215)

- Added a separate task to send payment notification in order to avoid blocking the user payment import.
- Increased the read timeout for the send message bulk API

### DIFF
--- a/commcare_connect/opportunity/tasks.py
+++ b/commcare_connect/opportunity/tasks.py
@@ -162,12 +162,10 @@ def _get_deliver_message(access: OpportunityAccess):
 
 
 @celery_app.task()
-def send_payment_notification(opportunity_id: int, access_ids: list[int]):
+def send_payment_notification(opportunity_id: int, payment_ids: list[int]):
     opportunity = Opportunity.objects.get(pk=opportunity_id)
     messages = []
-    for payment in Payment.objects.filter(opportunity_access__in=access_ids).select_related(
-        "opportunity_access__user"
-    ):
+    for payment in Payment.objects.filter(pk__in=payment_ids).select_related("opportunity_access__user"):
         messages.append(
             Message(
                 usernames=[payment.opportunity_access.user.username],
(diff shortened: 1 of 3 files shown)


commcare_connect/opportunity/tasks.py
```diff
@@ -158,3 +159,22 @@ def _get_deliver_message(access: OpportunityAccess):
             "To maximise your payout complete all the required service delivery."
         ),
     )
+
+
+@celery_app.task()
+def send_payment_notification(opportunity_id: int, access_ids: list[int]):
+    opportunity = Opportunity.objects.get(pk=opportunity_id)
+    messages = []
+    for payment in Payment.objects.filter(opportunity_access__in=access_ids).select_related(
```
calellowitz, line 168: This looks like it will send notifications for every payment the user has gotten in that opportunity, not just the new ones. I think you can fix this, by passing in the payment ids instead of the opportunity access ids. Then you will be able to get just the payments you want.

calellowitz (changes requested):

commcare_connect/opportunity/tasks.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/221/commits/712e701faecd76b5f7898a20d6cae094b4f7fc22

commcare_connect/opportunity/visit_import.py
```diff
@@ -191,7 +191,7 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
         raise ImportException(f"{len(invalid_rows)} have errors", invalid_rows)
 
     seen_users = set()
-    access_ids = set()
+    payment_ids = set()
```
calellowitz, line 194: nit: this doesn't have to be a set anymore since the payments are all newly created

calellowitz (approved):
nothing blocking, just a small note

commcare_connect/opportunity/visit_import.py
pxwxnvermx: Removed the set. https://github.com/dimagi/commcare-connect/pull/221/commits/87882d08db84332c75bc70f97c662802be3686e5

PR #226, opened 2023-12-21
move user invites to task

calellowitz:
This PR moves the invite messages to their own celery task. 

This has two benefits: 
1) Parallelization through celery, which should allow faster message sending
2) Exception sending messages won't cause the invite process to fail, which I believe was the cause of https://dimagi-dev.atlassian.net/browse/CCPC-209

### DIFF
--- a/commcare_connect/opportunity/tasks.py
+++ b/commcare_connect/opportunity/tasks.py
@@ -1,12 +1,14 @@
 import datetime
 
+from allauth.utils import build_absolute_uri
 from django.core.files.base import ContentFile
 from django.core.files.storage import default_storage
+from django.urls import reverse
 from django.utils.timezone import now
 from django.utils.translation import gettext
 from tablib import Dataset
 
-from commcare_connect.connect_id_client import fetch_users, send_message_bulk
+from commcare_connect.connect_id_client import fetch_users, send_message, send_message_bulk
 from commcare_connect.connect_id_client.models import Message
 from commcare_connect.opportunity.app_xml import get_connect_blocks_for_app, get_deliver_units_for_app
 from commcare_connect.opportunity.export import (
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/tasks.py
```diff
@@ -63,7 +65,31 @@ def add_connect_users(user_list: list[str], opportunity_id: str):
             username=user.username, defaults={"phone_number": user.phone_number, "name": user.name}
         )
         opportunity_access, _ = OpportunityAccess.objects.get_or_create(user=u, opportunity_id=opportunity_id)
-        invite_user(u, opportunity_access)
+        invite_user.delay(u, opportunity_access)
+
+
+@celery_app.task()
+def invite_user(user, opportunity_access):
```
pxwxnvermx, line 72: The user and opportunity_access params here needs to replaced with the pk's for user and access.

calellowitz: thanks. is that due to json serialization or some other reason?

pxwxnvermx: Yeah, the json serialization would error out for these django objects.

commcare_connect/opportunity/tasks.py
```diff
@@ -63,7 +65,33 @@ def add_connect_users(user_list: list[str], opportunity_id: str):
             username=user.username, defaults={"phone_number": user.phone_number, "name": user.name}
         )
         opportunity_access, _ = OpportunityAccess.objects.get_or_create(user=u, opportunity_id=opportunity_id)
-        invite_user(u, opportunity_access)
+        invite_user.delay(u.id, opportunity_access.id)
+
+
+@celery_app.task()
+def invite_user(user_id, opportunity_access_id):
```
pxwxnvermx, line 72: nit: user_id is not required since the user is already a part of the opportunity_access and the data for the user can be resolved from the `opportunity_access.user` property.

pxwxnvermx (approved):

PR #228, opened 2024-01-02
Refactor Export Code

pxwxnvermx:
There was a lot of duplication of code around the export code for payment export, user status export and deliver status export.

- This PR includes a export_modal template which can be used for exporting data in future.
- Refactored the code to use the save_export helper function

### DIFF
--- a/commcare_connect/templates/export_modal.html
+++ b/commcare_connect/templates/export_modal.html
@@ -13,7 +13,7 @@
               enctype="multipart/form-data">
           <div class="modal-body">
             <div class="mb-3">
-              {% crispy user_status_export_form %}
+              {% crispy export_form %}
             </div>
           </div>
           <div class="modal-footer">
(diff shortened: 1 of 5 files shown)


commcare_connect/templates/export_modal.html
```diff
@@ -0,0 +1,27 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <h1 class="modal-title fs-5">{{modal_title}}</h1>
+          <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+        </div>
+        <form action="{{export_url}}" method="post"
+              enctype="multipart/form-data">
+          <div class="modal-body">
+            <div class="mb-3">
+              {% crispy user_status_export_form %}
```
calellowitz, line 16: nit: we can probably rename this since it is used in a lot of exports now

calellowitz (approved):
This looks great, thanks for doing this

commcare_connect/templates/export_modal.html
pxwxnvermx: Thanks, i renamed this and refactored the code to use the same export form. It was already using the same form but the code was duplicated.

PR #231, opened 2024-01-04
fix tests to raise no warnings

calellowitz:
Goes from
Screenshot from 2024-01-04 10-46-06

to
Screenshot from 2024-01-04 14-24-32

This fixes the three sources of warnings in our test build, partially out of a general sense that we shouldn't let warnings pile up and because some of them were important deprecation warnings to fix, and partially because it was annoying to find test results in the long spew of warnings, especially when there were failures.

The three warnings I addressed were:
1) Using timestamps with timezones for models that have datetime fields. We were inconsistent about that in the past but this will do a good job both ensuring we catch issues and creating better consistency in our handling of datetimes. In the future we should continue to use `now` from `django.timezone.utils`, which we already used in some places.
2) Using `__` instead of `.` in django tables, which is the preferred way and the old one was going to be deprecated.
3) Updating usage of `post_generation` in our factories. Models will no longer be saved after `post_generation` steps in the next version so adding `skip_postgeneration_save = True` signals to factoryboy that we are already doing the right thing and will catch any future incompatibilities. As part of this, I got rid of the only real usage of `post_generation` we had, creating the user password, because there was already a builtin way to do so using `factory.django.Password`. That said, I wasn't able to completely verify that this had the same behavior as just setting password to a string (without any `make_password` calls to save it as a hash), still allowed the build to pass so it is possible I made a mistake and we just are not using this field. @snopoke was there a reason you did the password as an extra `post_generation` step rather than using the [built in field](https://factoryboy.readthedocs.io/en/stable/orms.html#factory.django.Password)?


pxwxnvermx (approved):
Thanks. 🙌🏽

PR #232, opened 2024-01-05
Add filtering for web users on add members

pxwxnvermx:
Added a filter to the add member dropdown on organization page to exclude all mobile users and show only the web users that are not already invited to the organization.

### DIFF
--- a/commcare_connect/organization/forms.py
+++ b/commcare_connect/organization/forms.py
@@ -1,6 +1,5 @@
 from crispy_forms import helper, layout
 from django import forms
-from django.db.models import Q
 from django.utils.translation import gettext
 
 from commcare_connect.organization.models import Organization, UserOrganizationMembership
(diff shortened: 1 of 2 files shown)


commcare_connect/organization/forms.py
```diff
@@ -30,8 +32,13 @@ class Meta:
         labels = {"user": "", "role": ""}
 
     def __init__(self, *args, **kwargs):
+        self.organization = kwargs.pop("organization")
         super().__init__(*args, **kwargs)
 
+        self.fields["user"].queryset = User.objects.exclude(
+            Q(email__isnull=True) | Q(memberships__organization=self.organization)
```
calellowitz, line 39: nit: I think this would be slightly clearer as `User.objects.filter(email__isnull=False).exclude(memberships__organization=self.organization)` (combining `NOT` and `OR` like that can be a bit less easy to reason about that a simple `AND`)

calellowitz (approved):
one nonblocking suggestion, but looks good

commcare_connect/organization/forms.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/232/commits/81d151fb86f62e7868d228e6661683b5ba0eb448

PR #233, opened 2024-01-05
Fix pagination warning on querysets without order_by

pxwxnvermx:
This PR fixes warnings for paginated views with unordered quersets.

`UnorderedObjectListWarning: Pagination may yield inconsistent results with an unordered object_list: <class 'commcare_connect.opportunity.models.Opportunity'> QuerySet`

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -2,7 +2,7 @@ from celery.result import AsyncResult
 from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
 from django.core.files.storage import storages
-from django.db.models import BooleanField, Case, F, Q, When
+from django.db.models import F
 from django.http import FileResponse, HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
@@ -73,18 +73,7 @@ class OpportunityList(OrganizationUserMixin, ListView):
     paginate_by = 10
 
     def get_queryset(self):
-        return (
-            Opportunity.objects.filter(organization=self.request.org)
-            .annotate(
-                is_active_val=Case(
-                    When(Q(end_date__gte=now()) & Q(active=True), then=True),
-                    default=False,
-                    output_field=BooleanField(),
-                )
-            )
-            .order_by("name")
-            .order_by("-is_active_val")
-        )
+        return Opportunity.objects.filter(organization=self.request.org).order_by("name")
 
 
 class OpportunityCreate(OrganizationUserMixin, CreateView):


commcare_connect/opportunity/views.py
```diff
@@ -172,7 +172,7 @@ def get_queryset(self):
         opportunity = get_object_or_404(Opportunity, organization=self.request.org, id=opportunity_id)
         access_id = self.kwargs["pk"]
         access = get_object_or_404(OpportunityAccess, opportunity=opportunity, pk=access_id)
-        return Payment.objects.filter(opportunity_access=access)
+        return Payment.objects.filter(opportunity_access=access).order_by("date_paid")
```
calellowitz, line 175: I think you want to do this in the opposite direction so the most recent payment is first

commcare_connect/opportunity/views.py
```diff
@@ -73,7 +73,7 @@ class OpportunityList(OrganizationUserMixin, ListView):
     paginate_by = 10
 
     def get_queryset(self):
-        return Opportunity.objects.filter(organization=self.request.org)
+        return Opportunity.objects.filter(organization=self.request.org).order_by("date_created")
```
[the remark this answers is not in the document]
pxwxnvermx: I replicated the logic for is_active method in Django annotate method to correct the sorting behavior. Now the sorting order is first sort by names and then show the active opportunities on top. https://github.com/dimagi/commcare-connect/pull/233/commits/8fb71ec3ef64b6ca040d0b5fa1e29b0d6cdd903b

commcare_connect/opportunity/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/233/commits/1080718825bd42758b3ffc0f05519e74cd96716c

commcare_connect/opportunity/views.py
```diff
@@ -73,7 +73,18 @@ class OpportunityList(OrganizationUserMixin, ListView):
...
+            Opportunity.objects.filter(organization=self.request.org)
+            .annotate(
+                is_active_val=Case(
+                    When(Q(end_date__gte=now()) & Q(active=True), then=True),
+                    default=False,
+                    output_field=BooleanField(),
+                )
+            )
+            .order_by("name")
+            .order_by("-is_active_val")
```
calellowitz, line 86: Is this sort name then active? If so, I think that won't do quite the right thing. I think we want the opposite order of terms. For example if there are three opportunities A, B, and C and C is inactive, we want the order to be A, C, B. I think this would order those A, B, C since activity would tiebreak name order rather than names tiebreaking activity. Let me know if I am misinterpreting how this code works. Otherwise looks good, although I forgot that active wasn't part of the model and would require an annotation, so if you would rather, I think it is fine to just order by name.

pxwxnvermx: the order currently is sort name then active. That ensures that opportunities that are active are displayed before the opportunities which are inactive and are sorted alphabetically, as you said in the earlier comment. I am happy to make this ordered by name only, as it would remove this annotation which adds computational overhead to the database.

calellowitz: I poked around in a shell and I don't think this works. It looks like using `order_by` twice like you are doing just overwrites the first with the second, so this is only sorting by `is_active_val`. You can verify this with `print(<my_orm_query>.query)`. I would be fine changing this to just name, otherwise you can sort on two values, by passing them both as arguments to ordery_by like `order_by("-is_active_val", "name")`

calellowitz (changes requested):

commcare_connect/opportunity/views.py
pxwxnvermx: Thanks that makes sense. I will make the change to sort this only by `name`.

pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/233/commits/da61df23463d1eaacf7675b9a7b3388cd40d0f1a

calellowitz (approved):

PR #235, opened 2024-01-05
do not prefetch on an instance

calellowitz:
One more dumb mistake


pxwxnvermx (approved):
A wise man once said "there are no mistakes only happy accidents".

PR #236, opened 2024-01-09
Rollback Payment

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-238)

image

image


commcare_connect/opportunity/views.py
```diff
@@ -453,3 +460,13 @@ def user_visits_list(request, org_slug=None, opp_id=None, pk=None):
         "opportunity/user_visits_list.html",
         context=dict(opportunity=opportunity, table=user_visits_table, user_name=opportunity_access.display_name),
     )
+
+
+@org_member_required
+@require_POST
+def payment_delete(request, org_slug=None, opp_id=None, access_id=None, pk=None):
+    opportunity = get_object_or_404(Opportunity, organization=request.org, pk=opp_id)
+    opportunity_access = get_object_or_404(OpportunityAccess, pk=access_id, opportunity=opportunity)
```
calellowitz, line 469: Why get the opportunity access here?

pxwxnvermx: It is to confirm that the opportunity_access exists for the user and to further fetch payment using that opportunity access.

calellowitz: I am not totally sure I understand. Why do we need to check the access? We have the PK of the payment we want to delete? Not blocking since I don't think it hurts. But it seems unnecessary unless we are gaining some value from it.

calellowitz (approved):

PR #238, opened 2024-01-11
duplicates

calellowitz:
This PR adds automatic marking of duplicate deliveries (those with the same entity id and deliver unit id) with a new form status. It also updates the status reports to include counts of those as well.

In addition, it modifies the payment accrual calculation to count duplicates rather than ignore them if those duplicates have both been approved.


pxwxnvermx (commented):
Thanks. The logic for payment accrued is great.

pxwxnvermx (approved):

PR #240, opened 2024-01-17
Send Message UI

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCPC-246)

- Added a page at the opportunity level to send messages (push notification or SMS) to users.
- Messages can be sent to the users who have accepted the invite.
- Only organization admins can send messages to users.
- The title field is optional for SMS and only needed for Push Notifications.
- Added a select all users checkbox.

image


calellowitz:
Could you do something like the checkboxes in the add budget page for user selection? I think that table looks a little cleaner. It might also be worth adding a note that title is only relevant for push notifications (and making it optional with a default value like "Notification from CommCare Connect" or something, since it won't show up in SMS anyway).

commcare_connect/opportunity/forms.py
```diff
@@ -303,3 +303,36 @@ def __init__(self, *args, **kwargs):
...
+            Row(Field("title")),
+            Row(Field("body")),
+            Row(Field("message_type")),
+            Submit(name="submit", value="Submit"),
+        )
+
+        choices = [(user.pk, f"{user.name} - {user.username}") for user in users]
+        self.fields["selected_users"] = forms.MultipleChoiceField(
+            choices=choices,
+            help_text="Use Ctrl-click to select multiple users",
```
calellowitz, line 336: I don't think this is necessary anymore

commcare_connect/opportunity/forms.py
```diff
@@ -303,3 +303,36 @@ def __init__(self, *args, **kwargs):
             self.fields["deliver_units"].initial = [
                 deliver_unit.pk for deliver_unit in self.instance.deliver_units.all()
             ]
+
+
+class SendMessageMobileUsersForm(forms.Form):
+    title = forms.CharField(
+        empty_value="Notification from CommCare Connect",
+        required=False,
+        widget=forms.TextInput(attrs={"placeholder": "Notification from CommCare Connect"}),
```
calellowitz, line 312: you can remove the placeholder since it gets set by default as "empty value"

commcare_connect/opportunity/views.py
```diff
@@ -453,3 +457,38 @@ def user_visits_list(request, org_slug=None, opp_id=None, pk=None):
...
         context=dict(opportunity=opportunity, table=user_visits_table, user_name=opportunity_access.display_name),
     )
+
+
+@org_admin_required
+def send_message_mobile_users(request, org_slug=None, pk=None):
+    opportunity = get_object_or_404(Opportunity, pk=pk, organization=request.org)
+    user_ids = (
+        OpportunityAccess.objects.filter(opportunity=opportunity, accepted=True)
+        .prefetch_related("user")
```
calellowitz, line 467: nit: i dont think you need to prefetch here since you only want the id. I would assume if you put "user_id" in the `values_list`, it will do the right thing without a join at all

calellowitz (changes requested):
Couple small things, not worried too much about the UI since we are about to do a bunch of work there, but the backend changes are probably worth making

commcare_connect/opportunity/forms.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/240/commits/6a97662b8cb04970ec2855fc71861bb6d491d60f

commcare_connect/opportunity/forms.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/240/commits/6a97662b8cb04970ec2855fc71861bb6d491d60f

calellowitz (approved):

PR #242, opened 2024-01-24
Migrate models to use OpportunityAccess as forieng key

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCCT-35)

This is Part 1 of the Migration which creates the field and populates it. All the code related to refactoring the usage of Opportunity Access will be in a second PR to keep the diff small enough to review.


commcare_connect/opportunity/models.py
```diff
@@ -209,6 +183,34 @@ def display_name(self):
...
+
+class Assessment(XFormBaseModel):
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        related_name="assessments",
+    )
+    app = models.ForeignKey(CommCareApp, on_delete=models.PROTECT)
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.PROTECT)
+    opportunity_access = models.ForeignKey(OpportunityAccess, on_delete=models.CASCADE, null=True)
```
pxwxnvermx, line 207: Only this line was added and the model was moved to be below OpportunityAccess to avoid any undefined warnings.

commcare_connect/opportunity/models.py
```diff
@@ -209,6 +183,34 @@ def display_name(self):
...
 
+class CompletedModule(XFormBaseModel):
+    user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        related_name="completed_modules",
+    )
+    module = models.ForeignKey(LearnModule, on_delete=models.PROTECT)
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.PROTECT)
+    opportunity_access = models.ForeignKey(OpportunityAccess, on_delete=models.CASCADE, null=True)
```
pxwxnvermx, line 194: Only this line was added and the model was moved to be below OpportunityAccess to avoid any undefined warnings.

calellowitz (approved):

sravfeyn (approved):

calellowitz (approved):

PR #243, opened 2024-01-25
Fix payment accrual bug

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-222)


calellowitz (approved):

calellowitz:
Can you add a test that includes multiple users

pxwxnvermx:
I have added a test for multiple users.

PR #245, opened 2024-01-29
Refactor to add opportunity access on models

pxwxnvermx:
[Ticket](https://dimagi-dev.atlassian.net/browse/CCCT-35)

This is Part 2 of the Migration which updates the code to use the opportunity access field on UserVisit, CompletedModules and Assessment models.


sravfeyn (approved):

commcare_connect/opportunity/models.py
```diff
@@ -152,7 +152,9 @@ def claimed_visits(self):
 
     @property
     def approved_visits(self):
-        return CompletedWork.objects.filter(opportunity_access__opportunity=self).count()
+        return CompletedWork.objects.filter(
+            opportunity_access__opportunity=self, status=CompletedWorkStatus.approved
```
calellowitz, line 156: is this change intentional?

commcare_connect/opportunity/models.py
```diff
@@ -243,9 +245,7 @@ def learn_progress(self):
         learn_modules_count = learn_modules.count()
         if learn_modules_count <= 0:
             return 0
-        completed_modules = CompletedModule.objects.filter(
-            opportunity=self.opportunity, module__in=learn_modules, user=self.user
-        ).count()
+        completed_modules = self.completedmodule_set.count()
```
calellowitz, line 248: same with this? is this equivalent behavior?

calellowitz (commented):
Few questions but otherwise looks good

commcare_connect/opportunity/models.py
pxwxnvermx: Yeah, this was found in the QA for this ticket where the Approved VIsits counts showed all visits.

commcare_connect/opportunity/models.py
pxwxnvermx: Yup, the filter for `module_in=learn_modules` is extra as the learn_modules could not be removed once the Opportunity is created so directly fetching completed_modules related to the opportunity_access for a user also gives the same output.

calellowitz (approved):

PR #252, opened 2024-01-31
Web UI Changes

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-233)

- Remove auto-generated usernames from Learn, Deliver, User, and Payment Status tables.
- Replaced date times with a popup that shows detailed time information on hovering.
- Clicking on the username popup icon copies the username to the system clipboard.
- Added duration in minutes and seconds on learn progress

image

image


commcare_connect/static/js/vendors.js
```diff
@@ -1,7 +1,23 @@
 import '@popperjs/core';
-import 'bootstrap';
+import * as bootstrap from 'bootstrap';
 import 'htmx.org';
 import Alpine from 'alpinejs';
 
 window.Alpine = Alpine;
 Alpine.start();
+
+function refreshTooltips() {
```
calellowitz, line 9: what does this function do?

calellowitz (changes requested):
One question about the javascript, but the only big thing is that we wanted to get rid of the username column, rather than replace it with a tooltip.

commcare_connect/static/js/vendors.js
pxwxnvermx: This function is used for loading the tooltips on page load or table load in this case.

calellowitz: why is it in `vendors/js`? this file seems to be around importing third party libraries

calellowitz (approved):
I think some of those functions are potentially in the wrong place, but nothing blocking

commcare_connect/static/js/vendors.js
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/252/commits/b5f33d9d99fea130444bce51fe5e71d09340100b

PR #261, opened 2024-02-05
Mapbox library config and load webpack config tokens from .env

pxwxnvermx:
- Added config for Mapbox
- Loading tokens for webpack bundles from `.env` files


calellowitz (approved):
Sorry, i left feedback for this on the other PR, but merging now

PR #248, opened 2024-01-30
Mobile User Profile Page

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-224)

- The Profile page is linked to the user's name on the Learn, Deliver, User, and Payment status tables on the opportunity details page.

image

PR #262, opened 2024-02-05
Mobile User Profile Visit Map

pxwxnvermx:
- Added map to show user visit locations with markers and popup with visit details

This PR contains code from #261 and is based on #248.

image


calellowitz (approved):
Approved so I can continue to build off the map work, but we may want to revisit soon.

PR #263, opened 2024-02-06
geopy

calellowitz:
I am not super happy with this code yet, but this is the basic approach I took. In the map view, it is green for the form being considered, blue for nearby forms from the same user, and red (none in this shot) for forms from other users.

Planning to adjust the padding on the buttons and add the tooltips tomorrow.

image


commcare_connect/opportunity/views.py
```diff
@@ -497,10 +498,38 @@ def visit_verification(request, org_slug=None, pk=None):
...
+        )
+        lat, lon, *_ = user_visit.location.split(" ")
+        for loc in locations:
+            other_lat, other_lon, *_ = loc["location"].split(" ")
+            dist = distance.distance((lat, lon), (other_lat, other_lon))
+            if user_visit.user_id == loc["user_id"]:
+                user_forms.append((loc, dist.m, other_lat, other_lon))
+            else:
+                other_forms.append((loc, dist.m, other_lat, other_lon))
+            user_forms.sort(key=lambda x: x[1])
```
pxwxnvermx, line 519: is this sorting for forms necessary for the UI?

pxwxnvermx (approved):
Just a query. Rest looks good.

commcare_connect/opportunity/views.py
calellowitz: yeah, we only want the 5 closest forms (you can see when we pass the list to the page that we only take the first five), so we sort them by distance here

ctsims: @calellowitz  - Note that we want the 5 closest forms within some threshold. I'm going to arbitrarily say that the threshold is 250 meters based on heuristically spot checking the data around. 

@pxwxnvermx one other thing I'm going to ask for at some point (possibly while Cal is offline) is the ability to change the actual marker images to be different (not just color) for the two types of geopoint, and to add a Legend to either the map view or next to the map view. If either of those is relevant for Cal's code here, good to plan for ahead.

calellowitz:
Made a few updates to the UI (popups, moved the buttons, fixed the duration). Going to leave it like this for now, but should be good to merge as is, or can incorporate more features

image

commcare_connect/opportunity/views.py
calellowitz: > Note that we want the 5 closest forms within some threshold. I'm going to arbitrarily say that the threshold is 250 meters based on heuristically spot checking the data around.

Updated to cap at 250 meters.

The other two will require a bit more work, but should be achievable. I can also pick up one or both on monday when I get back.

PR #267, opened 2024-02-09
Flag Visits with Close gps locations

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-253)

- Added flagging for visits with close GPS location

This PR uses the geopy library introduced in #263.

image


commcare_connect/form_receiver/processor.py
```diff
@@ -176,6 +178,16 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
     if xform.metadata.location is None:
         flags.append(["gps", "GPS data is missing"])
+    else:
+        cur_lat, cur_lon, *_ = xform.metadata.location.split(" ")
+        for visit in user_visits:
+            if visit.get("location") is None:
+                continue
+            lat, lon, *_ = visit["location"].split(" ")
+            dist = distance((lat, lon), (cur_lat, cur_lon))
+            if dist <= 10:
```
calellowitz, line 188: I think this does km by default rather than meters, which I imagine we want. You just need `dist.m` to get the right one.

calellowitz (changes requested):

commcare_connect/form_receiver/processor.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/267/commits/9bdc446ab6dbcfbf3cc092d676d5463c17798b21

calellowitz (approved):

PR #269, opened 2024-02-12
Fix payment accrued on visit approve from UI

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/QA-6124)

Fixed Payment not getting accrued on visit approval in the verification UI.


calellowitz (approved):
Oops, that was supposed to be an approval since this is fine on its own.

pxwxnvermx:
> This looks great, but I think it needs to be on the reject view as well just to be safe (in case a form moves from approved to rejected).

Added. https://github.com/dimagi/commcare-connect/pull/269/commits/c7c9f6bd2bba6104376f4e144368bd4f28c954ed

PR #270, opened 2024-02-13
Download form attachments on receiving

sravfeyn:
This downloads the form attachments from HQ whenever a form is received (skips if the media is already download). I have tested this end-to-end locally (apart from adding a unit test)


calellowitz:
I built on this PR to include the images in the visit verification page. Screenshot below

image

commcare_connect/opportunity/tasks.py
```diff
@@ -227,3 +231,31 @@ def send_sms_task(user_ids: list[int], body: str):
...
+    api_key = user_visit.opportunity.api_key
+    blobs = user_visit.form_json.get("attachments", {})
+    domain = user_visit.opportunity.deliver_app.cc_domain
+    form_id = user_visit.xform_id
+    for name, blob in blobs.items():
+        if name == "form.xml":
+            continue
+        url = f"{settings.COMMCARE_HQ_URL}/a/{domain}/api/form/attachment/{user_visit.xform_id}/{name}"
+
+        blob_meta = BlobMeta(
```
pxwxnvermx, line 248: nit: This can be replaced with the `BlobMeta.objects.get_or_create` method which internally executes the same code and returns the `object` and a `created` flag which can be used to continue the loop when the created flag is false.

pxwxnvermx (approved):
Just a nit. The rest looks good. Thanks 🙌🏽

commcare_connect/opportunity/tasks.py
```diff
@@ -227,3 +231,31 @@ def send_sms_task(user_ids: list[int], body: str):
...
+            try:
+                blob_meta.save()
+            except IntegrityError:
+                # attachment already exists
+                continue
+            response = httpx.get(
+                url,
+                headers={"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"},
+            )
+            default_storage.save(blob_meta.blob_id, ContentFile(response.content, name))
```
pxwxnvermx, line 261: There was an error on staging on this line due to the `blob_meta.blob_id` being a UUID object. That needs to be converted to a string representation before the file can be saved.

pxwxnvermx (changes requested):

commcare_connect/opportunity/tasks.py
sravfeyn: This probably is due to the difference being the local storage vs S3, though I wouldn't expect the implementation to differ. I have casted it to str now.

pxwxnvermx (approved):

PR #276, opened 2024-02-15
Ce/map improvements

calellowitz:
This addresses most of the feedback from https://app.asana.com/0/1206444985338835/1206604143294414/f

1) adds a bunch of new info to the popup including a link to the form verification page of that form
2) shrinks markers for the additional forms to further distinguish it from main form
3) add precision circles on mouse hover
4) adds toggle between satellite and streets
5) increases zoom

All features mentioned except zoom are shown in below screenshot (but scale is different since my local data is further spaced).
image

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -588,20 +588,20 @@ def visit_verification(request, org_slug=None, pk=None):
     if user_visit.location:
         locations = (
             UserVisit.objects.filter(opportunity=user_visit.opportunity)
-            .values("entity_id", "location", "user_id", "entity_name")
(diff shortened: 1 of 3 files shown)


pxwxnvermx (changes requested):

ctsims:
Hey Cal - functionality here looks good to me, assuming that the empty Name field above is just a test data peculiarity.

One tiny point of feedback is that I would love if the mouseover-accuracy circles popped up for all points on the screen and not just the one the mouseover hits (in the demo it would be great to see that two points have a concrete overlap), but that's very much "only worry about that if it's trivial to change" feedback for now.

ctsims (approved):

calellowitz:
Fixed feedback from Pawan. Merging so clayton can test.

PR #279, opened 2024-02-16
update request org usage

calellowitz:
This should fix the issues we have seen with the superuser access features, and a potential security issue.

1) Org was only being set on the request if the user was a member. Superusers are not, and the lack of `request.org` was causing an assortment of 500 errors creating a response. Now superusers actually should have access to all pages without an invite.

2) The org membership check in `OrganizationUserMixin` was incorrectly returning `True` always because `request.org_membership` was never `None`, but rather a `SimpleLazyObject` that evaluated to `None`.


pxwxnvermx (approved):
Thanks 🙌🏽

PR #291, opened 2024-03-14
Payment Verification

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-285)

- Added Export and Import for Verifying and Approving the completed work.
- Updated payment accrual logic to accrue payment after completed work approval.


calellowitz (approved):
Looks fine to me

commcare_connect/opportunity/visit_import.py
```diff
@@ -287,7 +287,7 @@ def get_status_by_completed_work_id(dataset):
     invalid_rows = []
     for row in dataset:
         row = list(row)
-        work_id = row[work_id_col_index]
+        work_id = str(row[work_id_col_index])
```
calellowitz, line 290: What is this change for?

pxwxnvermx: This is the fix for the broken payment verification import due to the work_id being a string when the import file was CSV and an integer when the import file was XLSX. 😢

PR #292, opened 2024-03-20
Deliver Progress API Update

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-278)

- Update deliver progress API to use the CompletedWork model


commcare_connect/opportunity/api/serializers.py
```diff
@@ -151,6 +151,27 @@ class Meta:
         ]
 
 
+# NOTE: this serializer is only required to avoid introducing breaking changes
+# to the deliver progress API
+class CompletedWorkSerializer(serializers.ModelSerializer):
+    deliver_unit_name = serializers.CharField(source="payment_unit.name")
+    deliver_unit_slug = serializers.CharField(default="dummy_payment_unit")
```
calellowitz, line 158: what is the default used for here?

pxwxnvermx: The default deliver_unit_slug here ensures that the API return does not change.

calellowitz: does that mean it will be the same for every work unit?

pxwxnvermx: Yeah, the slug would be same for every work unit.

calellowitz: just to make me feel better, can you make this the pk so it will be different on the off chance it matters

calellowitz (approved):
approved if you make that one change

commcare_connect/opportunity/api/serializers.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/292/commits/c46c4f253f02901631a891ffb89adc367a6daed0

calellowitz (approved):

PR #294, opened 2024-03-27
Auto Verification

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-290)

- Added auto-approve visits toggle to Opportunity Edit.
- Added logic to auto-approve unflagged visits and auto-reject flagged visits.
- Added logic to auto approve payments for completed and unflagged CompletedWork.

image


calellowitz:
> auto-reject flagged visits

I think this is incorrect. We want to leave flagged visits as pending for the user to reject or approve.

commcare_connect/opportunity/visit_import.py
```diff
@@ -114,12 +119,17 @@ def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
 
             UserVisit.objects.bulk_update(to_update, fields=["status", "reason"])
             missing_visits |= set(visit_batch) - seen_visits
+    if opportunity.auto_approve_payments:
+        approve_completed_work_and_update_payment_accrued.delay(list(seen_completed_works))
```
calellowitz, line 123: Is the accrual logic the same here? If so you can call the function from in the task. Or just call it outside the if/else block since it happens in either case.

Relatedly, why does this need to be a separate task? The upload is already happening in a celery task.

commcare_connect/opportunity/views.py
```diff
@@ -138,6 +138,10 @@ def form_valid(self, form):
         if end_date:
             opportunity.end_date = end_date
         response = super().form_valid(form)
+        if opportunity.auto_approve_visits:
```
calellowitz, line 141: I dont think you want to backfill this automatically. I think its fine to only have it affect forms going forward, but otherwise, it could be an option. I think people might accidentally do something they don't intend to if it always happens like this.

commcare_connect/opportunity/tasks.py
```diff
@@ -267,3 +269,21 @@ def generate_work_status_export(opportunity_id: int, export_format: str):
...
+def approve_completed_work_and_update_payment_accrued(completed_work_ids: list[int]):
+    """Approves and Updates Payment accrued for all completed (all deliver units submitted and any status) and
+    un-flagged completed work."""
+    for completed_work_id in completed_work_ids:
+        completed_work = CompletedWork.objects.get(pk=completed_work_id)
+        access = completed_work.opportunity_access
+        if completed_work.flags:
+            continue
+        payment_accrued = completed_work.payment_accrued
+        if payment_accrued > 0:
```
calellowitz, line 284: Even if there is no payment we want to still approve the completed work, right? In case it is necessary for other works, or just because they should show as approved, regardless of value?

On that note, does ordering matter here? So that payment units that depend on each other are approved in the right order?

commcare_connect/form_receiver/processor.py
```diff
@@ -206,7 +209,12 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     if flags:
         user_visit.flagged = True
         user_visit.flag_reason = {"flags": flags}
+
+    if opportunity.auto_approve_visits and user_visit.status == VisitValidationStatus.pending:
+        user_visit.status = VisitValidationStatus.rejected if user_visit.flagged else VisitValidationStatus.approved
```
calellowitz, line 214: I made this comment in the overall PR, but I think we want to auto approve only. Flags stay as pending, thats how we alert users they need to view it.

commcare_connect/form_receiver/processor.py
```diff
@@ -206,7 +209,12 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
     user_visit.save()
+    if opportunity.auto_approve_payments:
+        approve_completed_work_and_update_payment_accrued.delay([completed_work.id])
```
calellowitz, line 217: I would be a bit concerned about sequencing here. It may not matter but forms can be submitted many at once (since they will submit in bulk once the user gets data). We may want locking on this task, or to have it run on a schedule in bulk instead of immediately (which may make more sense anyway to do it as a batch)

calellowitz (changes requested):
Left a fair number of comments, let me know if you have any questions

commcare_connect/opportunity/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/294/commits/b1698ae22d0906989979e59cafd6dc51ab214630

commcare_connect/form_receiver/processor.py
pxwxnvermx: Fix. https://github.com/dimagi/commcare-connect/pull/294/commits/28a898593ab863404c5f4a79fce74f0c6d443491

commcare_connect/opportunity/tasks.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/294/commits/0d34603ed9262705f8f606e05302fbc577776f21

commcare_connect/opportunity/visit_import.py
pxwxnvermx: > Is the accrual logic the same here? If so you can call the function from in the task. Or just call it outside the if/else block since it happens in either case.
The payment accrual logic is same, the task only adds the logic to approve a completed work if it has no flags. I will simplify this code to run the update payment accrued function only once and do the completed work approval in the `if` block.

> Relatedly, why does this need to be a separate task? The upload is already happening in a celery task.
I checked the import does not happen in a task only the export for files is done through a celery task.

commcare_connect/form_receiver/processor.py
pxwxnvermx: The bulk processing for the payment approval makes sense. I will change the code to run this task periodically every day using celery schedules.

commcare_connect/opportunity/migrations/0041_bulk_approve_completed_work.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 4.2.5 on 2024-04-22 06:10
+
+from django.db import migrations
+from django_celery_beat.models import CrontabSchedule, PeriodicTask
+
+
+def create_bulk_apporve_completed_work_task(apps, schema_editor):
+    schedule, _ = CrontabSchedule.objects.get_or_create(
+        minute="00",
+        hour="12",
```
calellowitz, line 10: This will be right in the middle of the workday (12 UTC), It would be safer to do this overnight, maybe midnight UTC or a little earlier like 11pm)

commcare_connect/opportunity/tasks.py
```diff
@@ -286,3 +286,22 @@ def approve_completed_work_and_update_payment_accrued(completed_work_ids: list[i
...
+
+@celery_app.task()
+def bulk_approve_completed_work():
+    access_objects = OpportunityAccess.objects.filter(
+        opportunity__active=True,
+        opportunity__end_date__gte=datetime.date.today(),
+    )
+    for access in access_objects:
+        completed_works = access.completedwork_set.filter(status=CompletedWorkStatus.pending)
+        for completed_work in completed_works:
```
calellowitz, line 299: shouldn't this check if the setting to auto approve is enabled for that opportunity?

calellowitz (commented):
couple questions but looking better

commcare_connect/opportunity/tasks.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/294/commits/7635284b54b8c16e0bd12415bdc2ce54997fd4d9

commcare_connect/opportunity/migrations/0041_bulk_approve_completed_work.py
pxwxnvermx: https://github.com/dimagi/commcare-connect/pull/294/commits/3c2e32a1d3ec70733ff492b69c687f5c3daf6fe6

calellowitz (approved):
looks good once the conflict is fixed

sravfeyn (approved):

PR #296, opened 2024-04-02
Fix broken Completed Work migration

pxwxnvermx:
The completed work migration was broken due to out-of-sync model fields in the `populate_completed_work` command which uses models by directly importing them and caused the migration to fail with an error indicating that some fields were not yet created. I fixed the migration by using `apps.get_model()` method to get the correct model for the migration.


calellowitz (approved):
If this is just copied from the management command can we delete the command?

PR #298, opened 2024-04-03
Opportunity Creation Wizard and Start date

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-276 and 
https://dimagi.atlassian.net/browse/CCCT-14


commcare_connect/opportunity/models.py
```diff
@@ -411,6 +413,7 @@ class UserVisit(XFormBaseModel):
     location = models.CharField(null=True)
     flagged = models.BooleanField(default=False)
     flag_reason = models.JSONField(null=True, blank=True)
+    is_trial = models.BooleanField(default=False)
```
sravfeyn, line 416: I have added this instead of using `flag_reason`, because of two reasons.

1. The way the 'trial' flag is getting used is different to the way the existing `flag_reason`, their purpose seems different. The flags are used to show during a visit-verification, while the `trial` flag needs to be used to exclude the visits that were submitted before start-date
2. It's difficult to query `uservisits` with the way the `flag_reason` is setup now, it can't be queried easily (which needs to happen [here](https://github.com/dimagi/commcare-connect/pull/298/files#diff-0d916b95ee38daa9788eca7b8b3266f406c5c791cab6ff6bda1d5aeefc84d5d2R150) and some other places). We should migrate that to dictionary like below which should make it easier to query 

```
flags = [
    {"name": "duplicate", "description": "A beneficiary with the same identifier already exists"},
    {"name": "duration", "description": "The form was completed too quickly."},
    {"name": "gps", "description": "GPS data is missing"}
]
```

sravfeyn:
@calellowitz @pxwxnvermx This is now a PR into the main branch. I have fixed all the test failures as well.

commcare_connect/opportunity/api/views.py
```diff
@@ -80,23 +82,23 @@ def post(self, *args, **kwargs):
 
         if OpportunityClaim.objects.filter(opportunity_access=opportunity_access).exists():
             return Response(status=200, data="Opportunity is already claimed")
-        if opportunity.remaining_budget < opportunity.budget_per_visit:
+        if opportunity.remaining_budget < opportunity.minimum_budget_per_visit:
```
calellowitz, line 85: I think this needs to be max budget per visit (we dont want you claiming a job and not being able to get paid if you do the full delivery)

commcare_connect/opportunity/models.py
```diff
@@ -75,46 +78,95 @@ class Opportunity(BaseModel):
...
+        return CompletedWork.objects.filter(opportunity_access__opportunity=self).count()
+
+    @property
+    def number_of_users(self):
+        return self.total_budget / self.budget_per_user
 
     @property
     def allotted_visits(self):
-        return math.floor(self.total_budget / self.budget_per_visit)
+        payment_units = self.top_level_paymentunits.all()
```
calellowitz, line 156: why do all of these use `top_level_paymentunits`

commcare_connect/opportunity/models.py
```diff
@@ -75,46 +78,95 @@ class Opportunity(BaseModel):
     def __str__(self):
         return self.name
 
+    @property
+    def is_setup_complete(self):
+        return self.paymentunit_set.count() > 0 and self.total_budget
+
+    @property
+    def top_level_paymentunits(self):
+        # payment units that are prereqs of other paymentunits are ignored
```
calellowitz, line 87: why is this true? both payment units will contribute to the budget right?

commcare_connect/opportunity/models.py
calellowitz: That all makes sense over `flag_reason` but I am curious if you considered making this a status option like `OVER_LIMIT`. Both in how it is used and logically, it seems to fit nicely there. Plus it doesn't make a lot of sense to have another status attached to these (a pending trial form, or an over limit trial form). I think that might simplify the receiver logic as well

commcare_connect/opportunity/models.py
```diff
@@ -370,11 +425,41 @@ def images(self):
...
-    max_payments = models.IntegerField()
+    # to be removed
+    max_payments = models.IntegerField(null=True)
     end_date = models.DateField()
     date_claimed = models.DateField(auto_now_add=True)
 
 
+class OpportunityClaimLimit(models.Model):
+    opportunity_claim = models.ForeignKey(OpportunityClaim, on_delete=models.CASCADE)
+    payment_unit = models.ForeignKey(PaymentUnit, on_delete=models.CASCADE)
```
calellowitz, line 436: Is strictly on `OpportunityClaimLimit` per user per payment unit? If so, can that be enforced at the db level (a unique constraint on the claim/unit pair.)

commcare_connect/opportunity/models.py
```diff
@@ -370,11 +425,41 @@ def images(self):
...
+    max_visits = models.IntegerField()
+
+    @classmethod
+    def create_claim_limits(cls, opportunity: Opportunity, claim: OpportunityClaim):
+        claim_limits_by_payment_unit = defaultdict(list)
+        claim_limits = OpportunityClaimLimit.objects.filter(
+            opportunity_claim__opportunity_access__opportunity=opportunity
+        )
+        for claim_limit in claim_limits:
+            claim_limits_by_payment_unit[claim_limit.payment_unit].append(claim_limit)
```
calellowitz, line 446: related to the above question but when does this need to be a list?

commcare_connect/opportunity/models.py
```diff
@@ -370,11 +425,41 @@ def images(self):
...
     date_claimed = models.DateField(auto_now_add=True)
 
 
+class OpportunityClaimLimit(models.Model):
+    opportunity_claim = models.ForeignKey(OpportunityClaim, on_delete=models.CASCADE)
+    payment_unit = models.ForeignKey(PaymentUnit, on_delete=models.CASCADE)
+    max_visits = models.IntegerField()
+
+    @classmethod
+    def create_claim_limits(cls, opportunity: Opportunity, claim: OpportunityClaim):
```
calellowitz, line 440: Is there a high level description of what this method does?

calellowitz (commented):
Do you have screenshots of the new workflows? That might help in reviewing the front end code. Left a few questions while I try to figure this out, will do a further review once those are answered.

commcare_connect/opportunity/models.py
sravfeyn: This came after a discussion I had with Rama. The gist of it was that the payment-units which are prerequisites (the ones other than top-level paymentunits) would not be considered for budget calculations and any payment-amount you set for that is ignored. 

```
[10:28 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997106085829)
Sravan Reddy
 Per https://dimagi.atlassian.net/browse/CCCT-276, we are not having per user limits any more, is it?

[CCCT-276 D2ORG - Opportunity Setup Changes](https://dimagi.atlassian.net/browse/CCCT-276?atlOrigin=eyJpIjoiMDhmZmFkZjQzNDYwNDBiNmI0MTRhYjkzMzUyZTk1NmQiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)
Status: In Progress
Type: Story
Assignee: Sravan Reddy
Priority: P6
More actions...
Added by [Jira Cloud](https://dimagi.slack.com/services/B027GSXHAKD)
[10:28 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997119623709)
From 2.e and 2.f
Allow me to set a limit on the number of deliveries that are possible in total for that unit of work
Allow me to set a daily limit on the number of deliveries that are possible on any day before the Opportunity end date
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997142149099)
Rama Siva
 Let me check
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997143298559)
Sravan Reddy
 Or is that per user?
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997149409929)
Okay
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997152394819)
Rama Siva
 This has always been per user
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997176018529)
So no real change here other than making it per 'payment unit'
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997188441789)
Sravan Reddy
 I see, makes sense
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997193260479)
Rama Siva
 cool
[10:29 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997199850309)
Sravan Reddy
 Thanks for clarification.
[10:30 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997213800019)
Rama Siva
 :+1:
[10:33 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997437712669)
Sravan Reddy
 How do we change this section since it’s per Payment Unit now?
Screenshot 2024-03-21 at 10.33.37 AM.png
 
Screenshot 2024-03-21 at 10.33.37 AM.png
[10:34 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997461426479)
Meaning total visits now would be - X number of PU1 and Y number of PU2 (PU= payment unit)
[10:35 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997517347399)
Sravan Reddy
 Shall we just say X+Y?

1 reply
15 days agoView thread
[10:35 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997533763029)
Rama Siva
 great point. a 'visit' in this sense would map to a payment unit. so you could sum up the daily limit for all payment units
[10:35 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997550535169)
Sravan Reddy
 Cool
[10:36 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997576087669)
Rama Siva
 We do need to clean up our language in subsequent builds
[10:37 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997646311979)
Sravan Reddy
 Yeah
[10:37 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997678639769)
Rama Siva
 But in the case of sub-units of work this will not work.
[10:38 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997701665529)
So it will need to be the top level limit
[10:38 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710997723119449)
because we are really counting the number of beneficiaries in a day
[10:43 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998003210099)
Sravan Reddy
 Right, yeah
[10:43 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998009946669)
We will just do the top level limits
[10:44 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998064542939)
Rama Siva
 So the user can do 2 verified service deliveries and that is the limit shown here even if the sub-unit of work (verified screening) limit is set to 3
[10:45 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998116097949)
In reality, it is going to be the same and we should debate later if allowing a separate limit for a sub-unit of work is needed
[10:45 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998121338339)
Sravan Reddy
 So, we don’t take  the sub-units of work into account while calculating limits?
[10:45 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998129467199)
Yeah
[10:45 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998139458489)
Rama Siva
 Yeah, that is what I meant
[10:45 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998146013939)
Sravan Reddy
 Right now, the UI does allow specifying limits for sub unit of work as well


2 replies
Last reply 15 days agoView thread
[10:46 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998160109629)
Rama Siva
 they should just know how many verified service deliveries they should be doing
[10:46 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998209062339)
If the UI does not allow setting a separate limit then we should not do any addition
[10:47 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998235962019)
But if it is not a sub-unit, we do allow a separate limit?
[10:47 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998238945499)
on the UI?
[10:47 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998253583709?thread_ts=1710998146.013939&cid=D057T7G55DJ)
Sravan Reddy
 
replied to a thread:
Right now, the UI does allow specifying limits for sub unit of work as well
It does, but we can ignore them for calculating limits and budget, and let project-user know  that limits won’t be applicable if a payment-unit is made subunit of another payment unit
[10:47 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998268346029)
All, the paymentunits have a limit
[10:48 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998280196519)
Some payment-units can have sub-payment-units
[10:48 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998301825139)
Rama Siva
 great. i think we are aligned
[10:48 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998317183439)
Sravan Reddy
 Cool
[10:48 AM](https://dimagi.slack.com/archives/D057T7G55DJ/p1710998334325489)
Rama Siva
 great catch
```

commcare_connect/opportunity/models.py
sravfeyn: Refer to https://github.com/dimagi/commcare-connect/pull/298#discussion_r1553284344

commcare_connect/opportunity/models.py
sravfeyn: This is because we consider all users' claimlimits for an opportunity, so claimlimit for one paymentunit but for multiple users.

commcare_connect/opportunity/models.py
sravfeyn: This is basically the same as previous `claim.max_payments` ( I have renamed max_payments to max_visits since visit is what max_payments was also referring to), but because now the max_payments is per `paymentunit` and not just per user/opportunity, we need a per-paymentunit claiming. So, this method does exactly that. It creates per-paymentunit max visit limits

commcare_connect/opportunity/models.py
sravfeyn: Yeah. It is per paymentunit/claim. I guess we can, though we were not doing the same sort of db constraint for `OpportunityClaim` previously. We are (and were) using a `get_or_create` based on claim/paymentunit.

commcare_connect/opportunity/api/views.py
sravfeyn: The intention was to let user claim whatever the partial lowest budget visit if that was less than the remaining budget. But yeah, programmatically if want to allow only full delivery claiming, then I guess we should check that the remaining budget is more than the sum of budgets per all paymentunits (not just the Max).

commcare_connect/opportunity/models.py
```diff
@@ -64,9 +64,12 @@ class Opportunity(BaseModel):
         on_delete=models.CASCADE,
         null=True,
     )
+    # to be removed
     max_visits_per_user = models.IntegerField(null=True)
     daily_max_visits_per_user = models.IntegerField(null=True)
+    start_date = models.DateField(null=True, default=datetime.date.today)
```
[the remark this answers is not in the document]
sravfeyn: Will the existing opportunities without this field get auto-set to today, they won't right? That's the reason the null is there.

commcare_connect/opportunity/migrations/0037_add_max_visits.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 4.2.5 on 2024-03-20 04:27
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
sravfeyn: True, I will merge them once we are through the review

commcare_connect/opportunity/models.py
sravfeyn: Hmm, how did I miss that attribute. I will change it.

sravfeyn:
Here are screeshots. It's also live on staging.

Screenshot 2024-04-05 at 4 09 00 PM

Screenshot 2024-04-05 at 4 08 26 PM

commcare_connect/opportunity/models.py
calellowitz: `OpportunityClaim` has a `OneToOneField` foreign key to `OpportunityAccess` which provides that constraint (there can only be one per user per opportunity).

commcare_connect/opportunity/models.py
calellowitz: We chatted about this in the weekly meeting and I think there was a mis alignment somewhere previously, but we decided as a group that all payment units should contribute to the budget and visit counts, since for many projects (even if not all of them), the payment unit dependencies are true independent visits happening at different times that accrue different payments, and even for the chc use case here, the accrual happens in the "sub unit" not the top level one.

commcare_connect/opportunity/api/serializers.py
```diff
@@ -105,6 +114,15 @@ def get_deliver_progress(self, obj):
...
         return opp_access.visit_count
 
+    def get_max_visits_per_user(self, obj):
+        return obj.paymentunit_set.aggregate(max_total=Sum("max_total")).get("max_total", 0)
+
+    def get_daily_max_visits_per_user(self, obj):
+        return obj.paymentunit_set.aggregate(max_daily=Sum("max_daily")).get("max_daily", 0)
+
+    def get_budget_per_visit(self, obj):
+        return obj.paymentunit_set.aggregate(amount=Sum("amount")).get("amount", 0)
```
calellowitz, line 124: This seems like it will sum over all visits (since each unit is a different visit) not per visit. I think ideally this would be a weighted average (max_total*amount for each type / overall_total), but if thats too complicated this can be `max(amount)` for now (if there are two visit types worth 1 and 2 dollars, still better to say 2 than 3).

commcare_connect/opportunity/api/serializers.py
```diff
@@ -105,6 +114,15 @@ def get_deliver_progress(self, obj):
         opp_access = _get_opp_access(self.context.get("request").user, obj)
         return opp_access.visit_count
 
+    def get_max_visits_per_user(self, obj):
+        return obj.paymentunit_set.aggregate(max_total=Sum("max_total")).get("max_total", 0)
+
+    def get_daily_max_visits_per_user(self, obj):
+        return obj.paymentunit_set.aggregate(max_daily=Sum("max_daily")).get("max_daily", 0)
```
calellowitz, line 121: I think its safest to use `max(max_daily)` since payment units often represent sequential visits. This is all temporary until we update the mobile anyway, but it will be slightly more accurate.

calellowitz (changes requested):
I left a few small comments on the API changes. Beyond the crashes, the only other thing is that it looks like you may have missed the update to https://github.com/dimagi/commcare-connect/pull/298/files#r1558250600

Otherwise, this looks good to me. thanks for sticking through all this work

commcare_connect/opportunity/api/serializers.py
sravfeyn: Consider below
- PU1 (max_daily=6)
- PU2 (max_daily=5)

You are saying we should return 6 max_daily for now? I would expect 11 visits to be allowed per day. I guess that's fine temporarily (ideally we should send these as per payment units and mobile should handle the same)

commcare_connect/opportunity/api/serializers.py
sravfeyn: Cool, I will make it max for now.

commcare_connect/opportunity/api/serializers.py
calellowitz: Yeah, I agree that the real answer is definitely on mobile. In the meantime, it sounds like Rama agrees with you, so I hate to say this, but can you change it back before merging?

sravfeyn: Cool. Is it just here or the other place as well? https://github.com/dimagi/commcare-connect/pull/298#discussion_r1561887527

commcare_connect/opportunity/models.py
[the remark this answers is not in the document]
pxwxnvermx: According to django docs the default value specified on the models will be used for older table rows. 
https://docs.djangoproject.com/en/5.0/ref/migration-operations/#addfield

commcare_connect/opportunity/management/commands/populate_completed_work_user_visits.py
```diff
@@ -22,6 +22,7 @@ def handle(self, *args, **options):
                     opportunity=access.opportunity,
                     user=access.user,
                     deliver_unit__in=payment_unit.deliver_units.all(),
+                    is_trial=False,
```
pxwxnvermx, line 25: Can you add this change to the `completed_work` model migration also. The migration contains the code to populate the model and was recently changed.

commcare_connect/form_receiver/processor.py
```diff
@@ -146,7 +147,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     access = OpportunityAccess.objects.get(opportunity=opportunity, user=user)
     counts = (
         UserVisit.objects.filter(opportunity=opportunity, user=user)
-        .exclude(status=VisitValidationStatus.over_limit)
+        .exclude(status=VisitValidationStatus.over_limit, is_trial=True)
```
pxwxnvermx, line 150: This needs to be changed to the below suggestion. The exclude method does an AND on the filter kwargs so this would cause only Visits that have both `VisitValidationStatus=over_limit AND is_trial=true` to be excluded and will have the wrong daily count.
```suggestion
        .exclude(Q(status=VisitValidationStatus.over_limit) | Q(is_trial=True))
```

commcare_connect/opportunity/management/commands/populate_completed_work_user_visits.py
sravfeyn: Which migration are you referring to? The 0033 one? Isn't that already applied?

pxwxnvermx: Yeah, that one was already applied but I had to add a fix some days ago for that migration and I added code to migrate the completed work to the model itself. So this line of code will be required in that migration also.

https://github.com/dimagi/commcare-connect/blob/b15b4679dfdebfe1c71f85ed02d379666877ef8e/commcare_connect/opportunity/migrations/0033_completedwork_uservisit_completed_work.py#L7-L29

commcare_connect/templates/opportunity/opportunity_init.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+{% block title %}{{ request.org }} - Create Opportunity{% endblock %}
+
+{% block content %}
+  {% include "steps.html" with step='step1' %}
+{% include "partial_form.html" %}
```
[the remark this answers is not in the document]
sravfeyn: Good catch. Will add

commcare_connect/opportunity/management/commands/populate_completed_work_user_visits.py
sravfeyn: Ah, I see. Cool

commcare_connect/form_receiver/processor.py
sravfeyn: Right, updated.

commcare_connect/opportunity/models.py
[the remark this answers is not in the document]
calellowitz: If we do go with a default (which I think makes more sense than a nullable value because then we can be sure it always exists, unless this is necessary for the creation workflow), I think it makes more sense to use the opportunity creation date, rather than today if possible.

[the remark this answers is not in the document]
sravfeyn: This makes sense, I will remove null=True. 
>  I think it makes more sense to use the opportunity creation date, rather than today if possible.
Right, for old opportunities. For new ones, it would be today

commcare_connect/opportunity/models.py
sravfeyn: https://github.com/dimagi/commcare-connect/pull/298/files#diff-0718ce467ae87450bbb0b1d53245b67353a89cc48810cfb040f38a31e1bfd77f

commcare_connect/opportunity/models.py
sravfeyn: (looks like my old comment didn't stick)

Rama was of the opinion that we might care about the visit status of trial forms such as over-limit/duplicate. So `is_trial` becomes necessary.

pxwxnvermx (approved):
Looks good to me. Thanks 🙌🏽

calellowitz (approved):

PR #301, opened 2024-04-08
Verification Flags Config

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-291)

image


commcare_connect/form_receiver/processor.py
```diff
@@ -185,24 +183,42 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+                if visit.get("location") is None:
+                    continue
+                lat, lon, *_ = visit["location"].split(" ")
+                dist = distance((lat, lon), (cur_lat, cur_lon))
+                if dist.m <= 10:
+                    flags.append(["location", "Visit location is too close to another visit"])
+                    break
+    if (
+        opportunity_flags.form_submission_start
+        and opportunity_flags.form_submission_start.time() < xform.metadata.timeStart.time()
```
calellowitz, line 214: are these checking server timezone or phone timezone?

commcare_connect/opportunity/forms.py
```diff
@@ -418,3 +419,53 @@ def __init__(self, *args, **kwargs):
...
+class OpportunityVerificationFlagsConfigForm(forms.ModelForm):
+    class Meta:
+        model = OpportunityVerificationFlags
+        fields = ("duplicate", "duration", "gps", "location", "form_submission_start", "form_submission_end")
+        widgets = {
+            "form_submission_start": forms.TimeInput(attrs={"type": "time", "class": "form-control"}),
+            "form_submission_end": forms.TimeInput(attrs={"type": "time", "class": "form-control"}),
+        }
+        labels = {
+            "duplicate": "Check Duplicates",
```
calellowitz, line 433: It is pretty unclear what these mean, can you add a bit more explanation so users will understand what is being checked

calellowitz (commented):
Edited comment, but left a few questions

commcare_connect/opportunity/forms.py
pxwxnvermx: Fixed. Also added some text for GPS. Please suggest changes to help texts if any. https://github.com/dimagi/commcare-connect/pull/301/commits/842fe1a3788615fb8efd1a4d01e05511e6632159

commcare_connect/form_receiver/processor.py
pxwxnvermx: Thanks for pointing this out. Fixed here. https://github.com/dimagi/commcare-connect/pull/301/commits/fef6d6d915c80072d3ea01fc1bcd150cb9fa1501

pxwxnvermx: Also on a further note, we are using the UTC timezone right now. It would make sense to add a help text under the form submission section to indicate that the start and end times need to be adjusted according to the UTC time.

commcare_connect/opportunity/forms.py
```diff
@@ -616,6 +616,8 @@ class Meta:
         help_texts = {
             "duration": "Form completion duration threshold. (minutes)",
             "location": "Check form distance to other form submission. (metres)",
```
calellowitz, line 618: I would mention that these are the minimum duration and distance between forms. maybe something like

> Minimum time to complete form (minutes)

and 

> Minimum distance between form locations (metres)

commcare_connect/form_receiver/processor.py
calellowitz: Oops, I think you may not have been clear, and you took this the opposite way I mean. Because there can be a large delay between when a form is filled out and received by connect (for example if they need to wait for better data coverage to submit, or repeaters are slow), we need to use the time the form was filled not received to check whether it is flagged. I just wanted to make sure that the we were careful to make sure the entered time and form time were in the same timezone (not sure if what we get from the phone is in UTC or local time).

calellowitz (commented):
One small thing and some suggestions on language, otherwise looks like there is a merge conflict but then this is good to go

commcare_connect/form_receiver/processor.py
pxwxnvermx: The phone timezone is UTC.

pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/301/commits/66688cdd2e7e6621e0dae74f763ae7be821fa180

commcare_connect/opportunity/forms.py
pxwxnvermx: Thanks. Added. https://github.com/dimagi/commcare-connect/pull/301/commits/d7d5d30d3f689bc69747adba8928b3705815a2fc

calellowitz (approved):
looks good once the conflicts are fixed

calellowitz (approved):

calellowitz (approved):

PR #302, opened 2024-04-10
Add enhanced learn status for assessments

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-294)

image


commcare_connect/opportunity/models.py
```diff
@@ -223,6 +223,23 @@ def display_name(self):
         else:
             return "---"
 
+    @property
+    def assessment_count(self):
+        return Assessment.objects.filter(user=self.user, opportunity=self.opportunity).count()
+
+    @property
+    def assessment_status(self):
```
calellowitz, line 231: not blocking, but it would be a little better if we didnt have to hit the db twice. Both can be done from the same db query, but I see how this is simpler to implement.

calellowitz (approved):
one piece of not blocking feedback

commcare_connect/opportunity/models.py
pxwxnvermx: Added caching on assessment counts and used that property to get both total and passed/failed counts. https://github.com/dimagi/commcare-connect/pull/302/commits/4fb2f9af9e3a6749e5156814e5f59939a9362da7

commcare_connect/opportunity/models.py
```diff
@@ -276,6 +277,30 @@ def display_name(self):
         else:
             return "---"
 
+    @property
```
calellowitz, line 280: there is a `cached_property` decorator in django that should handle this https://docs.djangoproject.com/en/5.0/ref/utils/#module-django.utils.functional

pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/302/commits/8511d645903cf17689677fda64d2e561a18ff197

sravfeyn (approved):

calellowitz (approved):

PR #303, opened 2024-04-17
Fix all visits should count towards work completion

pxwxnvermx:
- Added code to make all visits make a CompletedWork complete.

This code is present in #294.

### DIFF
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -348,9 +348,7 @@ class CompletedWork(models.Model):
     @property
     def completed_count(self):
         """Returns the no of completion of this work. Includes duplicate submissions."""
-        visits = self.uservisit_set.filter(status=VisitValidationStatus.approved).values_list(
-            "deliver_unit_id", flat=True
-        )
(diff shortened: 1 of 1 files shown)


commcare_connect/opportunity/models.py
```diff
@@ -348,9 +348,7 @@ class CompletedWork(models.Model):
     @property
     def completed_count(self):
         """Returns the no of completion of this work. Includes duplicate submissions."""
-        visits = self.uservisit_set.filter(status=VisitValidationStatus.approved).values_list(
```
calellowitz, line 351: I can't tell if I asked this previously, but this property is used for accrual right? So it needs to limit to approved visits?

pxwxnvermx: [This document](https://docs.google.com/document/d/1XMjmBJPMOIYoQWUYpHFkY2JS2KluIr-oJ3ffV3288N8/edit#heading=h.6tqrqqcvmf19) has changed the criteria now. Any CompletedWork that is approved will accrue payment irrespective of the individual visits.

PR #305, opened 2024-04-18
User Invite Status

pxwxnvermx:
- Added User Invite model to store status for User Invites.

[Ticket](https://dimagi.atlassian.net/browse/CCCT-288)

image


commcare_connect/opportunity/helpers.py
```diff
@@ -1,42 +1,58 @@
...
     access_objects = (
-        OpportunityAccess.objects.filter(opportunity=opportunity)
-        .select_related("user", "opportunityclaim")
+        UserInvite.objects.filter(opportunity=opportunity)
+        .select_related("opportunity_access", "opportunity_access__opportunityclaim", "opportunity_access__user")
         .annotate(
-            last_visit_date_d=Max("user__uservisit__visit_date", filter=Q(user__uservisit__opportunity=opportunity)),
+            last_visit_date_d=Max(
+                "opportunity_access__user__uservisit__visit_date",
+                filter=Q(opportunity_access__user__uservisit__opportunity=opportunity),
```
sravfeyn, line 14: Should this exclude trial visits?

commcare_connect/opportunity/helpers.py
```diff
@@ -1,42 +1,58 @@
 from django.db.models import Case, Count, F, Max, Min, Q, Sum, When
 
-from commcare_connect.opportunity.models import Opportunity, OpportunityAccess, VisitValidationStatus
+from commcare_connect.opportunity.models import Opportunity, OpportunityAccess, UserInvite, VisitValidationStatus
 
 
 def get_annotated_opportunity_access(opportunity: Opportunity):
```
sravfeyn, line 6: Might be good to add test-coverage for this if not present already

pxwxnvermx: We have tests for this function in the `commcare_connect/opportunity/tests/test_export.py` file.

commcare_connect/opportunity/helpers.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/305/commits/501472374aac7b65bf31b50b88289bb05a519d8a

commcare_connect/opportunity/api/views.py
```diff
@@ -125,3 +127,18 @@ def post(self, *args, **kwargs):
         payment.confirmation_date = now()
         payment.save()
         return Response(status=200)
+
+
+class SMSStatusCallbackView(APIView):
```
calellowitz, line 132: nit: i don't think this belongs in opportunity API views, since this is currently only for the api the mobile uses to update its backend. I think this could just be in the normal views folder, either in users or opportunities, although i would lean toward users because that is where the other invite related views are.

calellowitz (approved):
One nit but this otherwise looks fine

commcare_connect/opportunity/tables.py
```diff
@@ -147,15 +153,24 @@ class Meta:
         empty_text = "No users invited for this opportunity."
         orderable = False
 
+    def render_display_name(self, record):
+        if record.opportunity_access is None:
```
calellowitz, line 157: This is obviously a different PR, but we may need to be careful about this when the credential workflow rolls out because this would expose the phone number of anyone with a given credential (which may not be what we want to do before they accept)

calellowitz (commented):
Sorry left one more comment

commcare_connect/opportunity/helpers.py
```diff
@@ -1,42 +1,65 @@
...
+    VisitValidationStatus,
+)
 
 
 def get_annotated_opportunity_access(opportunity: Opportunity):
     learn_modules_count = opportunity.learn_app.learn_modules.count()
     access_objects = (
-        OpportunityAccess.objects.filter(opportunity=opportunity)
-        .select_related("user", "opportunityclaim")
+        UserInvite.objects.filter(opportunity=opportunity)
```
calellowitz, line 15: Since one to one fields should be able to traverse the foreign key in either direction, it would be nice to leave this as based on the opportunity access, which would also make the diff much smaller and make the PR less risky

pxwxnvermx: The User Invites can exist for users who don't have a valid connect-id. Also in cases where users don't have a connect-id account, we do not create the opportunity access that would lead to no invite status showing for them in the UserStatusTable if the model that supports it is kept to be OpportunityAccess.

commcare_connect/opportunity/tables.py
pxwxnvermx: Thanks for pointing this out. I have a query regarding this. Can we show phone numbers for users who don't have a ConnectId account?

commcare_connect/opportunity/api/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/305/commits/1867c9ec34dbf50ff8a76c88051d3df054529e07

commcare_connect/opportunity/tables.py
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/305/commits/fe9a83e79dfe6161c811470f0781185ebb03c2ed

I updated the code to only show phone numbers for users who have not registered on ConnectID.

calellowitz (approved):

PR #306, opened 2024-04-18
Migrate trial to status

sravfeyn:
I have manually migrated data on prod for now. Before we deploy it we should migrate the data again.

Screenshot 2024-04-18 at 3 38 16 PM


calellowitz (changes requested):
Thanks for these changes. It would be great if you could add the data migration to the migration file so that data is updated to reflect the new status but the rest looks good

sravfeyn:
Updated.

calellowitz (approved):

PR #311, opened 2024-04-22
Fix stats

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-310

### DIFF
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -4,7 +4,7 @@ from uuid import uuid4
 
 from django.conf import settings
 from django.db import models
-from django.db.models import Count, F, Sum
+from django.db.models import Count, F, Max, Sum
 from django.utils.timezone import now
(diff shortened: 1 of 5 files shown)


commcare_connect/opportunity/models.py
```diff
@@ -146,8 +146,19 @@ def number_of_users(self):
 
     @property
     def allotted_visits(self):
-        payment_units = self.paymentunit_set.all()
-        return sum([pu.max_total or 0 for pu in payment_units]) * self.number_of_users
+        return self.max_visits_per_user_new * self.number_of_users
+
+    @property
+    def max_visits_per_user_new(self):
```
calellowitz, line 152: nit: `new` is not a great name for these since it doesn't explain why this is different (and if we add another one, will become out of date). ideally we would use a descriptive name, or delete the old column.

calellowitz (approved):
one nit but otherwise looks good. merging to get the fixes out

PR #312, opened 2024-04-22
Fix over_limit issue when multiple payment units are configured

sravfeyn:


commcare_connect/form_receiver/tests/test_receiver_endpoint.py
```diff
@@ -41,7 +41,12 @@ def test_process_xform_error(user: User, api_client: APIClient):
 
 
 def add_credentials(api_client: APIClient, user: User):
-    token = user.oauth2_provider_accesstoken.create(
-        expires=now() + datetime.timedelta(hours=1), token="token", scope="read write"
-    )
+    if not user.oauth2_provider_accesstoken.count():
```
calellowitz, line 44: What is the reason for this change?

commcare_connect/form_receiver/processor.py
```diff
@@ -146,7 +146,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     deliver_unit = get_or_create_deliver_unit(app, deliver_unit_block)
     access = OpportunityAccess.objects.get(opportunity=opportunity, user=user)
     counts = (
-        UserVisit.objects.filter(opportunity=opportunity, user=user)
+        UserVisit.objects.filter(opportunity=opportunity, user=user, deliver_unit=deliver_unit)
```
calellowitz, line 149: I think this logic might work, but it seems not quite right. Ideally we would be checking `CompletedWork` since that is where the limit is, but this seems fine for now.

calellowitz (approved):
Left a couple questions comments but fine for now

PR #315, opened 2024-04-25
Fix payment accrued

pxwxnvermx:
The payment accrued calculation does not calculate the payment correctly when an approved completed work is rejected.
Also combined the similar `update_payment_accrued` and `approve_completed_work_and_update_payment_accrued` function.s


calellowitz (approved):

pxwxnvermx (commented):
Looks good to me. Thanks

PR #316, opened 2024-04-26
ignore the same entity for location check

calellowitz:
I have been meaning to fix this for a while, but we shouldn't flag forms as being too close if they are for the same entity because we expect those to be close (ie same child should be in the same location).


sravfeyn:
You might want to consider doing this change on top of https://github.com/dimagi/commcare-connect/pull/306 instead of over main

calellowitz:
Great Idea!

pxwxnvermx (approved):
Thanks. 🙌🏽

PR #319, opened 2024-05-02
User Suspension

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-301)

image

image

image

image


commcare_connect/templates/opportunity/user_profile.html
```diff
@@ -71,3 +81,48 @@ <h1 class="display-4 mb-0 text-center">{{access.display_name}}</h1>
...
+                <td>{{ access.payment_accrued }}</td>
+              </tr>
+              <tr>
+                <th scope="row">{% translate "Payment Pending Disbursement" %}</th>
+                <td>{{ pending_payment }}</td>
+              </tr>
+            </tbody>
+          </table>
+          <div class="alert alert-danger" role="alert">
+            {% translate "You will be unable to approve payments or pay the user after suspension." %}
```
calellowitz, line 117: This could be a bit clearer that you won't be able to approve payments for that user, rather than everyone. it is a little unclear now and could be scary.

commcare_connect/form_receiver/processor.py
```diff
@@ -220,6 +220,9 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
             if dist.m <= 10:
                 flags.append(["location", "Visit location is too close to another visit"])
                 break
+    if access.suspended:
```
calellowitz, line 223: Do we even want these to show up? I wonder if we want to ignore them instead.

commcare_connect/opportunity/models.py
```diff
@@ -268,7 +271,7 @@ def last_visit_date(self):
 
     @property
     def total_paid(self):
-        return Payment.objects.filter(opportunity_access=self).aggregate(total=Sum("amount")).get("total", 0)
+        return Payment.objects.filter(opportunity_access=self).aggregate(total=Sum("amount")).get("total", 0) or 0
```
calellowitz, line 274: When does the `or` trigger here. It looks like the other thing should default to zero?

calellowitz (commented):
Couple questions but this looks good. One thing that would probably be ideal, would be to delink this user from their HQ worker. I suspect that would be tricky since the user might be reused in multiple opportunities for now, but a good reason for us to look at user per opportunity in the future. Because then we can actively prevent syncs and submissions rather than just rejecting them on receipt

commcare_connect/opportunity/models.py
pxwxnvermx: The thing should default to 0 but it was None in cases where no Payment objects existed, this prevents this method from returning None.

commcare_connect/opportunity/views.py
```diff
@@ -813,3 +826,31 @@ def update_completed_work_status_import(request, org_slug=None, pk=None):
...
+
+
+@org_member_required
+@require_POST
+def suspend_user(request, org_slug=None, opp_id=None, pk=None):
+    access = get_object_or_404(OpportunityAccess, opportunity_id=opp_id, id=pk)
+    access.suspended = True
+    access.suspension_date = now()
+    access.suspension_reason = request.POST.get("reason", "")
+    access.save()
```
sravfeyn, line 838: What happens to existing accrued and approved payments when user is suspended? Shouldn't they be rejected here?

commcare_connect/form_receiver/processor.py
sravfeyn: Yeah, should a suspended user be even able to submit the forms?

sravfeyn:
Does this have a corresponding mobile change? Shouldn't mobile know if a user is suspended?

commcare_connect/opportunity/views.py
pxwxnvermx: The existing accrued and approved payments are not touched when a user is suspended. The Spec mentioned that we should not make changes to data that is already on the platform and only focus on the data coming after the user is suspended.

commcare_connect/form_receiver/processor.py
pxwxnvermx: This change is completely based on the Spec, I can make the change to ignore the submissions.

sravfeyn: Makes sense, I guess, then it's a spec question for Rama.

commcare_connect/opportunity/views.py
sravfeyn: Cool.

pxwxnvermx:
> Does this have a corresponding mobile change? Shouldn't mobile know if a user is suspended?

This query came up recently in QA. I asked Nitin to create a corresponding ticket on the CCC Jira board.
https://dimagi.atlassian.net/browse/CCCT-333

commcare_connect/templates/opportunity/user_profile.html
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/319/commits/2946e237c6dea71eedbf5574e6d000ebdb7eef81

calellowitz (approved):
Pending the mobile work, this seems fine

calellowitz:
How do we communicate to the mobile that a user is suspended?

pxwxnvermx:
Added a field on the Opportunity API list to indicate if a user is suspended from an opportunity. https://github.com/dimagi/commcare-connect/pull/319/commits/667659ccc8d4b6c5c2a35a671c25842e2d8fb682.
@calellowitz

calellowitz (approved):

calellowitz (approved):

PR #320, opened 2024-05-02
Display start/enddate and make sortable table

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-267
Screenshot 2024-05-22 at 10 31 44 AM

### DIFF
--- a/commcare_connect/templates/opportunity/opportunity_list.html
+++ b/commcare_connect/templates/opportunity/opportunity_list.html
@@ -1,64 +1,8 @@
 {% extends "opportunity/base.html" %}
 {% load static %}
+{% load sort_link %}
 
 {% block title %}{{ request.org }} - Opportunities{% endblock %}
-
-{% block inline_javascript %}
-  {{ block.super }}
-  <script>
-    function sortableTable() {
-        return {
-            sortBy: "",
-            sortAsc: false,
-            sortByColumn($event) {
-                if (this.sortBy === $event.target.innerText) {
-                    if (this.sortAsc) {
-                        this.sortBy = "";
-                        this.sortAsc = false;
-                    } else {
-                        this.sortAsc = !this.sortAsc;
-                    }
-                } else {
-                    this.sortBy = $event.target.innerText;
-                }
-
-                let rows = this.getTableRows()
-                    .sort(
-                        this.sortCallback(
-                            Array.from($event.target.parentNode.children).indexOf(
-                                $event.target
-                            )
-                        )
-                    )
-                    .forEach((tr) => {
-                        this.$refs.tbody.appendChild(tr);
-                    });
-            },
-            getTableRows() {
-                return Array.from(this.$refs.tbody.querySelectorAll("tr"));
-            },
-            getCellValue(row, index) {
-                return row.children[index].innerText;
-            },
-            sortCallback(index) {
-                return (a, b) =>
-                    ((row1, row2) => {
-                        return row1 !== "" &&
-                            row2 !== "" &&
-                            !isNaN(row1) &&
-                            !isNaN(row2)
-                            ? row1 - row2
-                            : row1.toString().localeCompare(row2);
-                    })(
-                        this.getCellValue(this.sortAsc ? a : b, index),
-                        this.getCellValue(this.sortAsc ? b : a, index)
-                    );
-            }
-        };
-    }
-
-  </script>
-{% endblock inline_javascript %}
 
 {% block content %}
   <div class="container bg-white shadow">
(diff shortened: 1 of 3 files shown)


commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -3,6 +3,63 @@
 
 {% block title %}{{ request.org }} - Opportunities{% endblock %}
 
+{% block inline_javascript %}
+  {{ block.super }}
+  <script>
+    function sortableTable() {
```
sravfeyn, line 9: We could of course do this using jQuery datatables but that will add few JS dependencies. So this is done without any dependencies using AlpineJS.

sravfeyn:
@calellowitz FYI

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -3,6 +3,63 @@
 
 {% block title %}{{ request.org }} - Opportunities{% endblock %}
 
+{% block inline_javascript %}
+  {{ block.super }}
+  <script>
```
calellowitz, line 8: Is there a way to make this more reusable, since we will likely need sortable columns in other tables

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -15,19 +72,23 @@ <h1> Opportunities
         </span>
       </h1>
     </div>
-    <div class="pb-4">
+    <div class="pb-4" x-data="sortableTable()">
```
calellowitz, line 75: This looks like it will only sort the items displayed, but for paginated data, you won't get the true first X items. I think this might need to be implemented like the paginations controls, to rerender the table, and fetch the data from the backend. That likely makes the final implementation simpler (since you won't need much js, just click functions that make ajax calls). Let me know if I am misunderstanding how this works.

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -15,19 +72,23 @@ <h1> Opportunities
...
       <table class="table border table-responsive">
         <thead class="table-light">
         <tr>
-          <th>Name</th>
+          <th @click="sortByColumn" role="button">Name <i class="bi bi-arrow-down-short"></i></th>
+          <th @click="sortByColumn" role="button">Start Date <i class="bi bi-arrow-down-short"></i></th>
```
calellowitz, line 80: Does the arrow switch direction when the columns are sorted in the opposite order?

calellowitz (commented):
Left a couple questions but only one major one.

commcare_connect/templates/opportunity/opportunity_list.html
sravfeyn: Not really, apparently that requires good amount of Javascript and CSS. Best handled by libraries such as datatable

commcare_connect/templates/opportunity/opportunity_list.html
sravfeyn: > This looks like it will only sort the items displayed, but for paginated data, you won't get the true first X items.

Yes, that's right. This sorts client side data only. The server side sorting could be done, but it's entirely different approach and does need much rework. My understanding of the ticket was that they were looking for client side sorting.

sravfeyn:
@calellowitz I have updated the sorting to be done on server side. I have updated the screenshot too.

commcare_connect/web/templatetags/sort_link.py
```diff
@@ -0,0 +1,45 @@
...
+        new_sort = field
+        icon = "bi-arrow-down-square-fill"
+    else:
+        new_sort = field
+        icon = "bi-arrow-down-short"
+
+    # Construct the URL with the new sorting parameter
+    url = f"{request.path}?sort={new_sort}"
+    if request.GET.get("page"):
+        url += f"&page={request.GET.get('page')}"
```
calellowitz, line 26: nit: if you change the sorting it should probably go back to page 1 (it is unintuitive to start on page 3 of a different sort)

calellowitz (approved):
This looks much better. Thanks for making these changes. One potential change but nothing blocking

commcare_connect/web/templatetags/sort_link.py
sravfeyn: Good point [9dd2cd6120236f5672f3b9d678caf0434b5c9434](https://github.com/dimagi/commcare-connect/pull/320/commits/9dd2cd6120236f5672f3b9d678caf0434b5c9434)

PR #321, opened 2024-05-03
reuse apps

calellowitz:
This adds some safety features around reusable apps.
1) Prevents changing an opportunity from inactive to active if either of its apps are already in use.
2) Removes apps already in use in active opportunities from the list of available apps when setting up an opportunity


pxwxnvermx (commented):
We might need to think about the Deliver Units which are a part of the Payment Units in inactive opportunities, even if the user can reuse those apps they won't be able to create new payment units because the deliver units are already used. 

The filtering for already used app looks good to me.

calellowitz:
> We might need to think about the Deliver Units which are a part of the Payment Units in inactive opportunities

Great point. I will update the PR for that

sravfeyn (approved):
Looks good, may be some basic tests could make it easier for future modifications.

pxwxnvermx (approved):

PR #322, opened 2024-05-08
Multi Visit API changes to Opportunity List API.

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-300)

Changes:
- Opportunity List API now includes a new key `payment_units` which is an array of payment units for the opportunity.
```
payment_units: [
  {
    id: 1,
    name: "Payment Unit 1",
    total_max: 100,
    daily_max: 2
  }
]
```

- Opportunity List API `claim` key updated to include information from different payment unit configurations for the user.
- Note: the `payment_unit` key below is the payment_unit id.
```json
"claim": {
      "max_payments": 200,
      "end_date": "2024-05-06",
      "date_claimed": "2024-04-12",
      "payment_units": [
        {
          "max_visits": 100,
          "payment_unit": 1
        },
        {
          "max_visits": 100,
          "payment_unit": 2
        }
      ]
 }
```

### DIFF
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -13,6 +13,7 @@ from commcare_connect.opportunity.models import (
     Opportunity,
     OpportunityAccess,
     OpportunityClaim,
+    OpportunityClaimLimit,
     Payment,
     PaymentUnit,
     UserVisit,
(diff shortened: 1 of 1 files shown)


OrangeAndGreen (commented):
For the most part everything looks good to me. One question: should the payment_units in the opportunity call also include the actual payment amount (i.e. 1000 MWK)? I don't think we're displaying that info in mobile yet but it sounds like we probably will be soon

calellowitz (approved):

PR #323, opened 2024-05-15
User Credential Views

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-321)

Filter by Country and Credential, and Add Users:
image

Add Credential to Users:
image


commcare_connect/opportunity/views.py
```diff
@@ -152,8 +152,11 @@ def form_valid(self, form):
         opportunity = form.instance
         opportunity.modified_by = self.request.user.email
         users = form.cleaned_data["users"]
-        if users:
-            add_connect_users.delay(users, form.instance.id)
+        filter_country = form.cleaned_data["filter_country"]
+        filter_credential = form.cleaned_data["filter_credential"]
```
sravfeyn, line 156: Do these get validated per https://github.com/dimagi/commcare-connect/pull/323/files#diff-5eeea0bc14a923bfbb6d28553079e296782035b7dae74aa86fa1f0f644355135R91

commcare_connect/organization/views.py
```diff
@@ -25,10 +28,19 @@ def organization_home(request, org_slug):
     if not form:
         form = OrganizationChangeForm(instance=org)
 
+    credentials = connect_id_client.fetch_credentials()
+    if not len(list(c for c in credentials if c.name == f"Worked for {org.name}")):
```
sravfeyn, line 32:
```suggestion
    if not any(c.name == f"Worked for {org.name}" for c in credentials):
```

commcare_connect/organization/views.py
```diff
@@ -58,3 +70,19 @@ def accept_invite(request, org_slug, invite_id):
...
     messages.success(request, message=f"Accepted invite for joining {organization.slug} organization.")
     return redirect("organization:home", org_slug)
+
+
+@org_admin_required
+@require_POST
+def add_credential_view(request, org_slug):
+    org = get_object_or_404(Organization, slug=org_slug)
+    credentials = connect_id_client.fetch_credentials()
+    if not len(list(c for c in credentials if c.name == f"Worked for {org.name}")):
```
sravfeyn, line 80:
```suggestion
    if not any(c.name == f"Worked for {org.name}" for c in credentials):
```

commcare_connect/opportunity/views.py
pxwxnvermx: Yes, we only offer a choice on the select and do not let the user enter custom data.

commcare_connect/organization/views.py
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/pull/323/commits/f309779882048f4113df5559024342e746929f24 Thanks

commcare_connect/organization/views.py
pxwxnvermx: Added. https://github.com/dimagi/commcare-connect/commit/f309779882048f4113df5559024342e746929f24

sravfeyn (approved):

commcare_connect/organization/views.py
```diff
@@ -25,10 +28,19 @@ def organization_home(request, org_slug):
     if not form:
         form = OrganizationChangeForm(instance=org)
 
+    credentials = connect_id_client.fetch_credentials()
+    if not any(c.name == f"Worked for {org.name}" for c in credentials):
+        credentials.append(Credential(name=f"Worked for {org.name}", slug="default"))
```
calellowitz, line 33: what does it mean that slug is always default here?

commcare_connect/organization/views.py
```diff
@@ -58,3 +70,19 @@ def accept_invite(request, org_slug, invite_id):
...
     return redirect("organization:home", org_slug)
+
+
+@org_admin_required
+@require_POST
+def add_credential_view(request, org_slug):
+    org = get_object_or_404(Organization, slug=org_slug)
+    credentials = connect_id_client.fetch_credentials()
+    if not any(c.name == f"Worked for {org.name}" for c in credentials):
+        credentials.append(Credential(name=f"Worked for {org.name}", slug=f"Worked for {org.name}"))
```
calellowitz, line 81: why does the slug differ between the two places this occurs?

calellowitz: Additionally, slugs should be an actual slug not normal text (lowercased, underscore separated, etc).

calellowitz (commented):
A couple small questions but seems mostly good once the merge conflicts are fixed.

commcare_connect/organization/views.py
pxwxnvermx: This is here in case there are no credentials present in the database. If a credential with the same name is present the default credential is not added to the list.

commcare_connect/organization/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/323/commits/9aa3e50ceef5c1575cb6bf9733cdf91a96cef236

commcare_connect/organization/views.py
calellowitz: Is a credential added to the DB with a slug `default` ever?

pxwxnvermx: Nope, we only send the name of the credential that is to be added. The slugs are generated in ConnectID from the credential names.

calellowitz (approved):

PR #328, opened 2024-05-31
Change default behaviour for auto approval settings

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-335


pxwxnvermx (approved):

commcare_connect/opportunity/migrations/0045_alter_opportunity_auto_approve_payments_and_more.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.2.5 on 2024-05-30 11:14
+
+from django.db import migrations, models
+
+
+def change_defaults(apps, schema_editor):
```
calellowitz, line 6: This isn't changing the defaults, it changes everything right? Is that intentional? We want to change how existing projects work? If so, we should change the name of this method, but that seems overly aggressive since some of these projects are in progress.

commcare_connect/opportunity/models.py
```diff
@@ -75,8 +75,8 @@ class Opportunity(BaseModel):
     total_budget = models.IntegerField(null=True)
     api_key = models.ForeignKey(HQApiKey, on_delete=models.DO_NOTHING, null=True)
     currency = models.CharField(max_length=3, null=True)
-    auto_approve_visits = models.BooleanField(default=False)
-    auto_approve_payments = models.BooleanField(default=False)
+    auto_approve_visits = models.BooleanField(default=True)
```
calellowitz, line 78: If we are changing all values to true, and removing all ability to set it to false, why do we need this column still? It is impossible for this value to be false after the PR?

commcare_connect/opportunity/forms.py
```diff
@@ -43,7 +42,6 @@ def __init__(self, *args, **kwargs):
         self.helper.layout = Layout(
             Row(Field("name")),
             Row(Field("active", css_class="form-check-input", wrapper_class="form-check form-switch")),
-            Row(Field("auto_approve_visits", css_class="form-check-input", wrapper_class="form-check form-switch")),
```
calellowitz, line 46: Don't we need to remove both of these toggles based on the description? If the idea is to make it impossible to reject a payment when the forms have no flags, we would need to auto approve both the visit and the payment

calellowitz (changes requested):
Left some comments but I think this only does part of the ask, and separately I am a bit concerned how this affects ongoing projects

commcare_connect/opportunity/models.py
sravfeyn: Yeah, I deliberated on whether to remove it completely but kept it in case there are ideas to use it again for any specific project (or usecase)

commcare_connect/opportunity/migrations/0045_alter_opportunity_auto_approve_payments_and_more.py
sravfeyn: (sorry for the delay here).
> We want to change how existing projects work? 
Per my chat with Rama earlier, this seems what he wanted.

[3:33 PM](https://dimagi.slack.com/archives/D057T7G55DJ/p1717063383265839)
Sravan Reddy
 Should we turn Auto Approve Visits on for all existing opportunities (even the ones that have them OFF)?

3:33 PM
And for ‘Auto Approve Payments’ we don’t change for existing opportunities?
3:34 PM
Rama Siva
:palm_tree: You mean for existing Opportunities. Yes. There is currently only 1 live Opportunity that is real. They have already turned it on for both.
[3:34 PM](https://dimagi.slack.com/archives/D057T7G55DJ/p1717063445057319)
So no need to worry about what will happen to existing opportunities. Make it the same for past, current and future opportunities.

commcare_connect/opportunity/forms.py
sravfeyn: This is from Rama
> [3:30 PM](https://dimagi.slack.com/archives/D057T7G55DJ/p1717063216534179)
Rama Siva
:palm_tree: For payment approval, we should leave the toggle button there and turn it on by default

calellowitz (approved):

commcare_connect/opportunity/models.py
```diff
@@ -84,8 +84,8 @@ class Opportunity(BaseModel):
...
```
pxwxnvermx, line 87: The defaults for these columns need to be reverted back to false since the migration was removed. This will cause a new migration to be created when someone runs the makemigrations command.

sravfeyn: That's a good point. Though, I think we just need to add migration to change the defaults. I have added it.

pxwxnvermx (approved):

PR #329, opened 2024-06-04
Viewer Access

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-299)

Will update with screenshots soon.


calellowitz (approved):
Instead of disabling the buttons, it might be easier to just hide them, but this seems fine.

commcare_connect/organization/decorators.py
```diff
@@ -16,6 +18,10 @@ def _request_user_is_admin(request):
     ) or request.user.is_superuser
 
 
+def _request_user_is_viewer(request):
+    return (request.org and request.org_membership) or request.user.is_superuser
```
sravfeyn, line 22: I would have expected this to check on `request.org_membership.is_viewer`

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -8,10 +8,16 @@
...
+          {% if request.org_membership.is_viewer %}
+          <button class="btn btn-primary" disabled>
+            <span class="bi bi-plus"></span> Add new
+          </button>
+          {% else %}
           <a class="btn btn-primary"
              href="{% url 'opportunity:init' org_slug=request.org.slug %}" title="Add new opportunity"><span
             class="bi bi-plus"></span> Add new
           </a>
+          {% endif %}
```
sravfeyn, line 20: I think this might work be more concise way that works as well
```suggestion
          <a class="btn btn-primary {% if request.org_membership.is_viewer %}disabled{% else %}href="{% url 'opportunity:init' org_slug=request.org.slug %}" title="Add new opportunity"{% endif %}">
            <span class="bi bi-plus"></span> Add new
          </a>
```

commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -45,15 +51,23 @@ <h1> Opportunities
                    href="{% url 'opportunity:detail' org_slug=request.org.slug pk=opportunity.id %}">
                   <span class="bi bi-eye"></span><span class="d-none d-md-inline">&nbsp;View</span>
                 </a>
+              {% if request.org_membership.is_viewer %}
```
sravfeyn, line 54: Similar to https://github.com/dimagi/commcare-connect/pull/329/files#r1646916211

We should try to have a single HTML and disable condition within it, rather than branching on the condition to different HTML

commcare_connect/templates/opportunity/visit_verification.html
```diff
@@ -128,8 +128,13 @@ <h2 class="mb-2">Images</h2>
     </dl>
     {% endif %}
     <div id="action-buttons" style="margin-bottom: 25px;">
-    <a class="btn btn-success" href="{% url 'opportunity:approve_visit' org_slug=request.org.slug pk=visit.id  %}" role="button">Approve</a>
-    <a class="btn btn-danger" href="{% url 'opportunity:reject_visit' org_slug=request.org.slug pk=visit.id  %}" role="button">Reject</a>
+      {% if request.org_membership.is_viewer %}
```
sravfeyn, line 131: Similar to https://github.com/dimagi/commcare-connect/pull/329/files#r1646916211

commcare_connect/templates/opportunity/opportunity_list.html
pxwxnvermx: I tried to add the disabled attribute on the `<a>` tag but the disabled is not supported on links. The disabled attribute is supported by a button and mostly input related elements. : )

commcare_connect/organization/decorators.py
pxwxnvermx: There is a slight caveat here in naming, the viewer access is allowed for all roles i.e. MEMBER, ADMIN, or VIEWER. Hence this doesn't check if a user is a viewer.

calellowitz (approved):

PR #330, opened 2024-06-11
Fix delivery progress counts

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-347)

### DIFF
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -263,7 +263,9 @@ class OpportunityAccess(models.Model):
 
     @property
     def visit_count(self):
-        return self.completedwork_set.exclude(status=CompletedWorkStatus.over_limit).count()
+        return sum(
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/models.py
```diff
@@ -263,7 +263,9 @@ def learn_progress(self):
 
     @property
     def visit_count(self):
-        return self.completedwork_set.exclude(status=CompletedWorkStatus.over_limit).count()
+        return sum(
```
calellowitz, line 266: Do we want to exclude duplicates here? I forget a bit what we decided in the past.

calellowitz (commented):
One question, but otherwise seems fine.

commcare_connect/opportunity/models.py
pxwxnvermx: We did not exclude duplicates in the older code also. I am not sure what we decided for the duplicates.

calellowitz (approved):

PR #337, opened 2024-06-20
admin report

calellowitz:
This is a first pass at the admin report requested in https://docs.google.com/document/d/1eLmsbNKSTlGJ-qEsuOKQMbo-WBcek5Tgl6dMHJnPpNA/edit

There are a few unanswered questions in the doc, but wanted this out there so that we could start tagging the opportunities and playing with the report. Its also currently only accessible by manually specifying the URL, but once it is done we can add links for superusers

image


commcare_connect/reports/views.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@login_required
+@user_passes_test(lambda user: user.is_superuser)
+@require_GET
+def delivery_stats_report(request):
+    table_data = []
+    quarters = _get_quarters_since_start()
+    for q in quarters:
+        data = _get_table_data_for_quarter(q)
```
sravfeyn, line 82: Will this be fast enough for a synchronous request?

calellowitz: It will if it's an internal only report that is used quarterly for reporting. As our data scales, we can either make it generate an excel, or precompute into a dedicated table, but at our current size, this seems to work fine.

pxwxnvermx (approved):

calellowitz:
Fixed the migration so overriding the branch protection to get it deployed tonight, as that was the only feedback.

PR #342, opened 2024-06-24
Fix update user invite status on start learn app

pxwxnvermx:
This PR adds a change to accept the UserInvite when the user starts a learn app. This was missing in the User Invite PR.

### DIFF
--- a/commcare_connect/users/views.py
+++ b/commcare_connect/users/views.py
@@ -108,11 +108,15 @@ def start_learn_app(request):
         ConnectIDUserLink.objects.create(commcare_username=cc_username, user=request.user, domain=domain)
     try:
         access_object = OpportunityAccess.objects.get(user=request.user, opportunity=opportunity)
+        user_invite = UserInvite.objects.get(opportunity_access=access_object)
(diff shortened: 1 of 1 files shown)


commcare_connect/users/views.py
```diff
@@ -108,11 +108,15 @@ def start_learn_app(request):
         ConnectIDUserLink.objects.create(commcare_username=cc_username, user=request.user, domain=domain)
     try:
         access_object = OpportunityAccess.objects.get(user=request.user, opportunity=opportunity)
+        user_invite = UserInvite.objects.get(opportunity_access=access_object)
```
calellowitz, line 111: this should be out of the try/catch since the catch isn't relevant to it

calellowitz (changes requested):

commcare_connect/users/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/342/commits/08dacc3d2ab1940cc0e9a4d61c9e2daa7efc0603

calellowitz (approved):

PR #345, opened 2024-06-27
Added status date in UserVisit and CompletedWork

hemant10yadav:
[JIRA Ticket](https://dimagi.atlassian.net/browse/CCCT-363)


commcare_connect/opportunity/models.py
```diff
@@ -406,6 +406,7 @@ class CompletedWork(models.Model):
     entity_id = models.CharField(max_length=255, null=True, blank=True)
     entity_name = models.CharField(max_length=255, null=True, blank=True)
     reason = models.CharField(max_length=300, null=True, blank=True)
+    status_modification_date = models.DateTimeField(auto_now=True)
```
calellowitz, line 409: This will update every time the model is saved, rather than just when the status is updated. Also, I think it needs to be nullable since older records won't have this data

commcare_connect/opportunity/models.py
```diff
@@ -501,6 +502,7 @@ class UserVisit(XFormBaseModel):
     flagged = models.BooleanField(default=False)
     flag_reason = models.JSONField(null=True, blank=True)
     completed_work = models.ForeignKey(CompletedWork, on_delete=models.DO_NOTHING, null=True, blank=True)
+    status_modification_date = models.DateTimeField(auto_now=True)
```
calellowitz, line 505: same as above

calellowitz (changes requested):
Overall looks good, just a few small things on the model definition

commcare_connect/opportunity/models.py
hemant10yadav: Added null=True

commcare_connect/opportunity/models.py
hemant10yadav: Fixed

commcare_connect/form_receiver/processor.py
```diff
@@ -195,16 +195,20 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
             or datetime.date.today() > claim.end_date
         ):
             user_visit.status = VisitValidationStatus.over_limit
+            user_visit.status_modification_date = now()
```
sravfeyn, line 198: I would suggest to create an `update_status` method on the `UserVisit`.

```
def update_status(status):
     self.status = status
     self.status_modification_date = now()
```
Then you can replace most of the updates without having to update the date in each place.

There is also property setter way of doing this for e.g. https://stackoverflow.com/a/57523694

commcare_connect/opportunity/models.py
```diff
@@ -406,6 +406,7 @@ class CompletedWork(models.Model):
     entity_id = models.CharField(max_length=255, null=True, blank=True)
     entity_name = models.CharField(max_length=255, null=True, blank=True)
     reason = models.CharField(max_length=300, null=True, blank=True)
+    status_modification_date = models.DateTimeField(null=True)
```
sravfeyn, line 409: `nit` I know the name came from @calellowitz but we tend to use verb-based such as `status_modified_date`

commcare_connect/form_receiver/processor.py
hemant10yadav: I have reviewed the information provided in this link:  https://stackoverflow.com/questions/2898711/django-model-fields-getter-setter/57523694#57523694. If I use the property decorator, I will need to change the property name from status to _status in the model. This change will require updating any queries where the status property is used. Am I correct in this understanding? Should I proceed with this approach, or should I consider defining a method instead?

commcare_connect/opportunity/migrations/0050_completedwork_status_modification_date_and_more.py
pxwxnvermx, line 1: You can combine these migrations into single migrations. You can follow the below steps to compile the migrations.

1. Unapply the migrations locally using
`python manage.py migrate opportunity 0049`
2. Delete 0050 and 0051 migrations in the `opportunity/migrations`.
3. Regenerate the migration files using
`python manage.py makemigrations`

This will generate an updated migration which will contain the fields with the correct types.

hemant10yadav: Thanks @pxwxnvermx.
Done

commcare_connect/opportunity/models.py
hemant10yadav: Thanks @sravfeyn changed

sravfeyn (approved):
Thanks for addressing the comments. Looks good 💯

calellowitz (approved):

PR #348, opened 2024-07-01
Advanced Verification Flags

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-350)

Deliver Form Flags:
- Shows a row for every deliver unit in the opportunity.
image

Form Json Validation Rules:
- Can be added using the + icon.
- Can be removed using the delete button.
image


calellowitz:
Do you have screenshots for this?

pxwxnvermx:
@calellowitz Updated the PR with screenshots.

calellowitz:
Is there a way to make the attachments text styled in the same way as the other two columns? It may look best to have a header, and then something next to the toggle, like `Attachments` above, and then `require attachments` next to it. It will keep the page look much more consistent. For the others, is it possible to move the buttons, both add and delete to the bottom and style them more like our other buttons (the reject and approve ones for example?). Generally the code looks fine, but the styling doesn't quite fit in with the rest of the site.

Additionally, if there is one row per delivery unit anyway, how do you feel about just having the delivery unit name as static text, rather than a dropdown that you would select for each?

pxwxnvermx:
@calellowitz I changed the styling for the Deliver Unit Forms and the Buttons for Form Json forms.

calellowitz:
This looks great! Any chance you can make the deliver unit names just text instead of disabled selects? Otherwise, good to go, and if that will be a pain, I am happy to approve as is.

pxwxnvermx:
I did try to make the disabled select boxes just text, but that was not working as expected. @calellowitz

calellowitz (approved):

PR #352, opened 2024-07-04
Download and upload catchment area

hemant10yadav:
[Jira Ticket](https://dimagi.atlassian.net/browse/CCCT-377)


commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +315,122 @@ def get_status_by_completed_work_id(dataset):
...
+                    created = None
+                    if CATCHMENT_ID in headers and row[_get_header_index(headers, CATCHMENT_ID)]:
+                        catchment_id = row[_get_header_index(headers, CATCHMENT_ID)]
+                        catchment, created = CatchmentArea.objects.get_or_create(
+                            id=catchment_id,
+                            defaults={
+                                "latitude": latitude,
+                                "longitude": longitude,
+                                "radius": radius,
+                                "opportunity_accesses": opportunity_accesses[username],
```
pxwxnvermx, line 402: This fieldname should be `opportunity_access`.

pxwxnvermx (commented):
Can we add tests for the export and import? That will ensure that any future changes don't break the current functionality that we have.

commcare_connect/opportunity/models.py
```diff
@@ -580,3 +581,13 @@ class UserInvite(models.Model):
     opportunity_access = models.OneToOneField(OpportunityAccess, on_delete=models.CASCADE, null=True, blank=True)
     message_sid = models.CharField(max_length=50, null=True, blank=True)
     status = models.CharField(max_length=50, choices=UserInviteStatus.choices, default=UserInviteStatus.invited)
+
+
+class CatchmentArea(models.Model):
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.CASCADE)
+    latitude = models.DecimalField(max_digits=10, decimal_places=8)
```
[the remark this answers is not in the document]
hemant10yadav: 1. If both max_digits and decimal_places are set to 8 in a DecimalField, it means you are defining a field that can store numbers with exactly 8 digits in total, all of which are used for decimal places.

2. Latitude values have a narrower range (-90 to +90), so fewer total digits are needed to represent them 
accurately.
Longitude values have a wider range (-180 to +180), requiring more total digits to represent them accurately.

commcare_connect/opportunity/visit_import.py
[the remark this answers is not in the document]
hemant10yadav: Thanks, It was a typo.

commcare_connect/templates/opportunity/opportunity_detail.html
```diff
@@ -481,4 +499,37 @@ <h1 class="modal-title fs-5 text-white">{% translate "Import Verified Payments"
...
+            enctype="multipart/form-data">
+        <div class="modal-body">
+          <div class="mb-3">
+            {% csrf_token %}
+            <input class="form-control" type="file" id="importFile" name="catchments" required>
+            <div class="col-auto">
+              <span id="importFileHelp" class="form-text">
+                {% blocktrans %}
+                  The file must contain at least the "Area Name", "Active", "Latitude", "Longitude" and "Radius(meter)" column.
+                  The import is case-insensitive.
```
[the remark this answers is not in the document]
hemant10yadav: In all other imports flow "The import is case-insensitive." was mentioned.
Should I remove it?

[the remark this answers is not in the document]
pxwxnvermx: Yeah, that text can be removed here. In the other exports we were converting all the user input to lowercase.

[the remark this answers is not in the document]
hemant10yadav: Ok

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +315,122 @@ def get_status_by_completed_work_id(dataset):
...
+                        catchment.latitude = latitude
+                        catchment.longitude = longitude
+                        catchment.radius = radius
+                        catchment.active = active
+                        catchment.opportunity_accesses = opportunity_accesses[username]
+                        catchment.opportunity = opportunity
+                        catchment.name = area_name
+                        to_update.append(catchment)
+
+                    seen_catchments.add(catchment.name)
```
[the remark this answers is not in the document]
hemant10yadav: Changed

commcare_connect/opportunity/visit_import.py
```diff
@@ -27,6 +29,12 @@
 WORK_ID_COL = "instance id"
 PAYMENT_APPROVAL_STATUS_COL = "payment approval"
 REQUIRED_COLS = [VISIT_ID_COL, STATUS_COL]
+LATITUDE_COL = "latitude"
+LONGITUDE_COL = "longitude"
+RADIUS_COL = "radius"
+AREA_NAME_COL = "area name"
+ACTIVE_COL = "active"
+CATCHMENT_ID = "catchment id"
```
calellowitz, line 37: nit: i think this can just be id

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,121 @@ def get_status_by_completed_work_id(dataset):
...
+                        opportunity=opportunity,
+                        name=area_name,
+                        active=active,
+                    )
+                    missing_catchments += 1
+                    to_create.append(catchment)
+                elif row[username_index] in opportunity_access:
+                    username = row[username_index]
+                    catchment = None
+                    created = None
```
calellowitz, line 391: nit: since this is a boolean elsewhere, it can just be false

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,121 @@ def get_status_by_completed_work_id(dataset):
...
+                                "radius": radius,
+                                "opportunity_access": opportunity_access[username],
+                                "opportunity": opportunity,
+                                "name": area_name,
+                                "active": active,
+                            },
+                        )
+
+                    if not created:
+                        catchment.latitude = latitude
```
calellowitz, line 408: `catchment` can be `None` here. what happens in that case?

calellowitz (commented):
I will look at the core logic monday, but left a few comments and questions. Additionally, this would be easier to review if it were PRed into my branch, rather than just includes that commit, since my branch has also continued to evolve.

commcare_connect/opportunity/visit_import.py
hemant10yadav: Added this case too.

commcare_connect/opportunity/visit_import.py
hemant10yadav: Fixed

commcare_connect/opportunity/visit_import.py
hemant10yadav: Fixed

commcare_connect/utils/file.py
```diff
@@ -0,0 +1,15 @@
+import mimetypes
+import os
+
+
+def get_file_extension(file):
```
sravfeyn, line 5: 💯

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,133 @@ def get_status_by_completed_work_id(dataset):
...
+                    )
+                    missing_catchments += 1
+                    to_create.append(catchment)
+                elif row[username_index] in opportunity_access:
+                    username = row[username_index]
+                    catchment = None
+                    created = False
+                    if ID in headers and row[_get_header_index(headers, ID)]:
+                        catchment_id = row[_get_header_index(headers, ID)]
+                        catchment, created = CatchmentArea.objects.get_or_create(
```
sravfeyn, line 394: Why not just get it and if it's not there add to `to_create` so that it can take advantage of bulk creation?

hemant10yadav: @sravfeyn Update the logic as per your suggestions.

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,173 @@ def get_status_by_completed_work_id(dataset):
...
+
+def bulk_update_catchments(opportunity: Opportunity, file: UploadedFile):
+    file_format = get_file_extension(file)
+    if file_format not in ("csv", "xlsx"):
+        raise ImportException(f"Invalid file format. Only 'CSV' and 'XLSX' are supported. Got {file_format}")
+    imported_data = get_imported_dataset(file, file_format)
+    return _bulk_update_catchments(opportunity, imported_data)
+
+
+class RowDataError(Exception):
```
calellowitz, line 326: It would be nice to move all exceptions to the top of the file (or maybe even their own `exceptions.py` file) so they are in one place and easy to find.

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,173 @@ def get_status_by_completed_work_id(dataset):
...
+            return None
+        catchment_id = self.row[index]
+        return catchment_id if catchment_id else None
+
+
+def create_or_update_catchment(row_data: RowData, opportunity: Opportunity, opportunity_access: dict):
+    if row_data.id:
+        catchment = None
+        try:
+            catchment = CatchmentArea.objects.get(id=row_data.id)
```
calellowitz, line 416: It would be great if this enforced that the model was in the right opportunity, especially since the id is a simple int, and its easy to guess other people's ids.

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,173 @@ def get_status_by_completed_work_id(dataset):
...
+
+    return (
+        CatchmentArea(
+            latitude=row_data.latitude,
+            longitude=row_data.longitude,
+            radius=row_data.radius,
+            opportunity=opportunity,
+            name=row_data.area_name,
+            active=row_data.active,
+            opportunity_access=opportunity_access.get(row_data.username, None),
```
calellowitz, line 435: This is quite confusing because the variable `opportunity_access` isn't an opportunity access model, but the field `opportunity_access` is. It would be nice to rename the dict as something like `username_to_oa_map` or something similar that is a bit more descriptive.

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,173 @@ def get_status_by_completed_work_id(dataset):
...
+        raise ImportException("The uploaded file did not contain any headers")
+
+    with transaction.atomic():
+        to_create = []
+        to_update = []
+        invalid_rows = []
+        seen_catchments = set()
+        missing_catchments = 0
+
+        opportunity_access = {}
```
calellowitz, line 453: Same comment on naming this

commcare_connect/opportunity/visit_import.py
```diff
@@ -311,3 +313,173 @@ def get_status_by_completed_work_id(dataset):
...
+        if USERNAME_COL in headers:
+            opportunity_access = {
+                oa.user.username: oa
+                for oa in OpportunityAccess.objects.filter(opportunity=opportunity).select_related("user")
```
calellowitz, line 457: you could probably filter this to only the usernames that appear in the file

commcare_connect/opportunity/visit_import.py
```diff
@@ -77,14 +85,20 @@ def get_missing_message(self):
...
+@dataclass
+class CatchmentAreaImportStatus:
+    seen_catchments: set[str]
+    missing_catchments: int
+
+    def __len__(self):
+        return len(self.seen_catchments)
+
+    def get_missing_message(self):
+        return f"<br>{self.missing_catchments} catchment areas were not found."
```
calellowitz, line 97: This makes it seem like there is an error when one is missing, but that is the only way to create new ones, right? I think this can be reworded to say something like "X catchment areas created" and the variable can be renamed `new_catchments` to be clear its not a problem.

commcare_connect/opportunity/tables.py
```diff
@@ -315,6 +316,25 @@ def render_revoke_suspension(self, record, value):
         return format_html('<a class="btn btn-success" href="{}?next={}">Revoke</a>', revoke_url, page_url)
 
 
+class CatchmentAreaTable(tables.Table):
+    id = columns.Column(verbose_name="Id")
```
calellowitz, line 320: It might be nice to add the user phone number and name to these downloads (in addition to the username), because that will be much easier for users to identify the person than the username. It is ok to ignore those in the upload as we do in other downloads.

commcare_connect/opportunity/visit_import.py
calellowitz: On that note, I am a bit uncomfortable matching on id like this, since it is difficult to verify and easy to mess up. How would you feel about matching on name, or adding a new "site_code" to the model that would be a human readable unique identifier (like a slug)

calellowitz (changes requested):
Overall this looks great. A few small changes and questions

commcare_connect/opportunity/visit_import.py
hemant10yadav: Done

commcare_connect/opportunity/visit_import.py
hemant10yadav: Changed

commcare_connect/opportunity/visit_import.py
hemant10yadav: changed

commcare_connect/opportunity/visit_import.py
hemant10yadav: fixed

commcare_connect/opportunity/visit_import.py
hemant10yadav: changed

commcare_connect/opportunity/visit_import.py
hemant10yadav: @calellowitz We can match by name, but no two catchment areas can have the same name within an opportunity, right?

commcare_connect/opportunity/tables.py
hemant10yadav: Added

commcare_connect/opportunity/visit_import.py
```diff
@@ -478,6 +479,8 @@ def _bulk_update_catchments(opportunity: Opportunity, dataset: Dataset):
                 invalid_rows.append((row, f"Error in row {row}: {e}"))
 
         if to_create:
+            for ca in to_create:
+                ca.site_code = slugify_uniquely(ca.name, CatchmentArea, "site_code")
```
calellowitz, line 483: if the users specifies a site code, that should be used instead of just slug of the name

commcare_connect/opportunity/tables.py
```diff
@@ -326,20 +325,32 @@ class CatchmentAreaTable(tables.Table):
...
 
     class Meta:
         model = CatchmentArea
-        fields = ("username", "id", "name", "name_of_user", "phone_number" "active", "latitude", "longitude", "radius")
+        fields = (
+            "username",
+            "site_code",
+            "name",
+            "name_of_user",
+            "phone_number",  # Only include this once
```
calellowitz: what does this comment mean?

commcare_connect/opportunity/models.py
```diff
@@ -601,3 +601,9 @@ class CatchmentArea(models.Model):
     opportunity_access = models.ForeignKey(OpportunityAccess, null=True, on_delete=models.DO_NOTHING)
     active = models.BooleanField(default=True)
     name = models.CharField(max_length=255)
+    site_code = models.SlugField(max_length=255, unique=True, blank=True)
+
+    def save(self, *args, **kwargs):
```
calellowitz, line 606: I don't think we want this logic in the save function. I think the import is enough, and that it needs to be a bit different (as I mention there). Basically, use the provided code unless it doesn't exist in which case sluggify the name

commcare_connect/opportunity/tables.py
```diff
@@ -330,9 +332,18 @@ def render_active(self, value):
...
-        fields = ("username", "id", "name", "active", "latitude", "longitude", "radius")
+        fields = ("username", "id", "name", "name_of_user", "phone_number" "active", "latitude", "longitude", "radius")
         orderable = False
-        sequence = ("username", "id", "name", "active", "latitude", "longitude", "radius")
+        sequence = (
+            "username",
+            "id",
+            "name",
+            "name_of_user",
+            "phone_number" "active",
```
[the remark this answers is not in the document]
calellowitz: I see this is fixed

commcare_connect/opportunity/models.py
calellowitz: It is probably worth requiring a site code at this point, actually. Just raise an upload error if it is not included, then we don't have to guess.

hemant10yadav: @calellowitz Added the requested changes.

commcare_connect/opportunity/models.py
```diff
@@ -599,3 +600,14 @@ class UserInvite(models.Model):
...
+
+class CatchmentArea(models.Model):
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.CASCADE)
+    latitude = models.DecimalField(max_digits=11, decimal_places=8)
+    longitude = models.DecimalField(max_digits=11, decimal_places=8)
+    radius = models.IntegerField(default=1000)
+    opportunity_access = models.ForeignKey(OpportunityAccess, null=True, on_delete=models.DO_NOTHING)
+    active = models.BooleanField(default=True)
+    name = models.CharField(max_length=255)
+    site_code = models.SlugField(max_length=255)
```
calellowitz, line 613: This can have a DB constraint to ensure it is unique to an opportunity

calellowitz (approved):
This looks good pending adding the constraint to ensure the (opportunity, site_code) pairs are unique.

commcare_connect/opportunity/models.py
hemant10yadav: @calellowitz  Done

commcare_connect/form_receiver/processor.py
```diff
@@ -230,6 +230,18 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
                 if dist.m <= 10:
                     flags.append(["location", "Visit location is too close to another visit"])
                     break
+        if opportunity_flags.catchment_areas:
+            areas = access.catchmentarea_set.filter(active=True)
+            if areas:
+                cur_lat, cur_lon, *_ = xform.metadata.location.split(" ")
```
sravfeyn, line 236: It might be worth making this a separate method outside of `process_deliver_unit`

sravfeyn (approved):
Great work on this PR. LGTM pending any of Cal's comments

calellowitz (approved):

calellowitz:
This looks great! Let's get it out to our users.

PR #354, opened 2024-07-08
Map user catchment areas on the user profile page.

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-378

Screenshot 2024-07-09 at 14 28 25

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -41,6 +41,7 @@ from commcare_connect.opportunity.helpers import (
 )
 from commcare_connect.opportunity.models import (
     BlobMeta,
+    CatchmentArea,
     CompletedModule,
     CompletedWork,
     CompletedWorkStatus,
@@ -639,6 +640,7 @@ def payment_delete(request, org_slug=None, opp_id=None, access_id=None, pk=None)
 def user_profile(request, org_slug=None, opp_id=None, pk=None):
     access = get_object_or_404(OpportunityAccess, pk=pk, accepted=True)
     user_visits = UserVisit.objects.filter(user=access.user, opportunity=access.opportunity)
+    user_catchments = CatchmentArea.objects.filter(opportunity_access__user=access.user)
     user_visit_data = []
     for user_visit in user_visits:
         if not user_visit.location:
@@ -667,6 +669,16 @@ def user_profile(request, org_slug=None, opp_id=None, pk=None):
             if cw.approved_count
         ]
     )
+    user_catchment_data = [
+        {
+            "name": catchment.name,
+            "lat": float(catchment.latitude),
+            "lng": float(catchment.longitude),
+            "radius": catchment.radius,
+            "active": catchment.active,
+        }
+        for catchment in user_catchments
+    ]
     pending_payment = max(access.payment_accrued - access.total_paid, 0)
     return render(
         request,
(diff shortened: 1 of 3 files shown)


calellowitz:
Do you have screenshots for this? it would be helpful in reviewing the UI changes

commcare_connect/static/js/project.js
```diff
@@ -63,3 +63,36 @@ function addAccuracyCircles(map, visit_data) {
 }
 
 window.addAccuracyCircles = addAccuracyCircles;
+
+function addCatchmentAreas(map, catchments) {
+  map.on('load', () => {
+    dummyCatchments.forEach((catchment) => {
```
calellowitz, line 69: Is this able to be a single layer instead of one for each catchment area, like the forms above where they are one layer for all of them. Then we also dont have to do the dynamic naming

commcare_connect/opportunity/views.py
```diff
@@ -639,6 +640,7 @@ def payment_delete(request, org_slug=None, opp_id=None, access_id=None, pk=None)
 def user_profile(request, org_slug=None, opp_id=None, pk=None):
     access = get_object_or_404(OpportunityAccess, pk=pk, accepted=True)
     user_visits = UserVisit.objects.filter(user=access.user, opportunity=access.opportunity)
+    user_catchments = CatchmentArea.objects.filter(opportunity_access__user=access.user)
```
calellowitz, line 643: Can the filter just be `opportunity_access=access` or am I missing something?

commcare_connect/static/js/project.js
calellowitz: Where does `dummyCatchments` come from?

calellowitz (commented):
Few questions but the general idea looks right

commcare_connect/static/js/project.js
hemant10yadav: I forgot to rename the variable as i just added dummy content to test it. Will rename it.

commcare_connect/opportunity/views.py
hemant10yadav: Fetched using opportunity_access=access.

commcare_connect/static/js/project.js
```diff
@@ -63,3 +63,72 @@ function addAccuracyCircles(map, visit_data) {
...
+                            <span class="d-inline-block me-2" style="${activeStyle}"></span>
+                            <span>Active</span>
+                        </div>
+                        <div class="d-flex align-items-center">
+                            <span class="d-inline-block me-2" style="${inactiveStyle}"></span>
+                            <span>Inactive</span>
+                        </div>
+                    </div>
+                `;
+    map.getContainer().appendChild(legend);
```
sravfeyn, line 130: We generally try to avoid adding HTML via Javascript in this way (because it's hard to reason with the HTML if you have HTML scattered around various files). Have you explored if it would be possible to add this directly to the HTML and conditionally show it. We have `x-show` and `x-component` in AlpineJS that might be useful here.

calellowitz: Agree with sravan on this one, thanks for updating

hemant10yadav: @sravfeyn  and @calellowitz  Added the  changes.

sravfeyn (approved):

calellowitz (approved):

PR #356, opened 2024-07-08
Implement Deactivate user endpoint on Connect

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-375)

### DIFF
--- a/commcare_connect/users/urls.py
+++ b/commcare_connect/users/urls.py
@@ -4,9 +4,9 @@ from commcare_connect.users.views import (
     SMSStatusCallbackView,
     accept_invite,
     create_user_link_view,
+    deactivate_user,
     demo_user_tokens,
     start_learn_app,
-    user_deactivate,
     user_detail_view,
     user_redirect_view,
     user_update_view,
@@ -22,5 +22,5 @@ urlpatterns = [
     path("accept_invite/<slug:invite_id>/", view=accept_invite, name="accept_invite"),
     path("demo_users/", view=demo_user_tokens, name="demo_users"),
     path("sms_status_callback/", SMSStatusCallbackView.as_view(), name="sms_status_callback"),
-    path("user_suspend/", user_deactivate, name="user_suspend"),
+    path("deactivate_user/", deactivate_user, name="deactivate_user"),
 ]
--- a/commcare_connect/users/views.py
+++ b/commcare_connect/users/views.py
@@ -166,7 +166,7 @@ class SMSStatusCallbackView(APIView):
 @csrf_exempt
 @api_view(["POST"])
 @authentication_classes([OAuth2Authentication])
-def user_deactivate(request):
+def deactivate_user(request):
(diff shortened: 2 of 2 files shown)


commcare_connect/users/views.py
```diff
@@ -161,3 +161,19 @@ def post(self, *args, **kwargs):
                 user_invite.status = UserInviteStatus.sms_not_delivered
             user_invite.save()
         return Response(status=200)
+
+
+@csrf_exempt
+@api_view(["POST"])
+@authentication_classes([OAuth2Authentication])
+def user_deactivate(request):
```
calellowitz, line 169: is mobile supposed to call this or does connectid do it?

pxwxnvermx: Mobile wil call this. I spot the problem here. Thanks. : )

commcare_connect/users/urls.py
```diff
@@ -21,4 +22,5 @@
     path("accept_invite/<slug:invite_id>/", view=accept_invite, name="accept_invite"),
     path("demo_users/", view=demo_user_tokens, name="demo_users"),
     path("sms_status_callback/", SMSStatusCallbackView.as_view(), name="sms_status_callback"),
+    path("user_suspend/", user_deactivate, name="user_suspend"),
```
sravfeyn, line 25: `nit` `user_suspend` -> `suspend_user`

commcare_connect/users/views.py
pxwxnvermx: This is called by mobile.

commcare_connect/users/urls.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/356/commits/48a20697b5277b8744bbedeeffadc09f4ade48cb

commcare_connect/users/views.py
calellowitz: I think it would be safer and better if this were called by connectid. That way we don't need to rely on mobile, and as soon as a user account is deactivated we can be sure no new data will flow in

pxwxnvermx: Yeah, that makes sense. We need to think about the Authentication between Connect and ConnectID for making requests to this API.

calellowitz: Sravan made a good point against this on the spec. Let's chat about this offline and make a decision

commcare_connect/users/views.py
```diff
@@ -161,3 +161,19 @@ def post(self, *args, **kwargs):
                 user_invite.status = UserInviteStatus.sms_not_delivered
             user_invite.save()
         return Response(status=200)
+
+
+@csrf_exempt
+@api_view(["POST"])
+@authentication_classes([OAuth2Authentication])
+def deactivate_user(request):
```
sravfeyn, line 169: It would be a good practice to add a small note here as to where this is used.

PR #359, opened 2024-07-09
Status Modified Date Not Updating from Visit Verification View

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-392


pxwxnvermx (approved):

calellowitz (changes requested):
I hate to ask this since this design was my idea, but I think this illustrates a problem we are going to face again and again. It is currently too easy to update status and forget to call this method. Can we instead override the `save` method for this class to compare old and new values and automatically update it on save. Then we only need to call the normal `.save()` method and it is very difficult (bulk update and save methods would still be an issue) to accidentally update the status without updating the timestamp.

calellowitz:
This will need to hand `bulk_update` as well (or change the visit import code)  since that is how the visit import saves those models. This would also benefit from some tests

commcare_connect/opportunity/models.py
```diff
@@ -408,15 +408,16 @@ class CompletedWork(models.Model):
     reason = models.CharField(max_length=300, null=True, blank=True)
     status_modified_date = models.DateTimeField(null=True)
 
-    def save(self, *args, **kwargs):
-        if self.pk is not None:
-            old_instance = CompletedWork.objects.get(pk=self.pk)
-            if old_instance.status != self.status:
+    def __init__(self, *args, **kwargs):
```
hemant10yadav, line 411: @calellowitz , I've tried implementing this approach, but I'm uncertain whether it's advisable. What's your take on this?

FYI: 
This method sets default values for `status` and `status_modified_date` attributes  when the object is created. If `status` is provided in `kwargs`, it will override the default `CompletedWorkStatus.incomplete` value

The direct assignment to status here is necessary because the __setattr__ method, which handles attribute setting, might otherwise throw an error when attempting to check for the existence of the status property during initialization.

calellowitz: This looks good. I think it still misses some places such as queryset update methods, but it seems fine for now.

commcare_connect/opportunity/visit_import.py
```diff
@@ -107,7 +107,7 @@ def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
                 seen_completed_works.add(visit.completed_work_id)
                 status = status_by_visit_id[visit.xform_id]
                 if visit.status != status:
-                    visit.update_status(status)
```
calellowitz, line 110: I think you need to update [this line](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/visit_import.py#L117) to have the date saved

calellowitz (changes requested):
The approach looks good, but this would really benefit from some tests for the places that it updates

commcare_connect/opportunity/models.py
```diff
@@ -408,9 +408,16 @@ class CompletedWork(models.Model):
     reason = models.CharField(max_length=300, null=True, blank=True)
     status_modified_date = models.DateTimeField(null=True)
 
-    def update_status(self, status):
-        self.status = status
+    def __init__(self, *args, **kwargs):
+        self.status = CompletedWorkStatus.incomplete
```
sravfeyn, line 412: Doesn't the `default=CompletedWorkStatus.incomplete` on [model def](https://github.com/dimagi/commcare-connect/pull/359/files#diff-df90e317de0ae052504b3e88cc61aa8cf379effcb469ea2f74ac7357ac42832cR402) ensure this is set by default?

hemant10yadav: @sravfeyn When the values are initialised, the setattr method is used to set those values. During this process, the setattr method attempts to access the property (as implemented). However, since the property (like status) is not present during initialisation, it will throw an error. This behaviour is caused by the Django model.

sravfeyn (approved):

calellowitz (changes requested):
Just repeating my request to add tests for this code. I think what is here is good, but tests would make me much more confident in the changes

calellowitz (approved):

PR #361, opened 2024-07-10
Add column totals to the Deliver status tab.

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-390

Screenshot 2024-07-10 at 15 38 46

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -118,9 +118,7 @@ class AggregateColumn(columns.Column):
 
 class SumColumn(columns.Column):
     def render_footer(self, bound_column, table):
-        return sum(
-            getattr(x, bound_column.accessor) for x in table.data if getattr(x, bound_column.accessor) is not None
-        )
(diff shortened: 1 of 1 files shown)


commcare_connect/opportunity/tables.py
```diff
@@ -116,6 +116,13 @@ def render_footer(self, bound_column, table):
         return sum(1 if bound_column.accessor.resolve(row) else 0 for row in table.data)
 
 
+class SumColumn(columns.Column):
+    def render_footer(self, bound_column, table):
+        return sum(
+            getattr(x, bound_column.accessor) for x in table.data if getattr(x, bound_column.accessor) is not None
+        )
```
sravfeyn, line 123: Below might give you same result with more conciseness.
```suggestion
        return sum(
            getattr(x, bound_column.accessor, 0) for x in table.data
        )
```

hemant10yadav: @sravfeyn changed

calellowitz (approved):

sravfeyn (approved):

PR #365, opened 2024-07-16
Fix caching bug for suspended users

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-405

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -22,6 +22,7 @@ from django_tables2.export import TableExport
 from geopy import distance
 
 from commcare_connect.form_receiver.serializers import XFormSerializer
+from commcare_connect.opportunity.api.serializers import remove_opportunity_access_cache
 from commcare_connect.opportunity.forms import (
     AddBudgetExistingUsersForm,
     DateRanges,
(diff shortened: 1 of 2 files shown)


pxwxnvermx:
It would be better to remove the caching here and making the `get_opp_access` a method on the `OpportunitySerializer` class with the django [cached_property](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.functional.cached_property) decorator to avoid fetching the opportunity access multiple times. 

The caching here was intially added to accomplish the same goal of avoiding multiple fetches of opportunity access for getting the various properties required on the OpportunitySerializer which can be done better by using the `cached_property` decorator.

@hemant10yadav

pxwxnvermx (approved):

calellowitz (approved):

commcare_connect/opportunity/views.py
```diff
@@ -889,6 +890,10 @@ def suspend_user(request, org_slug=None, opp_id=None, pk=None):
     access.suspension_date = now()
     access.suspension_reason = request.POST.get("reason", "")
     access.save()
+
+    # Clear the cached opportunity access for the suspended user
+    remove_opportunity_access_cache(access.user, access.opportunity)
```
sravfeyn, line 895: You can call `_get_opp_access.clear` directly from within `access.save`

hemant10yadav: @sravfeyn  Changed

calellowitz:
Let's offline about that last commit. I prefer it the way it was before, because we only want to invalidate it at the moment the user is suspended, not every time we save if the user is suspended. This does make me realize we should also do it when they are unsuspended as well, but I would prefer to have that in the relevant method as well, rather than in the save function.

calellowitz (approved):

PR #368, opened 2024-07-22
update quarter date calculation

calellowitz:
This fixes a couple bugs in the admin report.
1) The `render_X` functions were not updated for new payment column names
2) Quarter start and end dates were wrong. Quarter Q starts on month (Q - 1) * 3 + 1. For example,
    Q1 starts on (1 - 1)*3 +1=1
    Q2 starts on (2 - 1)*3 +1=4

### DIFF
--- a/commcare_connect/reports/views.py
+++ b/commcare_connect/reports/views.py
@@ -34,9 +34,9 @@ def _get_quarters_since_start():
 
 
 def _get_table_data_for_quarter(quarter):
-    quarter_start = date(quarter[0], quarter[1] * 3, 1)
+    quarter_start = date(quarter[0], (quarter[1] - 1) * 3 + 1, 1)
     next_quarter = _increment(quarter)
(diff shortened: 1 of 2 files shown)


commcare_connect/reports/views.py
```diff
@@ -34,9 +34,9 @@ def _get_quarters_since_start():
 
 
 def _get_table_data_for_quarter(quarter):
-    quarter_start = date(quarter[0], quarter[1] * 3, 1)
+    quarter_start = date(quarter[0], (quarter[1] - 1) * 3 + 1, 1)
```
pxwxnvermx, line 37: nit: this can be simplified to `Q*3 - 2`

pxwxnvermx (approved):

hemant10yadav (approved):

commcare_connect/reports/views.py
calellowitz: great point! i will do that as followup so i can get this out today

PR #373, opened 2024-07-24
Management command to run one time auto approval for old opportunities

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-419

### DIFF
--- a/commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
+++ b/commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
@@ -1,0 +1,27 @@
+from django.core.management import BaseCommand
+
+from commcare_connect.opportunity.models import Opportunity, OpportunityAccess
+from commcare_connect.opportunity.visit_import import update_payment_accrued
+
+
+class Command(BaseCommand):
+    help = "Run auto-approval logic over opportunity"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--opp", type=int, required=True, help="ID of the opportunity to run auto-approval logic on"
+        )
+
+    def handle(self, *args, **options):
+        opp_id = options["opp"]
+        try:
+            opportunity = Opportunity.objects.get(id=opp_id)
+            access_records = OpportunityAccess.objects.filter(opportunity=opportunity)
+            users = [access.user for access in access_records]
+
+            update_payment_accrued(opportunity=opportunity, users=users)
+
+            self.stdout.write(self.style.SUCCESS(f"Successfully processed opportunity with id {opp_id}"))
+
+        except Opportunity.DoesNotExist:
+            self.stdout.write(self.style.ERROR(f"Opportunity with id {opp_id} does not exist."))
--- a/commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
+++ b/commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
@@ -1,7 +1,7 @@
 from django.core.management import BaseCommand
 
 from commcare_connect.opportunity.models import CompletedWorkStatus, Opportunity, OpportunityAccess
-from commcare_connect.opportunity.utils.completed_work import update_status_and_compute_payment
+from commcare_connect.opportunity.utils.completed_work import update_status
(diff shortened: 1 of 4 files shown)


sravfeyn (approved):

pxwxnvermx (approved):

commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Command(BaseCommand):
+    help = "Run auto-approval logic over opportunity"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--opp", type=int, required=True, help="ID of the opportunity to run auto-approval logic on"
+        )
+
+    def handle(self, *args, **options):
```
calellowitz, line 15: I think you can use `opp` as a named argument instead of getting it from options since it was added above.

commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
```diff
@@ -0,0 +1,27 @@
...
+        opp_id = options["opp"]
+        try:
+            opportunity = Opportunity.objects.get(id=opp_id)
+            access_records = OpportunityAccess.objects.filter(opportunity=opportunity)
+            users = [access.user for access in access_records]
+
+            update_payment_accrued(opportunity=opportunity, users=users)
```
calellowitz, line 22: The more I think about this, the more I think dont want to update `payment_accrued` just the status. In other words, I think just https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/visit_import.py#L158-L167

This also presents a good opportunity to break those out into some shared functions since much of that is duplicated in https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/tasks.py#L325

We can have both of those call the same utility functions (where behavior is shared), and have this call that function as well.

calellowitz (changes requested):
One tiny nit and one bigger change. Apologies for not being clear in the ticket but I think we only want to update the completed works, not the payment accrual (for now)

commcare_connect/opportunity/management/commands/auto_approval_opportunities.py
hemant10yadav: [fb7d9c1]([fb7d9c1](https://github.com/dimagi/commcare-connect/pull/373/commits/fb7d9c1836a80f8bbaff5ea17e8d7b0616cd7bdf))
Added

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -0,0 +1,31 @@
+from commcare_connect.opportunity.models import CompletedWorkStatus
+
+
+def update_status_and_compute_payment(completed_works, opportunity, compute_payment=True):
```
calellowitz, line 4: This is a really misleading function name if compute payment is `False`

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -0,0 +1,31 @@
...
+            continue
+
+        if opportunity.auto_approve_payments:
+            update_completed_work_status(completed_work)
+
+        if compute_payment:
+            approved_count = completed_work.approved_count
+            if approved_count > 0 and completed_work.status == CompletedWorkStatus.approved:
+                payment_accrued += approved_count * completed_work.payment_unit.amount
+    return payment_accrued
```
calellowitz, line 20: It is a bit confusing that this function modifies some fields but returns others. I think it would be clearer if it just modified everything rather than forcing the caller to assign some values. It is also a bit unintuitive that it saves some models but not others. I would lean toward it just saving everything, unless there is a good reason even though that is generally not best practice. It could be controlled by a variable if that feels better

calellowitz (changes requested):

commcare_connect/opportunity/utils/completed_work.py
hemant10yadav: @calellowitz Changed

calellowitz (approved):

PR #379, opened 2024-08-14
Network Manager Invite and Claim

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-456

Screenshot 2024-08-14 at 12 54 54
Screenshot 2024-08-14 at 13 00 29
Screenshot 2024-08-14 at 13 00 07
Screenshot 2024-08-14 at 12 55 14


commcare_connect/templates/opportunity/opportunity_list.html
```diff
@@ -89,53 +89,55 @@ <h1> Opportunities
...
+            <a class="btn btn-primary btn-sm"
+               href="{% url 'opportunity:apply_opportunity_invite' org_slug=request.org.slug pk=invitation.managed_opportunity.id application_id=invitation.id %}">
+              <span class="bi bi-check-circle-fill"></span><span class="d-none d-md-inline">&nbsp;Apply</span>
+            </a>
+          </div>
+        </td>
+      </tr>
+      {% endfor %}
+      </tbody>
+    </table>
```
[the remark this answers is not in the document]
calellowitz: strong plus one to this

commcare_connect/opportunity/views.py
```diff
@@ -114,11 +118,32 @@ def get_queryset(self):
...
+        # Opportunities that are linked to the organization through a managed opportunity application
+        # where the status is 'accepted'.
+        opportunities = Opportunity.objects.filter(
+            Q(organization=self.request.org, managed=False)
+            | Q(
+                id__in=ManagedOpportunityApplication.objects.filter(
+                    organization=self.request.org, status=ManagedOpportunityApplicationStatus.ACCEPTED
+                ).values("managed_opportunity_id")
+            )
+        ).order_by(ordering)
```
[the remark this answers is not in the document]
calellowitz: agree with this. I made a comment on the other one, thinking we had decided to get rid of the second model entirely, but agree that generally claimed opportunities should have the organization of the network manager.

commcare_connect/program/tests/test_views.py
```diff
@@ -0,0 +1,295 @@
...
+from commcare_connect.program.models import ManagedOpportunityApplication, ManagedOpportunityApplicationStatus, Program
+from commcare_connect.program.tests.factories import (
+    ManagedOpportunityApplicationFactory,
+    ManagedOpportunityFactory,
+    ProgramFactory,
+)
+from commcare_connect.users.models import User
+
+
+class BaseProgramTest:
```
calellowitz, line 20: It would be good to add tests for the forms, I know the managed opportunity one handles a lot, and especially any validation conditions so both good and bad data. It is great that this includes a lot of happy path tests, but tests that confirm expected failures can also be very useful.

calellowitz (commented):
generally looks good. left one comment but otherwise excited to see QA for all this

commcare_connect/program/tests/test_views.py
hemant10yadav: Added

commcare_connect/opportunity/views.py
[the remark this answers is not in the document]
hemant10yadav: Added [fc4ec11](https://github.com/dimagi/commcare-connect/pull/379/commits/fc4ec119c07c654a333af2f0cb15b8c65d07365b)

commcare_connect/templates/opportunity/opportunity_list.html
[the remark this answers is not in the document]
hemant10yadav: I have implemented this in all the other views, but this one is a bit tricky because there is already a table (Opportunity list) that includes sorting. If I create another table with Django-tables2, the pagination of the first table will affect the second one. I think I might need to change the entire view to implement Django-tables2. Should I do that?

commcare_connect/organization/decorators.py
```diff
@@ -18,6 +18,12 @@ def _request_user_is_admin(request):
     ) or request.user.is_superuser
 
 
+def _request_org_is_program_manager(request):
```
sravfeyn, line 21: Shouldn't this be`_request_user_is_program_manager`?

hemant10yadav: Done [84a04d7](https://github.com/dimagi/commcare-connect/pull/379/commits/84a04d76f25496165c9c7b6a5085bc87ccc5284a)

commcare_connect/opportunity/views.py
```diff
@@ -110,6 +110,16 @@ def test_func(self):
...
+def get_opportunity_or_404(pk, org_slug):
+    opp = get_object_or_404(Opportunity, id=pk)
+    if opp.organization is not None and opp.organization.slug == org_slug:
+        return opp
+    elif opp.managed and opp.managedopportunity.program.organization.slug == org_slug:
+        return opp.managedopportunity
+    else:
+        raise Http404("Opportunity not found.")
+
+
```
hemant10yadav: This is done to ensure that the same creation flow can be used for both opportunities and managed opportunities. Since a managed opportunity will not have an organization initially, we check the program's organization in the case of a managed opportunity to avoid 404 error.

we don't need to return `opp.managedopportunity` so changed the code little bit here [276fef0](https://github.com/dimagi/commcare-connect/pull/379/commits/276fef074507256f35a10eea4ce4f948cf5be67b)

commcare_connect/templates/opportunity/opportunity_list.html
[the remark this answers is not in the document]
pxwxnvermx: I think we are fine with leaving the Opportunity List table for now.

commcare_connect/program/models.py
```diff
@@ -15,7 +15,7 @@ class Program(BaseModel):
     currency = models.CharField(max_length=3)
     start_date = models.DateField()
     end_date = models.DateField()
-    organization = models.ForeignKey(Organization, on_delete=models.DO_NOTHING, null=True, blank=True)
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE)
```
calellowitz, line 18: I don't think we want this to cascade, because we might still want it to show up for the network managers. I think for now lets `PROTECT` and we can always change it later

commcare_connect/templates/program/program_list.html
```diff
@@ -18,42 +19,7 @@ <h1> Programs
...
-        </td>
-      </tr>
-      {% empty %}
-      <tr>
-        <td colspan="3">No programs yet.</td>
-      </tr>
-      {% endfor %}
-      </tbody>
-    </table>
+   {% render_table table %}
```
calellowitz, line 22: nice! this is so much better

commcare_connect/opportunity/forms.py
```diff
@@ -233,7 +235,12 @@ def save(self, commit=True):
         )
         self.instance.created_by = self.user.email
         self.instance.modified_by = self.user.email
-        self.instance.organization = organization
+
+        # In the case of a managed opportunity, the organization will be null initially,
+        # as it will be set to the network manager's organization when the invitation is accepted.
+        if not self.managed_opp:
```
calellowitz, line 241: This is no longer true, right? and we can remove the migration that makes org nullable?

commcare_connect/opportunity/views.py
```diff
@@ -340,7 +340,7 @@ def get_queryset(self):
 @org_member_required
 def export_user_visits(request, **kwargs):
     opportunity_id = kwargs["pk"]
-    get_object_or_404(Opportunity, organization=request.org, id=opportunity_id)
+    get_opportunity_or_404(org_slug=kwargs["org_slug"], pk=opportunity_id)
```
calellowitz, line 343: why does `request.org` no longer work? For all the views in this commit, it would be nice to add the `org_slug` arg to the function signature rather than pulling it from `kwargs`, so `def export_user_visits(request, org_slug, **kwargs):` or similar

calellowitz (commented):
left a few questions/comments but looks good

commcare_connect/opportunity/views.py
hemant10yadav: That was my mistake for not realizing earlier that req.org.org_slug is the same as org_slug. I apologize for that. I will make the necessary changes.

commcare_connect/opportunity/forms.py
hemant10yadav: Yeah this is changed in the later PR's

commcare_connect/program/models.py
hemant10yadav: Changed here ([0924f21](https://github.com/dimagi/commcare-connect/pull/385/commits/0924f21617da12af23fd9b877b3f13bc67d70a42))

As in program invite flow this script again got changed so to avoid conflict did this here

commcare_connect/opportunity/views.py
calellowitz: I won't block on this, because I see that we use this pattern elsewhere, but what I was trying to say was that the function based views would be a lot cleaner if they included the arg names explicitly in the signature, rather than using `kwargs` to get the org slug. Like these views for example https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/views.py#L348

There should also be a way to do that with the class based views but I don't have an easy example at hand

calellowitz (approved):
I dont think my last piece of feedback was really followed, but approving to not block release. It would be great if you could create a followup PR after these are merged with that change

commcare_connect/opportunity/views.py
hemant10yadav: Ok I will create a new PR for these improvements soon.

PR #383, opened 2024-08-22
Enterprise User Visit Review

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-464)

Network Manager View:
image

Program Manager View:
image


commcare_connect/templates/opportunity/visit_verification.html
```diff
@@ -135,10 +135,7 @@ <h2 class="mb-2">Images</h2>
     </dl>
     {% endif %}
     <div id="action-buttons" style="margin-bottom: 25px;">
-      {% if request.org_membership.is_viewer %}
```
calellowitz, line 138: why the change here around viewers?

calellowitz (commented):
Do you have screenshots? Tests would also be great here, since this is pretty key functionality we are changing. Left a couple questions but seems good

commcare_connect/templates/opportunity/visit_verification.html
pxwxnvermx: Fixed this. https://github.com/dimagi/commcare-connect/pull/383/commits/4c03128318ea563e84e8d138ecbac9303044870d

commcare_connect/opportunity/views.py
```diff
@@ -1050,7 +1050,11 @@ def user_visit_review(request, org_slug, opp_id):
     opportunity = get_opportunity_or_404(opp_id, org_slug)
     if not opportunity.managed:
         return redirect("opportunity:detail", org_slug, opp_id)
-    is_program_manager = request.org_membership.is_admin and request.org.program_manager
+    is_program_manager = (
+        request.org_membership != None  # noqa: E711
+        and request.org_membership.is_admin
```
calellowitz, line 1055: why does it need to be an admin? I think in the past we talked about it being a superuser but im not sure we want to restrict it to only admins. Since most of these orgs will act exclusively as program managers, it seems wrong to block the members from acting as program managers

calellowitz (commented):
Tests and screenshots would still be great

commcare_connect/opportunity/views.py
pxwxnvermx: Yeah, this is based off of the Program Spec doc that mentions "All features in the spec should be restricted to organization admins of organizations marked as “program_manager”". I think what you said above makes a lot more sense than what the document mentions.

commcare_connect/opportunity/tasks.py
```diff
@@ -332,12 +332,18 @@ def bulk_approve_completed_work():
...
+                if any(status == "rejected" for status, *_ in visits):
                     completed_work.update_status(CompletedWorkStatus.rejected)
-                    completed_work.reason = "\n".join(reason for _, reason in visits if reason)
-                elif all(status == "approved" for status, _ in visits):
+                    completed_work.reason = "\n".join(reason for _, reason, _ in visits if reason)
+                elif all(status == "approved" for status, *_ in visits):
                     completed_work.update_status(CompletedWorkStatus.approved)
+                if (
+                    access.opportunity.managed
+                    and not all(review_status == "agree" for *_, review_status in visits)
```
calellowitz, line 343: nit: this should reference a constant like we do for the other status

commcare_connect/opportunity/tasks.py
```diff
@@ -333,17 +332,18 @@ def bulk_approve_completed_work():
...
```
calellowitz, line 343: does this get set automatically somewhere for autoapproved forms?

commcare_connect/opportunity/visit_import.py
```diff
@@ -131,14 +131,11 @@ def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
 
                 if visit.status != status:
                     visit.update_status(status)
+                    visit.reason = reason
                     if opportunity.managed and status in ["approved", "rejected"]:
                         visit.review_created_on = now()
                     changed = True
 
-                if status == VisitValidationStatus.rejected and reason and reason != visit.reason:
```
calellowitz, line 135: I think this means changes to the reason after it has already been rejected will no longer be picked up. Is this intentional?

commcare_connect/opportunity/visit_import.py
```diff
@@ -132,7 +132,7 @@ def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
                 if visit.status != status:
                     visit.update_status(status)
                     visit.reason = reason
-                    if opportunity.managed and status in ["approved", "rejected"]:
```
calellowitz: is the review stll created, there just isn't a date on the visit anymore?

commcare_connect/templates/opportunity/visit_verification.html
```diff
@@ -103,7 +103,7 @@ <h4 class="mb-2">Flags</h4>
     </table>
     {% endif %}
     {% if visit.reason %}
-    <h4 class="mb-2">Rejection Reason</h4>
```
calellowitz, line 106: not blocking, but we talked about this on the call and I am not sure we want to remove this language completely

calellowitz (commented):
Couple questions, but generally looks good

commcare_connect/opportunity/visit_import.py
pxwxnvermx: This was a change to make the rejected reason field work as the justification. I reverted these changes in [7a3bfd8](https://github.com/dimagi/commcare-connect/pull/383/commits/7a3bfd8224a27efb62e59f98028645d5a44db573)

commcare_connect/opportunity/visit_import.py
pxwxnvermx: This was a change to make the rejected reason field work as the justification. I reverted these changes in [7a3bfd8](https://github.com/dimagi/commcare-connect/pull/383/commits/7a3bfd8224a27efb62e59f98028645d5a44db573)

commcare_connect/templates/opportunity/visit_verification.html
pxwxnvermx: Fixed. [7a3bfd8](https://github.com/dimagi/commcare-connect/pull/383/commits/7a3bfd8224a27efb62e59f98028645d5a44db573)

commcare_connect/opportunity/tasks.py
pxwxnvermx: Added. [4deb642](https://github.com/dimagi/commcare-connect/pull/383/commits/4deb642b2950301179b299d1ed8e227d08d72fac)

commcare_connect/opportunity/tasks.py
pxwxnvermx: I will make this change once I can merge the main into my current PR. Some changes around the status updates from the main branch are not present on this PR causing issues whenever I change something.

commcare_connect/opportunity/visit_import.py
```diff
@@ -112,41 +114,39 @@ def bulk_update_visit_status(opportunity: Opportunity, file: UploadedFile) -> Vi
 
 
 def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
-    status_by_visit_id, reasons_by_visit_id = get_status_by_visit_id(dataset)
```
calellowitz, line 114: is this old function still needed?

commcare_connect/opportunity/visit_import.py
```diff
@@ -208,17 +208,48 @@ def get_status_by_visit_id(dataset) -> dict[int, VisitValidationStatus]:
...
     return imported_data
 
 
-def _get_header_index(headers: list[str], col_name: str) -> int:
+def _get_header_index(headers: list[str], col_name: str, required=True) -> int:
     try:
         return headers.index(col_name)
     except ValueError:
+        if not required:
+            return -1
```
calellowitz, line 252: `-1` is a valid index, it might be better to return something like `None` that would raise an error if its improperly used. I see you check for this, so its not blocking, but might be safer in the future

commcare_connect/opportunity/visit_import.py
```diff
@@ -208,17 +208,48 @@ def get_status_by_visit_id(dataset) -> dict[int, VisitValidationStatus]:
     return status_by_visit_id, reason_by_visit_id
 
 
+def get_data_by_visit_id(dataset) -> dict[int, any]:
```
calellowitz, line 211: nit: there are enough keys and they are strict enough here that it might make sense to return a dataclass or named tuple for stronger typing

commcare_connect/form_receiver/processor.py
```diff
@@ -271,6 +272,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         and not user_visit.flagged
     ):
         user_visit.update_status(VisitValidationStatus.approved)
+        user_visit.review_status = VisitReviewStatus.agree
```
calellowitz, line 275: are we able to distinguish visits that were auto marked as agreement, and those manually approved by an enterprise user?

calellowitz (approved):
Left a few questions/suggestions but none blocking

hemant10yadav (approved):

commcare_connect/form_receiver/processor.py
pxwxnvermx: We can distinguish between auto marked as agreement and manually approved by using the review_created_on field. That field is not present on auto-marked as agreement visits. That field is also used to indicate that manual approval is required.

commcare_connect/opportunity/visit_import.py
pxwxnvermx: Removed this function and fixed the tests that use this. [737592c](https://github.com/dimagi/commcare-connect/pull/383/commits/737592c1941610838a0c164924504d85b7511dcc)

commcare_connect/opportunity/visit_import.py
pxwxnvermx: Added a dataclass to do this. [737592c](https://github.com/dimagi/commcare-connect/pull/383/commits/737592c1941610838a0c164924504d85b7511dcc)

commcare_connect/opportunity/visit_import.py
pxwxnvermx: yeah, the type system was very unhappy with None so I had to resort to -1 and added a check for it.

PR #384, opened 2024-08-29
Payment Accrual Report for Network Managers

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-471)

image


commcare_connect/opportunity/models.py
```diff
@@ -400,6 +401,8 @@ class Payment(models.Model):
     )
     confirmed = models.BooleanField(default=False)
     confirmation_date = models.DateTimeField(null=True)
+    # This is used to indicate Payments made to Network Manager organizations
+    organization = models.ForeignKey(Organization, on_delete=models.DO_NOTHING, null=True, blank=True)
```
calellowitz, line 405: I think this can just be a boolean like `org_payment` and it would be great to have a constraint to make sure  opportnity access is always null when it is True

commcare_connect/opportunity/views.py
```diff
@@ -1046,3 +1048,49 @@ def user_visit_review(request, org_slug, opp_id):
...
+def payment_report(request, org_slug, pk):
+    opportunity = get_opportunity_or_404(pk, org_slug)
+    if not opportunity.managed:
+        return redirect("opportunity:detail", org_slug, pk)
+    payment_units = PaymentUnit.objects.filter(opportunity=opportunity)
+    PaymentReportData = namedtuple(
+        "PaymentReportData", ["payment_unit", "approved", "user_payment_accrued", "nm_payment_accrued"]
+    )
+    data = []
+    total_paid_users = (
```
calellowitz, line 1063: nit: the queries may be cleaner in a helper function like the other big complex tables. I know those were single db queries, but i think its fine to have complex ones like this there too, becuase it separates it from the view code, making the view easier to read.

commcare_connect/opportunity/models.py
calellowitz: Also is this set in this PR or a different one. I couldn't find it.

calellowitz (commented):
My comment isn't blocking but curious how you feel about the model suggestion

commcare_connect/opportunity/models.py
pxwxnvermx: this is set in this PR. https://github.com/dimagi/commcare-connect/pull/387

commcare_connect/opportunity/views.py
pxwxnvermx: Fixed. a02109cce56287b6d6d6806e928542905a5f7626

calellowitz (approved):

hemant10yadav (approved):

PR #386, opened 2024-09-02
Changed Table arguments to avoid forbidden error in case when program manager view opportunity detail page.

hemant10yadav:
https://dimagi.atlassian.net/browse/QA-7046
Programs Manager, user is unable to access the status on opportunity detail page like payments, visits, deliver etc.

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -31,7 +31,7 @@ class LearnStatusTable(tables.Table):
         empty_text = "No learn progress for users."
 
     def __init__(self, *args, **kwargs):
-        self.org_slug = kwargs.pop("org_slug")
+        self.org_slug = kwargs.pop("org_slug", None)
         super().__init__(*args, **kwargs)
 
     def render_details(self, record):
@@ -166,7 +166,7 @@ class UserStatusTable(tables.Table):
         orderable = False
 
     def __init__(self, *args, **kwargs):
-        self.org_slug = kwargs.pop("org_slug")
+        self.org_slug = kwargs.pop("org_slug", None)
         super().__init__(*args, **kwargs)
 
     def render_display_name(self, record):
@@ -211,7 +211,7 @@ class PaymentUnitTable(tables.Table):
         orderable = False
 
     def __init__(self, *args, **kwargs):
-        self.org_slug = kwargs.pop("org_slug")
+        self.org_slug = kwargs.pop("org_slug", None)
         super().__init__(*args, **kwargs)
 
     def render_deliver_units(self, record):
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/views.py
```diff
@@ -291,6 +291,11 @@ def get_queryset(self):
         opportunity = get_opportunity_or_404(org_slug=self.request.org.slug, pk=opportunity_id)
         return OpportunityAccess.objects.filter(opportunity=opportunity).order_by("user__name")
 
+    def get_table_kwargs(self):
```
sravfeyn, line 294: Why not implement this on a class that subclasses `SingleTableView` and use that everywhere to be DRY?

commcare_connect/opportunity/tables.py
```diff
@@ -254,6 +251,13 @@ class Meta:
             "incomplete",
         )
 
+    def render_details(self, record):
```
calellowitz, line 254: why the rendering change? is it the same end result, because the `LinkColumn` is nice in that its defined with all the other columns rather than in a separate place

commcare_connect/opportunity/tables.py
```diff
@@ -86,15 +95,17 @@ class Meta:
...
     username = columns.Column(accessor="user__username", visible=False)
-    view_payments = columns.LinkColumn(
-        "opportunity:user_payments_table",
-        verbose_name="",
-        text="View Details",
-        args=[utils.A("opportunity__organization__slug"), utils.A("opportunity__id"), utils.A("pk")],
-    )
+    view_payments = columns.Column(verbose_name="", empty_values=())
+
+    def render_view_payments(self, record):
```
calellowitz, line 103: Same rendering question here. Why the move away from link columns?

calellowitz (changes requested):
Left a comment applicable everywhere, but I am a bit confused at the rationale for these changes.

Additionally all PRs should have a description that explains the change and why it was made (ideally with links to a ticket). That might also help clear up my confusion

commcare_connect/opportunity/views.py
hemant10yadav: Thanks @sravfeyn added the changes.

commcare_connect/opportunity/tables.py
hemant10yadav: The rendering has changed because earlier we were getting the `org_slug` from the record itself using `utils.A()`. Now, we're retrieving the org_slug from class-level properties. I think I tried using `utils.A()` to access class-level properties, but it didn't work.


FYI - This is primarily intended for Program Managers. It is not mandatory for them to be members of the Network Manager's organization, but they can still view all details related to opportunities created by the Program Manager.

calellowitz: So 
```
args=[utils.A(self.org_slug), utils.A("opportunity__id"), utils.A("pk")],
```
doesn't work? Is that because you can't reference self there?

hemant10yadav: yes

sravfeyn (approved):

PR #389, opened 2024-09-10
Program Management Feature Changes

pxwxnvermx:
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-477)

The commits have the information related to the change. Please ping me if any additional info is required about the implementation.


commcare_connect/opportunity/views.py
```diff
@@ -135,9 +135,7 @@ def get_context_data(self, **kwargs):
 
         program_invitation_table = None
         if self.request.org_membership and self.request.org_membership.is_admin or self.request.user.is_superuser:
-            program_invitations = ProgramApplication.objects.filter(
-                organization=self.request.org, status=ProgramApplicationStatus.INVITED
-            )
+            program_invitations = ProgramApplication.objects.filter(organization=self.request.org)
```
hemant10yadav, line 138: This will show the invitations even when they are accepted. The table was meant to show only the invitation.

commcare_connect/program/tables.py
```diff
@@ -30,6 +30,8 @@ def render_budget(self, record):
         return f"{record.program.budget} {record.program.currency}"
 
     def render_manage(self, record):
+        if record.status == ProgramApplicationStatus.ACCEPTED:
```
hemant10yadav, line 33: According to [line](https://github.com/dimagi/commcare-connect/pull/389/files#diff-1914cafdbf59040f7f77df353660e2634bcf3ec8e789b018d9864bc6ee7fded2R138) now even when NM decline the request the "Apply" and "Decline" will still appear.

commcare_connect/opportunity/views.py
pxwxnvermx: This change makes all the invites visible for Network Managers. This mitigates the problem of not knowing what Programs a network manager has applied to as the table filters only for Invited.

calellowitz (approved):

hemant10yadav (approved):

sravfeyn (approved):

commcare_connect/opportunity/views.py
```diff
@@ -436,7 +436,7 @@ def add_budget_existing_users(request, org_slug=None, pk=None):
         {
             "form": form,
             "opportunity_claims": opportunity_claims,
-            "budget_per_visit": opportunity.budget_per_visit,
+            "budget_per_visit": opportunity.budget_per_visit_new,
```
calellowitz, line 439: For exactly this reason, we really need to rename these properties (or get rid of the old one if its not in use), but `new` isn't helpful in telling you why it's different or that you should use it.

calellowitz (approved):

PR #390, opened 2024-09-12
Display payment data in USD

sravfeyn:
https://dimagi.atlassian.net/browse/CCCT-478

Note that the payment amounts (that occur on different dates) are summed and then converted to USD as of that day's exchange rate. It doesn't make much sense to sum them using each day's exchange rate without displaying the exchange rates for each of those dates to user.


sravfeyn:
I have modeled this pretty much after HQ. I am completely open to remove this app and simply cache the exchange rates in redis and lookup them from cache for calculations.

commcare_connect/accounting/migrations/0001_initial.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 4.2.5 on 2024-09-12 16:16
+
+from decimal import Decimal
+from django.db import migrations, models
+
+
+def create_task(apps, schema_editor):
+    schedule, _ = CrontabSchedule.objects.get_or_create(
```
sravfeyn, line 8: I have followed the existing way of defining periodic tasks. But I am wondering why are we using the DatabaseScheduler instead of simply defining the schedules via the settings.CELERY_BEAT_SCHEDULE?

calellowitz:
> https://dimagi.atlassian.net/browse/CCCT-478
> 
> Note that the payment amounts (that occur on different dates) are summed and then converted to USD as of that day's exchange rate. It doesn't make much sense to sum them using each day's exchange rate without displaying the exchange rates for each of those dates to user.



> https://dimagi.atlassian.net/browse/CCCT-478
> 
> Note that the payment amounts (that occur on different dates) are summed and then converted to USD as of that day's exchange rate. It doesn't make much sense to sum them using each day's exchange rate without displaying the exchange rates for each of those dates to user.

I think we want to do a one time conversion for the day of the visit. That way the numbers don't change for historical reporting, and they are much closer to the actual values paid out.

commcare_connect/opportunity/migrations/0055_payment_amount_usd.py
```diff
@@ -0,0 +1,46 @@
...
+    for payment in payments:
+        date_paid = payment['date_only']
+        currency = payment['opportunity_access__opportunity__currency']
+
+        if currency in ['USD', None]:
+            exchange_rate = 1
+        else:
+            exchange_rate = get_exchange_rate(currency, date_paid)
+        if not exchange_rate:
+            raise Exception(f"Invalid currency code {currency}")
```
sravfeyn, line 25: May be ignore the error and just print instead?

commcare_connect/opportunity/visit_import.py
```diff
@@ -242,14 +249,36 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
...
     missing_users = set(usernames) - seen_users
     send_payment_notification.delay(opportunity.id, payment_ids)
     return PaymentImportStatus(seen_users, missing_users)
 
 
+def _cache_key(currency_code, date=None):
+    return [currency_code, date.toordinal() if date else None]
+
+
+@quickcache(vary_on=_cache_key, timeout=60 * 60)
```
calellowitz, line 265: Since these are daily rates, we can probably cache for much longer

calellowitz (approved):

commcare_connect/opportunity/migrations/0055_payment_amount_usd.py
calellowitz: I think this is fine. Everything should be valid at this point, and we know what the currencies are. I think one opportunity has a typo in the currency, that we can fix before you deploy, but it would be good to have a hard failure to force us to do so.

calellowitz (approved):

PR #391, opened 2024-09-16
Network manager funnel performance

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-479

Screenshot 2024-09-16 at 16 21 15
Screenshot 2024-09-16 at 16 21 06


commcare_connect/program/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+            VisitValidationStatus.over_limit,
+            VisitValidationStatus.trial,
+        ]
+    )
+
+    earliest_visits = (
+        UserVisit.objects.filter(
+            opportunity_access=OuterRef("opportunityaccess"),
+        )
+        .exclude(status__in=[VisitValidationStatus.over_limit, VisitValidationStatus.trial])
```
calellowitz, line 19: can this use the filter you defined above?

commcare_connect/program/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+        .order_by("visit_date")
+        .values("visit_date")[:1]
```
calellowitz, line 21: nit: You can use `.first()` here instead of the slice

commcare_connect/program/helpers.py
```diff
@@ -0,0 +1,56 @@
...
+    managed_opportunities = (
+        ManagedOpportunity.objects.filter(program=program)
+        .order_by("start_date")
+        .annotate(
+            workers_invited=Count("opportunityaccess"),
+            workers_passing_assessment=Count(
+                "opportunityaccess__assessment",
+                filter=Q(
+                    opportunityaccess__assessment__passed=True,
+                    opportunityaccess__assessment__opportunity=F("opportunityaccess__opportunity"),
```
calellowitz, line 33: is this filter necessary? shouldn't the opportunity access guarantee its the right opportunity already?

calellowitz (commented):
generally looks good. a few questions and one small nit

commcare_connect/program/helpers.py
hemant10yadav: Thanks Removed [4f88e2a](https://github.com/dimagi/commcare-connect/pull/391/commits/4f88e2aa8a84e76736c6954c1e540b7857143581)

Also added some more test.

commcare_connect/program/helpers.py
hemant10yadav: I tried .first() but subquery here expects a queryset.

commcare_connect/program/helpers.py
```diff
@@ -0,0 +1,78 @@
...
+    Value,
+    When,
+)
+from django.db.models.functions import Cast, Round
+
+from commcare_connect.opportunity.models import UserVisit, VisitValidationStatus
+from commcare_connect.program.models import ManagedOpportunity, Program
+
+
+def calculate_safe_percentage(numerator, denominator):
```
sravfeyn, line 21: It's not worth for this to be a function, may be a private function inside the caller would be better

commcare_connect/program/tests/test_helpers.py
```diff
@@ -0,0 +1,139 @@
...
+        assert len(opps) == 1
+        annotated_opp = opps[0]
+        assert annotated_opp.organization.slug == self.nm_org.slug
+        assert annotated_opp.workers_invited == 5
+        assert annotated_opp.workers_passing_assessment == 5
+        assert annotated_opp.workers_starting_delivery == 3
+        assert annotated_opp.percentage_conversion == 60.0
+
+    def test_empty_scenario(self):
+        opps = get_annotated_managed_opportunity(self.program)
```
sravfeyn, line 49: You could consider using parametrize, to combine some of the tests like below (not tested). Allows to be more DRY and concise.

```
    @pytest.mark.parametrize(
        "visit_statuses, passing_assessments, expected_invited, expected_passing, expected_delivery, expected_conversion",
        [
            ([VisitValidationStatus.pending, VisitValidationStatus.pending, VisitValidationStatus.trial], 
             [True, True, True], 5, 5, 3, 60.0),  # Basic scenario
            ([], [], 0, 0, 0, 0.0),  # Empty scenario
            ([VisitValidationStatus.pending] * 3, [True] * 3, 1, 1, 1, 100.0),  # Multiple visits scenario
            ([VisitValidationStatus.over_limit, VisitValidationStatus.trial], 
             [True, True], 2, 2, 0, 0.0),  # Excluded statuses
            ([VisitValidationStatus.pending, VisitValidationStatus.pending], 
             [False, True], 2, 1, 2, 100.0),  # Failed assessments
        ]
    )
    def test_scenarios(self, visit_statuses, passing_assessments, expected_invited, expected_passing, expected_delivery, expected_conversion):
        for i, visit_status in enumerate(visit_statuses):
            self.create_user_with_access(visit_status=visit_status, passed_assessment=passing_assessments[i])

        opps = get_annotated_managed_opportunity(self.program)
        assert len(opps) == 1
        annotated_opp = opps[0]
        assert annotated_opp.workers_invited == expected_invited
        assert annotated_opp.workers_passing_assessment == expected_passing
        assert annotated_opp.workers_starting_delivery == expected_delivery
        assert annotated_opp.percentage_conversion == expected_conversion
```

hemant10yadav: Added [f87aa02](https://github.com/dimagi/commcare-connect/pull/391/commits/f87aa0223bd7fcd6b7a3857c1051e023e876068e)

pxwxnvermx (approved):
LGTM. Thanks 🙌🏽

calellowitz (approved):
Thanks for including tests, looks like those still only cover some of the calculations, so adding tests for the remaining ones would be great but otherwise looks good, and not blocking on that. Would also be good to also have QA validate the calculations if they haven't yet.

pxwxnvermx (approved):

PR #392, opened 2024-09-17
Use gevent workers

calellowitz:
We currently run gunicorn with a single worker and a single thread, which means all requests are handled sequentially. This changes our setup to run two gunicorn processes, each using gevent to service multiple requests concurrently.

### DIFF
--- a/docker/start
+++ b/docker/start
@@ -5,4 +5,4 @@ set -o pipefail
 set -o nounset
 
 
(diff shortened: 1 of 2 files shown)


pxwxnvermx (approved):
Thanks 🙌🏽

hemant10yadav (approved):

PR #397, opened 2024-09-20
change the default landing page content

czue:
I realize we're probably not expecting random people to land on this site, but it still seems better to have something basic in place at https://connect.dimagi.com


pxwxnvermx (approved):
Thanks

PR #405, opened 2024-10-03
Show payment metrics

hemant10yadav:
https://dimagi.atlassian.net/browse/CCCT-493


commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -35,3 +41,16 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
     if compute_payment:
         opportunity_access.payment_accrued = payment_accrued
         opportunity_access.save()
+
+
+def update_work_payment_date(payment: Payment):
+    completed_works = CompletedWork.objects.filter(opportunity_access=payment.opportunity_access)
```
calellowitz, line 47: These need to be ordered by approval date (probably `status_modified_date` is fine), and exclude ones that are already marked paid (otherwise, we will just keep updating the first X).

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -35,3 +41,16 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
     if compute_payment:
         opportunity_access.payment_accrued = payment_accrued
         opportunity_access.save()
+
+
+def update_work_payment_date(payment: Payment):
```
calellowitz, line 46: I don't think this logic is quite right, if the worker was paid for less than a full unit, this won't include the carryover. I think you have to iterate over all payments for the user in this function, not the migration, so that if they earned 6000 at one point, but were paid 5000, the 1000 is added to the next payment amount. It at least needs to be all payments that are not already attributed to a work.

The part that calls this in the upload needs to be able to handle that situation as well, it needs to be able to figure out if there is any carryover from previous payments.

calellowitz (changes requested):
Couple things but generally looks like its on the right track

commcare_connect/opportunity/utils/completed_work.py
hemant10yadav: Thanks for explaining it. [802e47b](https://github.com/dimagi/commcare-connect/pull/405/commits/802e47bfa002fde581ce5067751a2f4e60cc059d)

commcare_connect/opportunity/utils/completed_work.py
hemant10yadav: [802e47b](https://github.com/dimagi/commcare-connect/pull/405/commits/802e47bfa002fde581ce5067751a2f4e60cc059d)

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -35,3 +42,32 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
...
+    completed_works = CompletedWork.objects.filter(opportunity_access=access).order_by("status_modified_date")
+
+    paid = 0
+    works_to_update = []
+    completed_works_iter = iter(completed_works)
+
+    try:
+        current_work = next(completed_works_iter)
+    except StopIteration:
+        return
```
calellowitz, line 58: This could be simplified by just checking `if completed_works`

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -35,3 +42,32 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
...
+
+    for payment in payments:
+        paid += payment.amount
+
+        while paid > current_work.payment_accrued:
```
calellowitz, line 63: This is missing the logic deducts from the total paid when we assign that payment to a completed work. Right now this will run forever in most cases, because the total paid out is always greater than what is due for the next completed work. For example if each work accrues $5 and the first payment was $10, 10 is always greater than 5. We need to deduct what we are assigning to the work, or keep a similar running total for what we owe.

calellowitz (changes requested):
couple bugs, and as usual this would really benefit from tests, since this is pretty complicated work, and it will be difficult to visually determine if it is fully correct

calellowitz (approved):
Since this will be used internally only for now, going to merge this. I imagine there is a way to simplify that algorithm, which is quite difficult to follow with all the breaks and continues, but it looks correct, and this will allow people to start looking at the data

PR #409, opened 2024-10-08
Fix Invoice List Internationalization

pxwxnvermx:
Fixed translate tag not working in the invoice list template due to missing initialization for internationalization tags.


calellowitz (approved):
Do we want this in one of our base templates? Otherwise looks good

PR #412, opened 2024-10-09
adding consistent ordering to visit export

calellowitz:
I think this should fix the flaky test (and make sure that visit exports are always in the same order, which might be nice anyway).


hemant10yadav:
Ordering by pk may still fail because during bulk creation, the database does not guarantee sequential pk assignment based on the order of the objects in our code. This can result in the second record being inserted first and getting a lower pk than the first record. As a result, even if we order by pk, the output may not match the expected order (e.g., by visit_date), causing test to fail.

hemant10yadav (approved):

calellowitz:
Offlined with Hemant, and we are both convinced PK ordering should be sufficient. Previous failures were due to a different issue.

PR #413, opened 2024-10-09
include submission time slot in opportunity API

sravfeyn:
This adds form submission end/start flags into Opportunity API

### DIFF
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -1,3 +1,5 @@
+from functools import lru_cache
+
 from django.conf import settings
 from django.db.models import Sum
 from rest_framework import serializers
@@ -15,6 +17,7 @@ from commcare_connect.opportunity.models import (
     OpportunityAccess,
     OpportunityClaim,
     OpportunityClaimLimit,
+    OpportunityVerificationFlags,
     Payment,
     PaymentUnit,
     UserVisit,
@@ -105,6 +108,8 @@ class OpportunitySerializer(serializers.ModelSerializer):
     payment_units = serializers.SerializerMethodField()
     is_user_suspended = serializers.SerializerMethodField()
     catchment_areas = serializers.SerializerMethodField()
+    start_time_threshold = serializers.SerializerMethodField()
+    end_time_threshold = serializers.SerializerMethodField()
 
     class Meta:
         model = Opportunity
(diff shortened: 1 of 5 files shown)


calellowitz:
This PR needs a description

commcare_connect/opportunity/api/serializers.py
```diff
@@ -178,6 +185,18 @@ def get_catchment_areas(self, obj):
         catchments = CatchmentArea.objects.filter(opportunity_access=opp_access)
         return CatchmentAreaSerializer(catchments, many=True).data
 
+    @lru_cache
+    def _get_flags(self, obj):
+        return OpportunityVerificationFlags.objects.filter(opportunity=obj).first()
```
calellowitz, line 190: should this use `.get` instead or even `obj.opportunityverificationflags`? It's a `OneToOne` model, we can guarantee there is only one so `filter` and `first` seem unnecessary. Given that we can use normal accessor syntax, this whole cached get may be unnecessary, since it will be loaded the first time its accessed and already saved on the object the later times (rather than saved in redis)

commcare_connect/opportunity/api/serializers.py
```diff
@@ -105,6 +108,8 @@ class OpportunitySerializer(serializers.ModelSerializer):
     payment_units = serializers.SerializerMethodField()
     is_user_suspended = serializers.SerializerMethodField()
     catchment_areas = serializers.SerializerMethodField()
+    start_time_threshold = serializers.SerializerMethodField()
```
calellowitz, line 111: As an extension of my comment below, I am not sure this needs a method rather than just the `source` argument. Alternatively, I think this might be nicer in a new `verification_flags` object, rather than at the top level, since we are likely to add the others and it will help with organization.

commcare_connect/opportunity/api/serializers.py
```diff
@@ -133,6 +138,8 @@ class Meta:
             "payment_units",
             "is_user_suspended",
             "catchment_areas",
+            "start_time_threshold",
```
calellowitz, line 141: Is there a specific reason to change the names here? It obviously doesn't really matter but it does add a tiny bit of inconsistency between what the mobile and web team will call these which can be annoying.

calellowitz (commented):
Few questions, beyond the request for a description

commcare_connect/opportunity/api/serializers.py
sravfeyn: Good point, I totally overlooked it. I have updated the code and also addressed the rest of the comments too.

calellowitz (approved):

PR #415, opened 2024-10-16
Fix for historical model issue in opportunity 0060 migration.

hemant10yadav:
This PR updates the `update_work_payment_date` function to dynamically assign model references `(Payment and CompletedWork)` within the function. This change ensures compatibility with Django's migration system, where historical models must be used.

By avoiding top-level imports, the function can now:

Use historical models when called from migrations (via `apps.get_model()`).
Fallback to current models when called during normal app execution.
This approach prevents issues caused by schema changes between migration and runtime models, ensuring smoother migrations and consistent behaviour.

For reference- [Historical-models](https://docs.djangoproject.com/en/5.1/topics/migrations/#historical-models)

### DIFF
--- a/commcare_connect/opportunity/utils/completed_work.py
+++ b/commcare_connect/opportunity/utils/completed_work.py
@@ -44,9 +44,20 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
         opportunity_access.save()
 
 
-def update_work_payment_date(access: OpportunityAccess):
-    payments = Payment.objects.filter(opportunity_access=access).order_by("date_paid")
-    completed_works = CompletedWork.objects.filter(opportunity_access=access).order_by("status_modified_date")
+def update_work_payment_date(access: OpportunityAccess, payment_model_ref=None, completed_work_model_ref=None):
+    """
+    Import models dynamically within the function helps us avoid issues with historical models during migrations.
+    """
+    if not payment_model_ref:
+        payment_model_ref = Payment
+
+    if not completed_work_model_ref:
+        completed_work_model_ref = CompletedWork
+
+    payments = payment_model_ref.objects.filter(opportunity_access=access).order_by("date_paid")
+    completed_works = completed_work_model_ref.objects.filter(opportunity_access=access).order_by(
+        "status_modified_date"
+    )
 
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -44,9 +44,22 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
         opportunity_access.save()
 
 
-def update_work_payment_date(access: OpportunityAccess):
-    payments = Payment.objects.filter(opportunity_access=access).order_by("date_paid")
-    completed_works = CompletedWork.objects.filter(opportunity_access=access).order_by("status_modified_date")
+def update_work_payment_date(access: OpportunityAccess, payment_model_ref=None, completed_work_model_ref=None):
```
sravfeyn, line 47: For future note, utils like these can stay inside the migration file itself, if they aren't used anywhere else.

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -44,9 +44,22 @@ def update_status(completed_works, opportunity_access, compute_payment=True):
...
+    """
+    Dynamically assign models to avoid issues with historical models during migrations.
+    Top-level imports use the current model, which may not match the schema at migration
+    time. This ensures we use historical models during migrations and current models in normal execution.
+    """
+    if not payment_model_ref:
+        payment_model_ref = Payment
```
sravfeyn, line 54:
```suggestion
    payment_model_ref = payment_model_ref or Payment
```

commcare_connect/opportunity/utils/completed_work.py
sravfeyn, line 47:
```suggestion
def update_work_payment_date(access: OpportunityAccess, payment_model=None, completed_work_model=None):
```

hemant10yadav: Thanks! added

commcare_connect/opportunity/utils/completed_work.py
hemant10yadav: Thanks! added

calellowitz (approved):

PR #416, opened 2024-10-24
Update Completed Work status and Payment in form Processor

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-512)

This PR adds code to update the completed work status and payment accrued (through the `update_payment_accrued` function) for a user when a form is submitted.

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -1,5 +1,6 @@
 import datetime
 
+from django.core.cache import cache
 from django.db.models import Count, Q
 from django.utils.timezone import now
 from geopy.distance import distance
@@ -29,6 +30,7 @@ from commcare_connect.opportunity.models import (
     VisitValidationStatus,
 )
 from commcare_connect.opportunity.tasks import download_user_visit_attachments
+from commcare_connect.opportunity.visit_import import update_payment_accrued
 from commcare_connect.users.models import User
 
 LEARN_MODULE_JSONPATH = parse("$..module")
@@ -309,6 +311,8 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
             completed_work_needs_save = True
         if completed_work_needs_save:
             completed_work.save()
+    with cache.lock(access.id):
+        update_payment_accrued(opportunity, [access.user_id])
     download_user_visit_attachments.delay(user_visit.id)
 
 
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -303,15 +303,12 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         user_visit.status = VisitValidationStatus.approved
         user_visit.review_status = VisitReviewStatus.agree
     user_visit.save()
-    if (
-        completed_work is not None
-        and completed_work.completed_count > 0
-        and completed_work.status == CompletedWorkStatus.incomplete
-    ):
-        completed_work.status = CompletedWorkStatus.pending
-        completed_work_needs_save = True
-    if completed_work_needs_save:
-        completed_work.save()
+    if completed_work is not None:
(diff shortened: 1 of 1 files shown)


commcare_connect/form_receiver/processor.py
```diff
@@ -303,15 +305,14 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
-        completed_work_needs_save = True
-    if completed_work_needs_save:
-        completed_work.save()
+    if completed_work is not None:
+        if completed_work.completed_count > 0 and completed_work.status == CompletedWorkStatus.incomplete:
+            completed_work.status = CompletedWorkStatus.pending
+            completed_work_needs_save = True
+        if completed_work_needs_save:
+            completed_work.save()
+    with cache.lock(access.id):
```
sravfeyn, line 314: Should the key be more specific than just the ID? For e.g. if we have unrelated locking that also uses `ids` there is possibility for these two have same key.

commcare_connect/form_receiver/tests/test_receiver_integration.py
```diff
@@ -366,17 +369,10 @@ def test_auto_approve_payments_rejected_visit(
     rejected_reason.append(visit.reason)
     visit.save()
 
-    duplicate_json = deepcopy(form_json)
-    duplicate_json["id"] = str(uuid4())
-    make_request(api_client, duplicate_json, user_with_connectid_link)
-    visit = UserVisit.objects.get(xform_id=duplicate_json["id"])
-    visit.status = VisitValidationStatus.rejected
-    visit.reason = "duplicate"
```
sravfeyn, line 374: Is this not the behaviour any more?

pxwxnvermx: These test checks were causing tests to fail with these new additions. Adding duplicate visits causes completed works to be marked as duplicates. So checking for duplicates does not make sense in tests testing rejected behavior.

commcare_connect/opportunity/visit_import.py
```diff
@@ -168,10 +169,11 @@ def update_payment_accrued(opportunity: Opportunity, users):
     """Updates payment accrued for completed and approved CompletedWork instances."""
     access_objects = OpportunityAccess.objects.filter(user__in=users, opportunity=opportunity, suspended=False)
     for access in access_objects:
-        completed_works = access.completedwork_set.exclude(
-            status__in=[CompletedWorkStatus.rejected, CompletedWorkStatus.over_limit]
-        ).select_related("payment_unit")
-        update_status(completed_works, access, True)
+        with cache.lock(f"update_payment_accrued_lock_{access.id}"):
```
calellowitz, line 172: how does this handle failure? when does the lock timeout to avoid deadlocks, and if multiple forms come in quick succession will this block the for receiver view for an extended period?

commcare_connect/form_receiver/processor.py
pxwxnvermx: Fixed. [Add lock in update_payment_accrued](https://github.com/dimagi/commcare-connect/pull/416/commits/11d8e4ef9e94252e565bc0e9e6c992081ce01172)

commcare_connect/opportunity/visit_import.py
pxwxnvermx: This cache lock has blocking behavior, so if requests come in quick succession they will be blocked until the code inside the lock has completed executing. The timeouts on these locks are infinite by default i.e. they wait for the execution of code to complete before releasing the lock.

calellowitz: We should add a timeout to avoid deadlock. Longer than the work will actually take, but that way if there is an unhandled exception or unexpected failure, it wont block all other work forever

pxwxnvermx: Fixed. [d33b0af](https://github.com/dimagi/commcare-connect/pull/416/commits/d33b0af2d8a452dab6068aad6a37da20883a7acb)

calellowitz (approved):

PR #421, opened 2024-10-30
Payment Unit: Start/End Dates

pxwxnvermx:
[Ticket](https://dimagi.atlassian.net/browse/CCCT-522)

image


calellowitz (approved):

sravfeyn (approved):
Code looks good. But aren't there any larger implications when we introduce start/end dates for payment units like when claiming or when counting user progress etc?

pxwxnvermx:
@sravfeyn We let the users claim the full payment unit amount and max totals. Most opportunities are planned around a fixed number of users, so partial claiming based on the difference between the claiming date and the end date can be added later if a need comes up for that. The current spec needed start and end dates for payment units only.

hemant10yadav (approved):

commcare_connect/opportunity/forms.py
```diff
@@ -683,6 +692,8 @@ def clean(self):
                     "optional_deliver_units",
                     error=f"{deliver_unit_obj.name} cannot be marked both Required and Optional",
                 )
+            if cleaned_data["end_date"] < now().date():
```
hemant10yadav, line 695: If it's optional it will throw error. I recently faced this issue on staging.
I am guessing this is the same error [Sentry error](https://dimagi.sentry.io/issues/6077952930/?environment=staging&project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=90d&stream_index=5)

pxwxnvermx: Fixed. [fe18fc6](https://github.com/dimagi/commcare-connect/pull/421/commits/fe18fc6fd656d3f303f7bb4919a4461d4d68f9e2)

calellowitz (approved):

PR #422, opened 2024-10-31
Program Dashboard Updates

czue:
[Screencast from 31-10-2024 14:52:14.webm](https://github.com/user-attachments/assets/7c4d29a0-7a34-475f-863e-e8dcb543e0ca)

Major changes:

- Added filters
- Added top level indicators
- Moved query from SQL to the ORM

This is still very much in experimental mode, but hopefully this will help with loading issues + showing more realistic pictures.

### DIFF
--- a/commcare_connect/reports/queries.py
+++ b/commcare_connect/reports/queries.py
@@ -1,0 +1,58 @@
+from django.db.models import Case, DateTimeField, ExpressionWrapper, F, FloatField, Value, When
+from django.db.models.fields.json import KeyTextTransform
+from django.db.models.functions import Cast, Extract
+
+from commcare_connect.opportunity.models import UserVisit
+
+
+def get_visit_map_queryset_base():
+    return (
+        UserVisit.objects.annotate(
+            username_connectid=F("form_json__metadata__username"),
+            deliver_unit_name=F("deliver_unit__name"),
+            days_since_opp_start=ExpressionWrapper(
+                Extract(F("visit_date") - F("opportunity__start_date"), "day"), output_field=FloatField()
+            ),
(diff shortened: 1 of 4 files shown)


commcare_connect/reports/queries.py
```diff
@@ -0,0 +1,44 @@
...
+
+def get_visit_map_queryset(base_queryset):
+    return (
+        base_queryset.annotate(
+            username_connectid=F("form_json__metadata__username"),
+            deliver_unit_name=F("deliver_unit__name"),
+            days_since_opp_start=ExpressionWrapper(
+                Extract(F("visit_date") - F("opportunity__start_date"), "day"), output_field=FloatField()
+            ),
+            timestart_str=KeyTextTransform("timeStart", KeyTextTransform("metadata", F("form_json"))),
```
calellowitz, line 14: Can these use (https://docs.djangoproject.com/en/5.1/topics/db/queries/#module-django.db.models.fields.json
0)[KT] and not have to nest? Something like `KT("form_json__timeStart")`? In general curious why some of these need nested transforms and some have `F`, which is likely just my lack of familiarity with json field annotations

calellowitz (commented):
One question about the different ways this queries nested json data but otherwise looks good.

commcare_connect/reports/queries.py
czue: > In general curious why some of these need nested transforms and some have F, which is likely just my lack of familiarity with json field annotations

I mean, the honest answer is because that's what Claude decided to do. But I can dig into this further for consistency if you think it matters. I also don't have any experience with any of this

PR #426, opened 2024-11-07
Currency conversion table

hemant10yadav:
[CCCT-539](https://dimagi.atlassian.net/browse/CCCT-539)

[CCCT-539]: https://dimagi.atlassian.net/browse/CCCT-539?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +285,40 @@ def get_exchange_rate(currency_code, date=None):
...
     else:
         url = f"{base_url}/latest.json"
     url = f"{url}?app_id={settings.OPEN_EXCHANGE_RATES_API_ID}"
     rates = json.load(urllib.request.urlopen(url))
-    return rates["rates"].get(currency_code)
+
+    rate = rates["rates"].get(currency_code)
+
+    if rate:
+        rate_date = date if date else now().date()
```
sravfeyn, line 304: You can set `auto_now_add` to do this automatically.

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +285,40 @@ def get_exchange_rate(currency_code, date=None):
...
+
+    return rate
+
+
+def get_exchange_rate_from_db(currency_code, date=None):
+    if currency_code == "USD":
+        return 1
+
+    if not date:
+        date = now().date()
```
sravfeyn, line 315:
```suggestion
    date = date or now().date()
```

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +285,40 @@ def get_exchange_rate(currency_code, date=None):
...
+    rate = rates["rates"].get(currency_code)
+
+    if rate:
+        rate_date = date if date else now().date()
+        ExchangeRate.objects.create(currency_code=currency_code, rate=rate, rate_date=rate_date)
+
+    return rate
+
+
+def get_exchange_rate_from_db(currency_code, date=None):
```
sravfeyn, line 310: You can make this a class method on ExchangeRate

commcare_connect/opportunity/models.py
```diff
@@ -685,3 +685,10 @@ class CatchmentArea(models.Model):
 
     class Meta:
         unique_together = ("site_code", "opportunity")
+
+
+class ExchangeRate(models.Model):
+    currency_code = models.CharField(max_length=3)
+    rate = models.DecimalField(max_digits=10, decimal_places=6)
+    rate_date = models.DateField()
+    fetched_at = models.DateTimeField(auto_now_add=True)
```
sravfeyn, line 694: What's the use of this date?

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +285,40 @@ def get_exchange_rate(currency_code, date=None):
     if currency_code == "USD":
         return 1
```
sravfeyn, line 286: These two lines can be removed since `get_exchange_rate_from_db` does the same.

commcare_connect/opportunity/visit_import.py
```diff
@@ -277,11 +277,6 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
     return PaymentImportStatus(seen_users, missing_users)
 
 
-def _cache_key(currency_code, date=None):
-    return [currency_code, date.toordinal() if date else None]
-
-
-@quickcache(vary_on=_cache_key, timeout=12 * 60 * 60)
```
sravfeyn, line 284: One thing is, in this version, the API will be hit once, but in the modified version you are going to hit the same API separately for each currency_code. You might consider pulling the API call out into separate function and cache that.

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +285,40 @@ def get_exchange_rate(currency_code, date=None):
     if currency_code == "USD":
         return 1
 
+    rate = get_exchange_rate_from_db(currency_code, date)
```
sravfeyn, line 288: It may not be worth to make this a separate method as it's just a queryset call.

commcare_connect/opportunity/migrations/0061_exchangerate.py
```diff
@@ -0,0 +1,52 @@
...
+from commcare_connect.opportunity.visit_import import get_exchange_rate
+
+logger = logging.getLogger(__name__)
+
+
+def update_exchange_rate(apps, schema_editor):
+    Payment = apps.get_model("opportunity.Payment")
+    payments = (
+        Payment.objects.annotate(date_only=TruncDate("date_paid"))
+        .values("id", "date_only", "amount", "opportunity_access__opportunity__currency")
```
calellowitz, line 16: Payments without opportunity access (to program managers) will need to be accounted for as well, but wont have an opportunity access

commcare_connect/opportunity/migrations/0061_exchangerate.py
```diff
@@ -0,0 +1,52 @@
...
+        .distinct()
+    )
+
+    for payment in payments:
+        date_paid = payment["date_only"]
+        currency = payment["opportunity_access__opportunity__currency"]
+
+        if currency in ["USD", None]:
```
calellowitz, line 24: might be nicer not convert if currency is `None` rather than assume USD

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +305,24 @@ def get_exchange_rate(currency_code, date=None):
...
-    if date:
-        url = f"{base_url}/historical/{date.strftime('%Y-%m-%d')}.json"
-    else:
-        url = f"{base_url}/latest.json"
-    url = f"{url}?app_id={settings.OPEN_EXCHANGE_RATES_API_ID}"
-    rates = json.load(urllib.request.urlopen(url))
-    return rates["rates"].get(currency_code)
+    today = now().date()
+    rate = (
+        ExchangeRate.objects.filter(currency_code=currency_code, rate_date=date or today)
```
calellowitz, line 310: nit, `date or today` would be a little cleaner outside the arguments (like the line above)

commcare_connect/opportunity/visit_import.py
```diff
@@ -290,14 +305,24 @@ def get_exchange_rate(currency_code, date=None):
...
+        .values_list("rate", flat=True)
+        .first()
+    )
```
calellowitz, line 313: I think this can be a `get` and should catch the `DoesNotExist` exception

commcare_connect/opportunity/models.py
```diff
@@ -697,3 +697,10 @@ class CatchmentArea(models.Model):
 
     class Meta:
         unique_together = ("site_code", "opportunity")
+
+
+class ExchangeRate(models.Model):
```
calellowitz, line 702: it would be nice to have a `code, rate_date` uniqueness constraint so we only ahve one rate per currency per date

calellowitz (commented):
Generally looks good, left a few comments

commcare_connect/opportunity/migrations/0061_exchangerate.py
hemant10yadav: fixed - [63bc638](https://github.com/dimagi/commcare-connect/pull/426/commits/63bc6386f07df7e3e29c2989a6870cf511f32d15))

commcare_connect/opportunity/migrations/0065_exchangerate_exchangerate_unique_currency_code_date.py
```diff
@@ -11,17 +11,19 @@
 
 def update_exchange_rate(apps, schema_editor):
     Payment = apps.get_model("opportunity.Payment")
+
     payments = (
         Payment.objects.annotate(date_only=TruncDate("date_paid"))
-        .values("id", "date_only", "amount", "opportunity_access__opportunity__currency")
+        .filter(payment_unit__opportunity__currency__isnull=False)
```
calellowitz, line 17: do `Payments` have a `payment_unit`? I didn't think they were tied to one like that but to a user or an org?

calellowitz (commented):
One question on the migration but otherwise looks good

commcare_connect/opportunity/migrations/0065_exchangerate_exchangerate_unique_currency_code_date.py
hemant10yadav: Yes `Payments` have a `payment_unit`.- [Payment model](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/models.py#L423)

calellowitz: But that value is never used right? It will always be null, since it isn't set on payment (https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/visit_import.py#L275)[import] or (https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/views.py#L1225)[invoice payment]

calellowitz: I think we need to use the program for managed opps or the opportunity from the opportunity access as you were doing before.

calellowitz (approved):
approved if the query in the migration is tested

commcare_connect/opportunity/migrations/0066_exchangerate_exchangerate_unique_currency_code_date.py
```diff
@@ -0,0 +1,61 @@
...
+
+    for payment in payments:
+        date_paid = payment["date_only"]
+        currency = payment["opportunity_access__opportunity__currency"] or payment["invoice__opportunity__currency"]
+
+        if currency is "USD":
+            exchange_rate = 1
+        else:
+            exchange_rate = get_exchange_rate(currency, date_paid)
+            logger.info(
```
calellowitz, line 31: This logger is misleading right? This never actually changes the values? Or sets them when they are missing?

pxwxnvermx (approved):

commcare_connect/opportunity/migrations/0066_exchangerate_exchangerate_unique_currency_code_date.py
```diff
@@ -12,31 +12,46 @@
...
-            "id", "date_only", "amount", "opportunity_access__opportunity__currency", "invoice__opportunity__currency"
-        )
-        .distinct()
+    payments = Payment.objects.annotate(date_only=TruncDate("date_paid")).values(
+        "id",
+        "date_only",
+        "amount",
+        "amount_usd",
+        "opportunity_access__opportunity__currency",
+        "invoice__opportunity__currency",
```
calellowitz, line 21: why is distinct no longer necessary?

commcare_connect/opportunity/migrations/0066_exchangerate_exchangerate_unique_currency_code_date.py
```diff
@@ -12,31 +12,46 @@
...
             raise Exception(f"Invalid currency code {currency}")
 
+        old_usd = payment["amount_usd"]
+        new_usd = payment["amount"] / exchange_rate
+
+        # Log discrepancies if old_usd exists and is different from new_usd
+        if old_usd and old_usd != new_usd:
+            logger.info(f"Payment ID: {payment['id']}, original USD: {old_usd}, USD acc. to new rate: {new_usd}")
+
+        # Update only for org payments
```
calellowitz, line 48: why not update other payments if they are missing a conversion? There shouldn't be any, but it does seem worth doing if there are.

calellowitz (commented):
pending those two questions, looks good

commcare_connect/opportunity/migrations/0066_exchangerate_exchangerate_unique_currency_code_date.py
hemant10yadav: Added [22b86ed](https://github.com/dimagi/commcare-connect/pull/426/commits/22b86ed777ac9b7e90585498c96fb3454f69dc74)

calellowitz (approved):

pxwxnvermx (approved):

PR #429, opened 2024-11-12
admin dashboard updates

czue:
New Screenshot:

image

Changes:

1. **Change the current clusters to donuts, which show the approved/pending/rejected breakdown of the visits in that cluster.** This is mostly just changing from the previous cluster example to this one: https://docs.mapbox.com/mapbox-gl-js/example/cluster-html/, and then getting it working with our setup. 
2. **Add a loading state to the map.** And make a few other minor tweaks.
3. **Add some charts.** You can view that set of changes here if you want https://github.com/dimagi/commcare-connect/pull/431
4. **Migrate most of the JavaScript functions to a new external JS file.**

(recommend reviewing all at once)


czue:
I don't think the build failure is related, though let me know if I should look into it

commcare_connect/reports/views.py
```diff
@@ -419,3 +418,109 @@ def dashboard_stats_api(request):
             "percent_verified": f"{percent_verified:.1f}%",
         }
     )
+
+
+@login_required
+@user_passes_test(lambda u: u.is_superuser)
+def dashboard_charts_api(request):
+    filterset = DashboardFilters(request.GET)
+    queryset = UserVisit.objects.all()
```
calellowitz, line 427: How confident are you that you want `UserVisit` rather than `CompletedWork` (user visits can be combined to form a single completed work, which is the unit of payment). I think we typically view Completed Works as a unit of delivery instead of the form, since it can account for things a visit that includes 2 forms filled out at the same time that together mark a delivery.

calellowitz (approved):
I admittedly didn't look too closely at the JS but the rest looks fine. One question

calellowitz (changes requested):
Didn't mean to approve and I think "requesting changes" is the only way I can get back to a not approved state, but depending on the answer to the question will approve again.

commcare_connect/reports/views.py
czue: It's all using the same base queryset which I got from the superset dashboard's SQL, so I think it's right? Or at least, it's consistent with the superset view and itself.

But can definitely revisit this logic once the dashboard has been vetted by the team more.

calellowitz (approved):

PR #433, opened 2024-11-14
 Deadlock Detected in bulk update payments.

hemant10yadav:
[CCCT-549](https://dimagi.atlassian.net/browse/CCCT-549)

[Sentry link](https://dimagi.sentry.io/issues/6059769484/?environment=production&project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=7d&stream_index=11)

- The deadlock error indicates that two separate transactions were trying to access the same data in conflicting orders, even though the code is using a single transaction.atomic() block.
- This suggests there are concurrency issues beyond just the single transaction, likely due to the fact that this is a Celery task.
- The` transaction.atomic()` block only ensures atomicity within a single Celery task, but does not prevent deadlocks between multiple concurrent tasks.

To resolve this, I have used cache lock so that only one bulk update payment is allowed at a time.

[CCCT-549]: https://dimagi.atlassian.net/browse/CCCT-549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/visit_import.py
+++ b/commcare_connect/opportunity/visit_import.py
@@ -271,10 +271,16 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
             payment = Payment.objects.create(opportunity_access=access, amount=amount, amount_usd=amount_usd)
             seen_users.add(username)
             payment_ids.append(payment.pk)
-            update_work_payment_date(access)
+        process_work_payment_dates(users)
     missing_users = set(usernames) - seen_users
     send_payment_notification.delay(opportunity.id, payment_ids)
     return PaymentImportStatus(seen_users, missing_users)
+
+
+def process_work_payment_dates(accesses: list):
+    with transaction.atomic():
+        for oa in accesses:
+            update_work_payment_date(oa)
 
 
 def _cache_key(currency_code, date=None):
--- a/commcare_connect/opportunity/visit_import.py
+++ b/commcare_connect/opportunity/visit_import.py
@@ -6,6 +6,7 @@ from dataclasses import astuple, dataclass
 from decimal import Decimal, InvalidOperation
 
 from django.conf import settings
+from django.core.cache import cache
 from django.core.files.uploadedfile import UploadedFile
 from django.db import transaction
 from django.utils.timezone import now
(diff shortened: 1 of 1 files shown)


calellowitz:
How does this address the issue? If wrapping the whole thing in a transaction does not prevent locks, why would a second nested transaction?

commcare_connect/opportunity/visit_import.py
```diff
@@ -271,12 +271,18 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
             payment = Payment.objects.create(opportunity_access=access, amount=amount, amount_usd=amount_usd)
             seen_users.add(username)
             payment_ids.append(payment.pk)
-            update_work_payment_date(access)
+        process_work_payment_dates(users)
```
pxwxnvermx, line 274: Moving this function outside the transaction block would be a better solution to this problem. Reducing the amount of work done in a transaction over multiple tables will reduce the chances of a deadlock. Also `update_work_payment_date` calls a [`bulk_update/bulk_create`](https://github.com/django/django/blob/main/django/db/models/query.py#L911) function that also uses a transaction and would fail if it tries to access an object from the transaction before.

commcare_connect/opportunity/visit_import.py
```diff
@@ -259,19 +260,21 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
...
-        for access in users:
-            username = access.user.username
-            amount = payments[username]
-            amount_usd = amount / exchange_rate
-            payment = Payment.objects.create(opportunity_access=access, amount=amount, amount_usd=amount_usd)
-            seen_users.add(username)
-            payment_ids.append(payment.pk)
-            update_work_payment_date(access)
+    lock_key = f"bulk_update_payments_opportunity_{opportunity.id}"
+    with cache.lock(lock_key, timeout=300):
```
calellowitz, line 264: How confident are you this timeout is long enough? I would have guessed some of the really big ones, might exceed this, or at least come close.

calellowitz (approved):
Pending the answer to my question on lock timeout, looks good

commcare_connect/opportunity/visit_import.py
hemant10yadav: I recently faced this. I think we can aim for around 10 minutes that should be sufficient, I guess.

pxwxnvermx (approved):

PR #436, opened 2024-11-19
UserMessageStatus build method error

pxwxnvermx:
This fixes the UserMessageStatus.build method error when supplied with the `error` parameter.

[Error Link](https://dimagi.sentry.io/issues/5799897740/?environment=production&project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=4&utc=true)

### DIFF
--- a/commcare_connect/connect_id_client/models.py
+++ b/commcare_connect/connect_id_client/models.py
@@ -35,7 +35,7 @@ class UserMessageStatus:
 
     @classmethod
     def build(cls, username: str, status: str, *args, **kwargs):
-        error = kwargs.pop("error")
+        error = kwargs.get("error")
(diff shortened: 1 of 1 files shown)


pxwxnvermx:
@calellowitz This fix currently lets the UserMessageStatus.build accept any extra params passed through the `*args` and `**kwargs`. It does not do anything with the error param. Do you have any thoughts for what could be done with the error that is passed from ConnectID to Connect through the error param?

calellowitz:
Do you know why it is erroring? That may be more important than what we do with the error message.

PR #443, opened 2024-11-29
Check for None value

sravfeyn:
Should fix https://dimagi.sentry.io/issues/6098769063/?project=4505635339829248&referrer=slack [introduced here](https://github.com/dimagi/commcare-connect/pull/421/files)

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -262,7 +262,8 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     )
     completed_work_needs_save = False
     today = datetime.date.today()
-    if opportunity.start_date > today or (payment_unit.start_date and payment_unit.start_date > today):
+    paymentunit_startdate = payment_unit.start_date if payment_unit else None
+    if opportunity.start_date > today or (paymentunit_startdate and paymentunit_startdate > today):
         completed_work = None
         user_visit.status = VisitValidationStatus.trial
     else:


hemant10yadav (approved):

commcare_connect/form_receiver/processor.py
```diff
@@ -262,7 +262,8 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     )
     completed_work_needs_save = False
     today = datetime.date.today()
-    if opportunity.start_date > today or (payment_unit.start_date and payment_unit.start_date > today):
+    paymentunit_startdate = payment_unit.start_date if payment_unit else None
+    if opportunity.start_date > today or (paymentunit_startdate and paymentunit_startdate > today):
```
hemant10yadav, line 266: The issue arises when `payment_unit` is `null`. In cases where payment_unit is null and `opportunity.start_date > today` evaluates to `false`, the code proceeds to the else block. However, since `payment_unit` is required for `completed_work`, this results in an error.

sravfeyn: You are right, this seems to be a seperate issue for a while https://dimagi.sentry.io/issues/6098769063/?environment=production&project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=14d&stream_index=0

But let’s merge at least this fix, because that affects even where paymentunit is not NULL

sravfeyn: https://dimagi.atlassian.net/browse/CCCT-576

PR #444, opened 2024-12-04
Ignore the visit that comes again from hq with same form data

hemant10yadav:


calellowitz (commented):
This might be nicer as a database level constraint, and then we can catch that db exception in the processor code and respond with an appropriate message. That saves us the extra db hit (its normally just a write that succeeds, and occasionally a write that fails), and enforces the lack of duplicates at the db level, rather than needing to catch it in code each place.

At a minimum we can ensure `(xform_id, entity_id, deliver_unit)` are unique together.

PR #446, opened 2024-12-05
Command to delete duplicate visits

hemant10yadav:
[CCCT-571](https://dimagi.atlassian.net/browse/CCCT-571)
[CCCT-572](https://dimagi.atlassian.net/browse/CCCT-572)

Added a command to delete duplicate visits that were created when the same forms were re-forwarded from HQ.

[CCCT-571]: https://dimagi.atlassian.net/browse/CCCT-571?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-572]: https://dimagi.atlassian.net/browse/CCCT-572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/management/commands/delete_duplicate_visits.py
+++ b/commcare_connect/opportunity/management/commands/delete_duplicate_visits.py
@@ -1,0 +1,62 @@
+from django.core.management import BaseCommand
+from django.db import transaction
+from django.db.models import Count
+
+from commcare_connect.opportunity.models import UserVisit
+
+
+class Command(BaseCommand):
+    help = "Clean up duplicate visits."
+
+    def add_arguments(self, parser, *args, **kwargs):
+        parser.add_argument("--opp", type=int, help="Opportunity ID to clean up duplicate visits.")
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="If set, just print the visits that would be deleted without actually deleting them.",
+        )
+
+    def handle(self, *args, **options):
+        opportunity_id = options.get("opp")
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/management/commands/delete_duplicate_visits.py
```diff
@@ -0,0 +1,62 @@
...
+from commcare_connect.opportunity.models import UserVisit
+
+
+class Command(BaseCommand):
+    help = "Clean up duplicate visits."
+
+    def add_arguments(self, parser, *args, **kwargs):
+        parser.add_argument("--opp", type=int, help="Opportunity ID to clean up duplicate visits.")
+        parser.add_argument(
+            "--dry-run",
```
calellowitz, line 14: nit: dry run should probably be the default for safety reasons, and the added arg should be what causes it to delete (then you can't accidentally delete)

calellowitz (approved):
Approving so I can run, but left one comment. Thanks for including tests

commcare_connect/opportunity/management/commands/delete_duplicate_visits.py
calellowitz: It would probably also be nice to add a confirmation prompt when deleting

PR #449, opened 2024-12-06
capture user/org tags in sentry

sravfeyn:
This should make it easy to trace 500s when users report issues.

### DIFF
--- a/commcare_connect/utils/middleware.py
+++ b/commcare_connect/utils/middleware.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.contrib import messages
+from django.core.exceptions import MiddlewareNotUsed
 from django.http import HttpResponseRedirect
 from django.utils.safestring import mark_safe
 from rest_framework.settings import api_settings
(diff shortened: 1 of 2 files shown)


sravfeyn:
@calellowitz

pxwxnvermx (approved):

commcare_connect/utils/middleware.py
```diff
@@ -40,3 +42,29 @@ def __call__(self, request):
...
+        except ImportError:
+            raise MiddlewareNotUsed
+
+        if not getattr(settings, "SENTRY_DSN", None):
+            raise MiddlewareNotUsed
+
+    def process_view(self, request, view_func, view_args, view_kwargs):
+        from sentry_sdk import configure_scope
+
+        with configure_scope() as scope:
```
calellowitz, line 65: Why use this pattern instead of the one shown in [the docs](https://docs.sentry.io/platforms/python/enriching-events/scopes/) using `get_current_scope`? Is there a difference between them. I can't actually find this one in the docs.

calellowitz: On further look, it looks like this function is deprecated https://docs.sentry.io/platforms/python/migration/1.x-to-2.x#custom-instrumentation

sravfeyn: I have updated it to use set_user and set_tag. (Previously I adapted this from HQ which was using 2.8 )

sravfeyn:
@calellowitz Does this look good now?

pxwxnvermx (approved):

calellowitz (approved):

PR #450, opened 2024-12-09
Superuser page for all opportunities.

hemant10yadav:
[CCCT-569](https://dimagi.atlassian.net/browse/CCCT-569)

Screenshot 2024-12-09 at 16 33 58
Screenshot 2024-12-09 at 16 33 46

[CCCT-569]: https://dimagi.atlassian.net/browse/CCCT-569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -74,7 +74,6 @@ from commcare_connect.opportunity.tables import (
     DeliverStatusTable,
     LearnStatusTable,
     OpportunityPaymentTable,
-    OpportunityTable,
     PaymentInvoiceTable,
     PaymentReportTable,
     PaymentUnitTable,
@@ -1231,11 +1230,10 @@ def invoice_approve(request, org_slug, pk):
     return HttpResponse(headers={"HX-Trigger": "newInvoice"})
 
 
-class AllOpportunitiesView(SuperUserMixin, SingleTableView):
+class AllOpportunitiesView(SuperUserMixin, ListView):
     model = Opportunity
     paginate_by = 10
     template_name = "opportunity/all_opportunities_view.html"
-    table_class = OpportunityTable
 
     def get_queryset(self):
         search_query = self.request.GET.get("search", "")
@@ -1269,7 +1267,7 @@ class AllOpportunitiesView(SuperUserMixin, SingleTableView):
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
-        context["table"] = self.table_class(self.get_queryset())
+        context["table"] = self.get_queryset()
         return context
 
     def render_to_response(self, context, **response_kwargs):
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -12,6 +12,7 @@ from django.db.models import F, Q, Sum
 from django.forms import modelformset_factory
 from django.http import FileResponse, Http404, HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
+from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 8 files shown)


commcare_connect/templates/pagination.html
```diff
@@ -16,21 +16,67 @@
         </li>
       {% endif %}
 
-      {% for page_number in page_obj.paginator.page_range %}
```
hemant10yadav, line 19: I had to make this change because the previous pagination template displayed all the page numbers, and as the number of pages increased, it started overflowing.

commcare_connect/templates/opportunity/partial_opportunity_table.html
```diff
@@ -0,0 +1,74 @@
...
+        {% endif %}
+      </td>
+      <td>{% if opportunity.managed %} {{ opportunity.managedopportunity.program.name }} {% else %} - {% endif %}</td>
+      <td width="300">
+        <div>
+          <a class="btn btn-primary btn-sm"
+             href="{% url 'opportunity:detail' org_slug=request.org.slug pk=opportunity.id %}">
+            <span class="bi bi-eye"></span><span class="d-none d-md-inline">&nbsp;View</span>
+          </a>
+          {% if request.org_membership.is_viewer %}
```
pxwxnvermx, line 40: The "Edit" and "Add Budget" buttons can be removed from this view.

commcare_connect/opportunity/views.py
```diff
@@ -1222,3 +1228,50 @@ def invoice_approve(request, org_slug, pk):
         )
         payment.save()
     return HttpResponse(headers={"HX-Trigger": "newInvoice"})
+
+
+class AllOpportunitiesView(SuperUserMixin, ListView):
```
pxwxnvermx, line 1233: We can build the list using the `django-tables` and `django-filter` libraries. It would make it easier to create a Table of Opportunities with filters.

pxwxnvermx (changes requested):

commcare_connect/opportunity/views.py
sravfeyn: Yeah, this could be written with much less code (most of the HTML for filters and the table will be autogenerated, so you don't have to write it) if you go down that route. I have just made a [PR](https://github.com/dimagi/commcare-connect/pull/466) that makes it more easier.

Take a look at [DeliveryStatsReport](https://github.com/dimagi/commcare-connect/pull/466/files#diff-a2e3da6458e9408753340f022d496b12d8fb4334e2a1996bb475d4afec1b9761R378) which is implemented in this way.

commcare_connect/opportunity/views.py
```diff
@@ -122,6 +123,11 @@ def test_func(self):
         ) or self.request.user.is_superuser
 
 
+class SuperUserMixin(LoginRequiredMixin, UserPassesTestMixin):
```
sravfeyn, line 126: There is already `SuperUserRequiredMixin`

sravfeyn:
Is it not possible to use the existing view `OpportunityList` to implement this for both superuser and org-specific users?

hemant10yadav:
> Is it not possible to use the existing view `OpportunityList` to implement this for both superuser and org-specific users?

I am happy to use the same page, but doing so will make it much more complex, right? It is already being used for two cases: Opportunity and Managed Opportunity.

commcare_connect/opportunity/views.py
hemant10yadav: I’m not sure if we can create the exact same view using `django-tables`, as we have used custom sortable links in opportunity list page. However, the ticket clearly mentions that it should look the same as the normal Opportunity page. I’m open to discussing this further.

commcare_connect/opportunity/views.py
hemant10yadav: Thanks I will remove it and use the existing `SuperUserRequiredMixin`.

commcare_connect/opportunity/views.py
pxwxnvermx: Django Tables has Meta options to enable sorting and ordering by columns. https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#Table.Meta

PR #453, opened 2024-12-12
Negative average time to convert in funnel performance report.

hemant10yadav:
[CCCT-570](https://dimagi.atlassian.net/browse/CCCT-570)
[CCCT-585](https://dimagi.atlassian.net/browse/CCCT-585)

[CCCT-570]: https://dimagi.atlassian.net/browse/CCCT-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CCCT-585]: https://dimagi.atlassian.net/browse/CCCT-585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/program/helpers.py
+++ b/commcare_connect/program/helpers.py
@@ -64,7 +64,15 @@ def get_annotated_managed_opportunity(program: Program):
             percentage_conversion=calculate_safe_percentage("workers_starting_delivery", "workers_invited"),
             average_time_to_convert=Avg(
                 ExpressionWrapper(
-                    Subquery(earliest_visits) - F("opportunityaccess__invited_date"), output_field=DurationField()
+                    Case(
+                        When(
+                            Q(opportunityaccess__invited_date__isnull=False)
+                            & Q(opportunityaccess__uservisit__isnull=False),
+                            then=Subquery(earliest_visits) - F("opportunityaccess__invited_date"),
+                        ),
+                        default=None,
+                    ),
+                    output_field=DurationField(),
                 ),
                 filter=FILTER_FOR_VALID_VISIT_DATE,
                 distinct=True,
--- a/commcare_connect/program/helpers.py
+++ b/commcare_connect/program/helpers.py
@@ -14,7 +14,7 @@ from django.db.models import (
     Value,
     When,
 )
-from django.db.models.functions import Cast, Coalesce, Greatest, Round
+from django.db.models.functions import Cast, Coalesce, Round
 
 from commcare_connect.opportunity.models import UserVisit, VisitValidationStatus
 from commcare_connect.program.models import ManagedOpportunity, Program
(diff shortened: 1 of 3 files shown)


commcare_connect/program/helpers.py
```diff
@@ -50,7 +52,7 @@ def get_annotated_managed_opportunity(program: Program):
         .annotate(
             workers_invited=Count("opportunityaccess", distinct=True),
             workers_passing_assessment=Count(
-                "opportunityaccess__assessment",
+                "opportunityaccess__assessment__opportunity_access",
```
calellowitz, line 55: this looks like it traverses a foreign key in two directions, can we not just skip the join entirely and count opportunity accesses that meet the filter below

commcare_connect/program/helpers.py
```diff
@@ -62,12 +64,19 @@ def get_annotated_managed_opportunity(program: Program):
...
             ),
             percentage_conversion=calculate_safe_percentage("workers_starting_delivery", "workers_invited"),
-            average_time_to_convert=Avg(
-                ExpressionWrapper(
-                    Subquery(earliest_visits) - F("opportunityaccess__invited_date"), output_field=DurationField()
+            average_time_to_convert=Coalesce(
+                Avg(
+                    ExpressionWrapper(
+                        Greatest(
+                            Subquery(earliest_visits) - F("opportunityaccess__invited_date"),
```
calellowitz, line 71: how does this end up negative? I see that we floor it at zero, but why is that necessary? Zero seems equally wrong if we trigger that condition

calellowitz (commented):
couple questions, but mostly curious what the cause of the negative values was and if we can avoid that rather than just floor it at a different arbitrary value.

commcare_connect/program/helpers.py
hemant10yadav: This issue occurred because we recently added the `invited_date` field with `auto_now_add`=True. As a result, invited_date was set to `now(`) during the migration for earlier records.

commcare_connect/program/helpers.py
hemant10yadav: Fixed - [8b40e2d](https://github.com/dimagi/commcare-connect/pull/453/commits/8b40e2d04201a85b1c8ce3220b9b60102340216b)

sravfeyn (approved):

commcare_connect/program/helpers.py
calellowitz: that means the average will still be wrong right? since it includes a bunch of incorrect zeros mixed in with real data?

hemant10yadav: Fixed and added test also -[944785e](https://github.com/dimagi/commcare-connect/pull/453/commits/944785eb52aafa0bd2753857297a6f812ddc7830)

calellowitz (approved):

PR #455, opened 2024-12-13
Fix unique key violation error in tests

pxwxnvermx:
This PR fixes the unique key violation error that randomly occurs in the CI test runner. Using the [django_get_or_create](https://factoryboy.readthedocs.io/en/stable/orms.html#factory.django.DjangoOptions.django_get_or_create) meta option for OpportunityAccessFactory, we can specify to run the `get_or_create` method to create objects when certain fields are present, in this case, `user` and `opportunity`. This will help avoid creating duplicate OpportunityAccess for users.

[Failing CI Run and Test Description](https://github.com/dimagi/commcare-connect/actions/runs/12312178249/job/34363704785)

### DIFF
--- a/commcare_connect/opportunity/tests/factories.py
+++ b/commcare_connect/opportunity/tests/factories.py
@@ -68,6 +68,7 @@ class OpportunityAccessFactory(DjangoModelFactory):
 
     class Meta:
         model = "opportunity.OpportunityAccess"
+        django_get_or_create = ["opportunity", "user"]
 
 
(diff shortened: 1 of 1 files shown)


calellowitz:
Do you know how this happens? I would expect tests clean up state and new invocations to always have unique values. Generally we would probably want to know if tests were reusing objects or not, so I would prefer if we could figure out how those are getting repeated, although this is probably a fine change

calellowitz (approved):

hemant10yadav (approved):

PR #456, opened 2024-12-15
duration and utilized

calellowitz:
This fixes two separate issues, each in their own commit.

1) Duration was not visible in the visit export, which made verifying that flag very difficult. I added it to the model as a property in addition to the export so it would be easier to use in the future, and accessible outside the serializer function.
2) budget utilization for the opportunity was wrong in a number of ways. It included work that was not approved and didn't account for the fact that an individual completed work could accrue multiple payments. Using the completed work functions to get the correct data introduced major performance problems, so I switched to summing payment accrued for each worker, which should be identical, but is a much simpler calculation.

[CCCT-582](https://dimagi.atlassian.net/browse/CCCT-582)

[CCCT-582]: https://dimagi.atlassian.net/browse/CCCT-582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/models.py
```diff
@@ -129,15 +130,10 @@ def claimed_budget(self):
...
-            completed_count=Count("id"), amount=F("payment_unit__amount")
-        )
+        users = OpportunityAccess.objects.filter(opportunity=self)
         utilised = 0
-        for payment_unit_count in payment_unit_counts:
-            completed_count = payment_unit_count["completed_count"]
-            amount = payment_unit_count["amount"]
-            utilised += completed_count * amount
+        for u in users:
+            utilised += u.payment_accrued
```
pxwxnvermx, line 136: We might need to add the `org_pay` component here as well for the managed opportunities.

pxwxnvermx (approved):
For the budget, we might need to include the org pay as well. The rest looks good to me.

commcare_connect/opportunity/models.py
hemant10yadav: Thanks Pawan - Added the org_pay [c1822ac](https://github.com/dimagi/commcare-connect/pull/456/commits/c1822ac8d501d70fa120c330ae603ee0b5446216)

hemant10yadav:
@calellowitz - Used the saved values to calculate the budget utilisation.

pxwxnvermx (approved):
LGTM. Thanks

PR #457, opened 2024-12-16
Add unique_together on completed work model, migration

pxwxnvermx:
The CompletedWork model is missing a unique together option that is causing a [race condition in `get_or_create`](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#get-or-create) in the form processor when multiple forms come in quick succession for the same payment unit leading to the creation of multiple completed work instances with same `entity_id` and `payment_unit_id`. This race condition only comes up in the Payment Unit configuration with multiple deliver units needed for completion.

This PR adds the unique together constraint on the model for the `entity_id` and `payment_unit` fields.


calellowitz (approved):
Do we need to do anything in the form processor to handle this better? Locking or transactions?

pxwxnvermx:
https://github.com/dimagi/commcare-connect/pull/459 adds a transaction in the form processor code.

PR #459, opened 2025-01-07
fixed duplicate form submission

hemant10yadav:
## Product Description

This change ensures that when a form is submitted more than once, the system prevents duplicate submissions and raises an appropriate error. [CCCT-631](https://dimagi.atlassian.net/browse/CCCT-631) 

## Technical Summary

- The code was modified to handle the case where the same form is submitted twice. The UserVisit model now checks for duplicate submissions based on xform_id, entity_id, and deliver_unit_id. If the same combination already exists, a DuplicateFormException is raised.

### Automated test coverage

A test is added to ensure that if a user attempts to submit the same form twice, only one visit entry is created, and the second form submission raises the exception.

### QA Plan

NA

### Labels & Review

[CCCT-631]: https://dimagi.atlassian.net/browse/CCCT-631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/form_receiver/processor.py
```diff
@@ -248,67 +254,78 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
-            user_visit.status = VisitValidationStatus.over_limit
-            if not completed_work.status == CompletedWorkStatus.over_limit:
-                completed_work.status = CompletedWorkStatus.over_limit
+            user_visit.status = VisitValidationStatus.approved
+            user_visit.review_status = VisitReviewStatus.agree
+
+        try:
+            user_visit.save()
+        except IntegrityError as e:
+            if "unique_xform_entity_deliver_unit" in str(e):
```
calellowitz, line 317: Is there no other way to catch this? It is fine if not but kind of kludgy

calellowitz (approved):
One question, but not blocking, generally this looks great and thanks for the added transaction handling

commcare_connect/form_receiver/processor.py
pxwxnvermx: nit: I think we can remove this error-handling code here and let the unique constraint error be raised and propagated to the sentry.

pxwxnvermx (approved):
Just a small comment on the error handling. Rest looks good to me.

commcare_connect/form_receiver/processor.py
calellowitz: offlined with Hemant about this, but I prefer this to raising the error because it isn't an unexpected excpetion but one we are properly handling so there is no need to send to sentry. Instead we can log if we we want a record, but the key here is that we can return a meaningful error to HQ (and honestly probably want to return a 200, since we dont want them to keep retrying).

pxwxnvermx: Cool, that sounds good to me. I am happy with adding a log to show that a form was reforwarded from hq and returning a 200 status to HQ.

commcare_connect/form_receiver/views.py
```diff
@@ -16,5 +20,9 @@ def post(self, request):
         serializer = XFormSerializer(data=request.data)
         serializer.is_valid(raise_exception=True)
         xform = serializer.save()
-        process_xform(xform)
+        try:
+            process_xform(xform)
+        except DuplicateFormException:
```
pxwxnvermx, line 25: We can move the try-except block from the form processor here, capture the IntegrityError, and log the duplicate.
The django documentation for transactions has a note specifying not to handle any database-related exceptions inside the atomic block.
https://docs.djangoproject.com/en/5.1/topics/db/transactions/

hemant10yadav: Thanks added the change [9299b9b](https://github.com/dimagi/commcare-connect/pull/459/commits/9299b9b8c796c3eb8a9dd2fdb864ca7567f68110)

pxwxnvermx (approved):
Thanks 🙌🏽

czue:
@hemant10yadav @pxwxnvermx @calellowitz the migration introduced in this PR appears to crash on staging, making future deploys fail. any suggestions on how best to resolve it?

trace:

```
django.db.utils.IntegrityError: could not create unique index "unique_xform_entity_deliver_unit"
DETAIL:  Key (xform_id, entity_id, deliver_unit_id)=(09b62601-5f04-4a8d-849d-20fb916481d9, 76839, 62) is duplicated.

  Applying opportunity.0065_uservisit_unique_xform_entity_deliver_unit...
 ERROR {
  "Status": "unhealthy",
  "FailingStreak": 0,
  "Log": [

  ]
}
```

hemant10yadav:
> @hemant10yadav @pxwxnvermx @calellowitz the migration introduced in this PR appears to crash on staging, making future deploys fail. any suggestions on how best to resolve it?
> 
> 
> 
> trace:
> 
> 
> 
> ```
> 
> django.db.utils.IntegrityError: could not create unique index "unique_xform_entity_deliver_unit"
> 
> DETAIL:  Key (xform_id, entity_id, deliver_unit_id)=(09b62601-5f04-4a8d-849d-20fb916481d9, 76839, 62) is duplicated.
> 
> 
> 
>   Applying opportunity.0065_uservisit_unique_xform_entity_deliver_unit...
> 
>  ERROR {
> 
>   "Status": "unhealthy",
> 
>   "FailingStreak": 0,
> 
>   "Log": [
> 
> 
> 
>   ]
> 
> }
> 
> ```

This issue occurred because we recently added a unique constraint to the user_visit table. There are already some duplicate entries in the table that need to be removed, similar to what we did in production.

PR #460, opened 2025-01-07
Fix completed work unique constraint

pxwxnvermx:
## Product Description
This PR fixes the unique constraint to be user-specific for duplicate completed works.

## Technical Summary
The constraint earlier was failing on duplicate visits across multiple users which is not the correct scenario. This makes the constraint user specific by including the opportunity access in the unique constraint.

### Labels & Review


calellowitz (approved):

hemant10yadav (approved):

commcare_connect/opportunity/migrations/0064_alter_completedwork_unique_together.py
```diff
@@ -11,6 +11,6 @@ class Migration(migrations.Migration):
     operations = [
         migrations.AlterUniqueTogether(
             name="completedwork",
-            unique_together={("entity_id", "payment_unit")},
+            unique_together={("opportunity_access", "entity_id", "payment_unit")},
```
hemant10yadav, line 14: Just a piece of information -I was going through the Django docs recently and found that it suggests using [UniqueConstraint](https://docs.djangoproject.com/en/5.1/ref/models/options/#unique-together) instead of `unique_together`.

PR #461, opened 2025-01-08
Fix invite delete conditions

pxwxnvermx:
## Product Description
This PR fixes when the delete button is shown for User Invites. Currently, the button was shown for all invites that were not accepted, but it needs to be shown only in cases when a user was not found in ConnectID.

image

### Safety story

I have tested these changes on staging.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -1236,7 +1236,9 @@ def invoice_approve(request, org_slug, pk):
 @csrf_exempt
 def user_invite_delete(request, org_slug, opp_id, pk):
     opportunity = get_opportunity_or_404(opp_id, org_slug)
-    invite = get_object_or_404(UserInvite, pk=pk, opportunity=opportunity, status=UserInviteStatus.not_found)
+    invite = get_object_or_404(UserInvite, pk=pk, opportunity=opportunity)
+    if invite.status != UserInviteStatus.not_found:
+        return HttpResponse(status=403, data="User Invite cannot be deleted.")
     invite.delete()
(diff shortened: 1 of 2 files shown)


hemant10yadav (approved):

commcare_connect/opportunity/views.py
```diff
@@ -1236,7 +1236,7 @@ def invoice_approve(request, org_slug, pk):
 @csrf_exempt
 def user_invite_delete(request, org_slug, opp_id, pk):
     opportunity = get_opportunity_or_404(opp_id, org_slug)
-    invite = get_object_or_404(UserInvite, pk=pk, opportunity=opportunity)
+    invite = get_object_or_404(UserInvite, pk=pk, opportunity=opportunity, status=UserInviteStatus.not_found)
```
calellowitz, line 1239: I don't think we want to 404 here if the invite has a different status. 404 means we couldn't find the invite, but thats not the case here. At worst we want to 403, since they aren't allowed, but if this can actually be triggered by a user, we should probably return a normal response with a message telling them why the delete failed.

calellowitz (changes requested):
Small thing but otherwise looks fine. I think in general we overuse the get or 404 pattern when it isn't quite appropriate so I am trying to keep an eye out for it.

commcare_connect/opportunity/views.py
pxwxnvermx: I have added a 403 response on this view when the user invite is not ok for deletion. The user cannot trigger this for cases where the invites are valid as we hide the button.

calellowitz (approved):
Thanks

PR #465, opened 2025-01-15
add payment indicator to program dashboard

czue:
## Product Description

Adds an "amount paid to FLWs" indicator to the program dashboard.

## Technical Summary

https://docs.google.com/document/d/19Ds59vZ7PHRcNIEFjOoIQBIJSZ3r8t9idR-wZdr-SkM/edit?tab=t.0#heading=h.n59190s1qddt

### Safety story

Tested locally + dimagi internal only report.

### Automated test coverage

No tests.

### QA Plan

Not needed.

### Labels & Review

### DIFF
--- a/commcare_connect/reports/views.py
+++ b/commcare_connect/reports/views.py
@@ -441,7 +441,7 @@ def dashboard_stats_api(request):
 @user_passes_test(lambda u: u.is_superuser)
 def dashboard_charts_api(request):
     filterset = DashboardFilters(request.GET)
-    queryset = UserVisit.objects.all()
+    queryset = UserVisit.objects.filter(opportunity__is_test=False)
     # Use the filtered queryset if available, else use last 30 days
     if filterset.is_valid():
         queryset = filterset.filter_queryset(queryset)


commcare_connect/reports/views.py
```diff
@@ -401,22 +401,40 @@ def dashboard_stats_api(request):
     filterset = DashboardFilters(request.GET)
 
     # Use the filtered queryset to calculate stats
-    queryset = UserVisit.objects.all()
+    visit_queryset = UserVisit.objects.all()
+    payment_queryset = Payment.objects.all()
```
calellowitz, line 405: I might have missed this in a previous PR, but i think both of these need to filter for opportunities where `is_test` is `False`, and I think the payment one needs to filter for `opportunity_access` not being null (since that represents a payment to a user rather than an org)

czue: thanks, updated.

In the future we'll also want to add an indicator for org payments. is it clear how to filter out test opportunities there? Would it be through the invoice field?

czue:
@calellowitz anything blocking this from getting merged?

calellowitz:
> @calellowitz anything blocking this from getting approved/merged?

Nope, sorry about that.

commcare_connect/reports/views.py
calellowitz: Hm, I don't know that we have a good way to filter for that. I think we try to be clever about orgs with real opportunities, but not sure how efficient that query is (probably not horrible but its at least one extra join to get the opportunity for an org). We might want to eventually store that at the org level, although I can imagine eventually orgs will have some test payments and some real ones.

calellowitz (approved):

commcare_connect/reports/views.py
calellowitz: It looks like you figured this out, but yeah. That sounds right.

PR #467, opened 2025-01-17
Add accrued, date columns to payment export/import

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-583

`

## Technical Summary

Adds `Payment Accrued`, `Payment Completed`, `Paid Date` columns to payment import/export workbooks. First two are current values which are just for information, and are not editable. `Paid Date` can now be specified.

### Safety story

Added tests and tested locally.

### Labels & Review


sravfeyn:
@calellowitz

commcare_connect/opportunity/models.py
```diff
@@ -426,7 +426,7 @@ class Meta:
 class Payment(models.Model):
     amount = models.PositiveIntegerField()
     amount_usd = models.DecimalField(max_digits=10, decimal_places=2, null=True)
-    date_paid = models.DateTimeField(auto_now_add=True)
+    date_paid = models.DateTimeField(default=datetime.datetime.utcnow)
```
sravfeyn, line 429: Without this it's not possible to override `date_paid` while doing `Payment.objects.create()`

commcare_connect/opportunity/visit_import.py
```diff
@@ -256,7 +260,16 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
...
             else:
-                payments[username] = amount
+                payments[username] = {"amount": amount}
+                try:
+                    if date_paid_raw:
+                        date_paid = datetime.datetime.strptime(date_paid_raw, "%Y-%m-%d").date()
+                    else:
+                        date_paid = None
+                except ValueError:
+                    invalid_rows.append((row, "paid date must be in YYYY-MM-DD format"))
```
calellowitz, line 270: will need to update this language when you update the column header

commcare_connect/opportunity/visit_import.py
```diff
@@ -256,7 +260,16 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
             except ValueError:
                 invalid_rows.append((row, "amount must be an integer"))
             else:
-                payments[username] = amount
+                payments[username] = {"amount": amount}
+                try:
+                    if date_paid_raw:
+                        date_paid = datetime.datetime.strptime(date_paid_raw, "%Y-%m-%d").date()
```
calellowitz, line 266: is there a reason to use this instead of one of the more flexible date parsers?

commcare_connect/opportunity/visit_import.py
```diff
@@ -256,7 +260,16 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
             except ValueError:
                 invalid_rows.append((row, "amount must be an integer"))
             else:
-                payments[username] = amount
+                payments[username] = {"amount": amount}
+                try:
```
calellowitz, line 264: I think we want to pull this out of the `else` block so we can surface errors in both fields at once. Its also just a bit hard to follow all the nesting

commcare_connect/opportunity/export.py
```diff
@@ -98,12 +99,33 @@ def _schema_sort(item):
...
 def export_empty_payment_table(opportunity: Opportunity) -> Dataset:
-    headers = ["Username", "Phone Number", "Name", "Payment Amount"]
+    headers = [
+        "Username",
+        "Phone Number",
+        "Name",
+        "Payment Accrued",
+        "Payment Completed",
+        "Payment Amount",
+        "Paid date",
```
calellowitz, line 109: nit: this should be "Date of Payment" or "Payment Date" to be consistent with the other headers

commcare_connect/opportunity/tests/test_visit_import.py
```diff
@@ -290,17 +290,44 @@ def test_bulk_update_payments(opportunity: Opportunity):
...
+
     assert payment_import_status.seen_users == {user.username for user in mobile_user_seen}
     assert payment_import_status.missing_users == {user.username for user in mobile_user_missing}
+
     assert Payment.objects.filter(opportunity_access__opportunity=opportunity).count() == 5
+
     for access in access_objects:
         payment = Payment.objects.get(opportunity_access=access)
         assert payment.amount == 50
+        assert payment.date_paid.strftime("%Y-%m-%d") == payment_date
```
calellowitz, line 330: It would be nice to test the default value as well

commcare_connect/opportunity/visit_import.py
sravfeyn: What other flexible date parsers are you referring to?

commcare_connect/opportunity/visit_import.py
sravfeyn: The issue with both of them being inside one block is that we won't know on which field the error occurred.

commcare_connect/opportunity/tests/test_visit_import.py
sravfeyn: Done

commcare_connect/opportunity/export.py
sravfeyn: Done

sravfeyn:
@calellowitz Updated

commcare_connect/opportunity/visit_import.py
calellowitz: I was thinking [this django util](https://docs.djangoproject.com/en/5.0/ref/utils/#module-django.utils.dateparse) which is mostly the same but slightly nicer to use, and a bit more flexible (mostly around digits for months and days)

commcare_connect/opportunity/visit_import.py
calellowitz: Yeah, agreed. I meant the opposite. I was thinking of pulling [this block](https://github.com/dimagi/commcare-connect/blob/sr/payex/commcare_connect/opportunity/visit_import.py#L264-L276) out two indentation levels (so its at the same level as `if amount`). We would still parse them sequentially, but would parse the date even if the parsing of the amount failed, so that we could show the date error when the amount is bad too. I can see an argument for only pulling it out one level, so we only check the date if there is something in the `amount_raw` field, but I think that is less helpful to the user, since something in the date field probably implies they meant to include a payment. Either way, I don't think it should be in the `else` block of the `try...except` for amount parsing, but outside of that, since we want it to run even when the amount `try` fails.

I would also structure it more like the amount one, moving everything out of the `try` except the actual date parsing expression.

commcare_connect/opportunity/export.py
```diff
@@ -103,10 +103,8 @@ def export_empty_payment_table(opportunity: Opportunity) -> Dataset:
         "Username",
         "Phone Number",
         "Name",
-        "Payment Accrued",
-        "Payment Completed",
         "Payment Amount",
-        "Paid date",
+        "Payment Date (YYYY-MM-DD)",
     ]
```
calellowitz, line 108: I think including the format in the header title is a little gross, especially if we have good error messages.

commcare_connect/opportunity/tests/test_visit_import.py
```diff
@@ -324,10 +325,13 @@ def test_bulk_update_payments(opportunity: Opportunity):
 
     assert Payment.objects.filter(opportunity_access__opportunity=opportunity).count() == 5
 
-    for access in access_objects:
+    for index, access in enumerate(access_objects):
         payment = Payment.objects.get(opportunity_access=access)
         assert payment.amount == 50
-        assert payment.date_paid.strftime("%Y-%m-%d") == payment_date
+        if index == 4:
```
calellowitz, line 331: instead of checking the index like this, the test would be more robust if we passed the uploaded data. it would also be easier to reason about, rather than trying to understand why 4 is a magic number. Ideally that would use something like the parametrized tests we use elsewhere, but if that is tricky with the rest of this test setup, storing the list from line 307 as you pass it to the dataset, would allow you to cross reference it here

calellowitz (approved):
As discussed offline, fine to followup on this feedback tomorrow, and happy to discuss it further then. In the interest of urgency, merging and deploying now

PR #471, opened 2025-01-20
add computed earnings fields to CompletedWork

czue:
## Product Description

For now there are no product-visible changes, though this may unlock better, faster reporting on "amount earned" indicators in the longer term once the fields are confirmed working.

## Technical Summary

https://docs.google.com/document/d/19Ds59vZ7PHRcNIEFjOoIQBIJSZ3r8t9idR-wZdr-SkM/edit?tab=t.0#heading=h.n59190s1qddt

This adds several "saved" fields to the `CompletedWork` model to make it easier to efficiently determine the amount of work completed/approved as well as payments earned to FLWs/orgs in both local currency and USD.

### Safety story

I was reasonably careful to try and preserve all existing functionality and only add new functionality. However, since this does touch a function that is called regularly and involved in generating critical reporting values, I think it's worth paying careful attention to.

For the new properties I think it's very low risk, since nothing uses them. So the concern would be:

1. If this accidentally changes the behavior of the `update_status` function. Either by causing a crashing error where previously there wasn't one, modifying how the value of `completed_work.status` or `opportunity_access.payment_accrued` is calculated (I'm reasonably confident this wasn't done, but mentioning it anyways).
2. If this has a dramatic impact on the performance of that function. It does introduce a few new database queries to get needed fields and exchange rates. The exchange rate lookup also has the potential to do external API calls which could be slow.
3. If somehow the process of "backfilling" the data (`./manage.py populate_completed_work_saved_fields`) accidentally modifies the `status` or `payment_accrued` fields in an incorrect way. Again, I don't see this as likely, but there could be something I've overlooked.

There does appear to be pretty decent test coverage of the function that computes these values. I've updated the tests I could find that were checking it, and once those are all passing and someone has reviewed the code, I'll feel reasonably good about the safety of this.

### Automated test coverage

There were tests for a lot of the `update_status` logic. All those tests are still passing, and I've updated them to also test for the new computed fields that were relevant to the existing tests.

### QA Plan

I don't really have a plan for this, apart from running it on staging (including the management command) before merging it to production. As long as it doesn't crash or mess up data it should be low risk to impact anything else, as it's all isolated to a single function.

### Labels & Review


commcare_connect/opportunity/management/commands/populate_completed_work_saved_fields.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Command(BaseCommand):
+    help = (
+        "Populates the saved_* fields on CompletedWork objects "
+        "(and also the payment_accrued field on OpportunityAccess)"
+    )
+
+    def handle(self, *args, **options):
+        for opportunity_access in OpportunityAccess.objects.all():
```
czue, line 14: The current task applies a bunch of filters on the opportunities it checks, so we will need this management command to backfill the existing data.

czue (commented):
Ok I think this is ready to go

commcare_connect/opportunity/utils/completed_work.py
```diff
@@ -11,37 +11,74 @@
...
+            completed_work.status = CompletedWorkStatus.approved
+
+        if (
+            opportunity_access.opportunity.managed
+            and not all(review_status == VisitReviewStatus.agree for *_, review_status in visits)
+            and completed_work.status == CompletedWorkStatus.approved
+        ):
+            completed_work.status = CompletedWorkStatus.pending
+
+        made_changes = True
```
calellowitz, line 51: nit: this is only true if  one of the above `if`s is true, right? In this implementation `made_changes == opportunity_access.opportunity.auto_approve_payments` rather than any actual changes?

calellowitz (approved):
One small question, but not blocking. Looks good

commcare_connect/opportunity/utils/completed_work.py
czue: Yeah that's right, which I did intentionally since that's also [how the previous version's save logic worked](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/utils/completed_work.py#L20-L35). We could theoretically make it more specific, but I was worried about accidentally changing behavior.

calellowitz (approved):

PR #474, opened 2025-01-21
Payment dashboard updates

czue:
This is a PR into https://github.com/dimagi/commcare-connect/pull/471 as it's dependent on that work. It also has everything in https://github.com/dimagi/commcare-connect/pull/465, since that hasn't been merged yet.

## Product Description

Adds FLW/Org earnings and payments to the program dashboard report.

image

## Technical Summary

The reporting side of this is pretty straightforward.

### Safety story

This only touches an admin report that very few people know about. So should be safe.

### Automated test coverage

No tests.

### QA Plan

Will throw on staging, but shouldn't require any QA before being merged.

### Labels & Review

### DIFF
--- a/commcare_connect/reports/views.py
+++ b/commcare_connect/reports/views.py
@@ -400,7 +400,8 @@ def dashboard_stats_api(request):
 
     # Use the filtered queryset to calculate stats
     visit_queryset = UserVisit.objects.filter(opportunity__is_test=False)
-    payment_queryset = Payment.objects.filter(opportunity_access__opportunity__is_test=False)
+    flw_payment_queryset = Payment.objects.filter(opportunity_access__opportunity__is_test=False)
+    org_payment_queryset = Payment.objects.filter(invoice__opportunity__is_test=False)
     if filterset.is_valid():
         visit_queryset = filterset.filter_queryset(visit_queryset)
         raw_filters = filterset.form.cleaned_data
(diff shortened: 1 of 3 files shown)


commcare_connect/reports/views.py
```diff
@@ -401,22 +399,68 @@ def dashboard_stats_api(request):
...
-    queryset = UserVisit.objects.all()
+    visit_queryset = UserVisit.objects.filter(opportunity__is_test=False)
+    flw_payment_queryset = Payment.objects.filter(opportunity_access__opportunity__is_test=False)
+    org_payment_queryset = Payment.objects.filter(invoice__opportunity__is_test=False)
+    completed_work_queryset = CompletedWork.objects.filter(opportunity_access__opportunity__is_test=False)
     if filterset.is_valid():
-        queryset = filterset.filter_queryset(queryset)
+        visit_queryset = filterset.filter_queryset(visit_queryset)
+        raw_filters = filterset.form.cleaned_data
+        program = raw_filters.get("program")
```
calellowitz, line 409: nit: we use `program` to mean something different elsewhere in the codebase (a collection of opportunities managed together, see the `programs` directory). It might be clearer if this were `delivery_type` unless we mean `program` in which case it doesn't appear to be filtering correctly.

calellowitz (approved):
One naming nit, but otherwise looks good.

commcare_connect/reports/views.py
czue: hmm, I was mainly using it to reflect the language on the filters in the front end, which was terminology that I believe I got from Ali. Do you think that was referring to the type of program you're talking about? Or is it a difference between the user-facing terminology and the dev-facing terminology?

czue: (I'm going to merge this for now since it'd be great to see the data on the next deploy, but happy to follow up on this if we think we got something wrong)

PR #478, opened 2025-01-23
Learning time does not include all attempts

hemant10yadav:
## Technical Summary

Implemented functionality to allow users to complete the learn module multiple times. This enables capturing the learning time for each module. [CCCT-638](https://dimagi.atlassian.net/browse/CCCT-638)

### Automated test coverage

Comprehensive test coverage has been added to ensure that a completed module is created for every submission and that only unique modules are sent to the device. This guarantees accurate progress calculation and eliminates redundancy.

### Labels & Review

[CCCT-638]: https://dimagi.atlassian.net/browse/CCCT-638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pxwxnvermx:
@hemant10yadav Are you planning to get this PR QA'ed?

commcare_connect/form_receiver/processor.py
```diff
@@ -88,24 +88,24 @@ def process_learn_modules(user: User, xform: XForm, app: CommCareApp, opportunit
...
+            date=xform.received_on,
+            duration=xform.metadata.duration,
+            app_build_id=xform.build_id,
+            app_build_version=xform.metadata.app_build_version,
         )
+        completed_modules.append(completed_module)
 
-        if not created:
-            raise ProcessingError("Learn Module is already completed")
+    if completed_modules:
```
calellowitz, line 107: This will need logic to make sure we dont create duplicates if the same form is forwarded twice (as happened with visits). Ideally that would be enforced at the DB level.

calellowitz (changes requested):

commcare_connect/form_receiver/tests/test_receiver_integration.py
```diff
@@ -90,6 +92,59 @@ def test_form_receiver_learn_module_create(
     ).exists()
 
 
+@pytest.mark.django_db
+def test_form_receiver_multiple_module_submissions(
```
sravfeyn, line 96: Making this test use parameters might make it easy to understand various cases.

hemant10yadav:
> @hemant10yadav Are you planning to get this PR QA'ed?

Yeah I have raised the QA form for it.

commcare_connect/form_receiver/tests/test_receiver_integration.py
hemant10yadav: Changed [d3fcc0e](https://github.com/dimagi/commcare-connect/pull/478/commits/d3fcc0e39fae078890d1042f79d2645bb03a9315)

hemant10yadav:
@calellowitz @sravfeyn @pxwxnvermx 
QA has been completed, and all issues have been resolved. Could you please review this PR further?

commcare_connect/form_receiver/processor.py
calellowitz: If this is violated, do all fail to create or just one? Either way, we likely want to catch this exception and return a reasonable message to hq (probably a 200, since there isn't an issue, but if an error code, one with a clear message like existed previously)

hemant10yadav: Added - [f9ff20c](https://github.com/dimagi/commcare-connect/pull/478/commits/f9ff20c5389448703b6ab6041df07a87b0f99cf5)

commcare_connect/form_receiver/processor.py
```diff
@@ -88,24 +88,24 @@ def process_learn_modules(user: User, xform: XForm, app: CommCareApp, opportunit
...
+        CompletedModule.objects.bulk_create(completed_modules)
```
sravfeyn, line 108: Wouldn't this result in IntegrityError when the same form is forwarded again?

hemant10yadav: Yeah we are catching that at outer level.

commcare_connect/form_receiver/processor.py
calellowitz: I think to ensure this is the case, this update should be in a transaction. That provides certainty regardless of how django handles the underlying queries, which is key since otherwise we could process the first of multiple and then keep rejecting the following ones.

hemant10yadav: Added-[8fd2c04](https://github.com/dimagi/commcare-connect/pull/478/commits/8fd2c04e7b9ff121e2602689e4bf47d0a07bfb5a)

PR #479, opened 2025-01-24
Fix for Network Manager exceed program's total budget.

hemant10yadav:
## Technical Summary

On the "Add Budget" page, for managed opportunities, the NM cannot increase the opportunity's budget but can add additional visits if they remain within the opportunity's budget. If the additional visits exceed the opportunity's budget, the user will see the error message:
`Additional visits exceed the opportunity's budget.`

 [CCCT-639](https://dimagi.atlassian.net/browse/CCCT-639)

### Safety story

This functionality has been tested on the staging environment.

### Automated test coverage

Added test scenarios to validate both cases:

- When the total budget exceeds the program budget for a managed opportunity.
- When the total budget remains within the program budget.

### Labels & Review

[CCCT-639]: https://dimagi.atlassian.net/browse/CCCT-639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/views.py
```diff
@@ -455,16 +455,43 @@ def add_budget_existing_users(request, org_slug=None, pk=None):
         if form.is_valid():
             selected_users = form.cleaned_data["selected_users"]
             additional_visits = form.cleaned_data["additional_visits"]
-            if form.cleaned_data["end_date"]:
-                OpportunityClaim.objects.filter(pk__in=selected_users).update(end_date=form.cleaned_data["end_date"])
+            end_date = form.cleaned_data["end_date"]
```
calellowitz, line 458: I assume this isn't an optional field?

commcare_connect/opportunity/views.py
```diff
@@ -455,16 +455,43 @@ def add_budget_existing_users(request, org_slug=None, pk=None):
...
+                if opportunity.managed:
+                    managed_opp = opportunity.managedopportunity
+                    program = managed_opp.program
+                    total_budget_sum = (
+                        ManagedOpportunity.objects.filter(program=program)
+                        .exclude(id=managed_opp.id)
+                        .aggregate(total=Sum("total_budget"))["total"]
+                        or 0
+                    )
+                    if total_budget_sum + opportunity.total_budget > program.budget:
```
calellowitz, line 476: Shouldn't the opportunity have a predefined budget limit separate from the program? I would have thought we don't need to check the program total, because it's limited at the opportunity level (which is guaranteed to be within program limits at the time of opportunity creation)

commcare_connect/opportunity/views.py
```diff
@@ -455,16 +455,43 @@ def add_budget_existing_users(request, org_slug=None, pk=None):
...
+                        form.add_error("additional_visits", "Additional visits exceed the program budget.")
```
calellowitz, line 477: If we are adding errors to the form, it feels like that should be in the form validation itself

commcare_connect/opportunity/views.py
```diff
@@ -455,16 +455,43 @@ def add_budget_existing_users(request, org_slug=None, pk=None):
...
+                        return render(
```
calellowitz, line 478: I am hoping this is made unnecessary by adding it to the form validation, but if not this is quite confusing, especially since it returns the same as the end of the function. It would be great if this method could be restructured to only return at the end rather than an early return here.

calellowitz (changes requested):
One question about the logic, but this feels like form validation, can it not be handled at that level? it feels wrong that `form.is_valid` would return `True` in this case, and that handles the error messaging automatically, without the special handling you added here

commcare_connect/opportunity/views.py
hemant10yadav: This view allows us to increase the opportunity's total budget by adding additional visits to the user, which the creator fixed during opportunity creation, as you mentioned. That's why I checked the program budget, as the opportunity budget will increase if we try to add additional visits. Let me know if I missed anything here.

commcare_connect/opportunity/views.py
hemant10yadav: Yeah

hemant10yadav:
> Who has permission to use this page? This gets back to a comment I left earlier, but if this is for network managers, they should not be allowed to take additional budget from the program, only program managers should be able to do that. However, network managers should have a way to add additional users against the opp budget limit (previously set by the program manager).
> 
> I see the need for the workflow this seems to be for, but I think there is an issue if managed opps are allowed to set their own budgets

I checked that this page is accessible to an org_member, and in this case, it is accessible to the Network Manager.

commcare_connect/opportunity/forms.py
```diff
@@ -634,18 +634,12 @@ def _validate_budget(self, selected_users, additional_visits):
...
-                .aggregate(total=Sum("total_budget"))["total"]
-                or 0
-            )
-
-            if budget_used + self.opportunity.total_budget + budget_increase > program.budget:
-                raise forms.ValidationError({"additional_visits": "Additional visits exceed the program budget."})
+            # NM cannot increase the opportunity budget they can only
+            # assign new visits if the opportunity has remaining budget.
+            print("budget_increase", budget_increase)
+            print("self.opportunity.remaining_budget", self.opportunity.remaining_budget)
```
pxwxnvermx, line 640: nit: print statements

sravfeyn (approved):

pxwxnvermx (approved):

calellowitz (approved):

PR #480, opened 2025-01-27
Fix UnboundLocalError when accessing amount

pxwxnvermx:
## Product Description
This fixes an error when the user tries to submit payments with non-integer amounts i.e INR 1000, USD 100.

## Technical Summary
This error was due to the Unbound amount variable when the ValueError exception occurred for non-integer amounts.

[Sentry Link](https://dimagi.sentry.io/issues/6248248829/?alert_rule_id=14580978&alert_timestamp=1737889101584&alert_type=email&environment=production&notification_uuid=2f06f11a-7b8c-4640-83a9-c2f79eb3bf64&project=4505635339829248&referrer=alert_email)

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/visit_import.py
+++ b/commcare_connect/opportunity/visit_import.py
@@ -253,9 +253,10 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
                 invalid_rows.append((row, "username required"))
             try:
                 amount = int(amount_raw)
-                payments[username] = amount
             except ValueError:
                 invalid_rows.append((row, "amount must be an integer"))
+            else:
+                payments[username] = amount
 
     if invalid_rows:
         raise ImportException(f"{len(invalid_rows)} have errors", invalid_rows)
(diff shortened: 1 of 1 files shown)


commcare_connect/opportunity/visit_import.py
```diff
@@ -253,9 +253,9 @@ def _bulk_update_payments(opportunity: Opportunity, imported_data: Dataset) -> P
                 invalid_rows.append((row, "username required"))
             try:
                 amount = int(amount_raw)
+                payments[username] = amount
             except ValueError:
                 invalid_rows.append((row, "amount must be an integer"))
-            payments[username] = amount
```
calellowitz, line 258: I think the "correct" way to do this is put it in an `else` clause. We don't want to catch errors with that statement (although I am not sure that is possible here), but only want it to run if the the `try` succeeds, and `else` is the standard way to do that.

pxwxnvermx: Thanks. Added. [14828cf](https://github.com/dimagi/commcare-connect/pull/480/commits/14828cfdb2c7ad103c12da26982781d1e8ce1c11)

calellowitz (approved):

PR #482, opened 2025-01-28
User Visit Review Page: Filters and Sortable fields

pxwxnvermx:
## Product Description
This PR adds filters and sortable fields for the User Visit Review page.

image

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-577)

### Safety story
The features included in this PR are deployed and tested on staging.

### Automated test coverage
None

### QA Plan
None

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -59,12 +59,13 @@ def show_warning(record):
 
 class UserVisitReviewFilter(FilterSet):
     review_status = ChoiceFilter(choices=VisitReviewStatus.choices, empty_label="All Reviews")
-    user = ModelChoiceFilter(queryset=User.objects.none(), empty_label="All Users")
+    user = ModelChoiceFilter(queryset=User.objects.none(), empty_label="All Users", to_field_name="username")
     visit_date = DateRangeFilter()
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.filters["user"].queryset = User.objects.filter(id__in=self.queryset.values_list("user_id", flat=True))
+        self.filters["user"].field.label_from_instance = lambda obj: obj.name
 
         self.form.helper = FormHelper()
         self.form.helper.disable_csrf = True
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -3,7 +3,7 @@ from crispy_forms.layout import Column, Layout, Row
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
-from django_filters import ChoiceFilter, FilterSet, ModelChoiceFilter
+from django_filters import ChoiceFilter, DateRangeFilter, FilterSet, ModelChoiceFilter
 from django_tables2 import columns, tables, utils
 
 from commcare_connect.opportunity.models import (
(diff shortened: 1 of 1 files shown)


commcare_connect/opportunity/tables.py
```diff
@@ -58,15 +59,26 @@ def show_warning(record):
...
-        self.form.fields["review_status"].widget.attrs.update({"@change": "$refs.reviewFilterForm.submit()"})
+        self.form.helper.layout = Layout(
+            Row(
+                Column("review_status", css_class="col-md-3"),
+                Column("user", css_class="col-md-3"),
+                Column("visit_date", css_class="col-md-3"),
+            )
+        )
+        for field_name in self.form.fields.keys():
+            self.form.fields[field_name].widget.attrs.update({"@change": "$refs.reviewFilterForm.submit()"})
```
calellowitz, line 81: does this submit the form or just refresh it?

commcare_connect/opportunity/tables.py
```diff
@@ -58,15 +59,26 @@ def show_warning(record):
 
 class UserVisitReviewFilter(FilterSet):
     review_status = ChoiceFilter(choices=VisitReviewStatus.choices, empty_label="All Reviews")
+    user = ModelChoiceFilter(queryset=User.objects.none(), empty_label="All Users", to_field_name="username")
```
calellowitz, line 62: Im pretty confused by this user filter queryset behavior. Can you explain a bit more what is occurring? It is set to empty, then overwritten on `init`, but where do the values in the queryset for `id__in` come from, and why does it need to be done in such a way?

calellowitz (commented):
Generally looks good, but a few questions about behavior

commcare_connect/opportunity/tables.py
pxwxnvermx: this submits the review filter form whenever any filter in that form is changed.

commcare_connect/opportunity/tables.py
pxwxnvermx: The `ModelChoiceFilter` filter requires a query set of objects to be instantiated with, its a filter wrapper for `ModelChoiceField`, so I have used a `.none()` query set object to instantiate that filter as none query set don't make any queries to database. Further the `self.queryset` used in the `id__in` is the query set for UserVisits that this `UserVisitReviewFilter` works on and is passed into the filter when it is instantiated in the `user_visit_review` view. I hope this sounds good and please suggest any improvements you have to improve this.

sravfeyn (approved):
LGTM

commcare_connect/opportunity/tables.py
```diff
@@ -58,15 +59,26 @@ def show_warning(record):
...
+    visit_date = DateRangeFilter()
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.filters["user"].queryset = User.objects.filter(id__in=self.queryset.values_list("user_id", flat=True))
```
sravfeyn, line 67: Don't you need to pass any org/program filtering on this queryset?

commcare_connect/opportunity/tables.py
calellowitz: ah filters are their own form?

commcare_connect/opportunity/tables.py
calellowitz: Thanks, it feels like there should be a better way, but not one I can think of.

commcare_connect/opportunity/tables.py
pxwxnvermx: Yup. The filters are in their own form.

commcare_connect/opportunity/tables.py
pxwxnvermx: This is using the query set passed for the filter initially, so the user choices will be limited to the User Visits passed in the queryset.

hemant10yadav (approved):

PR #488, opened 2025-02-12
Ensure Case-Insensitive Currency Code Comparison

hemant10yadav:
## Technical Summary

Currently, when the currency code "USD" is in lowercase, the system incorrectly attempts to fetch the rate from the rates API. This update ensures that currency code comparisons are case-insensitive by converting currency_code to uppercase before checking against "USD", preventing unnecessary API calls due to case mismatches.

[Sentry Error](https://dimagi.sentry.io/issues/6289210916/?project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&statsPeriod=24h&stream_index=2)

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/forms.py
+++ b/commcare_connect/opportunity/forms.py
@@ -237,6 +237,8 @@ class OpportunityInitForm(forms.ModelForm):
         deliver_app = self.cleaned_data["deliver_app"]
         learn_app_domain = self.cleaned_data["learn_app_domain"]
         deliver_app_domain = self.cleaned_data["deliver_app_domain"]
+
+        self.instance.currency = self.cleaned_data["currency"].upper()
         self.instance.learn_app, _ = CommCareApp.objects.get_or_create(
             cc_app_id=learn_app["id"],
             cc_domain=learn_app_domain,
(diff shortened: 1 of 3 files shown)


hemant10yadav:
> @hemant10yadav I don't think this error should be raised in the form processor. When this code raises an exception, we return a 500 response to HQ, which is incorrect as the form was processed correctly.

@pxwxnvermx Yeah, you are right. Ideally, this should not raise any exceptions since the form has already been processed. However, if an exception occurs during currency conversion, we should be aware of the issue rather than allowing a silent failure. In this case, the issue arose because an opportunity had its currency value as "usd" in lowercase, which caused a problem.

calellowitz (changes requested):
Thanks for catching this! Two quick thoughts:

I think we want to just always upper case currency codes (maybe on first save, but also in all comparisons), not just this place. Since if it wasn't USD we would still have the same problem.

Longer term, I think we want a prepopulated list of currencies, but we don't have to do that now.

sravfeyn (approved):

calellowitz (approved):
This looks fine, but does it address if we have other lowercase currencies already? I guess we would already see those errors, but it might be nice to add it to the code that fetches the rate just to be safe

PR #491, opened 2025-02-14
Other Expense related Invoices

pxwxnvermx:
## Product Description
This PR adds a checkbox to mark invoices for payments that are other expenses and not related to service delivery.

image

image

### Safety story

I have tested these changes locally.

### Labels & Review


commcare_connect/opportunity/forms.py
```diff
@@ -860,6 +860,7 @@ def __init__(self, *args, **kwargs):
             Row(Field("amount")),
             Row(Field("date")),
             Row(Field("invoice_number")),
+            Row(Field("other_expense")),
```
calellowitz, line 863: I think we want to invert all this and default to true and call it "Service Delivery," it is unclear otherwise what "other" refers to (ie what it is other than)

calellowitz (changes requested):
This needs to be in the admin report as well (and we might as well do both parts in this PR)

commcare_connect/opportunity/forms.py
pxwxnvermx: Done. [05468b3](https://github.com/dimagi/commcare-connect/pull/491/commits/05468b328d14ff3b9114b83855d3a75b4cfc19b4)

calellowitz (approved):

PR #492, opened 2025-02-14
Address Payments Import issues

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-653

- Makes payment import async to make it clear to users when it's processing
   - By adding a celery util to propagate task progress
- Fixes exchange-rate calc to be based on payment-date
- Handles multiple payments to same user

### Safety story

Tested locally

### Automated test coverage

Tests exist

### QA Plan

NA

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -123,7 +123,7 @@ from commcare_connect.organization.decorators import org_admin_required, org_mem
 from commcare_connect.program.models import ManagedOpportunity, ProgramApplication
 from commcare_connect.program.tables import ProgramInvitationTable
 from commcare_connect.users.models import User
-from commcare_connect.utils.celery import get_task_progress_message
+from commcare_connect.utils.celery import CELERY_TASK_SUCCESS, get_task_progress_message
 from commcare_connect.utils.commcarehq_api import get_applications_for_user_by_domain, get_domains_for_user
 from commcare_connect.utils.file import get_file_extension
 
@@ -423,7 +423,7 @@ def export_status(request, org_slug, task_id):
     task = AsyncResult(task_id)
     task_meta = task._get_task_meta()
     status = task_meta.get("status")
-    progress = {"complete": status == "SUCCESS", "message": get_task_progress_message(task)}
+    progress = {"complete": status == CELERY_TASK_SUCCESS, "message": get_task_progress_message(task)}
     if status == "FAILURE":
         progress["error"] = task_meta.get("result")
     return render(
--- a/commcare_connect/utils/celery.py
+++ b/commcare_connect/utils/celery.py
@@ -1,5 +1,9 @@
+CELERY_TASK_SUCCESS = "SUCCESS"
+CELERY_TASK_IN_PROGRESS = "PROGRESS"
+
+
 def set_task_progress(task, message, is_complete=False):
-    task.update_state(state="SUCCESS" if is_complete else "PROGRESS", meta={"message": message})
+    task.update_state(state=CELERY_TASK_SUCCESS if is_complete else CELERY_TASK_IN_PROGRESS, meta={"message": message})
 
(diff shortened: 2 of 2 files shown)


sravfeyn:
@calellowitz

sravfeyn:
@calellowitz Bumping this for review

commcare_connect/utils/celery.py
```diff
@@ -0,0 +1,7 @@
+def set_task_progress(task, message, is_complete=False):
+    task.update_state(state="SUCCESS" if is_complete else "PROGRESS", meta={"message": message})
```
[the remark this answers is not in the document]
calellowitz: :heavy_plus_sign:

commcare_connect/opportunity/tasks.py
```diff
@@ -343,3 +344,19 @@ def generate_catchment_area_export(opportunity_id: int, export_format: str):
     export_tmp_name = f"{now().isoformat()}_{opportunity.name}_catchment_area.{export_format}"
     save_export(dataset, export_tmp_name, export_format)
     return export_tmp_name
+
+
+@celery_app.task(bind=True)
+def bulk_update_payments_task(self, opportunity_id: int, headers: list[str], rows: list[list]):
+    from commcare_connect.opportunity.visit_import import ImportException, bulk_update_payments
+
+    set_task_progress(self, "Payment Record Import is in porgress.")
```
calellowitz, line 353: nit: progress

commcare_connect/opportunity/tasks.py
```diff
@@ -343,3 +344,19 @@ def generate_catchment_area_export(opportunity_id: int, export_format: str):
...
+    try:
+        status = bulk_update_payments(opportunity_id, headers, rows)
+        messages = [f"Payment status updated successfully for {len(status)} users."]
+        if status.missing_users:
+            messages.append(status.get_missing_message())
+    except ImportException as e:
+        messages = [f"Payment Import failed: {e}"] + getattr(e, "invalid_rows", [])
+
+    set_task_progress(self, f"<br>".join(messages), is_complete=True)
```
calellowitz, line 362: how do these messages get shown to the user? are they maintained if the user navigates away?

commcare_connect/opportunity/views.py
```diff
@@ -512,14 +512,15 @@ def export_users_for_payment(request, org_slug, pk):
...
-    else:
-        message = f"Payment status updated successfully for {len(status)} users."
-        messages.success(request, mark_safe(message))
-    return redirect("opportunity:detail", org_slug, pk)
+
+    file_format = get_file_extension(file)
+    if file_format not in ("csv", "xlsx"):
+        raise ImportException(f"Invalid file format. Only 'CSV' and 'XLSX' are supported. Got {file_format}")
+    imported_data = get_imported_dataset(file, file_format)
+    rows = list(imported_data)
```
calellowitz, line 520: If the import is quite large this will take a lot of memory in the workers and redis (our celery broker). It would be good to add some kind of length limit here, since it is so easy to break the sheet into multiple, and ideally add locking so that only one task per opportunity can be run. I wouldn't expect payments to hit the limit (unlike visit import), but better to be safe. Alternatively, you can save the sheet in s3, and pass a key to the celery task so that we don't need to store the info in redis at all.

calellowitz (approved):
Left a couple comments but generally looks good. Nothing blocking, but would be good to fix

commcare_connect/opportunity/views.py
sravfeyn: Implemented storing it as file and passing ref.

sravfeyn:
@calellowitz @pxwxnvermx I have addressed your changes, please have a review

pxwxnvermx (approved):

calellowitz (approved):

PR #494, opened 2025-02-15
add secondary db to env

calellowitz:
## Technical Summary

I think this was missing from https://github.com/dimagi/commcare-connect/pull/481/

Waiting on review to let me know if I missed any other way this was being updated.


pxwxnvermx (approved):
Thanks

sravfeyn (approved):

PR #496, opened 2025-02-17
Fix filter only valid payment dates

pxwxnvermx:
## Product Description
This fixes the negative average time to payment dates generated due to invalid payment dates set for completed work records that are in the past.

## Technical Summary
[Ticket](https://dimagi.atlassian.net/browse/CCCT-769)

### Automated test coverage
Already Covered in test included earlier.

### Labels & Review

### DIFF
--- a/commcare_connect/reports/helpers.py
+++ b/commcare_connect/reports/helpers.py
@@ -61,6 +61,7 @@ def get_table_data_for_year_month(
             saved_approved_count__gt=0,
             saved_payment_accrued_usd__gt=0,
             saved_org_payment_accrued_usd__gt=0,
+            payment_date__gte=F("date_created"),
         )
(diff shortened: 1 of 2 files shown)


calellowitz (commented):
Does this exclude the data from all metrics not just time to payment? I think we want to know the visits happened even in that case and count them in the relevant metrics

pxwxnvermx:
> Does this exclude the data from all metrics not just time to payment? I think we want to know the visits happened even in that case and count them in the relevant metrics

Yeah the filter before was excluding data from all metrics. I moved the filtering to be on Avg and Max aggregation functions to exclude the invalid dates only from the avg and max time to payment metrics.

sravfeyn (approved):

PR #497, opened 2025-02-18
Add Total ConnectID Accounts Metric

pxwxnvermx:
## Product Description
This PR adds the Total ConnectID Accounts metric to the report.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-776)

### Labels & Review

### DIFF
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -70,6 +70,12 @@ def filter_users(country_code: str, credential: list[str]):
     return [ConnectIdUser(**user_dict) for user_dict in data["found_users"]]
 
 
+def fetch_user_counts() -> dict[str, int]:
+    response = _make_request(GET, "/users/fetch_user_counts")
+    data = response.json()
+    return data
+
+
 def _make_request(method, path, params=None, json=None, timeout=5) -> Response:
     if json and not method == "POST":
         raise ValueError("json can only be used with POST requests")
--- a/commcare_connect/reports/helpers.py
+++ b/commcare_connect/reports/helpers.py
@@ -6,6 +6,7 @@ from statistics import mean
 from django.db.models import Avg, Count, DurationField, ExpressionWrapper, F, Max, OuterRef, Q, Subquery, Sum
 from django.utils.timezone import make_aware
 
+from commcare_connect.connect_id_client.main import fetch_user_counts
 from commcare_connect.opportunity.models import (
     CompletedWork,
     CompletedWorkStatus,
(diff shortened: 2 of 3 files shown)


hemant10yadav (approved):

commcare_connect/connect_id_client/main.py
```diff
@@ -70,6 +70,12 @@ def filter_users(country_code: str, credential: list[str]):
     return [ConnectIdUser(**user_dict) for user_dict in data["found_users"]]
 
 
+def fetch_user_counts() -> dict[str, int]:
```
sravfeyn, line 73: We should cache this

sravfeyn: Also, it doesn't seem very ideal to fetch this data from CID, is there no way to infer it on CCC?

commcare_connect/reports/helpers.py
```diff
@@ -78,6 +80,7 @@ def get_table_data_for_year_month(
         date_paid__lt=end_date,
     )
 
+    connectid_user_count_data = fetch_user_counts()
```
sravfeyn, line 83: Doesn't seem ideal that we are fetching all months data but we need only the selected month data, so may be pass a month param to the API?

commcare_connect/connect_id_client/main.py
calellowitz: There is not a way to infer it on CCC, but I agree we should cache it. Honestly, we could just save it in postgres. We only need to fetch the current month (and maybe the last one if we haven't fetched since the month switched). We might even be able to do that nightly on a task so loading the table never causes a network call.

calellowitz (changes requested):
Beyond the fetch comment, I think this metric isn't quite right. It should be cumulative to that point (the total number created through that month, not in that month). I don't have strong opinions whether to do that on the connect id side or here, but I believe that is the requirement.

calellowitz (approved):
This looks fine, but I do think long run we will want to save this in postgres. Since past data wont change over time, its silly to refetch it on page view

commcare_connect/reports/helpers.py
```diff
@@ -169,5 +176,23 @@ def get_table_data_for_year_month(
...
+                "delivery_type_name": delivery_type_name,
+                "flw_amount_paid": flw_amount_paid,
+                "avg_top_paid_flws": avg_top_paid_flws,
+            }
+        )
+
+    connectid_user_count = fetch_user_counts()
+    total_connectid_user_count = 0
+    total_eligible_user_count = 0
+    for group_key in visit_data_dict.keys():
```
calellowitz, line 191: If it is broken out by delivery type, what value is in each row? The same for a given month, or something else?

PR #500, opened 2025-02-25
Program manager should be able to increase the managed opportunity budget.

hemant10yadav:
## Product Description

The Program Manager should be able to modify the managed opportunity budget.

## Technical Summary

- The budget option for both adding a new user and updating an existing user has been moved to the same page.
- A new field, Opportunity Total Budget, has been added as requested in the ticket comments.
- Additionally, the "Add Budget for New User" option will be disabled for NM.
- The `additional user` field is removed from Edit opportunity page and moved in Add budget for new users as `Number of users`.

[CCCT-778](https://dimagi.atlassian.net/browse/CCCT-778)

Screenshot 2025-02-26 at 12 13 57
Screenshot 2025-02-26 at 12 14 07

### Safety Assurance

- Wrote test cases covering all scenarios.
- Tested locally.
- Tested on staging.

### QA Plan
NA

### Labels & Review

[CCCT-778]: https://dimagi.atlassian.net/browse/CCCT-778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/forms.py
```diff
@@ -131,6 +127,9 @@ def __init__(self, *args, **kwargs):
         self.initial["end_date"] = self.instance.end_date.isoformat()
         self.currently_active = self.instance.active
 
+        if self.instance.managed:
```
hemant10yadav, line 130: Disabled currency edit in case of managed opp as It is picked initially from program.

commcare_connect/opportunity/forms.py
```diff
@@ -103,9 +103,8 @@ def __init__(self, *args, **kwargs):
             Row(Field("delivery_type")),
             Row(Field("description")),
             Row(Field("short_description")),
-            Row(Field("currency")),
             Row(
-                Field("additional_users", wrapper_class="form-group col-md-6 mb-0"),
+                Field("currency", wrapper_class="form-group col-md-6 mb-0"),
```
calellowitz, line 107: Why move currency to be inline here?

calellowitz (commented):
This generally seems fine, but can you speak a bit more about the rationale behind some of these changes?

> The budget option for both adding a new user and updating an existing user has been moved to the same page.
> Additionally, the "Add Budget for New User" option will be disabled for NM.

Why are they able to add budget some ways but not other ways? Or do those actually do slightly different things. I can't tell how equivalent these options are and am wondering a bit if moving them closer together will be more confusing.

commcare_connect/opportunity/forms.py
```diff
@@ -663,6 +662,94 @@ def save(self):
             OpportunityClaim.objects.filter(pk__in=selected_users).update(end_date=end_date)
 
 
+class AddBudgetNewUsersForm(forms.Form):
+    add_users = forms.IntegerField(
+        required=False,
+        label="Number Of Users",
+        help_text="New Budget = Existing Budget + ∑ (Amount × Max Total × No. of Users) across all payment units.",
```
calellowitz, line 669: nit: the summation symbol is probably not known to everyone hard to look up. It might be clearer to just use the word sum.

calellowitz (commented):
Left one more code comment. One overall question is about the other form, "add budget existing users." Does that allow network managers to still add budget to the program? Or does it only allow them to assign already allocated budget?

hemant10yadav:
> This generally seems fine, but can you speak a bit more about the rationale behind some of these changes?
> 
> > The budget option for both adding a new user and updating an existing user has been moved to the same page.
> > Additionally, the "Add Budget for New User" option will be disabled for NM.
> 
> Why are they able to add budget some ways but not other ways? Or do those actually do slightly different things. I can't tell how equivalent these options are and am wondering a bit if moving them closer together will be more confusing. Additionally, if we do want to do it this way, can we hide the option rather than disabling it, and is there a good reason you can't do both at once?

Add Budget for Existing Users – This option allows NM to allocate additional budget to a specific user if the opportunity budget permits.
Add Budget for New Users – This option allows the PM to allocate additional budget to opportunity if the program budget permits.

The "Add Budget for New Users" option adds more budget to the opportunity, which can only be done by the PM. That's why I have disabled this option.

commcare_connect/opportunity/forms.py
hemant10yadav: The "Additional Users" field has been moved to the budget page and it was inline with the "End Date" field. Since the "Currency" field is only three characters long, I placed it in the same row where "Additional Users" used to be. Placing the "Currency" field on the next line and taking up an entire row wouldn't be ideal, especially when there is still available space.

calellowitz:
Thanks for the explanation. I think it would still be better to hide rather than disable (unless there is a good reason to have done it this way), and the help text comment is still valid, but nothing blocking if those are problematic for some reason.

commcare_connect/opportunity/forms.py
hemant10yadav: Changed [10bdc92](https://github.com/dimagi/commcare-connect/pull/500/commits/10bdc921f56794f74bd71305ceff59d8da2e8b64)

pxwxnvermx (approved):

calellowitz (approved):

PR #503, opened 2025-02-26
Add Date Range filtering to Dashboard

pxwxnvermx:
## Product Description
This PR removes the year and month filters and adds support for filtering using custom date ranges.

image

image

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-779)

This PR also adds refactors for the way grouping in queryset works.

### Safety story
- Tested on staging

### Automated test coverage

- Added test to cover the new changes.

### Labels & Review

### DIFF
--- a/commcare_connect/reports/helpers.py
+++ b/commcare_connect/reports/helpers.py
@@ -1,19 +1,6 @@
 from collections import defaultdict
 
-from django.db.models import (
-    Avg,
-    Case,
-    Count,
-    DurationField,
-    ExpressionWrapper,
-    F,
-    Max,
-    OuterRef,
-    Q,
-    Subquery,
-    Sum,
-    When,
-)
+from django.db import models
 from django.db.models.functions import ExtractDay, TruncMonth
 from django.utils.timezone import now
 
@@ -62,16 +49,18 @@ def get_table_data_for_year_month(
         group_values.append("delivery_type_name")
 
     max_visit_date = (
-        UserVisit.objects.filter(completed_work_id=OuterRef("id"), status=VisitValidationStatus.approved)
+        UserVisit.objects.filter(completed_work_id=models.OuterRef("id"), status=VisitValidationStatus.approved)
         .values_list("visit_date")
         .order_by("-visit_date")[:1]
     )
-    time_to_payment = ExpressionWrapper(F("payment_date") - Subquery(max_visit_date), output_field=DurationField())
+    time_to_payment = models.ExpressionWrapper(
+        models.F("payment_date") - models.Subquery(max_visit_date), output_field=models.DurationField()
+    )
     visit_data = (
         CompletedWork.objects.annotate(
-            filter_date=Case(
-                When(status_modified_date__isnull=True, then=F("date_created")),
-                default=F("status_modified_date"),
+            filter_date=models.Case(
+                models.When(status_modified_date__isnull=True, then=models.F("date_created")),
+                default=models.F("status_modified_date"),
             )
         )
         .filter(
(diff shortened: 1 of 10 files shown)


pxwxnvermx:
@calellowitz @sravfeyn bumping for review.

commcare_connect/reports/helpers.py
```diff
@@ -48,143 +64,106 @@ def get_table_data_for_year_month(
...
-                "nm_amount_paid": nm_amount_paid,
-                "nm_other_amount_paid": sum(nm_other_amount_paid_data.values()),
-                "avg_top_paid_flws": sum(avg_top_flw_amount_paid_data.values()),
-            }
-        )
-    return data
+        flw_amount_paid = sum(sum_total_users.values())
+        avg_top_paid_flws = sum(sorted(sum_total_users.values(), reverse=True)[:top_five_percent_len])
+        visit_data_dict[group_key].update({"flw_amount_paid": flw_amount_paid, "avg_top_paid_flws": avg_top_paid_flws})
+    return list(visit_data_dict.values())
```
sravfeyn, line 169: This method is getting quite longer, it becomes difficult to read/modify latter. It will add value to refactor it into a sort of class with separate methods for various colulms/queries.

For e.g. 
```
class VisitDataHelper:

def get_data(**filters):

def get_filter_kwargs()

def get_payment_data()

def get_visit_data()
```

Not exactly like that, but whatever to break it into smaller pieces will improve readability.

commcare_connect/static/js/vendors.js
```diff
@@ -3,5 +3,15 @@ import Alpine from 'alpinejs';
 import './htmx';
 import 'htmx.org/dist/ext/loading-states';
 
+import flatpickr from 'flatpickr';
+import monthSelectPlugin from 'flatpickr/dist/esm/plugins/monthSelect';
+import 'flatpickr/dist/flatpickr.css';
+import 'flatpickr/dist/plugins/monthSelect/style.css';
+window.monthPickr = (id, flatPickrConfig = {}, monthPluginConfig = {}) =>
```
sravfeyn, line 10: Have you tried altering the default style to look more similar to our site?

commcare_connect/reports/helpers.py
```diff
@@ -13,22 +11,40 @@
...
     opportunity=None,
     country_currency=None,
 ):
-    year = year or datetime.now().year
-    _, month_end = calendar.monthrange(year, month or 1)
-    start_date = make_aware(datetime(year, month or 1, 1))
-    end_date = make_aware(datetime(year, month or 12, month_end))
+    from_date = from_date or now().date()
+    to_date = to_date if to_date and to_date <= now().date() else now().date()
+    timeseries = get_month_series(from_date, to_date)
```
sravfeyn, line 29: `timeseries` is just `months`, right?

commcare_connect/reports/helpers.py
pxwxnvermx: I will try to address this in a future PR.

sravfeyn (approved):

hemant10yadav (approved):

commcare_connect/reports/helpers.py
calellowitz: plus one to this comment. not blocking but this is a very hard method to follow and review

commcare_connect/reports/views.py
```diff
@@ -359,15 +359,32 @@ def __init__(self, *args, **kwargs):
...
-                Column("year", css_class="col-md-4"),
-                Column("month", css_class="col-md-4"),
+                Column("from_date", css_class="col-md-4"),
+                Column("to_date", css_class="col-md-4"),
             ),
-            Row(Column("by_delivery_type", css_class="col-md-4")),
+            Row(Column("group_by_delivery_type", css_class="col-md-4")),
         )
 
+        if not self.data:
```
calellowitz, line 368: what is this doing and why is it now necessary?

commcare_connect/reports/views.py
```diff
@@ -359,15 +359,32 @@ def __init__(self, *args, **kwargs):
...
+            self.data = self.data.copy() if self.data else {}
+            today = date.today()
+            default_from = today - timedelta(days=30)
+            self.data["to_date"] = today.strftime("%Y-%m")
+            self.data["from_date"] = default_from.strftime("%Y-%m")
+            self.form.is_bound = True
```
calellowitz, line 374: where is `self.form` defined and can we use it to do the validation that the `to_date` must be later rather than just ignoring it in the data fetch method if it is invalid

calellowitz (approved):
nothing blocking but one question on validation and agree with Sravan's suggestion to refactor

commcare_connect/reports/views.py
pxwxnvermx: This is used to configure the filter's defaults. It detects that the filter was not set in a request and gives it a default from and to date.

commcare_connect/reports/views.py
pxwxnvermx: This form is configured automatically from the fields provided on the FilterSet.

PR #508, opened 2025-03-06
Ensure the Transaction is Committed Before Celery Runs download_user_visit_attachments.

hemant10yadav:
## Technical Summary

We have been seeing a high volume of UserVisit.DoesNotExist errors on [Sentry](https://dimagi.sentry.io/issues/6302069322/?project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=0)

This PR ensures that `download_user_visit_attachments.delay(user_visit.id)` is only triggered after the transaction is successfully committed using `transaction.on_commit()`. This guarantees that the Celery task only runs once the `UserVisit` record is available in the database.

## Safety Assurance

This change does not alter any data or logic beyond ensuring the function is called at the right time.

### Labels & Review

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -318,7 +318,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
                 completed_work.save()
 
     update_payment_accrued(opportunity, [user.id])
-    download_user_visit_attachments.delay(user_visit.id)
+    transaction.on_commit(lambda: download_user_visit_attachments.delay(user_visit.id))
 
 
 def get_or_create_deliver_unit(app, unit_data):
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -1,4 +1,5 @@
 import datetime
+from functools import partial
 
 from django.db import transaction
 from django.db.models import Count, Q
@@ -318,7 +319,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
                 completed_work.save()
 
     update_payment_accrued(opportunity, [user.id])
-    transaction.on_commit(lambda: download_user_visit_attachments.delay(user_visit.id))
+    transaction.on_commit(partial(download_user_visit_attachments.delay, user_visit.id))
 
 
 def get_or_create_deliver_unit(app, unit_data):


commcare_connect/form_receiver/processor.py
```diff
@@ -318,7 +318,7 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
                 completed_work.save()
 
     update_payment_accrued(opportunity, [user.id])
-    download_user_visit_attachments.delay(user_visit.id)
+    transaction.on_commit(lambda: download_user_visit_attachments.delay(user_visit.id))
```
pxwxnvermx, line 321: The import for the transaction module is missing.

hemant10yadav: It is already present [here](https://github.com/dimagi/commcare-connect/blob/ce6054e39de99eeda7d34b2ff5814d735336a3ad/commcare_connect/form_receiver/processor.py#L3C5-L3C34)

pxwxnvermx (approved):
LGTM, Do you have insights on why the user visits are not getting created and the failing transactions?

hemant10yadav:
> LGTM, Do you have insights on why the user visits are not getting created and the failing transactions?

I don't think the transaction is failing in these cases, as I don't see any exception or the log message we added in the view. Instead, I believe the task is running before the transaction commits, possibly due to nested transaction blocks behaving unpredictably in this scenario.
Whatever the case, this is the safer option, as it ensures everything commits before the function is called.

commcare_connect/form_receiver/processor.py
calellowitz, line 321: nit: it looks like `functools.partial` is the preferred way to handle passing arguments based on [the docs](https://docs.djangoproject.com/en/5.1/topics/db/transactions/#performing-actions-after-commit)

calellowitz (approved):
This looks great, the commit callback is really nifty

commcare_connect/form_receiver/processor.py
hemant10yadav: Thanks added [e59fc2f](https://github.com/dimagi/commcare-connect/pull/508/commits/e59fc2f23b1154faa87a7668a64497e613ba5b9f)

pxwxnvermx (approved):

PR #513, opened 2025-03-10
Make visit import async

sravfeyn:
## Product Description

This import has failed due to timeout on gunicorn worker. ([sentry](https://dimagi.sentry.io/issues/6257795631/?project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=6)). Likely it's taking a bit more time, so I have made this async

PRing into similar PR where I added celery util needed by this (https://github.com/dimagi/commcare-connect/pull/492)

## Technical Summary
Simple sync to async process. 

### Safety story

Will test on staging

### Automated test coverage

Existing tests are modified to test.

### QA Plan
NA

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tasks.py
+++ b/commcare_connect/opportunity/tasks.py
@@ -360,3 +360,19 @@ def bulk_update_payments_task(self, opportunity_id: int, headers: list[str], row
         messages = [f"Payment Import failed: {e}"] + getattr(e, "invalid_rows", [])
 
     set_task_progress(self, f"<br>".join(messages), is_complete=True)
+
+
+@celery_app.task(bind=True)
+def bulk_update_visit_status_task(self, opportunity_id: int, headers: list[str], rows: list[list]):
+    from commcare_connect.opportunity.visit_import import ImportException, bulk_update_visit_status
+
+    set_task_progress(self, "Visit Verification Import is in porgress.")
+    try:
+        status = bulk_update_visit_status(opportunity_id, headers, rows)
+        messages = [f"Visit status updated successfully for {len(status)} visits."]
+        if status.missing_users:
+            messages.append(status.get_missing_message())
+    except ImportException as e:
+        messages = [f"Visit status import failed: {e}"] + getattr(e, "invalid_rows", [])
+
+    set_task_progress(self, f"<br>".join(messages), is_complete=True)
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -88,6 +88,7 @@ from commcare_connect.opportunity.tables import (
 from commcare_connect.opportunity.tasks import (
     add_connect_users,
     bulk_update_payments_task,
+    bulk_update_visit_status_task,
     create_learn_modules_and_deliver_units,
     generate_catchment_area_export,
     generate_deliver_status_export,
@@ -104,7 +105,6 @@ from commcare_connect.opportunity.visit_import import (
     ImportException,
     bulk_update_catchments,
     bulk_update_completed_work_status,
-    bulk_update_visit_status,
     get_exchange_rate,
     get_imported_dataset,
     update_payment_accrued,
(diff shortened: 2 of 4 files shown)


sravfeyn:
@calellowitz Bumping for review

pxwxnvermx (approved):

commcare_connect/opportunity/views.py
```diff
@@ -429,18 +429,18 @@ def download_export(request, org_slug, task_id):
...
-        message = f"Visit status updated successfully for {len(status)} visits."
-        if status.missing_visits:
-            message += status.get_missing_message()
-        messages.success(request, mark_safe(message))
-    if opportunity.managed:
-        return redirect("opportunity:user_visit_review", org_slug, pk)
-    return redirect("opportunity:detail", org_slug, pk)
+        imported_data = get_imported_dataset(file, file_format)
+        rows = list(imported_data)
+        result = bulk_update_visit_status_task.delay(opportunity.pk, imported_data.headers or [], rows)
```
calellowitz, line 442: same comment about passing rows here as payment, but blocking in this case. I think given the size visit imports can be, having a hard limit, and ideally not passing the rows at all is important.

commcare_connect/opportunity/tasks.py
```diff
@@ -360,3 +360,19 @@ def bulk_update_payments_task(self, opportunity_id: int, headers: list[str], row
         messages = [f"Payment Import failed: {e}"] + getattr(e, "invalid_rows", [])
 
     set_task_progress(self, f"<br>".join(messages), is_complete=True)
+
+
+@celery_app.task(bind=True)
+def bulk_update_visit_status_task(self, opportunity_id: int, headers: list[str], rows: list[list]):
```
calellowitz, line 366: same comment on the other one about locking. I think its important only one of these is running per opportunity to avoid race conditions in the accrual calculations

calellowitz (changes requested):
few small things but looks fine overall

commcare_connect/templates/tailwind/components/upload_progress_bar.html
```diff
@@ -14,25 +14,79 @@
...
-           style="width: 80%; background-image: linear-gradient(to right, #8b5cf6, #a78bfa, #8b5cf6); background-size: 200% 100%;"></div>
+    <div class="w-full">
+      <div class="w-full h-2 bg-gray-200 rounded-full overflow-hidden mb-4">
+        <div
+          class="h-full bg-purple-600 rounded-full animate-[progress_2s_linear_infinite]"
+          style="width: 80%; background-image: linear-gradient(to right, #8b5cf6, #a78bfa, #8b5cf6); background-size: 200% 100%;">
+        </div>
+      </div>
+      <p class="w-full text-sm text-gray-700">
+        {% if progress.message %}
```
sravfeyn, line 25: I think this change got missed in tailwind migration. Originally added [here](https://github.com/dimagi/commcare-connect/pull/492/files#diff-5de9a1b7e069a39ac2833ba3a23b329db52860d819b9e25901900bc7c4caeb74)

zandre-eng (approved):
Code overall looks fine, though I am in agreement with having a hard limit set for import size.

commcare_connect/opportunity/views.py
```diff
@@ -504,17 +503,17 @@ def download_export(request, org_slug, task_id):
...
     opportunity = get_opportunity_or_404(org_slug=org_slug, pk=pk)
     file = request.FILES.get("visits")
-    redirect_url = reverse("opportunity:worker_list", args=(org_slug, pk)) + "?active_tab=delivery"
+    redirect_url = reverse("opportunity:worker_list", args=(org_slug, pk))
 
     file_format = get_file_extension(file)
     if file_format not in ("csv", "xlsx"):
         messages.error(request, f"Invalid file format. Only 'CSV' and 'XLSX' are supported. Got {file_format}")
-        return redirect(redirect_url)
+        return redirect(redirect_url + "?active_tab=delivery")
```
calellowitz, line 511: nit: it is probably slightly cleaner to use url constructors rather than appending strings

commcare_connect/opportunity/views.py
```diff
@@ -504,17 +503,17 @@ def download_export(request, org_slug, task_id):
...
     else:
-        imported_data = get_imported_dataset(file, file_format)
-        rows = list(imported_data)
-        result = bulk_update_visit_status_task.delay(opportunity.pk, imported_data.headers or [], rows)
-        return redirect(f"{redirect_url}?export_task_id={result.id}")
+        file_path = f"{opportunity.pk}_{datetime.datetime.now().isoformat}_visit_import"
+        saved_path = default_storage.save(file_path, ContentFile(file.read()))
```
calellowitz, line 514: why do we need to read from one file like object into another here?

commcare_connect/opportunity/views.py
```diff
@@ -504,17 +503,17 @@ def download_export(request, org_slug, task_id):
...
+        result = bulk_update_visit_status_task.delay(opportunity.pk, saved_path, file_format)
+        return redirect(f"{redirect_url}?active_tab=delivery&export_task_id={result.id}")
```
calellowitz, line 516: same comment about url constructurs especially as they get longer and have more query parameters. it is a little easier to read.

calellowitz (approved):
I had one questions and a few suggestions. None blocking, but it does look like there are some merge conflicts that will need to be addressed as well

commcare_connect/opportunity/views.py
sravfeyn: Updated to save directly

PR #514, opened 2025-03-10
Add demo_users permissions for page access

pxwxnvermx:
## Product Description
This moves the demo users page access to be behind a permission and removes the need for assigning the superuser role to users to them access this page.

### Labels & Review


sravfeyn:
It doesn't seem ideal to have to create a dedicated permission to whitelist a specific page. Are there any other ways? (like a similar existing permission that this falls under or staff role or something else)

calellowitz (approved):
I assume you tested this locally? I wasn't familiar with the django permissions models, but this looks pretty clean

calellowitz:
> It doesn't seem ideal to have to create a dedicated permission to whitelist a specific page. Are there any other ways? (like a similar existing permission that this falls under or staff role or something else)

I think we are intentionally moving in the opposite direction as that. Giving people access only to what they need, rather than needing to make people superusers or django admins to get access to restricted pages. We could potentially make this a bit more broad if we think we will want to include other special pages, but given how simple this is, erring on the side of more granularity doesn't seem like a bad thing. We can always rename or expand it the next time a use case comes up.

Charl1996 (approved):

Charl1996:
[This PR](https://github.com/dimagi/commcare-connect/pull/728) contains the content of this PR, so this PR can technically be closed.

PR #515, opened 2025-03-12
Make 'Justification' field mandatory for NMs

hemant10yadav:
## Product Description

Screenshot 2025-03-12 at 10 27 33

Screenshot 2025-03-13 at 11 39 08

## Technical Summary

Added mandatory justification for flagged visits in managed opportunity cases. This applies both when manually approving visits and when importing visits via file [CCCT-836.](https://dimagi.atlassian.net/browse/CCCT-836)

### Safety story

Tested locally

### Automated test coverage

Existing tests are passing. Added new test cases for different scenarios.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tests/test_views.py
+++ b/commcare_connect/opportunity/tests/test_views.py
@@ -192,11 +192,9 @@ def test_approve_visit(
         opportunity=opportunity, opportunity_access=access, flagged=True, status=VisitValidationStatus.pending
     )
     user = MembershipFactory.create(organization=opportunity.organization).user
-    print("opportunity.organization.slug", opportunity.organization.slug)
     approve_url = reverse("opportunity:approve_visit", args=(opportunity.organization.slug, visit.id))
     client.force_login(user)
     response = client.post(approve_url, {"justification": justification}, follow=True)
-    print(response.status_code)
     visit.refresh_from_db()
     assert visit.status == VisitValidationStatus.approved
     expected_redirect_url = None
--- a/commcare_connect/opportunity/tests/test_views.py
+++ b/commcare_connect/opportunity/tests/test_views.py
@@ -23,6 +23,7 @@ from commcare_connect.opportunity.tests.factories import (
 from commcare_connect.organization.models import Organization
 from commcare_connect.program.tests.factories import ManagedOpportunityFactory, ProgramFactory
 from commcare_connect.users.models import User
+from commcare_connect.users.tests.factories import MembershipFactory
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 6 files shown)


commcare_connect/opportunity/tests/test_views.py
```diff
@@ -169,3 +170,50 @@ def test_add_budget_existing_users_for_managed_opportunity(
...
+    organization,
+    opportunity,
+):
+    justification = "Justification test."
+    access = OpportunityAccessFactory(opportunity=opportunity)
+    visit = UserVisitFactory.create(
+        opportunity=opportunity, opportunity_access=access, flagged=True, status=VisitValidationStatus.pending
+    )
+    user = MembershipFactory.create(organization=opportunity.organization).user
+    print("opportunity.organization.slug", opportunity.organization.slug)
```
sravfeyn, line 195: Should the print statements be removed?

commcare_connect/opportunity/visit_import.py
```diff
@@ -150,16 +151,24 @@ def _bulk_update_visit_status(opportunity: Opportunity, dataset: Dataset):
...
                 if justification and justification != visit.justification:
                     visit.justification = justification
                     changed = True
+
                 if changed:
                     to_update.append(visit)
                 user_ids.add(visit.user_id)
+
+            if missing_justifications:
+                raise ImportException(get_missing_justification_message(missing_justifications))
```
sravfeyn, line 170: A separate method may not be warranted, since it's just one line and it's not reused anywhere.

commcare_connect/templates/opportunity/opportunity_detail.html
```diff
@@ -436,10 +436,17 @@ <h1 class="modal-title fs-5 text-white">{% translate "Import Verified Visits" %}
             <input class="form-control" type="file" id="importFile" name="visits" required>
             <div class="col-auto">
               <span id="importFileHelp" class="form-text">
-                {% blocktrans %}
-                  The file must contain at least the "Visit ID" and "Status" column.
-                  The import is case-insensitive.
-                {% endblocktrans %}
+                {% if opportunity.managed %}
```
sravfeyn, line 439: Could be simplified
{% blocktrans %}
    The file must include "Visit ID", "Status"{% if opportunity.managed %}, and "Justification"{% endif %}. Import is case-insensitive.
{% endblocktrans %}

commcare_connect/opportunity/visit_import.py
hemant10yadav: I have also used this function in [tests](https://github.com/dimagi/commcare-connect/pull/515/files#diff-1a7f59351bf1d9a918232b5795c79ea61f9cdf6c0f424e6428f5da540afedfcdR626) also to assert the error message.

commcare_connect/templates/opportunity/opportunity_detail.html
hemant10yadav: 'blocktrans' doesn't allow other block tags, but I have modified it in the way you suggested.

calellowitz: nit: a normal `{% trans %}` tag is probably more appropriate here since these arent multiline strings (blocks) anymore. You could also render the full string on the backend to a variable and then just add the variable, which might clean up the template a bit. This is a bit hard to read

calellowitz (approved):
non blocking comment to maybe help clean up that translated text section. otherwise looks good

commcare_connect/templates/opportunity/opportunity_detail.html
hemant10yadav: Thanks moved the string in the view.

PR #526, opened 2025-03-24
Async update_payment_accrued for Imports

pxwxnvermx:
## Product Description
Added a celery task for the update_payment_accrued function for bulk imports. The update_payment_accrued task times out in a web worker. Doing this task in a celery task will let the import process go without timing out, and the payment accrual can happen asynchronously independent of the upload.

Payment accrual code performance is very bad due to multiple queries made during the `approved_count` and `completed_count` calculations for CompletedWork. A better solution would be to improve those functions but to unblock the users right now, we have to settle for an intermediate solution while we figure out how to improve that calculation.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-870)

### Automated test coverage
- This code involves turning a well-tested function into a task.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tasks.py
+++ b/commcare_connect/opportunity/tasks.py
@@ -4,6 +4,7 @@ import logging
 import httpx
 from allauth.utils import build_absolute_uri
 from django.conf import settings
+from django.core.cache import cache
 from django.core.files.base import ContentFile
 from django.core.files.storage import default_storage
 from django.db import transaction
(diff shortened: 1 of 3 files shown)


calellowitz:
test failure looks real

commcare_connect/opportunity/tasks.py
```diff
@@ -357,3 +358,15 @@ def generate_catchment_area_export(opportunity_id: int, export_format: str):
...
     save_export(dataset, export_tmp_name, export_format)
     return export_tmp_name
+
+
+@celery_app.task()
+def bulk_update_payment_accrued(opportunity_id, user_ids: list):
+    """Updates payment accrued for completed and approved CompletedWork instances."""
+    access_objects = OpportunityAccess.objects.filter(opportunity=opportunity_id, user__in=user_ids, suspended=False)
+    for access in access_objects:
+        with cache.lock(f"update_payment_accrued_lock_{access.id}", timeout=900):
```
calellowitz, line 368: thank you for including a lock

commcare_connect/opportunity/tasks.py
```diff
@@ -357,3 +358,15 @@ def generate_catchment_area_export(opportunity_id: int, export_format: str):
...
+            completed_works = access.completedwork_set.exclude(status=CompletedWorkStatus.rejected).select_related(
```
calellowitz, line 369: is this modeled off another query that does the same?

calellowitz (approved):
assuming the answer to the question is yes, looks good

commcare_connect/opportunity/tasks.py
pxwxnvermx: yes, this is the same as the `update_payment_accrued` function.

hemant10yadav (approved):

PR #533, opened 2025-03-27
Program Manager Permission Issue

hemant10yadav:
## Product Description

We had a security issue where an NM organization was also a PM organization in another program. Our current code allowed the NM organization to access all PM functionalities in an opportunity where it was only an NM. This PR fixes that issue.

## Technical Summary
[CCCT-880](https://dimagi.atlassian.net/browse/CCCT-880)

### Safety story

Tested the views on local.

### Automated test coverage
All previous tests are passing

### QA Plan
Raised a [ticket](https://dimagi.atlassian.net/browse/QA-7602) for QA

### Labels & Review

[CCCT-880]: https://dimagi.atlassian.net/browse/CCCT-880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/program/helpers.py
+++ b/commcare_connect/program/helpers.py
@@ -16,6 +16,7 @@ from django.db.models import (
 )
 from django.db.models.functions import Cast, Coalesce, Round
 
+from commcare_connect.cache import quickcache
 from commcare_connect.opportunity.models import UserVisit, VisitValidationStatus
 from commcare_connect.program.models import ManagedOpportunity, Program
 
@@ -130,3 +131,16 @@ def get_delivery_performance_report(program: Program, start_date, end_date):
     )
 
     return managed_opportunities
+
+
+@quickcache(vary_on=["current_org_slug", "opp_id"], timeout=24 * 60 * 60)
+def is_org_manages_opportunity_via_program(current_org_slug, opp_id: int):
+    """
+    Check if the current organization manages the given managed opportunity
+    through any of its programs.
+    """
+    programs = Program.objects.filter(organization__slug=current_org_slug)
+    if programs:
+        return ManagedOpportunity.objects.filter(program__in=programs, id=opp_id).exists()
+
+    return False
--- a/commcare_connect/program/helpers.py
+++ b/commcare_connect/program/helpers.py
@@ -139,8 +139,4 @@ def is_org_manages_opportunity_via_program(current_org_slug, opp_id: int):
     Check if the current organization manages the given managed opportunity
     through any of its programs.
     """
-    programs = Program.objects.filter(organization__slug=current_org_slug)
-    if programs:
-        return ManagedOpportunity.objects.filter(program__in=programs, id=opp_id).exists()
-
-    return False
+    return ManagedOpportunity.objects.filter(program__organization__slug=current_org_slug, id=opp_id).exists()
--- a/commcare_connect/templates/opportunity/invoice_list.html
+++ b/commcare_connect/templates/opportunity/invoice_list.html
@@ -3,6 +3,7 @@
 {% load static %}
 {% load crispy_forms_tags %}
 {% load i18n %}
+{% load permissions %}
 
 {% block title %}{{ request.org }} - Invoices{% endblock title %}
 
--- a/commcare_connect/templates/opportunity/invoice_list.html
+++ b/commcare_connect/templates/opportunity/invoice_list.html
@@ -19,7 +19,7 @@
 {% block content %}
   <h2>Invoices</h2>
   <hr />
-  {% if not request.org_membership.is_program_manager %}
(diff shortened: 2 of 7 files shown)


commcare_connect/program/helpers.py
```diff
@@ -130,3 +131,16 @@ def get_delivery_performance_report(program: Program, start_date, end_date):
...
+def is_org_manages_opportunity_via_program(current_org_slug, opp_id: int):
+    """
+    Check if the current organization manages the given managed opportunity
+    through any of its programs.
+    """
+    programs = Program.objects.filter(organization__slug=current_org_slug)
+    if programs:
+        return ManagedOpportunity.objects.filter(program__in=programs, id=opp_id).exists()
+
+    return False
```
pxwxnvermx, line 146: This can be simplified to just a single query to check if the managed opportunity exists through a program managed by current organization.
```suggestion
    return ManagedOpportunity.objects.filter(program__organization__slug=current_org_slug, id=opp_id).exists()
```

hemant10yadav: Thanks - [8b6514f](https://github.com/dimagi/commcare-connect/pull/533/commits/8b6514f19ed01aff9e82ea3054a9f5e367d0ba25)

commcare_connect/program/helpers.py
```diff
@@ -130,3 +131,12 @@ def get_delivery_performance_report(program: Program, start_date, end_date):
     )
 
     return managed_opportunities
+
+
+@quickcache(vary_on=["current_org_slug", "opp_id"], timeout=24 * 60 * 60)
```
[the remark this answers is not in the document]
hemant10yadav: This function will be called each time we use `request|is_program_manager_for_opportunity:opportunity.id` in templates or in views. Given its potential frequency of execution, caching could still provide performance benefits by reducing redundant database queries.

commcare_connect/web/templatetags/permissions.py
```diff
@@ -0,0 +1,20 @@
...
+
+@register.filter
+def is_program_manager_for_opportunity(request, opp_id):
+    if request.user.is_superuser:
+        return True
+
+    current_org = getattr(request, "org", None)
+    org_membership = getattr(request, "org_membership", None)
+
+    is_admin = bool(current_org and org_membership and org_membership.is_admin)
```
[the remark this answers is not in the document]
hemant10yadav: I wanted to ensure that this function consistently returns a `boolean` value. Otherwise, it could introduce subtle bugs in places where a boolean is expected, but the function returns `None` for some cases.

commcare_connect/templates/opportunity/user_visit_review.html
```diff
@@ -21,7 +22,7 @@ <h2 class="mb-2">User Visit Review</h2>
 <form method="get" x-data="" x-ref="reviewFilterForm">
   {% crispy review_filter.form %}
 </form>
-{% if request.org_membership.is_admin and request.org.program_manager %}
```
pxwxnvermx, line 24: Same comment as above.

commcare_connect/templates/opportunity/visit_verification.html
```diff
@@ -145,7 +146,7 @@ <h2 class="mb-2">Images</h2>
     </dl>
     {% endif %}
     <div id="action-buttons" style="margin-bottom: 25px;">
-      {% if request.org_membership.is_admin and request.org.program_manager or request.org_membership.is_viewer %}
+      {% if request.org_membership.is_admin and request|is_program_manager_for_opportunity:visit.opportunity.id or request.org_membership.is_viewer %}
```
pxwxnvermx, line 149: Same comment as above.

pxwxnvermx (commented):
Just some comments related to the filter. We can reduce the number of times the filter is called by evaluating the permissions once in the views and then passing the value down through the view context.

commcare_connect/templates/opportunity/invoice_list.html
```diff
@@ -19,7 +20,7 @@
 {% block content %}
   <h2>Invoices</h2>
   <hr />
-  {% if not request.org_membership.is_program_manager %}
+  {% if not request|is_program_manager_for_opportunity:opportunity.id %}
```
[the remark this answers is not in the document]
hemant10yadav: I see your point, and I partially agree. However, the main reason I implemented this as a template filter was to avoid explicitly passing the value from the view. This allows it to be accessed directly within the template when needed. Additionally, I think calling this function shouldn’t have a performance impact since the request object properties are already pre-cached—or do you see any potential concerns in this regard?

hemant10yadav:
Closing this one as a new PR has been raised for the same issue: https://github.com/dimagi/commcare-connect/pull/677

PR #539, opened 2025-04-02
IntegrityError: Duplicate Key Violation in Currency Exchange Rates

hemant10yadav:
## Technical Summary

This PR resolves a race condition that was causing an `IntegrityError` due to duplicate currency exchange rate entries. The issue has been addressed by replacing the `create` method with `get_or_create` to prevent duplicate inserts.

[Sentry Error](https://dimagi.sentry.io/issues/6494300041/?project=4505635339829248&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=10)

[CCCT-912](https://dimagi.atlassian.net/browse/CCCT-912)

### Safety story
This change is low risk as it only modifies the way exchange rate entries are created, ensuring that duplicates are avoided without altering existing functionality.

### Automated test coverage

A test has been added to confirm that the issue was caused by a race condition. The test now passes successfully after the fix in the get_exchange_rate function.

### QA Plan
NA

### Labels & Review

[CCCT-912]: https://dimagi.atlassian.net/browse/CCCT-912?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/visit_import.py
+++ b/commcare_connect/opportunity/visit_import.py
@@ -376,7 +376,8 @@ def get_exchange_rate(currency_code, date=None):
         rate = rates.get(currency_code)
         if not rate:
             raise ImportException("Rate not found for opportunity currency")
-        ExchangeRate.objects.create(currency_code=currency_code, rate=rate, rate_date=rate_date)
+        # get_or_create to avoid race condition.
+        ExchangeRate.objects.get_or_create(currency_code=currency_code, rate=rate, rate_date=rate_date)
 
     return rate
 
--- a/commcare_connect/opportunity/tests/test_visit_import.py
+++ b/commcare_connect/opportunity/tests/test_visit_import.py
@@ -1,9 +1,11 @@
 import datetime
 import random
 import re
+from concurrent.futures import ThreadPoolExecutor
 from datetime import timedelta
 from decimal import Decimal
 from itertools import chain
+from unittest.mock import patch
 
 import pytest
 from django.db import transaction
@@ -16,6 +18,7 @@ from commcare_connect.opportunity.models import (
     CatchmentArea,
     CompletedWork,
     CompletedWorkStatus,
+    ExchangeRate,
     Opportunity,
     OpportunityAccess,
     Payment,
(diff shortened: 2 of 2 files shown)


pxwxnvermx (approved):

commcare_connect/opportunity/visit_import.py
```diff
@@ -376,7 +376,8 @@ def get_exchange_rate(currency_code, date=None):
         rate = rates.get(currency_code)
         if not rate:
             raise ImportException("Rate not found for opportunity currency")
-        ExchangeRate.objects.create(currency_code=currency_code, rate=rate, rate_date=rate_date)
+        # get_or_create to avoid race condition.
+        ExchangeRate.objects.get_or_create(currency_code=currency_code, rate=rate, rate_date=rate_date)
```
calellowitz, line 380: Won't this error if the rate changes? We want to at least catch that and give a message (and possibly catch it and continue the upload).

calellowitz (changes requested):
One quick question, but I think this might be missing the handling logic for some cases

commcare_connect/opportunity/visit_import.py
hemant10yadav: I think it's better to let this fail hard here, because the key question is: which rate should we return?

1. If we return the rate from the external API, there might be no corresponding entry in our database.
2. If we return the rate from our database, we're effectively ignoring the inconsistency.

Even if we attempt to return `None`, it would still likely raise an `ImportException` or an invalid operand error, since we're performing` amount / rate.` So returning None doesn't really help—the process would still fail.

A hard failure here makes the issue immediately visible, with full context captured in Sentry. That gives us the opportunity to properly investigate and make an informed decision on how we want to handle such cases going forward.

calellowitz: Yeah, definitely think we should fail, I just think it would be good to show the user a message that says effectively "import failed, try again." If that happens in the hard failure scenario already, that's fine. I just don't want them to know there was an issue so they can reupload.

PR #555, opened 2025-04-10
Implement header dropdowns

sravfeyn:
## Product Description

This implements the top header menus.

### Labels & Review


calellowitz (commented):
This looks like it is fixing bugs in the first PR. If that is the case, it might be easier to combine them

PR #557, opened 2025-04-11
Base tailwind table and styling and Opportunity List Table

sravfeyn:
This implements the base utils required to create a tailwind table


commcare_connect/templates/tailwind/components/pagination.html
```diff
@@ -0,0 +1,18 @@
+<div class="flex items-center gap-2 max-w-fit text-brand-deep-purple text-sm">
```
calellowitz, line 1: how is this used? basically how would we make a paginated table?

commcare_connect/utils/tables.py
```diff
@@ -0,0 +1,9 @@
+import django_tables2 as tables
+
+
+class BaseTailwindTable(tables.Table):
```
calellowitz, line 4: nit: this applies to most of the code, and can be done at the end if easier, but I don't think we want everything named tailwind like it is now. Much like we don't call existing tables "BootstrapTable," I don't think it adds anything, because when this gets merged, it will be the only table type

calellowitz (commented):
Hard to fully understand without seeing it in use in an example table, but this looks reasonable

sravfeyn:
> Hard to fully understand without seeing it in use in an example table, but this looks reasonable

Should I club this with an OpportunityList page for the above reason?

commcare_connect/templates/tailwind/components/pagination.html
sravfeyn: I have updated the code. This is now removed with pagination now going into base_table itself

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+        <div class="relative inline-flex justify-center items-center group cursor-default">
+            <span>{label}</span>
+            <i class="fa-regular fa-circle-question text-xs text-slate-400 ml-1 cursor-help"></i>
+            <div class="fixed hidden group-hover:block z-50 pointer-events-none -translate-x-[15%] -translate-y-[70%] transform">
+                <div class="absolute top-full left-1/2 -translate-x-1/2 w-0 h-0 border-l-8 border-r-8 border-t-8 border-l-transparent border-r-transparent border-t-white"></div>
+                <div class="relative bg-white w-28 rounded p-2 text-slate-500 text-xs whitespace-normal break-words">
+                    {tooltip_text}
+                </div>
+            </div>
+        </div>
```
sravfeyn, line 570: This could potentially be turned into a component by introducing classes like `tooltip-wrapper, tooltip-content` etc. But this is the only place its used.

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+        self.use_view_url = False
+
+    index = IndexColumn()
+    opportunity = tables.Column(accessor="name")
+    entity_type = tables.TemplateColumn(
+        verbose_name="",
+        orderable=False,
+        template_code="""
+                <div class="flex justify-start text-sm font-normal text-brand-deep-purple w-fit"
+                     x-data="{
```
sravfeyn, line 610: This could potentially be turned into an alpinejs/tailwind plugin, but this is also the only place it's used.

commcare_connect/templates/tailwind/pages/opportunities_list.html
```diff
@@ -0,0 +1,29 @@
+{% extends 'tailwind/base.html' %}
```
calellowitz, line 1: nit: template name change (opportunity -> opportunities) seems unnecessary

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,92 @@ def get_payment_report_data(opportunity: Opportunity):
             PaymentReportData(payment_unit.name, completed_work_count, user_payment_accrued, nm_payment_accrued)
         )
     return data, total_user_payment_accrued, total_nm_payment_accrued
+
+
+def get_opportunity_list_data(organization, program_manager=False):
```
calellowitz, line 177: is this performant? just making sure we aren't building something that will get too slow very quickly

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+    def render_actions(self, record):
+        actions = [
+            {
+                "title": "View Opportunity",
+                "url": reverse("opportunity:detail", args=[record.organization.slug, record.id]),
+            },
+            {
+                "title": "View Workers",
+                "url": reverse("opportunity:detail", args=[record.organization.slug, record.id]),
+                # "url": reverse("opportunity:tw_worker_list", args=[record.organization.slug, record.id]),
```
calellowitz, line 734: do we need the commented out code?

commcare_connect/opportunity/views.py
```diff
@@ -156,28 +158,25 @@ def get_table_kwargs(self):
...
 
 
-class OpportunityList(OrganizationUserMixin, ListView):
-    model = Opportunity
-    paginate_by = 10
+class OpportunityList(OrganizationUserMixin, SingleTableMixin, TemplateView):
+    template_name = "tailwind/pages/opportunities_list.html"
+    paginate_by = 15
+    base_columns = ["program", "start_date", "end_date", "status", "opportunity"]
+    pm_columns = ["active_workers", "total_deliveries", "verified_deliveries", "worker_earnings"]
```
calellowitz, line 165: why are these for PMs only? They seem useful to everyone

calellowitz (commented):
one nit and a few questions but this generally seems fine.

commcare_connect/opportunity/tables.py
sravfeyn: Yeah, this is updated in the followup PR [here](https://github.com/dimagi/commcare-connect/pull/573/files#diff-04b178fa6d9c4a6615700a161f3de90ed4ba7650547e3590dafd5eb4e17019a2R735)

commcare_connect/opportunity/helpers.py
sravfeyn: Yes, it's likely this is not performant enough. We are setting aside time to look into this and other places with these kind of queries

commcare_connect/templates/tailwind/pages/opportunities_list.html
sravfeyn: Agreed, I will try to be wary of these for next pages

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,96 @@ def get_payment_report_data(opportunity: Opportunity):
...
+            "uservisit",
+            filter=Q(uservisit__status=VisitValidationStatus.pending),
+            distinct=True,
+        ),
+        total_accrued=Coalesce(
+            Sum("opportunityaccess__payment_accrued", distinct=True), Value(0), output_field=DecimalField()
+        ),
+        total_paid=Coalesce(
+            Sum(
+                "opportunityaccess__payment__amount_usd",
```
sravfeyn, line 202: Todo: To be converted to local currency.

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,96 @@ def get_payment_report_data(opportunity: Opportunity):
...
+            When(Q(active=True) & Q(end_date__gte=today), then=Value(0)),  # Active
+            When(Q(active=True) & Q(end_date__lt=today), then=Value(1)),  # Ended
+            default=Value(2),  # Inactive
+            output_field=IntegerField(),
+        ),
+    )
+
+    if program_manager:
+        queryset = queryset.annotate(
+            active_workers=Count(
```
sravfeyn, line 241: QFR: Why not include active_workers for NM as well?

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,96 @@ def get_payment_report_data(opportunity: Opportunity):
...
+                "opportunityaccess",
+                filter=Q(
```
sravfeyn, line 243: Nit: Can we do `total - inactive` to get the active to avoid this queries.

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,96 @@ def get_payment_report_data(opportunity: Opportunity):
...
+                    | Exists(
+                        CompletedModule.objects.filter(
+                            opportunity_access=OuterRef("opportunityaccess"),
+                            date__gte=three_days_ago,
+                        )
+                    )
+                ),
+                distinct=True,
+            ),
+            total_deliveries=Count("opportunityaccess__completedwork", distinct=True),
```
sravfeyn, line 259: Could be. `Sum("opportunityaccess__completedwork__saved_completed_count", distinct=True)`

commcare_connect/opportunity/helpers.py
```diff
@@ -151,3 +172,96 @@ def get_payment_report_data(opportunity: Opportunity):
...
+            verified_deliveries=Count(
```
sravfeyn, line 260: Could be. Sum("opportunityaccess__completedwork__saved_approved_count", distinct=True)

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+        all_classes = f"{base_classes} {extra_classes}".strip()
+        return format_html('<div class="{}">{}</div>', all_classes, value)
+
+    def render_opportunity(self, value):
+        return self._render_div(value, extra_classes="justify-start")
+
+    def render_program(self, value):
+        return self._render_div(value if value else "--", extra_classes="justify-start")
+
+    def render_start_date(self, value):
```
sravfeyn, line 688: Todo; Turns out this doesn't render the date format consistently.

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+    def render_inactive_workers(self, value):
+        return self._render_div(value, extra_classes=self.stats_style)
+
+    def render_pending_approvals(self, value):
+        return self._render_div(value, extra_classes=self.stats_style)
+
+    def render_payments_due(self, value):
+        if value is None:
+            value = 0
+        return self._render_div(f"${value}", extra_classes=self.stats_style)
```
sravfeyn, line 723: Todo; should use the local currency.

commcare_connect/opportunity/tables.py
```diff
@@ -549,3 +553,285 @@ def date_with_time_popup(table, date):
...
+        verbose_name=header_with_tooltip("Active Workers", tooltip_text="Delivered forms in the last 3 days.")
+    )
+    total_deliveries = tables.Column(
+        verbose_name=header_with_tooltip("Total Deliveries", "Total services delivered so far.")
+    )
+    verified_deliveries = tables.Column(
+        verbose_name=header_with_tooltip("Verified Deliveries", "Approved service deliveries.")
+    )
+    worker_earnings = tables.Column(
+        verbose_name=header_with_tooltip("Worker Earnings", "Approved service deliveries"), accessor="total_accrued"
```
sravfeyn, line 769: Todo; update text.

commcare_connect/opportunity/helpers.py
sravfeyn: AFR: NM needs to see action oriented columns, PM needs to see performance columns

PR #562, opened 2025-04-15
Opportunity List Page

hemant10yadav:
### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tw_views.py
+++ b/commcare_connect/opportunity/tw_views.py
@@ -2,11 +2,18 @@ from django import forms
 from django.http import HttpResponse
 from django.shortcuts import redirect, render
 from django.template import Template, Context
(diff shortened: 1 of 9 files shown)


commcare_connect/opportunity/tw_views.py
```diff
@@ -3717,4 +3724,45 @@ def opportunity_worker_payment(request, org_slug=None, opp_id=None):
...
+        table = super().get_table(**kwargs)
+
+        validated_order = self.get_validated_order_by()
+
+        if validated_order:
+            table.order_by = validated_order
+        else:
+            table.order_by = ("-status", "start_date", "end_date")
+
+        return table
```
sravfeyn, line 3768: @hemant10yadav Why do we need to manage this sorting? Django-tables already does.

PR #563, opened 2025-04-15
Opportunity Dashboard

hemant10yadav:
### Labels & Review


commcare_connect/opportunity/tw_views.py
```diff
@@ -3766,3 +3790,155 @@ def get_table(self, **kwargs):
             table.order_by = ("-status", "start_date", "end_date")
 
         return table
+
+
+def opportunity_dashboard(request, org_slug=None, opp_id=None):
```
sravfeyn, line 3795: You will need correct permission decorators here.

PR #565, opened 2025-04-15
Worker Tables

hemant10yadav:
### Labels & Review


commcare_connect/web/templatetags/sort_link.py
```diff
@@ -46,3 +49,51 @@ def update_query_params(context, **kwargs):
...
+
+
+@register.simple_tag(takes_context=True)
+def sortable_header(context, field, label, use_htmx=False):
+    request = context["request"]
+    current_sort = next_sort = None
+    icon_element = '<i class="fa-solid ml-1 {}"></i>'
+
+
+    if use_htmx:
```
sravfeyn, line 61: Are you adding HTMX implementation? I am confused since we decided we are only going to use server side pagination/sorting for now

hemant10yadav: Just an option if we needed at some place by default it uses normal refresh page.

commcare_connect/web/templatetags/sort_link.py
```diff
@@ -46,3 +49,51 @@ def update_query_params(context, **kwargs):
         updated[key] = value
 
     return updated.urlencode()
+
+
+@register.simple_tag(takes_context=True)
+def sortable_header(context, field, label, use_htmx=False):
```
sravfeyn, line 55: I don't understand why we need to change the logic from `sort_link`, all you will need to is just update the icons/css-classes. What am I missing?

commcare_connect/opportunity/tw_views.py
```diff
@@ -3857,7 +2694,211 @@ def opportunity_worker_payment(request, org_slug=None, opp_id=None):
...
+        if validated_order:
+            table.order_by = validated_order
+        else:
+            table.order_by = ("-status", "start_date", "end_date")
+
+        return table
+
+@org_member_required
+def opportunity_dashboard(request, org_slug=None, opp_id=None):
+    opp = get_opportunity_dashboard_data(opp_id=2).first()
```
sravfeyn, line 2742: I think you forgot to fix this after your local testing?

commcare_connect/web/templatetags/sort_link.py
hemant10yadav: There are cases where the table is initially loaded via HTMX and then sorting and pagination are handled on the server side. That’s why we don’t use request.path, since it returns the path of the HTMX request itself.

For tab switching, we definitely need to use an HTMX call—otherwise, we’d have to load all the tables on the page at once or refresh the entire page when switching tabs, which isn’t ideal.

To handle these scenarios more gracefully, I’ve updated the implementation.

commcare_connect/opportunity/views.py
```diff
@@ -420,14 +420,23 @@ def review_visit_export(request, org_slug, pk):
 def export_status(request, org_slug, task_id):
     task_meta = AsyncResult(task_id)._get_task_meta()
     status = task_meta.get("status")
+    print(status)
```
sravfeyn, line 423: @hemant10yadav Please be careful to not add these statements in the PR

commcare_connect/templates/tailwind/components/upload_progress_bar.html
```diff
@@ -0,0 +1,39 @@
...
+  {% if progress.error %}
+    <div class="rounded-lg bg-red-100 text-red-700 px-4 py-3">
+      <p class="font-medium">Export Failed</p>
+      <p class="text-sm mt-1">{{ progress.error }}</p>
+    </div>
+
+  {% elif not progress.complete %}
+    <div class="w-full h-2 bg-gray-200 rounded-full overflow-hidden mb-4">
+      <div class="h-full bg-purple-600 rounded-full animate-[progress_2s_linear_infinite]"
+           style="width: 80%; background-image: linear-gradient(to right, #8b5cf6, #a78bfa, #8b5cf6); background-size: 200% 100%;"></div>
```
sravfeyn, line 18: We shouldn't be using raw styles. If we aren't able to style these, we can get yume-labs help

commcare_connect/templates/tailwind/components/upload_progress_bar.html
```diff
@@ -0,0 +1,39 @@
...
+    </div>
+    <p class="text-sm text-gray-700">Preparing export, hang tight...</p>
+
+    <style>
```
sravfeyn, line 22: Same as [above](https://github.com/dimagi/commcare-connect/pull/565/files#r2050122678)

commcare_connect/organization/views.py
```diff
@@ -107,3 +110,64 @@ def add_credential_view(request, org_slug):
...
+        add_credential_task.delay(org.pk, credential, users)
+    return redirect("organization:org_member_view", org_slug)
+
+@org_admin_required
+def org_member_view(request, org_slug=None, opp_id=None):
+    org = get_object_or_404(Organization, slug=org_slug)
+    org_form = TwOrganizationChangeForm(instance=request.org)
+    form = TwMembershipForm(None, organization=org)
+
+    print(request.POST)
```
sravfeyn, line 137: To be removed

commcare_connect/opportunity/views.py
```diff
@@ -420,14 +420,23 @@ def review_visit_export(request, org_slug, pk):
...
     progress = {
         "complete": status == "SUCCESS",
     }
     if status == "FAILURE":
         progress["error"] = task_meta.get("result")
+
+    template = (
+        "tailwind/components/upload_progress_bar.html"
+        if request.GET.get("ui") == "new"
```
sravfeyn, line 432: Let's not mix old and new views like this, we better just create a separate view for the new UI.

commcare_connect/opportunity/tests/test_views.py
```diff
@@ -215,3 +221,280 @@ def test_approve_visit(
...
+    CompletedModuleFactory(
+        xform_id=uuid.uuid4(),
+        opportunity=opportunity,
+        opportunity_access=access,
+        user=access.user,
+        module=module2,
+        date=three_days_ago,
+        duration=timedelta(hours=2),
+    )
+    print(OpportunityAccess.objects.filter(id=access.id).count())
```
sravfeyn, line 321: To be removed

commcare_connect/opportunity/tw_views.py
```diff
@@ -1619,9 +1666,123 @@ def worker_payments(request, org_slug=None, opp_id=None):
...
+    file = request.FILES.get("visits")
+    try:
+        status = bulk_update_visit_status(opportunity, file)
+    except ImportException as e:
+        messages.error(request, e.message)
+    else:
+        message = f"Visit status updated successfully for {len(status)} visits."
+        if status.missing_visits:
+            message += status.get_missing_message()
+        messages.success(request, mark_safe(message))
```
sravfeyn, line 1732: @hemant10yadav It looks like you have removed this check from tailwind version of the view, what's the reason for it?

```
if status.over_limit_count:
            messages.warning(request, mark_safe(status.get_over_limit_count_message()))
```

hemant10yadav: @sravfeyn I didn’t understand the main branch view, and this view seems almost the same.

PR #583, opened 2025-05-05
Implement Payment History, Rollback and Delivered count breakup

sravfeyn:
## Product Description

Implements
- Pulldown table for payment-history/rollback feature
- Pulldown table for delivered column breakup

### Labels & Review


commcare_connect/opportunity/helpers.py
```diff
@@ -150,6 +150,14 @@ def get_annotated_opportunity_access_deliver_status(opportunity: Opportunity):
                     distinct=True,
                 ),
                 completed=F("approved") + F("rejected") + F("pending") + F("over_limit"),
+                duplicate=Count(
```
sravfeyn, line 153: @pxwxnvermx I am not sure if this is the right way to calculate duplicate count, in terms of the spec. Could you please validate this?

pxwxnvermx: Yeah, that looks good to me. Another way the duplicates can be calculated is to filter for CompletedWork which have `saved_completed_count` greater than 1.

commcare_connect/opportunity/tables.py
```diff
@@ -164,10 +164,10 @@ class OpportunityPaymentTable(OrgContextTable):
 
     def render_view_payments(self, record):
         url = reverse(
-            "opportunity:user_payments_table",
-            kwargs={"org_slug": self.org_slug, "opp_id": record.opportunity.id, "pk": record.pk},
+            "opportunity:worker_list",
+            kwargs={"org_slug": self.org_slug, "opp_id": record.opportunity.id},
         )
-        return mark_safe(f'<a href="{url}">View Details</a>')
+        return mark_safe(f'<a href="{url}&active_tab=payments">View Details</a>')
```
calellowitz, line 170: is the url field guaranteed to have query params? otherwise this could be an invalid url

commcare_connect/opportunity/tables.py
```diff
@@ -176,15 +176,6 @@ class Meta:
         empty_text = "No user have payments accrued yet."
 
 
-class UserPaymentsTable(tables.Table):
```
calellowitz, line 179: why don't we need a payments table anymore?

commcare_connect/templates/tailwind/components/worker_page/deliver_column.html
```diff
@@ -0,0 +1,42 @@
+{% load i18n %}
+{% load static %}
+{% load crispy_forms_tags %}
+
+<div class="relative underline"
+     x-data="{
+         isOpen: false,
+         positionMenu() {
```
calellowitz, line 8: This appears to be used in multiple templates in this PR (and I imagine more outside this PR). Can it be pulled into some kind of shared js?

calellowitz (commented):
left a few questions and suggestsions

commcare_connect/opportunity/tables.py
sravfeyn: That was a mistake! Fixed it.

commcare_connect/opportunity/tables.py
sravfeyn: This list is now displayed as payment-history pulldown. You can click on 'Last Paid' to see the history in this [staging link](https://connect-staging.dimagi.com/a/nsaxenaccc/opportunity/160/worker_list/?active_tab=payments). Though, it doesn't display the additional columns (payment operator)

There is some out of scope work that displays the rest of the columns as well as part of bulk payments UI.

commcare_connect/templates/tailwind/components/worker_page/deliver_column.html
sravfeyn: Yeah, this is a good one to be pulled out, there are few more things like this. I am hoping to do these JS components in a post UI work PR.

sravfeyn:
@calellowitz Updated PR per your comments.

commcare_connect/templates/tailwind/components/worker_page/deliver_column.html
calellowitz: Is there a reason not to do this now?

calellowitz:
New changes look good, but still curious about the old comment

commcare_connect/opportunity/tables.py
```diff
@@ -1141,15 +1145,15 @@ class WorkerDeliveryTable(OrgContextTable):
 
     id = tables.Column(visible=False)
     index = IndexColumn()
-    user = tables.Column(orderable=False, verbose_name="Name")
+    user = tables.Column(orderable=False, verbose_name="Name", footer="Total")
     suspended = SuspendedIndicatorColumn()
     last_active = tables.Column()
-    payment_unit = tables.Column(orderable=False)
+    payment_unit = tables.Column(accessor="payment_unit__name", orderable=False)
```
hemant10yadav, line 1151: Is this accessor correct here? It looks like payment unit name is directly accessible using `payment_unit`, based on [this change](https://github.com/dimagi/commcare-connect/pull/583/files#diff-396037c378e64671cea6adc7185fa4f3aee28145c1fcf11f27bd597e494cd9a5R163)

sravfeyn: Fixed

commcare_connect/templates/tailwind/components/worker_page/deliver_column.html
sravfeyn: There are some open questions that needs to be answered to do these better, that I think could be part of latter refactor.
- Where should these common utils be placed in?
- Should they be included in the base-template?
- But since they aren't used in all pages, may be we will need a require_js setup. What are our options for this setup.

etc.

sravfeyn:
@calellowitz I believe [this](https://github.com/dimagi/commcare-connect/pull/583#discussion_r2094944090) was the only pending comment. Let me know if this is good to go, or we should rather address that comment as part of this PR

PR #588, opened 2025-05-06
Prevent Null Strings in Serialization

zandre-eng:
## Product Description

No user-facing changes.

## Technical Summary

Relevant Slack discussion thread [here](https://dimagi.slack.com/archives/C05APTXCTSM/p1746453947390889?thread_ts=1746430482.305919&cid=C05APTXCTSM).

This PR resolves an issue where the `send_delivery` endpoint would sometimes serialize an `OpportunityAccess` instance and send a "null" string for the `reason` field in the JSON data. A likely cause for this could be that a file gets imported when doing a bulk update for work/visit status and the file contains null for the `reason` field. This then gets converted into a string, causing `reason` to be saved as "null".

A check has now been added to ensure that a "null" string doesn't get sent in the JSON payload. If the instance contains a "null" string then this simply gets set to `None`. Additionally, the imports for user visits has also been tweaked so as to default to `None` if a "null" string is included for the `reason` field.

### Safety story

- Local testing

Small formatting change to the JSON payload for the `send_delivery` endpoint.

### Automated test coverage

Unit tests do exist.

### QA Plan

No QA planned.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -247,6 +247,7 @@ class CompletedWorkSerializer(serializers.ModelSerializer):
     deliver_unit_slug = serializers.CharField(source="payment_unit.pk")
     visit_date = serializers.DateTimeField(source="completion_date")
     flags = serializers.SerializerMethodField()
+    reason = serializers.SerializerMethodField()
 
     class Meta:
         model = CompletedWork
@@ -272,6 +273,11 @@ class CompletedWorkSerializer(serializers.ModelSerializer):
             for slug, reason in visit.get("flags", []):
                 flags[slug] = reason
         return flags
+
+    def get_reason(self, obj):
+        if obj.reason in ["None", "null"]:
+            return None
+        return obj.reason
 
 
 class PaymentSerializer(serializers.ModelSerializer):
--- a/commcare_connect/opportunity/api/serializers.py
+++ b/commcare_connect/opportunity/api/serializers.py
@@ -247,7 +247,6 @@ class CompletedWorkSerializer(serializers.ModelSerializer):
     deliver_unit_slug = serializers.CharField(source="payment_unit.pk")
     visit_date = serializers.DateTimeField(source="completion_date")
     flags = serializers.SerializerMethodField()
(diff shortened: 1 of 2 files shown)


pxwxnvermx (approved):

commcare_connect/opportunity/api/serializers.py
```diff
@@ -273,6 +274,11 @@ def get_flags(self, obj):
                 flags[slug] = reason
         return flags
 
+    def get_reason(self, obj):
+        if obj.reason in ["None", "null"]:
```
calellowitz, line 278: One of the goals of making this web change was to remove exactly these checks from the mobile side. Do you think we can clean up existing data (since we will be ignoring it anyway), and then be able to remove this check?

calellowitz (commented):
One question/suggestion

commcare_connect/opportunity/api/serializers.py
zandre-eng: I've made an interesting discovery that I've now confirmed with mobile. The mobile code is saving this field as a string [here](https://github.com/dimagi/commcare-android/blob/5667bceb4a5871e36f7052e00e2d7d4ae7f1fe62/app/src/org/commcare/android/database/connect/models/ConnectJobDeliveryRecord.java#L93). This change was made as part of the current merge work. So I believe the server is actually sending the field correctly, but then mobile is converting `null` to `"null"`.

Given the above, I don't think we need to do any data clean up, and the function added here in the serializer can safely be removed.

@calellowitz what are your thoughts on the above? We could probably still do a manual sanity check on the server to verify that this holds true for all `OpportunityAccess` the instances if we need to.

calellowitz: I agree with both of those conclusions

zandre-eng: The commit has been reverted ([48cd625](https://github.com/dimagi/commcare-connect/pull/588/commits/48cd62517747caa34d878dd0fd5c8cc42ed927c0)). I'll still keep the other changes in this PR as they're still beneficial to have.

calellowitz (approved):

PR #589, opened 2025-05-07
[New UI] Opp Create and Edit Forms

sravfeyn:
## Product Description

Forms for Create/Edit Opportunity
Screenshot 2025-05-09 at 12 12 38 PM
Screenshot 2025-05-09 at 12 12 54 PM

Screenshot 2025-05-09 at 12 12 23 PM

### Labels & Review


sravfeyn:
@calellowitz

calellowitz:
This PR is missing a description and screenshots

sravfeyn:
Added screenshots, and update title

commcare_connect/opportunity/forms.py
```diff
@@ -103,35 +103,54 @@ class Meta:
...
+                    Field("filter_country", wrapper_class="w-full"),
+                    Field("filter_credential", wrapper_class="w-full"),
+                    css_class="flex gap-2"
                 ),
+                Row(Field("users", wrapper_class="w-full"), css_class="col-span-2"),
+                css_class="grid grid-cols-2 gap-4 p-6 card_bg"
             ),
-            Submit("submit", "Submit"),
+            Row(
+                Submit("submit", "Submit", css_class="button button-md primary-dark"),
```
pxwxnvermx, line 146: Adding float-end to the button style will have the same effect as flex justify-end, also you can remove the row here since it contains a single element and it will be wrapped in a div by default.

commcare_connect/opportunity/forms.py
```diff
@@ -178,26 +194,39 @@ def __init__(self, *args, **kwargs):
...
-                Row(Field("learn_app")),
-                Row(Field("learn_app_description")),
-                Row(Field("learn_app_passing_score")),
-                data_loading_states=True,
+            Row(
+                HTML("<div class='col-span-2'><h6 class='title-sm'>Opportunity Details</h6>  <span class='hint'>Add the details of the opportunity. All fields are mandatory.</span> </div>"),
+                Column(
+                    Field("name"),
+                    Field("short_description"),
+                    Field("description"),
```
pxwxnvermx, line 202: Can we make the description a full width on the form?

commcare_connect/opportunity/forms.py
```diff
@@ -842,10 +897,10 @@ def __init__(self, *args, **kwargs):
                 </button>
                 """
             ),
-            Row(Field("payment_units")),
-            Field("max_total", wrapper_class="form-group col-md-4 mb-0"),
-            Field("max_daily", wrapper_class="form-group col-md-4 mb-0"),
-            Submit(name="submit", value="Submit"),
+            Row(
+                Submit(name="submit", value="Submit",css_class="button button-md primary-dark"),
```
pxwxnvermx, line 901: same thing about `float-end`

commcare_connect/opportunity/forms.py
```diff
@@ -103,35 +103,54 @@ class Meta:
...
-                Field("end_date", wrapper_class="form-group col-md-6 mb-0"),
+                HTML("<div class='col-span-2'><h6 class='title-sm'>Opportunity Details</h6>  <span class='hint'>Edit the details of the opportunity. All fields are mandatory.  </span> </div>"),
+                Column(
+                    Field("name", wrapper_class="w-full"),
+                    Field("short_description", wrapper_class="w-full"),
+                    Field("description", wrapper_class="w-full"),
+                ),
+                Column(
+                    Field("delivery_type"),
+                    Field("active", css_class=CHECKBOX_CLASS, wrapper_class="bg-slate-100 flex items-center justify-between p-4 rounded-lg"),
```
calellowitz, line 119: is there a way to pull out these styles that get repeated so often, there are a few that show up multiple times and it would be good to have an easy way to keep them in sync

commcare_connect/opportunity/forms.py
```diff
@@ -822,15 +855,37 @@ def __init__(self, *args, **kwargs):
...
+            Row(
+                Column(
+                    Field("name"),
+                    Field("description"),
+                ),
+                Column(
+                    Row(
+                        Field("start_date"),
+                        Field("end_date"),
+                        css_class="grid grid-cols-2 gap-4"),
```
calellowitz, line 867: same with these classes

calellowitz (approved):
no blocking comments, but a few suggestions and would be good to address Pawan's feedback as well

commcare_connect/opportunity/forms.py
sravfeyn: These are all full width already.

commcare_connect/opportunity/forms.py
sravfeyn: I don't think it gets wrapped in Div without Row. I have used float-end though, that's good idea.

commcare_connect/opportunity/forms.py
sravfeyn: `flex justify-end` is actually better than `float-end` and is supposed to work better in responsive layouts and interacts better with other elements than float-end.

PR #590, opened 2025-05-07
[New UI] Payment Unit And Budget Forms

hemant10yadav:
## Product Description
This Pr adds the Following Forms.
1. Payment unit Form
2. Add budget Form (Existing user and new users both)

Screenshot 2025-05-08 at 01 54 18
Screenshot 2025-05-08 at 01 53 42
Screenshot 2025-05-08 at 01 53 52
Screenshot 2025-05-08 at 01 54 04

https://dimagi.atlassian.net/browse/CCCT-1112

### Labels & Review


commcare_connect/templates/opportunity/add_visits_existing_users.html
```diff
@@ -1,147 +1,220 @@
...
-        </div>
-      {% endif %}
+        // set tab on url
+        const url = new URL(window.location);
+        url.searchParams.set('active_tab', tab);
+        window.history.pushState({}, '', url);
+
+         // initially load the form once
+        if (tab === 'new_users' && !this.newUsersLoaded) {
+          console.log(tab)
```
sravfeyn, line 28: Remove this?

commcare_connect/opportunity/forms.py
```diff
@@ -822,31 +819,35 @@ def __init__(self, *args, **kwargs):
...
-                event.detail.successful && location.reload();
-                this.removeAttribute('disabled'); this.innerHTML='Sync Deliver Units';""
-                hx-disabled-elt="this"
-                hx-on:click="this.innerHTML=&quot;<span class=\\
-                'spinner-border spinner-border-sm'></span> Syncing...&quot;;">
-                <span id="sync-text">Sync Deliver Units</span>
-                </button>
+            Div(
+                Row(
+                    Column(Field("name"), Field("description")),
```
pxwxnvermx, line 824: This `Column` class can be removed as this will only add a wrapping div with no style classes assigned to it.

commcare_connect/opportunity/forms.py
```diff
@@ -822,31 +819,35 @@ def __init__(self, *args, **kwargs):
...
                 """
-            ),
-            Row(Field("payment_units")),
-            Field("max_total", wrapper_class="form-group col-md-4 mb-0"),
-            Field("max_daily", wrapper_class="form-group col-md-4 mb-0"),
-            Submit(name="submit", value="Submit"),
-        )
+                    ))),
+                    css_class="grid grid-cols-2 gap-4 p-6 card_bg"),
+                Row(Submit("submit", "Submit", css_class="button button-md primary-dark"), css_class="flex justify-end")
```
pxwxnvermx, line 849: Adding `float-end` to the button style will have the same effect as `flex justify-end`.

commcare_connect/templates/partial_form.html
```diff
@@ -1,5 +1,6 @@
 {% load static %}
 {% load crispy_forms_tags %}
+{% load tailwind_filters %}
```
pxwxnvermx, line 3: is this needed here?

commcare_connect/opportunity/forms.py
```diff
@@ -822,31 +819,35 @@ def __init__(self, *args, **kwargs):
...
+                , css_class="flex flex-col gap-4"))
```
pxwxnvermx, line 850: Can we check if this is needed? The form fields have margins already added to them so they should look fine without this.

commcare_connect/opportunity/forms.py
```diff
@@ -730,9 +728,8 @@ def __init__(self, *args, **kwargs):
 
         self.helper = FormHelper(self)
         self.helper.layout = Layout(
-            Row(Field("add_users")),
-            Row(Field("total_budget")),
-            Submit(name="submit", value="Submit"),
+            Row(Field("add_users"), Field("total_budget"), css_class="grid grid-cols-2 gap-4"),
+            Row(Submit("submit", "Submit", css_class="button button-md primary-dark"), css_class="flex justify-end")
```
pxwxnvermx, line 732: same comment about using `float-end`

commcare_connect/opportunity/forms.py
hemant10yadav: I don't think it will work here as it is inside row not a Div which is a block element.

commcare_connect/opportunity/forms.py
hemant10yadav: No then the gap will not work between two cards.

commcare_connect/opportunity/forms.py
pxwxnvermx: I think the Row here might not be necessary, crispy adds divs around form elements, so you can remove the row and then add float-end to make the button float to the right side.

hemant10yadav: I think that will still be an issue—the button will be full width because of the flex properties on the outer elements, and we'll have to add an extra class to fix that.

pxwxnvermx: cool, that makes sense. I think other places that don't have `flex` on the parent can still benefit from using `float-end`, instead of a container with `flex justify-end` specified on them.

calellowitz (approved):
assuming pawan's comments are resolved this looks good

pxwxnvermx (approved):

pxwxnvermx (approved):

PR #593, opened 2025-05-09
[New UI] Optimize worker-deliver query

sravfeyn:
## Product Description

I have tested the previous query on superset, which took about 2 minutes, I broke this up into subqueries to mirror CTE (CTEs aren't supported in django without django-cte) avoiding previously bigger and complex joins. The optimized one took about 4-5 seconds.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -3,7 +3,9 @@ from datetime import timedelta
 
 from django.db.models import (
     Case,
+    CharField,
     Count,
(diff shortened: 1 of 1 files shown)


calellowitz (approved):
I wasn't able to totally confirm correctness, but it looks right, and the performance improvements sound great

calellowitz:
This query might benefit from tests as well, as it is quite complex

sravfeyn:
There are tests for this, already!

PR #594, opened 2025-05-12
[New UI] Optimize Dashboard Queries

hemant10yadav:
## Technical Summary
This PR introduces a `completed_learn_date` field on `OpportunityAccess` and includes a data migration to backfill this field for existing records.

In addition, query performance has been optimized by separating out the UserVisit-related logic. The join with the UserVisit table was identified as a performance bottleneck, so the queries have been restructured to avoid this expensive join. These optimized queries were tested on Superset and demonstrated significantly improved performance.

### Safety story
Tested changes locally to ensure correctness. Also validated the performance and accuracy of the updated queries using Superset.

### Automated test coverage

Added tests to verify that the completed_learn_date is correctly updated as expected.

### Labels & Review


sravfeyn:
Given that this introduces data and model migrations, and only impacts production performance, let's keep this PR for the end and exclude it from merging into ui-main for now and staging deploy workflow?

commcare_connect/opportunity/migrations/0074_opportunityaccess_completed_learn_date.py
```diff
@@ -0,0 +1,66 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def back_fill_completed_learn_date(apps, schema_editor):
+    Opportunity = apps.get_model("opportunity", "Opportunity")
+    OpportunityAccess = apps.get_model("opportunity", "OpportunityAccess")
+    CompletedModule = apps.get_model("opportunity", "CompletedModule")
+
+    BATCH_SIZE = 50
```
calellowitz, line 17: This batch seems unnecessarily small but fine for now. Next time I would think 100 or even 1000.

calellowitz (approved):
nothing blocking just for the next bulk update we can be a bit more aggressive

hemant10yadav:
Closing this one as I have separate oot the commits for this PR in two PRs #621 and #616

PR #601, opened 2025-05-14
[New UI] Avatar using initials

sravfeyn:
## Product Description

Implements initial based avatars

Screenshot 2025-05-14 at 5 55 24 PM

### Labels & Review

### DIFF
--- a/commcare_connect/web/templatetags/avatar.py
+++ b/commcare_connect/web/templatetags/avatar.py
@@ -1,0 +1,54 @@
+from django import template
+
+register = template.Library()
+
+
+def _calculate_initials_internal(name):
+    if not name:
+        # Guest
+        return "GU"
+
+    parts = name.split()
+    initials = ""
+    if len(parts) >= 2 and parts[0] and parts[-1]:
+        initials = (parts[0][0] + parts[-1][0]).upper()
+    elif len(parts) == 1 and parts[0]:
+        if len(parts[0]) >= 2:
+            initials = parts[0][:2].upper()
+        else:
+            initials = parts[0][0].upper()
+    elif name:
+        if len(name) >= 2:
+            initials = name[:2].upper()
+        else:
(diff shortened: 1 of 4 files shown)


commcare_connect/web/templatetags/avatar.py
```diff
@@ -0,0 +1,54 @@
...
+        if len(name) >= 2:
+            initials = name[:2].upper()
+        else:
+            initials = name[0].upper()
+
+    return initials if initials else "GU"
+
+
+@register.inclusion_tag("tailwind/components/avatar.html")
+def user_avatar(user, size="small", color_classes="bg-brand-mango text-white"):
```
calellowitz, line 30: not blocking but it feels more natural to have the template tag just handle the string part and the rest of the logic in the template itself

calellowitz (approved):
looks fine, but could be much simpler if we use one letter icons

commcare_connect/web/templatetags/avatar.py
sravfeyn: You mean each time repeat the full avatar HTML and just use `{{ request.user|initials}}`. That will not be DRY, right? 

The way we have it here, it's kind of a component now.

commcare_connect/web/templatetags/avatar.py
```diff
@@ -0,0 +1,54 @@
...
+    if size == "large":
+        size_classes = "w-16 h-16 text-xl font-bold tracking-widest"
+    else:
+        size_classes = "w-10 h-10 text-xs font-medium tracking-wider"
+
+    return {
+        "initials_text": initials_text,
+        "display_name": display_name,
+        "size_classes": size_classes,
+        "color_classes": color_classes,
```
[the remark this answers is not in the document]
sravfeyn: It was a requirement from Rama

PR #616, opened 2025-05-22
Added completed learn date

hemant10yadav:
## Technical Summary
This PR introduces a completed_learn_date field on OpportunityAccess and includes a data migration to backfill this field for existing records.
### Safety story
Tested locally

### Automated test coverage

Added test the update completed learn date function.

### Labels & Review


calellowitz (approved):
How does this interact with https://github.com/dimagi/commcare-connect/pull/594#

The code looks good but it appears to have some of the same work and seems likely to cause conflicts

hemant10yadav:
> How does this interact with #594
I will close that PR and raise a new one that includes only the query changes. We want to merge this into main so that the ui-main changes contain only the UI updates. This way, if we need to revert the ui-main changes, we can do so without affecting the migration work.

pxwxnvermx (approved):
LGTM. 🙌🏽

PR #619, opened 2025-05-22
Optimize by changing joins to subqueries

sravfeyn:
## Product Description

This optimizes opp-list queries by converting joins to subqueries. I have tested this on a domain that took around a minute, it takes a second or so with subqueries.

https://dimagi.atlassian.net/browse/CCCT-1260
### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -34,12 +34,13 @@ from commcare_connect.opportunity.models import (
     LearnModule,
     Opportunity,
     OpportunityAccess,
+    OpportunityClaimLimit,
     PaymentUnit,
     UserInvite,
     UserInviteStatus,
     UserVisit,
     VisitReviewStatus,
-    VisitValidationStatus, OpportunityClaimLimit,
+    VisitValidationStatus,
 )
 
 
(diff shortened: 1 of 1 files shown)


sravfeyn:
@calellowitz

commcare_connect/opportunity/helpers.py
```diff
@@ -308,17 +313,28 @@ def get_opportunity_list_data(organization, program_manager=False):
...
+            .annotate(total=Sum("saved_completed_count", distinct=True))
+            .values("total")[:1],
+            output_field=IntegerField(),
+        )
+
+        verified_deliveries_sq = Subquery(
+            CompletedWork.objects.filter(opportunity_access__opportunity_id=OuterRef("pk"))
+            .values("opportunity_access__opportunity_id")
+            .annotate(total=Sum("saved_approved_count", distinct=True))
+            .values("total")[:1],
```
calellowitz, line 328: This question is relevant to all the subqueries, but do these error if there are no completed works?

commcare_connect/opportunity/helpers.py
```diff
@@ -280,18 +281,22 @@ def get_opportunity_list_data(organization, program_manager=False):
     today = now().date()
     three_days_ago = now() - timedelta(days=3)
 
+    pending_approvals_sq = Subquery(
+        UserVisit.objects.filter(opportunity_access__opportunity_id=OuterRef("pk"), status="pending")
+        .values("opportunity_access__opportunity_id")
```
calellowitz, line 286: why is the value the opportunity id, rather than the visit id? won't this always count one if we count distinct  values in the annotation (since they all have the same opportunity id)

calellowitz (commented):
Couple questions on this one

commcare_connect/opportunity/helpers.py
sravfeyn: Since this comes before the annotate, the opp-id is only used for grouping by. And the following `count=Count("id", distinct=True)` counts distinct visit IDs, grouped by opp-id.

snopoke (approved):
Looks OK to me except for the lack of tests.

sravfeyn:
There is existing tests for this already, which is passing

PR #638, opened 2025-06-02
Fix a bug; add loading indicator

sravfeyn:
## Product Description

- Fixes a 500 bug
- Removes calculated columns from opp-list since it's slow
- Adds loading spinner to tab switching

Screenshot 2025-06-02 at 10 51 02 AM

### Labels & Review


pxwxnvermx (approved):
LGTM. Thanks

PR #654, opened 2025-06-04
Optimize location calculations

sravfeyn:
## Product Description

This fixes the long time taken around 44 seconds for visit detail loading when there are lot of visits in an opportunity like [this one](https://connect.dimagi.com/a/dimagi-gw-chc-2024-2025/opportunity/411/user_visits/2425/?).

The main strategy used is to filter out all the visits that are outside of 250m radius using boundingbox lat/long filtered on location field.

This improved performance for above visit to 1.5 seconds from 44 seconds.

### Labels & Review


pxwxnvermx (approved):
LGTM. Thanks

PR #657, opened 2025-06-09
Optimize Payment report; Move python cacls to db

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-1314

The existing version does the calcs in Python by fetching uservisit set for each completework to calculate payment_accrued  (via cw.payment_accrued), this gets extremely slow when there are lot of visits/cws. 

The optimization pushes these calcs by doing all aggs in SQL and using `saved_approved_count`.

### Labels & Review


commcare_connect/opportunity/helpers.py
```diff
@@ -315,25 +314,36 @@ def completed_work_status_total_subquery(status_value):
...
+        data.append(
+            PaymentReportData(
+                group["payment_unit__name"],
+                group["completed_work_count"],
+                user_payment_accrued,
+                nm_payment_accrued,
+            )
         )
-        completed_work_count = len(completed_works)
-        user_payment_accrued = sum([cw.payment_accrued for cw in completed_works])
```
sravfeyn, line 330: This is the key cause of slowness, each call `payment_accrued` fetches all its uservisits to calculate `approved_count` which has been replaced with saved_approved_count above and in DB

snopoke (approved):
A test case would be useful

sravfeyn:
Added a test, and realized that the method can be further simplified by making use of saved payment fields

sravfeyn:
@pxwxnvermx @calellowitz

PR #666, opened 2025-06-19
Multiple HQ's on Connect

pxwxnvermx:
## Product Description

https://github.com/user-attachments/assets/56b366f8-6958-43c8-b8d0-b3aa154decb7

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-1349)

### Automated test coverage
- Updated all automated tests to include the new additions.

### QA Plan
- Going to request QA for this ticket.
[QA Ticket Link](https://dimagi.atlassian.net/browse/QA-7878)

### Labels & Review

### DIFF
--- a/commcare_connect/utils/commcarehq_api.py
+++ b/commcare_connect/utils/commcarehq_api.py
@@ -55,7 +55,9 @@ def get_domains_for_user(api_key):
         f"{api_key.hq_server.url}/api/v0.5/user_domains/?limit=100",
         headers={"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"},
     )
-    if response.status_code != 200:
+    try:
+        response.raise_for_status()
+    except httpx.HTTPStatusError:
         raise CommCareHQAPIException(f"Failed to fetch domains: {response.text}")
     data = response.json()
     domains = [domain["domain_name"] for domain in data["objects"]]
@@ -85,7 +87,9 @@ async def _get_applications_for_domain(user_email, api_key, domain, hq_server_ur
 async def _get_commcare_app_json(client, domain):
     applications = []
     response = await client.get(f"/a/{domain}/api/v0.5/application/")
-    if response.status_code != 200:
+    try:
+        response.raise_for_status()
+    except httpx.HTTPStatusError:
         raise CommCareHQAPIException(f"Failed to fetch applications: {response.text}")
     data = response.json()
 


commcare_connect/opportunity/forms.py
```diff
@@ -408,170 +458,6 @@ def clean(self):
             return cleaned_data
 
 
-class OpportunityCreationForm(forms.ModelForm):
```
pxwxnvermx, line 411: This flow has not been maintained since we got the new Opportunity Creation flow. I have removed it in this PR, also removed any tests that were related to this.

commcare_connect/opportunity/migrations/0078_create_prod_hq_server.py
pxwxnvermx, line 1: This migration is used to create the default HQServer for the Opportunities that were created before this change.

config/settings/base.py
```diff
@@ -367,3 +367,5 @@
 MAPBOX_TOKEN = env("MAPBOX_TOKEN", default=None)
 
 OPEN_EXCHANGE_RATES_API_ID = env("OPEN_EXCHANGE_RATES_API_ID", default=None)
+
+OAUTH2_PROVIDER_APPLICATION_MODEL = "oauth2_provider.Application"
```
pxwxnvermx, line 371: This was added to fix a migration issue related to using the oauth2_provider Application model as a foreign key.

https://github.com/jazzband/django-oauth-toolkit/issues/240

commcare_connect/opportunity/migrations/0078_create_prod_hq_server.py
```diff
@@ -0,0 +1,36 @@
...
+from oauth2_provider.models import Application
+
+
+def create_prod_hq_server_and_update_objects(apps, schema_editor):
+    HQServer = apps.get_model("opportunity", "HQServer")
+    Opportunity = apps.get_model("opportunity", "Opportunity")
+    HQApiKey = apps.get_model("opportunity", "HQApiKey")
+    CommCareApp = apps.get_model("opportunity", "CommCareApp")
+
+    oauth_application = Application.objects.first()
```
calellowitz, line 15: I don't think ordering is technically guaranteed by postgres. If this is an existing application, it would be safer to fetch by `id`

commcare_connect/opportunity/views.py
```diff
@@ -228,11 +210,15 @@ def get_success_url(self):
...
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
-        kwargs["domains"] = get_domains_for_user(self.request.user)
         kwargs["user"] = self.request.user
         kwargs["org_slug"] = self.request.org.slug
         return kwargs
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["api_key_form"] = HQApiKeyCreateForm(auto_id="api_key_form_id_for_%s")
```
calellowitz, line 219: Is the `%s` literal or is something being interpolated there?

commcare_connect/opportunity/views.py
```diff
@@ -967,11 +953,62 @@ def send_message_mobile_users(request, org_slug=None, pk=None):
     )
 
 
+# used for loading api key dropdown
+@org_member_required
```
calellowitz, line 957: This should probably be in the `users` app since it is tied to the user rather than the opportunity. That is a bit weird given that it is in an opportunity specific page, but still makes sense to me since the data wouldn't change with different opportunities in the URL.

commcare_connect/opportunity/views.py
```diff
@@ -967,11 +953,62 @@ def send_message_mobile_users(request, org_slug=None, pk=None):
...
+def get_api_keys(request, org_slug=None):
+    hq_server = request.GET.get("hq_server")
+    if not hq_server:
+        return HttpResponse(
```
calellowitz, line 961: the strings here should be marked for translation (and elsewhere in the PR, just seeing it here first)

commcare_connect/opportunity/views.py
```diff
@@ -967,11 +953,62 @@ def send_message_mobile_users(request, org_slug=None, pk=None):
...
+                api_key.id,
+                api_key_hidden,
+            )
+        )
+    return HttpResponse("\n".join(options))
+
+
+# used for loading domain dropdown
+@org_member_required
+def get_domains(request, org_slug=None):
```
calellowitz, line 985: It's a little weird this view is in the opportunities as well. It should probably be whereever we move the server model.

commcare_connect/opportunity/views.py
```diff
@@ -967,11 +953,62 @@ def send_message_mobile_users(request, org_slug=None, pk=None):
...
+    hq_server = request.GET.get("hq_server")
+    api_key_id = request.GET.get("api_key")
+    if not hq_server or not api_key_id:
+        return HttpResponse(format_html("<option value='{}'>{}</option>", None, "Select an API Key to load domains."))
+
+    options = []
+    api_key = HQApiKey.objects.get(id=api_key_id, hq_server=hq_server, user=request.user)
```
calellowitz, line 992: is this guaranteed to be unique? I would expect some users to have multiple api keys for one server since they both expire and are often domain scoped

commcare_connect/opportunity/views.py
```diff
@@ -967,11 +953,62 @@ def send_message_mobile_users(request, org_slug=None, pk=None):
...
+    domains = get_domains_for_user(api_key)
```
calellowitz, line 993: Since API keys are domain scoped, this doesn't seem reliable

commcare_connect/utils/commcarehq_api.py
```diff
@@ -48,32 +50,39 @@ def refresh_access_token(user, force=False):
     return social_token
 
 
-def get_domains_for_user(user):
-    social_token = refresh_access_token(user)
+def get_domains_for_user(api_key):
     response = httpx.get(
-        f"{settings.COMMCARE_HQ_URL}/api/v0.5/user_domains/?limit=100",
-        headers={"Authorization": f"Bearer {social_token}"},
+        f"{api_key.hq_server.url}/api/v0.5/user_domains/?limit=100",
```
calellowitz, line 55: does this api allow api key? if so, what does it do when the key is domain scoped?

commcare_connect/opportunity/models.py
```diff
@@ -3,20 +3,31 @@
...
 from django.utils.timezone import now
 from django.utils.translation import gettext
+from oauth2_provider import settings
 
 from commcare_connect.organization.models import Organization
 from commcare_connect.users.models import User
 from commcare_connect.utils.db import BaseModel, slugify_uniquely
 
 
+class HQServer(models.Model):
```
calellowitz, line 20: This doesn't feel like it is an opportunity based model (and this file is already massive, so wouldn't hurt to break out anyway)

commcare_connect/opportunity/forms.py
calellowitz: nice

calellowitz (commented):
A few questions and suggestions, generally concerned about how we get full info from HQ with API keys that often don't have full access for a user (and we encourage the use of limited ones generally for security reasons, especially for superusers)

commcare_connect/utils/commcarehq_api.py
pxwxnvermx: yeah, that API allows api keys auth. When a domain scoped API key is given it returns all the domains for the user.  🤦🏽

commcare_connect/opportunity/views.py
pxwxnvermx: The %s is needed for automatic id generation of the field. There were two forms on the same page with similar fields that caused issues so this ensures that the ids generated are unique due to the prefix added here.

commcare_connect/opportunity/migrations/0078_create_prod_hq_server.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/666/commits/e8c25ff58ed227d01292032b64b7e0c375d77e1c

commcare_connect/opportunity/views.py
pxwxnvermx: https://github.com/dimagi/commcare-connect/pull/666/commits/586d38642657d3e203e4cc48c685d3e4078151be

commcare_connect/opportunity/models.py
pxwxnvermx: https://github.com/dimagi/commcare-connect/pull/666/commits/e8c25ff58ed227d01292032b64b7e0c375d77e1c

commcare_connect/users/views.py
```diff
@@ -164,8 +163,8 @@ def post(self, *args, **kwargs):
 
 
 # used for loading api key dropdown
-@org_member_required
-def get_api_keys(request, org_slug=None):
+@login_required
```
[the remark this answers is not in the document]
pxwxnvermx: This was earlier under the opportunities app where all urls get a `org_slug` parameter, it is not needed here. Also the `org_member_required` decorator assumes that the request contains the org_slug.

PR #668, opened 2025-06-23
Add Cloudwatch to Ansible Setup

zandre-eng:
## Product Description

No user-facing changes.

## Technical Summary

Link to ticket [here](https://dimagi.atlassian.net/browse/CCCT-1253).
Link to investigation doc [here](https://docs.google.com/document/d/1lR6oJQGpyXtOlBZbxGYnSLq7B0pXfBp97jKEA07uYuM/edit?tab=t.0).

This PR adds the necessary commands in Ansible for setting up and configuring Cloudwatch on an EC2 machine. The config has been set up to track the following items:
- CPU
- Memory
- Disk
- Processes

### Safety story

The Ansible commands for setting up Cloudwatch do have a block, so if something goes wrong then the deploy won't be blocked.

### Automated test coverage

N/A

### QA Plan

No QA planned.

### Labels & Review

### DIFF
--- a/deploy/roles/connect/tasks/main.yml
+++ b/deploy/roles/connect/tasks/main.yml
@@ -107,50 +107,46 @@
   tags:
     - django_settings
 
-- name: Download CloudWatch agent
-  get_url:
-    url: https://s3.amazonaws.com/amazoncloudwatch-agent/ubuntu/amd64/latest/amazon-cloudwatch-agent.deb
-    dest: /tmp/amazon-cloudwatch-agent.deb
-    mode: '0644'
-  tags:
-    - cloudwatch
+- name: Cloudwatch agent setup
+  block:
+    - name: Download CloudWatch agent
+      get_url:
+        url: https://s3.amazonaws.com/amazoncloudwatch-agent/ubuntu/amd64/latest/amazon-cloudwatch-agent.deb
+        dest: /tmp/amazon-cloudwatch-agent.deb
+        mode: '0644'
 
-- name: Install CloudWatch agent
-  apt:
-    deb: /tmp/amazon-cloudwatch-agent.deb
-    state: present
-  tags:
-    - cloudwatch
+    - name: Install CloudWatch agent
+      apt:
+        deb: /tmp/amazon-cloudwatch-agent.deb
+        state: present
 
-- name: Remove CloudWatch agent file
-  file:
-    path: /tmp/amazon-cloudwatch-agent.deb
-    state: absent
-  tags:
-    - cloudwatch
+    - name: Remove CloudWatch agent file
+      file:
+        path: /tmp/amazon-cloudwatch-agent.deb
+        state: absent
 
-- name: Create CloudWatch agent config directory
-  file:
-    path: /opt/aws/amazon-cloudwatch-agent/etc
-    state: directory
-    mode: 0755
-  tags:
-    - cloudwatch
+    - name: Create CloudWatch agent config directory
+      file:
+        path: /opt/aws/amazon-cloudwatch-agent/etc
+        state: directory
+        mode: 0755
 
-- name: Configure CloudWatch agent
-  template:
-    src: cloudwatch-config.json.j2
-    dest: /opt/aws/amazon-cloudwatch-agent/etc/config.json
-    mode: 0644
-  notify: restart cloudwatch agent
-  tags:
-    - cloudwatch
+    - name: Configure CloudWatch agent
+      template:
+        src: cloudwatch-config.json.j2
+        dest: /opt/aws/amazon-cloudwatch-agent/etc/config.json
+        mode: 0644
+      notify: restart cloudwatch agent
 
-- name: Start CloudWatch agent
-  systemd:
-    name: amazon-cloudwatch-agent
-    state: started
-    enabled: yes
-    daemon_reload: yes
+    - name: Start CloudWatch agent
+      systemd:
+        name: amazon-cloudwatch-agent
+        state: started
+        enabled: yes
+        daemon_reload: yes
+  rescue:
+    - name: Cloudwatch setup failed
+      debug:
+        msg: 'Cloudwatch agent setup failed, continuing with deployment'
   tags:
     - cloudwatch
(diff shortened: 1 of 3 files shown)


deploy/roles/connect/tasks/main.yml
```diff
@@ -105,3 +106,47 @@
     mode: 0644
   tags:
     - django_settings
+
+- name: Cloudwatch agent setup
+  block:
```
zandre-eng, line 111: My motivation for adding a block here is that the Cloudwatch setup is purely for monitoring and isn't critical for a successful deploy, so if something does go wrong it shouldn't block the deploy.

The biggest thing I imagine that could go wrong is if the file download doesn't succeed then that would essentially block the rest of the Cloudwatch commands.

Curious to hear from other folks if this is perhaps overkill, or if we would rather prefer to fail hard if something went wrong with setting up Cloudwatch.

deploy/roles/connect/templates/cloudwatch-config.json.j2
```diff
@@ -0,0 +1,60 @@
+{
+    "agent": {
+        "metrics_collection_interval": 60,
+        "run_as_user": "cwagent"
+    },
+    "metrics": {
+        "namespace": "PersonalID/{{ deploy_environment }}",
```
pxwxnvermx, line 7: It would be nice to have these metrics under CommCare Connect namespace.

zandre-eng: Updated in [bc6542c](https://github.com/dimagi/commcare-connect/pull/668/commits/bc6542cc5961887c800c308e7a343fbc243ce69a).

deploy/roles/connect/tasks/main.yml
calellowitz: I think we want to fail hard. These are only run when setting up the machine initially. We wouldn't want to run a machine without this monitoring, would rather know and try again. It won't affect anything running because code isn't even deployed on the machines by this point

deploy/roles/connect/tasks/main.yml
```diff
@@ -105,3 +106,47 @@
...
+    - name: Download CloudWatch agent
```
calellowitz, line 112: is this the recommend way to install rather than something like apt?

deploy/roles/connect/templates/cloudwatch-config.json.j2
```diff
@@ -0,0 +1,60 @@
+{
+    "agent": {
```
calellowitz, line 2: If this config is used for logshipping as well, it would be great to add syslog so we get OS logging as well, not just the django app in the container for debugging

calellowitz:
In addition to the other questions, are you able to go into the machines and make sure there is not an existing cloudwatch configuration file this will overwrite? If so, we will probably want to merge them

deploy/roles/connect/tasks/main.yml
zandre-eng: For installing it in the terminal there are two available approaches:
- Download the file directly.
- Use `sudo yum install amazon-cloudwatch-agent` which is a redhat command, so we can't use that.

Testing locally I am able to run `sudo apt install amazon-cloudwatch-agent` successfully but I don't see this method getting mentioned anywhere in [the docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/download-cloudwatch-agent-commandline.html), so I'm a bit hesitant to go for that approach.

Alternatively, we can also use the AWS Systems Manager to install and manage the Cloudwatch agent, but this also requires either a direct download or a snap install, so I opted for the "installing in the terminal" approach which looked more straightforward.

deploy/roles/connect/templates/cloudwatch-config.json.j2
zandre-eng: I've added logging for `syslog` as well ([ccaf338](https://github.com/dimagi/commcare-connect/pull/668/commits/ccaf338cd61682e0d171a792d3d84fbd5f785876)).

zandre-eng:
> In addition to the other questions, are you able to go into the machines and make sure there is not an existing cloudwatch configuration file this will overwrite? If so, we will probably want to merge them

@calellowitz I took a look at the machines and the `/opts/aws` directory doesn't exist so there doesn't seem to be an existing config setup.

deploy/roles/connect/templates/cloudwatch-config.json.j2
```diff
@@ -3,6 +3,20 @@
...
         "run_as_user": "cwagent"
     },
+    "logs": {
+        "logs_collected": {
+            "files": {
+                "collect_list": [
+                    {
+                        "file_path": "/var/log/syslog",
+                        "log_group_name": "{{ deploy_environment }}_syslog.log",
+                        "log_stream_name": "{{ deploy_environment }}_syslog.log",
```
calellowitz, line 13: Does this need some way to make sure different instances logs don't get mixed? It looks like for HQ, the log stream name includes the instance id and instance name within the same log group, to separate them. I think if we don't do that, the syslogs of all instances will mix. It appears we can get that info using special variables `{instance_id}` and/or `{hostname}`

calellowitz (changes requested):
looks good except for the log stream name

sravfeyn:
Would be nice to update [this wiki](https://dimagi.atlassian.net/wiki/spaces/commcareconnect/pages/3178430747/CCC+Server+Logs) in case it needs (after deploy)

sravfeyn (approved):

deploy/roles/connect/templates/cloudwatch-config.json.j2
```diff
@@ -0,0 +1,74 @@
...
+                        "file_path": "/var/log/syslog",
+                        "log_group_name": "{{ deploy_environment }}_syslog",
+                        "log_stream_name": "{instance_id}",
+                        "timezone": "UTC"
+                    }
+                ]
+            }
+        }
+    },
+    "metrics": {
```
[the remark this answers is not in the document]
zandre-eng: Ah good catch. I've removed the extra invalid fields in [61bf20c](https://github.com/dimagi/commcare-connect/pull/668/commits/61bf20c668639bc4b075637f3a5fd73b9f0e06d3).

PR #669, opened 2025-06-24
add invite views

calellowitz:
## Product Description

Connect part of https://github.com/dimagi/connect-id/pull/131

Adds two views, one to check if a phone number has been invited to a connect opportunity. One to resend invites to a phone number. Only invites that were not previously sent, because the user was not found, are sent.

### Safety story

Both views are unused except in new workflows

### Labels & Review


pxwxnvermx (approved):
LGTM. Thanks

hemant10yadav (approved):

PR #678, opened 2025-07-08
Duplicate Completed Work Handling

pxwxnvermx:
## Product Description
Added unique constraints and form processor code to create separate duplicate completed works on the database for duplicate user visits. 
Added migration to create duplicate completed works for older duplicate user visit submissions as well.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-1412)

### Labels & Review


pxwxnvermx:
@calellowitz @sravfeyn bumping for review before i deploy this on staging.
The migration here are easily revertable, so it would be nice to have those reviewed.

commcare_connect/form_receiver/processor.py
```diff
@@ -271,59 +274,77 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
-            ):
-                user_visit.status = VisitValidationStatus.over_limit
-                if not completed_work.status == CompletedWorkStatus.over_limit:
-                    completed_work.status = CompletedWorkStatus.over_limit
-                    completed_work_needs_save = True
-            elif counts["entity"] > 0:
-                user_visit.status = VisitValidationStatus.duplicate
+            user_visit.save()
+    except IntegrityError as e:
+        if "unique_completed_work_deliver_unit_access" not in str(e):
```
sravfeyn, line 322: Probably better to pull this out to a constant `UNIQUE_COMPLETED_WORK_CONSTRAINT='unique_completed_work_deliver_unit_access'`

sravfeyn:
Added few comments. This likely needs QA as well some sort of performance testing, right?

commcare_connect/form_receiver/processor.py
```diff
@@ -272,59 +275,77 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
-                entity_id=entity_id,
-                payment_unit=payment_unit,
-                defaults={"entity_name": entity_name},
+        if completed_work is None:
+            completed_work = CompletedWork(
+                opportunity_access=access, entity_id=entity_id, payment_unit=payment_unit, entity_name=entity_name
             )
+            completed_work.save()
+
+    try:
```
hemant10yadav, line 317: Would it make sense to catch this exception [here](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/form_receiver/views.py#L32), since all similar exceptions are already being handled in this section?

Also, I believe the explicit transaction.atomic block can be removed, as our requests are already atomic by [default](https://github.com/dimagi/commcare-connect/blob/main/config/settings/base.py#L54). It might be cleaner to remove all the transaction blocks from this function unless there's a specific reason to keep them.

commcare_connect/form_receiver/processor.py
```diff
@@ -271,59 +274,77 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+        location=xform.metadata.location,
+    )
+    today = datetime.date.today()
+    paymentunit_startdate = payment_unit.start_date if payment_unit else None
+    if opportunity.start_date > today or (paymentunit_startdate and paymentunit_startdate > today):
+        completed_work = None
+        user_visit.status = VisitValidationStatus.trial
+    else:
+        completed_work = (
+            CompletedWork.objects.filter(opportunity_access=access, entity_id=entity_id, payment_unit=payment_unit)
```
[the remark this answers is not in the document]
pxwxnvermx: Since we are filtering by user and entity id. This query will be limited to a very small (less than 3) number of rows being returned, so the performance for this was not a big concern.

commcare_connect/form_receiver/processor.py
pxwxnvermx: > Would it make sense to catch this exception [here](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/form_receiver/views.py#L32), since all similar exceptions are already being handled in this section?

That will make it a little harder to do the logic operation that are being done in the except block here.

> Also, I believe the explicit transaction.atomic block can be removed, as our requests are already atomic by [default](https://github.com/dimagi/commcare-connect/blob/main/config/settings/base.py#L54). It might be cleaner to remove all the transaction blocks from this function unless there's a specific reason to keep them.

I will try to remove the transaction blocks and check if the code works as expected.

commcare_connect/form_receiver/processor.py
```diff
@@ -271,59 +274,77 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+            .annotate(rank=Window(Rank(), order_by="date_created"))
```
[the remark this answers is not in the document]
pxwxnvermx: this is related to the [comment above](https://github.com/dimagi/commcare-connect/pull/678#discussion_r2213572749), since we are not fetching a lot of rows I have done the Rank query through postgres.

commcare_connect/form_receiver/processor.py
pxwxnvermx: Added. [0e01bad](https://github.com/dimagi/commcare-connect/pull/678/commits/0e01badfbe15b4f2771a3079d5156c22fb8464b9)

hemant10yadav (approved):

commcare_connect/opportunity/models.py
```diff
@@ -540,7 +540,12 @@ class CompletedWork(models.Model):
     )
 
     class Meta:
-        unique_together = ("opportunity_access", "entity_id", "payment_unit")
+        constraints = [
+            models.UniqueConstraint(
+                fields=["opportunity_access", "entity_id", "payment_unit", "date_created"],
```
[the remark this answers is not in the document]
calellowitz: Do we gain anything from this constraint, rather than just deleting the original. I am struggling to see where the `date_created` check excludes anything

commcare_connect/opportunity/migrations/0081_create_duplicated_completed_works.py
```diff
@@ -0,0 +1,89 @@
...
+                "entity_id": entity_id,
+                "entity_name": user_visit.entity_name,
+                "opportunity_access_id": access_id,
+                "payment_unit_id": payment_unit_id,
+            }
+        )
+
+    to_update = []
+    for key, details in completed_work_details_map.items():
+        completed_work = CompletedWork(**details)
```
calellowitz, line 54: what status do these have? is there a way to determine a "correct" status based on the underlying visits?

commcare_connect/form_receiver/processor.py
```diff
@@ -272,83 +276,99 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+        access.save()
 
-        if completed_work is not None:
-            if completed_work.completed_count > 0 and completed_work.status == CompletedWorkStatus.incomplete:
-                completed_work.status = CompletedWorkStatus.pending
-                completed_work_needs_save = True
-            if completed_work_needs_save:
-                completed_work.save()
+    if user_visit.completed_work is not None:
+        if completed_work.completed_count > 0 and user_visit.completed_work.status == CompletedWorkStatus.incomplete:
```
calellowitz, line 367: are `user_visit.completed_work` and `completed_work` ever different? If so, when/how?

calellowitz (commented):
a couple small questions, trying to think through the explicit transactions as well. I am not positive either way, yet, but in theory they were there to make sure partial failures are cleaned up. Specifically we catch most exceptions to return meaningful error messages, and I am not sure the automatic request transactions would handle that, since the request didn't raise an exception.

commcare_connect/form_receiver/processor.py
pxwxnvermx: They are the same. I will update this code to use `completed_work`.

commcare_connect/opportunity/migrations/0081_create_duplicated_completed_works.py
pxwxnvermx: These did not have a status. I have added a Duplicate status for completed works now which will be assigned here.

commcare_connect/form_receiver/processor.py
```diff
@@ -272,83 +276,99 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+    try:
+        user_visit.completed_work = completed_work
+        user_visit.save()
+    except IntegrityError as e:
+        if UNIQUE_USER_VISIT_CONSTRAINT not in str(e):
+            raise e
+        completed_work = CompletedWork(
+            opportunity_access=access, entity_id=entity_id, payment_unit=payment_unit, entity_name=entity_name
+        )
+        completed_work.save()
```
sravfeyn, line 327: Does't this end up creating two completed_work objects, first one unlinked ([above](https://github.com/dimagi/commcare-connect/pull/678/files#diff-0d916b95ee38daa9788eca7b8b3266f406c5c791cab6ff6bda1d5aeefc84d5d2R316)), second one linked here?

commcare_connect/form_receiver/processor.py
```diff
@@ -272,83 +276,99 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
...
+        counts["daily"] >= payment_unit.max_daily
+        or counts["total"] >= claim_limit.max_visits
+        or (today > claim.end_date or (claim_limit.end_date and today > claim_limit.end_date))
+    ):
+        user_visit.status = VisitValidationStatus.over_limit
+        if not user_visit.completed_work.status == CompletedWorkStatus.over_limit:
+            user_visit.completed_work.status = CompletedWorkStatus.over_limit
+            completed_work_needs_save = True
+    elif counts["entity"] > 0:
+        user_visit.status = VisitValidationStatus.duplicate
```
sravfeyn, line 342: Shouldn't this affect existing `saved_completed_count` calculation?

calellowitz:
> in theory they were there to make sure partial failures are cleaned up. Specifically we catch most exceptions to return meaningful error messages, and I am not sure the automatic request transactions would handle that, since the request didn't raise an exception.

Based on this understanding, it seems like the removed transaction is still necessary, but I see that it has not be added back. Do you think this is a misunderstanding? Otherwise, I think the form processor will no longer be atomic.

commcare_connect/opportunity/models.py
```diff
@@ -684,6 +685,40 @@ def __setattr__(self, name, value):
                 self.status_modified_date = now()
         super().__setattr__(name, value)
 
+    def save(self, **kwargs):
+        # The code below automatically switches/create CompletedWork when the UNIQUE_USER_VISIT_CONSTRAINT
+        # is triggered.
+        rank = kwargs.pop("rank", 0)
+        try:
+            with transaction.atomic():
```
calellowitz: why wrap a transaction here?

commcare_connect/opportunity/models.py
```diff
@@ -684,6 +685,40 @@ def __setattr__(self, name, value):
                 self.status_modified_date = now()
         super().__setattr__(name, value)
 
+    def save(self, **kwargs):
+        # The code below automatically switches/create CompletedWork when the UNIQUE_USER_VISIT_CONSTRAINT
+        # is triggered.
+        rank = kwargs.pop("rank", 0)
```
calellowitz: nit: can this just be handled with a default arg? or is the `pop` necessary for the `super` call

calellowitz (commented):
I think it looks good. Couple questions

commcare_connect/opportunity/models.py
pxwxnvermx: Without this transaction, any other transaction will exit due to the IntegrityError exception, this basically helps as a wrapper to localize that error to only this transaction instead of any other parent transactions.

calellowitz (approved):
Has QA taken a look at this? Given how central this is to our form processing and accrual counting, it would be good to have it well tested.

pxwxnvermx:
@calellowitz Nope, QA has not yet started. I will request QA for this now and also deploy these changes on staging.

PR #683, opened 2025-07-14
Extend command to update tables

sravfeyn:
## Product Description

Ticket https://dimagi.atlassian.net/browse/CI-40
This change enables adding more tables to RDS replica. 

## Technical Summary

Instead of simply adding a subscriotion/publication if it doesn't exist, this change sets the tables even if they exist and then refreshes them. 

### Safety story

I have tested this locally by adding new table, verifying it populates, and that main db updates get propagated for the newly added table

### Labels & Review


sravfeyn:
@calellowitz 

We need to run the management command to update the tables. However, you had a [pending comment](https://docs.google.com/document/d/1r9uFUQK8cOreaUm6E8nx9nyeKZL8tcNLJwJiWXh8kGI/edit?disco=AAABdB0ohmI) on the doc related to running that command as postgres-superser or not. I have just responded on that, we need to decide some strategy to address that.

sravfeyn:
@calellowitz Updated the code to get superuser creds from command line

calellowitz (approved):
assuming you were able to test this, looks good to me

PR #692, opened 2025-07-23
Sr/oplop

sravfeyn:
## Product Description

This adds various calculated columns to opportunity list page by optimizing them. (was [removed](https://github.com/dimagi/commcare-connect/pull/638) previously for performance reasons)

## Technical Summary

This optimizes at three levels
1. Adds DB index on `UserVisit.status` and `OpportunityAccess.date_learn_started` as these queries need to query on these fields. (should also help in other places where this queries are getting used)
2. Django-tables does count, order_by, page on the queryset that we return. If this is a large queryset it becomes really slow as it has to calculate pretty much all pages instead of the requested page. This is optimized by counting, ordering, paging and then getting IDs on just the opportunities without calculating the expensive calculations first, then uses only that page IDs to get calculated annotations. Though due to this, we remove sortability on these calculated fields.
3. Optimizes the queries themselves by pre-aggregating in one query instead of multiple queries triggered for each opportunity in `OuterRef`. For completedwork calcs

I have extended current data generation command to generate large amount of data efficiently. I have generated about 1 million visits/works, and then ran the report before and after above optimizations. I have observed about 10x speed improvements (~30 seconds to ~3seconds).

### Safety story

To be QAed.

### Automated test coverage

Tests exist.

### Labels & Review


zandre-eng:
> we remove sortability on these calculated fields

@sravfeyn Have we confirmed that this is something we are fine losing? It sounds okay given the tradeoff for better performance, but I'm not completely sure how often users use the sorting features.

sravfeyn:
@zandre-eng I think I have originally checked with Rama on this and we were okay doing this. The sort-ability on all other metadata fields of opportunity still works. (Like, Name, start-date, end-date, program, status etc)

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
+    def get_data(self):
+        base_qs = self.get_opportunity_list_id_qs(self.org, self.is_program_manager)
+
+        def data_qs(ids):
+            return self.get_opportunity_list_data(ids, self.is_program_manager)
+
+        return TieredQueryset(base_qs, data_qs)
+
+    @staticmethod
+    def get_opportunity_list_id_qs(organization, program_manager=False):
```
Charl1996, line 409: Is this method name right? From the name I get the impression that it's returning ids (or a queryset with id's), but the actual return is an Opportunity queryset it seems.

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
+    def __getitem__(self, key):
+        if isinstance(key, slice):
+            page_qs = self._ordered_qs[key]
+            ids = list(page_qs.values_list("id", flat=True))
+            return list(self.data_qs_fn(ids))
+        else:
+            obj = self._ordered_qs[key]
+            result = list(self.data_qs_fn([obj.id]))
+            if not result:
+                raise IndexError("Index out of range.")
```
calellowitz, line 384: If the index was out of range, line 381 would raise that error right? This is a different error caused when an ID can't be found in the data queryset.

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
+
+    @staticmethod
+    def get_opportunity_list_id_qs(organization, program_manager=False):
+        today = now().date()
+        base_filter = Q(organization=organization)
+        if program_manager:
+            base_filter |= Q(managedopportunity__program__organization=organization)
+        return Opportunity.objects.filter(base_filter).annotate(
+            program=F("managedopportunity__program__name"),
+            status=Case(
```
calellowitz, line 416: why does the base queryset need these annotations? Is it for sorting?

commcare_connect/opportunity/helpers.py
calellowitz: Also interested in this question. The description mentions only returning IDs, but this seems like a normal queryset, plus some extra annotations

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
-            .values("total")[:1],
-            output_field=IntegerField(),
+class TieredQueryset:
+    # This queryset enables targetted count, order_by, pagination
+    #   for large querysets on which it may not be efficient to run
+    #   them.
+    #
+    # This works by passing a base_queryset that just returns ids
+    #   and data_qs_fn which returns the data for those ids
+    def __init__(self, base_qs, data_qs_fn):
```
calellowitz, line 367: Is there any enforcement that the two querysets are equivalent in any way (or filtered equivalently)?

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
+        self.base_qs = base_qs
+        self.data_qs_fn = data_qs_fn
```
calellowitz, line 369: why use a function rather than a queryset here?

commcare_connect/opportunity/helpers.py
```diff
@@ -357,85 +357,148 @@ def get_payment_report_data(opportunity: Opportunity, usd=False):
...
+            status=Case(
+                When(Q(active=True) & Q(end_date__gte=today), then=Value(0)),  # Active
+                When(Q(active=True) & Q(end_date__lt=today), then=Value(1)),  # Ended
+                default=Value(2),  # Inactive
+                output_field=IntegerField(),
+            ),
         )
 
-    return queryset
+        if program_manager:
```
calellowitz, line 457: can you explain more about these changes (going from a subquery to python). Naievely I would expect postgres to be more efficient than doing the same thing in python, especially with the filtering.

calellowitz (commented):
Generally looks ok. Asked a few questions. Do we have tests that confirm the new calculations match he old calculations? Would also be good to add tests for `TieredQueryset` since thats a totally new thing.

commcare_connect/opportunity/helpers.py
sravfeyn: Only the base queryset does filtering (as well as ordering and pagination), the data queryset simply takes the IDs and returns matching data, so the filtering is automatically consistent.

commcare_connect/opportunity/helpers.py
sravfeyn: Yeah, this could be removed.

This is handling the case where the base queryset is able to return a matching object ID, but the data queryset is not, which shouldn't really happen.

commcare_connect/opportunity/helpers.py
sravfeyn: This indeed is used to get just the IDs [above](https://github.com/dimagi/commcare-connect/pull/692/files#diff-396037c378e64671cea6adc7185fa4f3aee28145c1fcf11f27bd597e494cd9a5R378), though I could see how that's not clear from the name. I will rename it to `get_base_qs` which is more apt.

commcare_connect/opportunity/helpers.py
sravfeyn: Exactly, it's for sorting and then paging on that.

commcare_connect/opportunity/helpers.py
sravfeyn: Sure, the main parts that were slow were various subqueries starting from [here](https://github.com/dimagi/commcare-connect/pull/692/files#diff-396037c378e64671cea6adc7185fa4f3aee28145c1fcf11f27bd597e494cd9a5L414). 

The reason it's slow is because it's running a lot of (and a bit inefficient) subqueries for each of the opportunity on UserVisit and CompletedWork.

Below is an example old query, which is run for each opportunity ID triggering many internal subqueries.

```
Subquery(
            CompletedWork.objects.filter(opportunity_access__opportunity_id=OuterRef("pk"))
            .values("opportunity_access__opportunity_id")
            .annotate(total=Sum("saved_completed_count"))
            .values("total")[:1],
            output_field=IntegerField(),
        )
```

Now, it's a single query

```
            deliveries_agg = (
                CompletedWork.objects.filter(opportunity_access__opportunity_id__in=opp_ids)
                .values("opportunity_access__opportunity_id")
                .annotate(
                    total_deliveries=Sum("saved_completed_count"),
                    verified_deliveries=Sum("saved_approved_count"),
                )
            )
```

Effectively, the old queryset looks like a single queryset, but runs a lot of subqueries and JOINs. While the new one has more querysets, it avoids those expensive subqueries and JOINS.

Because the updated queries are faster and this queryset is [meant](https://github.com/dimagi/commcare-connect/pull/692/files#diff-396037c378e64671cea6adc7185fa4f3aee28145c1fcf11f27bd597e494cd9a5R426) to be operated on a smaller set of opp-ids, the python JOIN of already aggregated data ends up faster as well.

calellowitz:
Were you able to test on staging at all?

calellowitz (approved):
Nice, this should give us a good chance to see how it behaves in prod, especially in the really bad organizations (not sure how much data your local had there).

PR #699, opened 2025-08-04
Bulk visit approval

hemant10yadav:
## Product Description
We've introduced a new enhancement to the worker visit verification page that allows users to select multiple visible visits and approve or reject them in bulk. This streamlines the review process, saving time when validating a high volume of visits.

_Key highlights:_
- Users can select all visible visits on the page.
- Bulk Approve or Reject actions are now supported. PM functionality for Agree and Disagree will still be handled individually.
- Individual visit actions remain unaffected.

Screenshot 2025-08-04 at 18 39 27
Screenshot 2025-08-04 at 18 39 34
<p align="center">
Screenshot 2025-08-04 at 18 39 45
Screenshot 2025-08-04 at 18 39 39
</p>

## Technical Summary
This update enhances the visit verification workflow by enabling bulk approval and rejection of visits directly from the UI. The implementation includes the following key changes:  [CCCT-1529](https://dimagi.atlassian.net/browse/CCCT-1529)

**Frontend (Alpine.js & HTMX):**
A checkbox system was added to allow selection of multiple visible visits.
A dynamic status summary shows the count of selected visits by status.
Bulk Approve and Reject buttons trigger modal confirmations with `justification`  and r`ejection reason` input.

**Backend (Django Views)**
`approve_visit` and `reject_visit` views now handle multiple visit IDs via POST (visit_ids[]).
Already approved/rejected visits are excluded from update logic.
Justification or Rejected reason is captured and stored where applicable.

### Safety story
Tested locally and on the staging environment. A QA request has also been raised.

### Automated test coverage
Existing automated tests already cover this functionality, and all tests are passing.

### QA Plan
[QA-7965](https://dimagi.atlassian.net/browse/QA-7965) - Completed

### Labels & Review

[QA-7965]: https://dimagi.atlassian.net/browse/QA-7965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CCCT-1529]: https://dimagi.atlassian.net/browse/CCCT-1529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -16,7 +16,7 @@ from django.core.files.storage import default_storage, storages
 from django.db.models import Count, DecimalField, FloatField, Func, Max, OuterRef, Q, Subquery, Sum, Value
 from django.db.models.functions import Cast, Coalesce
 from django.forms import modelformset_factory
-from django.http import FileResponse, Http404, HttpResponse
+from django.http import FileResponse, Http404, HttpResponse, HttpResponseBadRequest
 from django.middleware.csrf import get_token
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
@@ -845,6 +845,12 @@ def approve_visits(request, org_slug, opp_id):
         .prefetch_related("opportunity")
         .only("status", "review_status", "flagged", "justification", "review_created_on")
     )
+
+    if len({visit.user_id for visit in visits}) > 1:
+        return HttpResponseBadRequest(
+            "All visits must belong to the same user.",
+            headers={"HX-Trigger": "form_error"},
+        )
 
     today = now()
     for visit in visits:


commcare_connect/static/js/project_alpine.js
```diff
@@ -0,0 +1,94 @@
+/**
```
hemant10yadav, line 1: Including the script directly via a `<script>` tag in the HTML file was not feasible, as Alpine.js was being initialized before the script was loaded. To address this, a separate file was created for custom Alpine.js logic. All Alpine.js initialization code can now be managed within this file.

commcare_connect/opportunity/views.py
```diff
@@ -836,37 +836,55 @@ def send_message_mobile_users(request, org_slug=None, opp_id=None):
...
-    if user_visit.status != VisitValidationStatus.approved or user_visit.review_status == VisitReviewStatus.disagree:
-        user_visit.status = VisitValidationStatus.approved
-        if user_visit.opportunity.managed:
-            user_visit.review_created_on = now()
-            if user_visit.review_status == VisitReviewStatus.disagree:
-                user_visit.review_status = VisitReviewStatus.pending
-
-            if user_visit.flagged:
+def approve_visits(request, org_slug, opp_id):
+    visit_ids = request.POST.getlist("visit_ids[]")
```
calellowitz, line 840: what is the `[]` at the end of the name?

commcare_connect/opportunity/views.py
```diff
@@ -836,37 +836,55 @@ def send_message_mobile_users(request, org_slug=None, opp_id=None):
...
+    visit_ids = [int(vid) for vid in visit_ids if vid.isdigit()]
```
calellowitz, line 841: why is this necessary?

commcare_connect/opportunity/views.py
```diff
@@ -836,37 +836,55 @@ def send_message_mobile_users(request, org_slug=None, opp_id=None):
...
+                        status=400,
+                        headers={"HX-Trigger": "form_error"},
+                    )
+                visit.justification = justification
 
-        user_visit.save()
-        update_payment_accrued(opportunity=user_visit.opportunity, users=[user_visit.user], incremental=True)
+    UserVisit.objects.bulk_update(visits, ["status", "review_created_on", "review_status", "justification"])
+    if visits:
+        update_payment_accrued(opportunity=visits[0].opportunity, users=[visits[0].user], incremental=True)
```
calellowitz, line 869: maybe we don't need to do this yet, but very soon we will have pages that have visits from multiple users, so we will need to be able to accomodate that in this code

commcare_connect/opportunity/views.py
```diff
@@ -836,37 +836,55 @@ def send_message_mobile_users(request, org_slug=None, opp_id=None):
...
-            user_visit.review_created_on = now()
-            if user_visit.review_status == VisitReviewStatus.disagree:
-                user_visit.review_status = VisitReviewStatus.pending
-
-            if user_visit.flagged:
+def approve_visits(request, org_slug, opp_id):
+    visit_ids = request.POST.getlist("visit_ids[]")
+    visit_ids = [int(vid) for vid in visit_ids if vid.isdigit()]
+
+    visits = (
```
calellowitz, line 843: Does the user have any idea which visits were skipped because they were already approved or rejected? Also does this mean there is no way to change the status of a visit anymore?

hemant10yadav: > Does the user have any idea which visits were skipped because they were already approved or rejected?

Yes, the modal says in the approve modal that already approved visits will be ignored, and vice versa in the reject modal.
 
> Also does this mean there is no way to change the status of a visit anymore?

No, it only skips already approved visits when an approval request is made, and skips already rejected ones in the rejected_visit request. And they still have the ability to change the status.

commcare_connect/templates/opportunity/user_visit_verification_table.html
```diff
@@ -71,3 +62,127 @@
...
+              <textarea required class="base-textarea" id="justification" rows="3" name="justification"
+                placeholder="Please provide a reason for approving flagged visits.">{{ justification }}</textarea>
+            {% else %}
+              <input type="hidden" name="justification" value="">
+            {% endif %}
+             <div x-show="selected?.length > 0"
+              class="mt-3 p-4 rounded-lg border bg-blue-50 border-blue-200" >
+                <p class="text-sm text-gray-800 leading-relaxed flex flex-wrap items-center gap-2">
+                  <strong><span x-text="`${selected.length} visit${selected.length !== 1 ? 's' : ''} selected —`"></span></strong>
+                  <span x-show="selectedStatusCounts.approved" class="px-2 py-0.5 text-xs font-medium rounded-full bg-green-100 text-green-800">
```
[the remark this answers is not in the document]
hemant10yadav: Done - [3ee8c3c](https://github.com/dimagi/commcare-connect/pull/699/commits/3ee8c3c6f272c47a9fe1e5028aa4bba2225a0dd3)

commcare_connect/opportunity/views.py
hemant10yadav: Thanks — it wasn't needed. I was just being cautious during the implementation and then forgot to check it afterward.

commcare_connect/static/js/project_alpine.js
```diff
@@ -0,0 +1,98 @@
...
+      url.search = params.toString();
+      window.history.pushState({}, '', url.toString());
+      this.$dispatch('reload_table');
+
+      this.$nextTick(() => {
+        this.updateVisitStatusMap();
+      });
+    },
+
+    toggleSelectAll() {
```
sravfeyn, line 79: I am sorry for a delayed review, but have you considered this approach https://htmx.org/examples/bulk-update/ which doesn't require most of the alpinejs?

commcare_connect/static/js/project_alpine.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+document.addEventListener('alpine:init', () => {
+  // used in user_visit_verification_table.html
+  Alpine.data('visitVerification', (initialTab, defaultPage) => ({
+    activeTab: initialTab,
+    page: defaultPage,
+    selected: [],
+    showBulkApproveModal: false,
+    showBulkRejectModal: false,
```
sravfeyn, line 16: If you move the modal html, right below the button that toggles these modals, then we don't need to push these two vars into js here. For e.g. see [this](https://github.com/dimagi/commcare-connect/pull/703/files#diff-ec7cf771fc1b6b0ab618b95da0597a648282fb8995dadd312029c74e142f8523R6)

commcare_connect/templates/opportunity/user_visit_verification_table.html
```diff
@@ -71,3 +62,115 @@
   <input type="hidden" name="page" value="{{ request.GET.page|default:1 }}" x-model="page" />
   <input type="hidden" name="sort" value="{{ request.GET.sort|default:'' }}" />
 </form>
+
+
+
+
+<!--Bulk Approve Modal -->
+<div x-cloak x-show="showBulkApproveModal" x-transition.opacity class="modal-backdrop">
```
sravfeyn, line 70: Both these modals look a lot similar, so they could be DRYed up, similar to import/export modals that we have.

commcare_connect/static/js/project_alpine.js
hemant10yadav: This uses a form, and adding another would only make the template more complex, especially since there’s already a form on the table itself. That’s why I opted for this approach.

commcare_connect/templates/opportunity/user_visit_verification_table.html
hemant10yadav: I partially agree with your point. These modals rely on several Alpine vars and include conditional logic for managed vs. regular opportunities, in addition to handling approval and rejection flows. To keep the logic manageable and avoid excessive if-else blocks, I extracted the common parts and kept the modals separate.

commcare_connect/static/js/project_alpine.js
hemant10yadav: The modal contains forms, and there's already a form on the table.

commcare_connect/templates/opportunity/user_visit_verification_table.html
sravfeyn: It could look something like below and similar one for Accept, without the full HTML for both.

```
{% include "components/bulk_modal.html" with
  show_modal="showBulkRejectModal"
  title="Reason for Rejection"
  post_url=reject_url
  show_textarea=True
  textarea_id="rejection-reason"
  textarea_name="reason"
  textarea_placeholder="Please provide a reason for rejecting visit."
  textarea_value=reason
  note="Note: Visits that are already rejected will not be affected by this action."
  action_label="Reject"
  action_class="negative-dark"
  on_submit="'showBulkRejectModal = false'"
%}
```

commcare_connect/static/js/project_alpine.js
sravfeyn: Okay, it's a bit late for this, I guess. But for future, I would recommend checking out anyway of doing interactivity within htmx when possible, if it's easier that way. For e.g. you can see [this refactor](https://github.com/dimagi/commcare-connect/pull/703/files) that removes most of tab management javascript and uses a declarative style htmx to simplify it.

commcare_connect/templates/opportunity/user_visit_verification_table.html
hemant10yadav: Thanks added

commcare_connect/templates/opportunity/user_visit_verification.html
```diff
@@ -233,8 +233,97 @@
           this.flatpickrInstance.clear();
         }
       }));
+      Alpine.data('visitVerification', (initialTab, defaultPage) => ({
```
sravfeyn, line 236: There is bunch of separate things inside this that are all listed under single Alpine.data model

I would recommend breaking them into their own models like

```
Alpine.data('selectionManager', 
...
selectedStatusCounts()
toggleSelectAll()
updateSelectAll()
```
It's okay to not break down the existing logic into separate models, but at least the bulk select code can be in a separate model.

commcare_connect/templates/opportunity/user_visit_verification.html
```diff
@@ -233,8 +233,97 @@
...
+        tableRows.forEach((row) => {
+          const visitId = row.getAttribute('data-visit-id');
+          const status = row.getAttribute('data-visit-status');
+          if (visitId && status) {
+            this.visitStatusMap[visitId] = status;
+          }
+        });
+      },
+
+      get selectedStatusCounts() {
```
sravfeyn, line 262: Can we get this counts directly from server when loading this file, does it change while on client side?

hemant10yadav: The individual visit status may change based on the individual approval and rejection flow.

commcare_connect/templates/opportunity/user_visit_verification.html
hemant10yadav: Moved - [08490be](https://github.com/dimagi/commcare-connect/pull/699/commits/08490bed7d24102c01947d81144068ca179f46ca)

PR #702, opened 2025-08-05
currency handling improvements

calellowitz:
## Technical Summary

This task is only supposed to run once per month to get the rates, but when it fails and a rate is missing, it gets called over and over causing us to hit our limits on openexchangerates. This PR does two things to handle failures a bit better

1) limits our requests to once per day (per URL, which changes monthly) by caching the result.
2) reports currencies that aren't found to sentry, but continues processing the rest to prevent a bad code from causing overall failure. 

It also fixes a bug with the loop variables.
 
I had to switch the test runner to use prod requirements in order to avoid an import error for `sentry_sdk`. We could probably also move it to a different file that isn't called during tests, but there doesn't seem to be a huge downside in making our tests closer to the prod environment.
### Safety story

This task is already broken and at worst can only affect fetching exchange rates, rather than any core functionality.

### Labels & Review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements.txt
+          pip install -r requirements-dev.txt
 
       - name: Test with pytest
         run: pytest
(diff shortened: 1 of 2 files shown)


Charl1996:
> when it fails and a rate is missing, it gets called over and over 

Where does this retry happen in the code?

calellowitz:
> > when it fails and a rate is missing, it gets called over and over
> 
> Where does this retry happen in the code?

It is not quite a retry in the code, but if a rate is missing, this gets called on form submission. So each time a form came in from a misconfigured opportunity, we were hitting the API

PR #710, opened 2025-08-13
Add "staff" view for manually requesting OTP 

Charl1996:
## Product Description
This PR adds a view at `<base_url>/users/connect_user_otp/` which allows a staff user to request an OTP for a given phone number.

Here's the page:
image

When an OTP is received it will be displayed to the user using django's built-in messaging alerts:
image

## Technical Summary
[Ticket](https://dimagi.atlassian.net/browse/CCCT-1522)
Also relates to [this PersonalID PR](https://github.com/dimagi/connect-id/pull/163).

### Safety story
Did local testing (except making the call to ConnectID server). Will test on staging as well.

### Automated test coverage
No automated tests added.

### QA Plan
No formal QA planned.

### Labels & Review

### DIFF
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -1,10 +1,8 @@
 import httpx
-import sentry_sdk
 from django.conf import settings
 from httpx import BasicAuth, Response
 
 from commcare_connect.cache import quickcache
-from commcare_connect.connect_id_client.exceptions import ConnectIDClientError
 from commcare_connect.connect_id_client.models import (
     ConnectIdUser,
     DemoUser,
@@ -91,9 +89,7 @@ def get_user_otp(phone_number):
     except httpx.HTTPStatusError as e:
         if e.response.status_code == 404:
             return None
-        else:
-            sentry_sdk.capture_message(message=f"ConnectID client error: {str(e)}", level="error")
-            raise ConnectIDClientError(f"Failed to fetch OTP: HTTP {e.response.status_code}")
+        raise
 
 
 def _make_request(method, path, params=None, json=None, timeout=5) -> Response:
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -1,4 +1,5 @@
 import httpx
+import sentry_sdk
 from django.conf import settings
 from httpx import BasicAuth, Response
 
@@ -94,6 +95,7 @@ def get_user_otp(phone_number):
                 "that the user has started their device seating process.",
             )
         else:
+            sentry_sdk.capture_message(message=f"ConnectID client error: {str(e)}", level="error")
             raise ConnectIDClientError(f"Failed to fetch OTP: HTTP {e.response.status_code}")
 
 
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -90,10 +90,7 @@ def get_user_otp(phone_number):
         return data.get("otp")
     except httpx.HTTPStatusError as e:
         if e.response.status_code == 404:
-            raise ConnectIDClientError(
-                "Failed to fetch OTP. Please make sure the number is correct and "
-                "that the user has started their device seating process.",
-            )
+            return None
         else:
             sentry_sdk.capture_message(message=f"ConnectID client error: {str(e)}", level="error")
             raise ConnectIDClientError(f"Failed to fetch OTP: HTTP {e.response.status_code}")
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -1,5 +1,4 @@
 import httpx
-import sentry_sdk
 from django.conf import settings
 from httpx import BasicAuth, Response
 
@@ -90,10 +89,6 @@ def get_user_otp(phone_number):
     except httpx.HTTPStatusError as e:
         if e.response.status_code == 404:
             return None
-        sentry_sdk.capture_exception(e)
-        raise
-    except httpx.HTTPError as e:
-        sentry_sdk.capture_exception(e)
         raise
 
 
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -81,6 +81,15 @@ def fetch_user_counts() -> dict[str, int]:
     return data
 
 
+def fetch_otp(phone_number):
+    try:
+        response = _make_request(GET, "/users/generate_manual_otp", params={"phone_number": phone_number})
+        data = response.json()
+        return data.get("otp")
+    except httpx.HTTPError:
+        return None
+
+
 def _make_request(method, path, params=None, json=None, timeout=5) -> Response:
     if json and not method == "POST":
         raise ValueError("json can only be used with POST requests")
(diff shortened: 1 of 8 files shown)


commcare_connect/users/views.py
```diff
@@ -209,3 +211,35 @@ def post(self, request, *args, **kwargs):
         for opp_id in opps:
             update_user_and_send_invite(user, opp_id)
         return HttpResponse(status=200)
+
+
+class RetrieveUserOTPView(TemplateView):
+    template_name = "pages/connect_user_otp.html"
+
+    def dispatch(self, request, *args, **kwargs):
+        if self.request.user.is_authenticated and self.request.user.is_staff:
```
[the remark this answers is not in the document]
Charl1996: For some reason I thought it was the other way around...thanks for clarifying

commcare_connect/users/forms.py
```diff
@@ -52,3 +54,25 @@ class OrganizationCreationForm(forms.ModelForm):
     class Meta:
         model = Organization
         fields = ["name", "program_manager"]
+
+
+class ManualUserOTPForm(forms.Form):
+    phone_number = forms.CharField(
+        max_length=15,
+        label="Phone Number",
+        help_text="Enter the phone number of the user you which to retreive the OTP for.",
```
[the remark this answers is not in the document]
Charl1996: Ah, nice spotting!

commcare_connect/users/views.py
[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/commcare-connect/pull/710/commits/91b7c903f0b64e7fa50fc7db11f2755e591be381

commcare_connect/connect_id_client/main.py
```diff
@@ -81,6 +81,15 @@ def fetch_user_counts() -> dict[str, int]:
     return data
 
 
+def fetch_otp(phone_number):
```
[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/commcare-connect/pull/710/commits/f9dfd81739bb5845b1dfafa2c5e95520b07e9938

commcare_connect/users/views.py
```diff
@@ -209,3 +211,35 @@ def post(self, request, *args, **kwargs):
         for opp_id in opps:
             update_user_and_send_invite(user, opp_id)
         return HttpResponse(status=200)
+
+
+class RetrieveUserOTPView(TemplateView):
```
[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/commcare-connect/pull/710/commits/254378629b585571f5ac52591a8a5803beb14dee

sravfeyn:
It's totally fine to keep it as it is, something to keep in mind when we do another.

zandre-eng (approved):
Thanks for the changes @Charl1996. Small nit on tests, but LGTM!

hemant10yadav (approved):
Thanks! Looks good.

commcare_connect/connect_id_client/main.py
```diff
@@ -86,8 +87,14 @@ def get_user_otp(phone_number):
         response = _make_request(GET, "/users/generate_manual_otp", params={"phone_number": phone_number})
         data = response.json()
         return data.get("otp")
-    except httpx.HTTPError:
-        return None
+    except httpx.HTTPStatusError as e:
+        if e.response.status_code == 404:
```
calellowitz, line 92: This doesn't really feel like a `ClientError`, but the expected result when a user types in an unknown number. I would recommend not raising an error in this part of the `if` block, just returning `None`, checking for that in the view and showing user not found in that case, then only raising for unexpected errors. That way, we know that we raise an exception when something went wrong in the code, not an expected status when the user makes a mistake.

commcare_connect/users/views.py
```diff
@@ -225,13 +226,10 @@ def test_func(self):
...
-        if otp is None:
-            messages.error(
-                self.request,
-                "Failed to fetch OTP. Please make sure the number is correct and"
-                " that the user has started their device seating process.",
-            )
+        try:
+            otp = get_user_otp(form.cleaned_data["phone_number"])
+        except ConnectIDClientError as e:
+            messages.error(self.request, str(e))
```
calellowitz, line 232: I don't think we need to show the user the status code when there is an actual error, since there isn't anything they can do with that. It might be fine to just let it fail hard, but at worst a generic "something unexpected went wrong" message should be sufficient for the user.

commcare_connect/connect_id_client/main.py
```diff
@@ -94,6 +95,7 @@ def get_user_otp(phone_number):
                 "that the user has started their device seating process.",
             )
         else:
+            sentry_sdk.capture_message(message=f"ConnectID client error: {str(e)}", level="error")
```
calellowitz, line 98: nit: I think it makes more sense to add this to the view where we catch the exception so that it logs the actual exception class to sentry, like a normal exception there rather than a custom message

calellowitz (changes requested):
I think the error handling could be improved a bit to draw a greater distinction between unexpected errors in the system, and expected results for certain user inputs.

commcare_connect/connect_id_client/main.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/710/commits/6a4d84baacd443f52e7bc1404c4c0b379abbce4d

commcare_connect/connect_id_client/main.py
Charl1996: Somehow I don't like my own solution here: https://github.com/dimagi/commcare-connect/commit/c87b6f2d1e19819e426136256a6c1b5f62145769

I do think exceptions should be logged further up-stream, closer to the source (i.e. inside `_make_request` where we raise for status). I can create another PR for that if you agree in the interest of not getting side-tracked in this PR.

zandre-eng (approved):

commcare_connect/connect_id_client/main.py
```diff
@@ -89,6 +90,10 @@ def get_user_otp(phone_number):
     except httpx.HTTPStatusError as e:
         if e.response.status_code == 404:
             return None
+        sentry_sdk.capture_exception(e)
+        raise
+    except httpx.HTTPError as e:
+        sentry_sdk.capture_exception(e)
```
calellowitz, line 96: if we are raising the error and not handling it, it will be logged to sentry automatically

calellowitz (approved):
I think this double reports to sentry (in the call, and by when the error is uncaught). Even if it is smarter than that, the capture is unnecessary if we aren't catching the execption. Sentry handles that automatically (that's how we get all our unexpected exceptions).

Not blocking because I don't think it's harmful is this is urgent but would be good to follow up with

commcare_connect/connect_id_client/main.py
calellowitz: I think since we aren't catching the exception this has become a moot point (and explicit sending is no longer necessary.

commcare_connect/connect_id_client/main.py
Charl1996: I really don't know why I thought that sentry would not capture these exceptions

Charl1996:
I'm going to merge this PR since [the last commit](https://github.com/dimagi/commcare-connect/pull/710/commits/de6887562cf230c079d226ea1d22998f1dd5a236) is simply a followup to a suggestion and the PR has been approved otherwise.

PR #725, opened 2025-08-29
Add filtering to opportunity list report

sravfeyn:
## Product Description

PRing on top of base filters PR  https://github.com/dimagi/commcare-connect/pull/715/files

https://dimagi.atlassian.net/browse/CCCT-1407

Adds filters to Opportunity List page

Screenshot 2025-08-29 at 3 43 01 PM

### Automated test coverage

Added

### Labels & Review


calellowitz (commented):
curious about the new dependency but otherwise looks good

commcare_connect/opportunity/filters.py
```diff
@@ -57,6 +58,13 @@ def get_filter_context(self):
         }
 
 
+class CSRFExemptForm(forms.Form):
```
calellowitz, line 61: Why disable CSRF instead of adding the token?

commcare_connect/static/css/tom-select.css
```diff
@@ -0,0 +1,429 @@
+/**
```
calellowitz, line 1: does this not come with the package somewhere? it is weird we have to commit their whole css

calellowitz (commented):
Two questions on the new stuff but still looks fine overall

sravfeyn:
Hemant, I have added x-cloak now, that should fix it

commcare_connect/opportunity/views.py
```diff
@@ -2256,7 +2256,7 @@ def opportunity_delivery_stats(request, org_slug, opp_id):
             "name": "Services Delivered",
             "status": "Total",
             "value": header_with_tooltip(stats.total_deliveries, "Total delivered so far excluding duplicates"),
-            "url": delivery_url,
+            "url": delivery_url + "?sort=-last_active",
```
calellowitz, line 2259: same comment I always make in these cases but can we use a url generation library to pass these params in rather than using string concatenation. I know it recently caused a bug when we changed a base url, and that would make it much harder to break (and the code feel a bit cleaner). This applies to all the urls where we add params by string concatenation, including the one in the template

calellowitz (approved):
nothing blocking but would like the url generation fixed, especially since it recently caused a bug

hemant10yadav (approved):
Thanks!

sravfeyn:
Thanks for all the review

PR #728, opened 2025-08-29
Manual OTP page permissions

Charl1996:
## Technical Summary
Based on [this PR](https://github.com/dimagi/commcare-connect/pull/514/files), the PR adds another permission that's required when a user wants to access the `/users/connect_user_otp` page. [This PR](https://github.com/dimagi/commcare-connect/pull/514/files) can technically be closed now.

[This doc](https://docs.google.com/document/d/1gmAvc9C_HStEuWV5FMH_8XOB56fzm026dvquxJ8YBUg/edit?tab=t.0#heading=h.ybvahbh4ifho) outlines some notes on using granular permission rather than granting superuser access to users.

Once this is deployed, no users will have access to this page except superusers.

### Safety story
Tested locally

### Automated test coverage
Tests for the relevant view is included.

### QA Plan
N.A

### Labels & Review


commcare_connect/users/tests/test_views.py
```diff
@@ -138,7 +139,8 @@ def test_no_otp_returned(self, get_user_otp_mock, user, client):
         assert str(messages[0]) == expected_failure_message
 
     def _get_superuser_response(self, client, user):
-        user.is_superuser = True
-        user.save()
+        perm = Permission.objects.get(codename="otp_access")
```
zandre-eng, line 142: It would be great if we could have an extra test here to validate that a non-superuser is able to access the page with the new permission.

zandre-eng (approved):
Just one suggestion on a testing improvement, but otherwise looks good.

commcare_connect/users/tests/test_views.py
zandre-eng: I was thinking more about the other way around - testing that a non-superuser is able to access the page if they have the necessary permission. 

There is a test to validate that a superuser can access this page, but it would be great if we could also validate that the same holds true for a normal user with the right permission.

Charl1996: > it would be great if we could also validate that the same holds true for a normal user with the right permission.

Actually, looking at the tests now again I don't set users to be superusers anymore, but I just add the permission, which means that is what I'm testing for.

The test names weren't correct though, so I updated that.

zandre-eng: Thank you!

zandre-eng (approved):

pxwxnvermx (approved):
Changes look good to me. I am not sure though if my review counts, since I also have commits in this PR. 😄

calellowitz (approved):
Looks great. thanks!

PR #756, opened 2025-10-02
Permissions

hemant10yadav:
## Technical Summary

This pull request introduces two new user-level permissions:

`kpi_report_access` – Grants individual users access to KPI reports.
`all_org_access` – Grants individual users access to all organizations without requiring superuser privileges.
`view_commcarehq_form_link`- Can view CommCareHQ form link.

Previously, users needed superuser status to access reports or organizations. These new permissions allow fine-grained access control, enabling users to perform specific actions without granting full superuser rights.

[CCCT-1736](https://dimagi.atlassian.net/browse/CCCT-1736)

## Safety Assurance

Tested locally and on staging.

### Automated test coverage
 Written tests for all the permission decorators and mixins we use, covering all cases, and all previous tests are passing.

### QA Plan
QA - [completed](https://dimagi.atlassian.net/browse/QA-8134)

### Labels & Review

[CCCT-1736]: https://dimagi.atlassian.net/browse/CCCT-1736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/organization/decorators.py
+++ b/commcare_connect/organization/decorators.py
@@ -2,6 +2,7 @@ from functools import wraps
 
 from django.http import Http404, HttpResponseRedirect
 from django.urls import reverse
+from django.utils.decorators import method_decorator
 
 from commcare_connect.utils.permission_const import ALL_ORG_ACCESS
 
@@ -64,12 +65,14 @@ def _get_decorated_function(view_func, permission_test_function):
 class OrganizationUserMixin:
     """Mixin version of org_viewer_required decorator"""
 
+    @method_decorator(org_viewer_required)
     def dispatch(self, request, *args, **kwargs):
-        return _get_decorated_function(super().dispatch(request, *args, **kwargs), _request_user_is_viewer)
+        return super().dispatch(request, *args, **kwargs)
 
 
 class OrganizationUserMemberRoleMixin:
     """Mixin version of org_member_required decorator"""
 
+    @method_decorator(org_member_required)
     def dispatch(self, request, *args, **kwargs):
-        return _get_decorated_function(super().dispatch(request, *args, **kwargs), _request_user_is_member)
+        return super().dispatch(request, *args, **kwargs)
--- a/commcare_connect/data_export/views.py
+++ b/commcare_connect/data_export/views.py
@@ -80,10 +80,14 @@ class BaseStreamingCSVExportView(BaseDataExportView):
 
 def _get_opportunity_or_404(user, opp_id):
     try:
-        return Opportunity.objects.get(
-            Q(organization__memberships__user=user)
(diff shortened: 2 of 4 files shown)


commcare_connect/conftest.py
```diff
@@ -145,3 +146,39 @@ def program_manager_org_user_member(program_manager_org) -> User:
 @pytest.fixture
 def program_manager_org_user_admin(program_manager_org) -> User:
     return program_manager_org.memberships.filter(role="admin").first().user
+
+
+@pytest.fixture
+def check_basic_permissions(client):
```
calellowitz, line 152: Curious why you decided to do this as a fixture, rather than just a function in the test file. We don't have any other "test functions" as fixtures, just objects.

commcare_connect/organization/decorators.py
```diff
@@ -52,7 +55,7 @@ def _inner(request, *args, **kwargs):
             return HttpResponseRedirect("{}?next={}".format(reverse("account_login"), request.path))
 
         if not permission_test_function(request):
-            raise Http404
+            raise PermissionDenied()
```
calellowitz, line 58: I think we typically 404 in these situations to not reveal if the org exists, and it was likely intentional

commcare_connect/organization/decorators.py
```diff
@@ -1,31 +1,34 @@
 from functools import wraps
 
-from django.http import Http404, HttpResponseRedirect
+from django.core.exceptions import PermissionDenied
+from django.http import HttpResponseRedirect
 from django.urls import reverse
 
 from .models import UserOrganizationMembership
 
+ALL_ORG_ACCESS = "users.all_org_access"
```
calellowitz, line 9: It might be nice to make a constant with all of these permissions so we can reference them everywhere consistently.

calellowitz (commented):
couple questions/suggestions, otherwise looks good

commcare_connect/conftest.py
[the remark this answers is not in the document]
hemant10yadav: I specifically combined these into one due to the order in which they should be executed.

pxwxnvermx (approved):

commcare_connect/organization/decorators.py
calellowitz: > all other permissions decorators that also raise a 404 error ... to keep things consistent, I went ahead with this update

Can you explain a bit more? It sounds like you are saying others use 404, if so, how does this keep things consistent?

calellowitz (commented):
curious about the answer to the two question but still looks good otherwise

pxwxnvermx (approved):

PR #761, opened 2025-10-07
toggles

calellowitz:
## Technical Summary

This adds a library to manage release toggles in connect. It also includes a short readme with guidance for using these features.

### Safety story

This does not modify any existing behavior. I tested use of switches locally, both in templates and views successfully.

### Labels & Review

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -8,7 +8,6 @@ import environ
 BASE_DIR = Path(__file__).resolve(strict=True).parent.parent.parent
 # commcare_connect/
 APPS_DIR = BASE_DIR / "commcare_connect"
-397
 
 env = environ.Env()
 


config/settings/base.py
```diff
@@ -361,12 +367,19 @@
...
 TWILIO_ACCOUNT_SID = env("TWILIO_SID", default=None)
 TWILIO_AUTH_TOKEN = env("TWILIO_TOKEN", default=None)
 TWILIO_MESSAGING_SERVICE = env("TWILIO_MESSAGING_SERVICE", default=None)
 MAPBOX_TOKEN = env("MAPBOX_TOKEN", default=None)
 
 OPEN_EXCHANGE_RATES_API_ID = env("OPEN_EXCHANGE_RATES_API_ID", default=None)
 
-OAUTH2_PROVIDER_APPLICATION_MODEL = "oauth2_provider.Application"
+# Waffle Settings
+WAFFLE_FLAG_MODEL = "flags.Flag"
```
calellowitz, line 382: Despite not using flags yet, we need to specify a custom model before first deploy if we will want one later, so this is to keep our options open, but is not to be used.

requirements/dev.txt
```diff
@@ -9,8 +9,6 @@ anyio==3.7.1
     #   -c requirements/base.txt
     #   httpcore
     #   watchfiles
-appnope==0.1.4
```
calellowitz, line 12: I am not sure where this change came from. It was done automatically by the requirements compiling command. It does seem safe, but happy to remove if we prefer.

calellowitz (commented):
left a few comments to guide review, but its a pretty low stakes PR

PR #764, opened 2025-10-07
Remove sms option from "send message" view

Charl1996:
## Product Description
Removes SMS as an option from "send message" view.

Note that I removed the `message_type` select option entirely because there's only one option at this point, i.e. push notification. A note has been added for the user making it clear that a push notification will be sent.

image

## Technical Summary
Ticket: https://dimagi.atlassian.net/browse/CCCT-1350

This PR removes the SMS option when users go to the "Send Message" view. 

### Safety story
Tested locally what I could

### Automated test coverage
No new tests

### QA Plan
No QA planned

### Labels & Review

### DIFF
--- a/commcare_connect/templates/opportunity/send_message.html
+++ b/commcare_connect/templates/opportunity/send_message.html
@@ -64,7 +64,9 @@
             {{ form.title|as_crispy_field }}
             {{ form.body|as_crispy_field }}
             <p class="text-sm text-gray-600">
-                Note: A push notification will be sent to the selected users.
+              {% blocktranslate %}
+              Note: A push notification will be sent to the selected users.
+              {% endblocktranslate %}
             </p>
             <div class="flex justify-end">
               <button class="button button-md primary-dark" type="submit">Confirm</button>


zandre-eng (commented):
> Should we tell users on the UI that "sending a message" means sending a push notification?

To me that sounds like a good idea. Since users have had two options in the past they might be confused as to which option they have now (push notifications or SMS), so a little description at the top of the page noting that this will send a push notification could be nice.

pxwxnvermx (approved):
+1 for showing to the user that these messages are "Push Notifications" on the UI.

commcare_connect/templates/opportunity/send_message.html
```diff
@@ -63,6 +63,9 @@
           <div class="flex flex-col gap-4 p-4 border border-gray-200 rounded-lg">
             {{ form.title|as_crispy_field }}
             {{ form.body|as_crispy_field }}
+            <p class="text-sm text-gray-600">
+                Note: A push notification will be sent to the selected users.
```
zandre-eng, line 67: Nit: Could you please wrap this in a trans block? We don't currently do translations but it's a good habit for future-proofing.

zandre-eng (approved):

commcare_connect/templates/opportunity/send_message.html
Charl1996: https://github.com/dimagi/commcare-connect/pull/764/commits/c2c7dbe2ebb19e9d83299ec0755805d76e1e3f45

zandre-eng: Thanks!

zandre-eng (approved):

pxwxnvermx (approved):

PR #868, opened 2025-11-26
Incorrect Accrued Payment Calculation Due to exclude Usage on CompletedWork

hemant10yadav:
## Technical Summary

While debugging an interrupt issue, I identified a bug in how accrued payments are calculated.

When a visit is initially marked as rejected, the corresponding CompletedWork record is also marked as rejected. However, if the visit is later approved through the visit import workflow, the CompletedWork record is not updated accordingly. Because the queryset excludes rejected records.

This results in inconsistencies between the visit status and the CompletedWork status, leading to inaccurate payment accruals. The bulk approval/reject UI flow does not apply this exclusion logic, which further introduces inconsistent behavior between the two flows.

To ensure consistency across both flows and to correctly include updated records in the accrual calculation, I have removed the exclusion condition from the queryset.

### Safety story
I’ve removed the exclude filter, as the payment accrual method already handles these scenarios correctly. Given its existing logic and safeguards, I’m confident this change is safe and maintains the intended behaviour.

### Automated test coverage
Already exisit

### QA Plan
No QA

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tasks.py
+++ b/commcare_connect/opportunity/tasks.py
@@ -429,9 +429,7 @@ def bulk_update_payment_accrued(opportunity_id, user_ids: list):
     access_objects = OpportunityAccess.objects.filter(opportunity=opportunity_id, user__in=user_ids, suspended=False)
     for access in access_objects:
         with cache.lock(f"update_payment_accrued_lock_{access.id}", timeout=900):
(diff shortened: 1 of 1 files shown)


calellowitz:
Can you hold off on merging this until I get back next week? I want to make sure we aren't overriding manually rejected completed works. I am also curious how this came up. Once a delivery is rejected we wouldn't expect it to be approved later, that is meant to be a terminal status.

PR #1098, opened 2026-04-01
Filters for Assigned Tasks Page

zandre-eng:
## Product Description

The assigned task list page now has a filter button (sliders icon) next to the task table.
Screenshot_20260416_072440

Clicking it opens a modal with fields to filter by worker name, task status, task type, date assigned range, and due date range. 
Screenshot_20260416_072500

When one or more filters are active, a badge on the button shows the count of applied filters.
Screenshot_20260416_072532-1

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2178)

The existing `FilterMixin` and `FilterSet` pattern (already used by `WorkerTaskView`) was extended to `AssignedTaskListView`. A new `AssignedTaskFilterSet` was added rather than reusing `TasksFilterSet` because the latter filters a worker-aggregated queryset while this one filters `AssignedTask` rows directly. The metric card counts (total/open/complete) intentionally use unfiltered aggregates.

- **New `AssignedTaskFilterSet`** (`filters.py`): Filters by `worker_name` (scoped to workers with at least one `AssignedTask`), `task_status` (custom "To Do"/"Completed" labels, not raw enum values), `task_type` (scoped to task types with at least one `AssignedTask`), `date_assigned_after/before` (using `date__gte/lt` cast to avoid timezone fragility on the `DateTimeField`), and `due_date_after/before`.
- **`AssignedTaskListView` updated** (`views.py`): Added `FilterMixin` to MRO, declared `filter_class = AssignedTaskFilterSet`, implemented `get_filter_kwargs()` to pass `opportunity` for dynamic choices, and delegated `get_queryset()` to `self._get_filter().qs`.
- **Filter modal partial** (`assigned_task_filter_modal.html`): Alpine.js toggle + htmx form submission (`hx-get`, `hx-target="#task-list-table"`, `hx-push-url="true"`) with a badge indicator for active filter count.

### Safety story

All `AssignedTask` queries in the filterset and view remain scoped to `opportunity_access__opportunity=opportunity`, so no cross-opportunity data leakage is possible. The filter form uses the existing CSRF session mechanism (no new surface). Dynamic choices for worker name and task type are populated from the same opportunity scope. Metric counts use a separate unfiltered aggregate query — they are read-only and cannot be corrupted by filter state. No schema changes or data mutations are involved.

### Automated test coverage

- **`TestAssignedTaskFilterSet`** (`test_filters.py`): Covers no-filter baseline, single-field filters for worker name, task status, task type, date assigned range (in-range and out-of-range), due date range, and combined multi-field filters.
- **`TestAssignedTaskListView`** (`test_views.py`): `test_filter_by_status_returns_filtered_table` — verifies the table is filtered by status; `test_filters_applied_count_in_context` — verifies the badge count is passed in context; `test_metric_counts_unaffected_by_filters` — verifies total/open/complete counts reflect full opportunity totals regardless of active filters.

### QA Plan

- [ ] Navigate to an opportunity's assigned task list page
- [ ] Verify the sliders filter button appears near the top of the table
- [ ] Click the filter button and confirm the modal opens with all filter fields (Worker Name, Task Status, Task Type, Date Assigned From/Before, Due Date From/Before)
- [ ] Apply a Task Status filter and confirm the table updates and the URL reflects the filter param
- [ ] Confirm the filter button badge shows the correct count of active filters
- [ ] Confirm the Total Tasks / Open Tasks / Complete metric cards are unchanged by the filter
- [ ] Clear filters (submit with empty fields) and confirm all rows return
- [ ] Reload the page with a filter param in the URL and confirm the filter is pre-applied

### Labels & Review


commcare_connect/opportunity/tests/test_filters.py
```diff
@@ -178,3 +180,75 @@ def test_tasks_filterset_task_type_excludes_inactive():
...
+
+    def test_filter_by_date_assigned_excludes_outside_range(self):
+        future = (date.today() + timedelta(days=10)).isoformat()
+        result = self.filter_assigned_tasks({"date_assigned_after": future})
+        assert result.count() == 0
+
+    def test_filter_by_due_date_range(self):
+        soon = (date.today() + timedelta(days=14)).isoformat()
+        result = self.filter_assigned_tasks({"due_date_before": soon})
+        assert result.count() == 2
```
pxwxnvermx: nit: these tests can be parameterized for the filter conditions and the count returned.

pxwxnvermx (commented):
Just some comments about the test. Rest of the changes look good to me.

commcare_connect/opportunity/filters.py
```diff
@@ -222,3 +223,87 @@ def __init__(self, *args, **kwargs):
...
+            "task_type",
+            Row(Column("date_assigned_after"), Column("date_assigned_before")),
+            Row(Column("due_date_after"), Column("due_date_before")),
+        )
+
+    def _get_task_type_choices(self):
+        """
+        Fetch task types with at least one `AssignedTask` for this opportunity
+        """
+        tasks = Task.objects.filter(assignedtask__opportunity_access__opportunity=self.opportunity).distinct()
```
Charl1996: Should this include both active and inactive tasks? I know for the "Tasks" tab on the Connect Workers page we only show active tasks.

zandre-eng: Good question. If we do want to only show active tasks then we'll need to update the other filters on this page to match as the entire table + metrics currently allow for both active and inactive tasks. Let me follow up on the ticket.

zandre-eng: @Charl1996 I have confirmed that it's fine to show both active and inactive tasks on this page, but I've added an extra filter so users can filter between the two statuses if needed.

Addressed in [979d01d](https://github.com/dimagi/commcare-connect/pull/1098/commits/979d01d5b338c9207aeac9d024c4c200df8c49ef)

commcare_connect/templates/opportunity/assigned_task_list.html
```diff
@@ -60,10 +60,11 @@ <h1 class="text-2xl font-medium text-white">{% trans "Task List" %}</h1>
 
         <div id="task-list-table" class="bg-white rounded-lg shadow-xs p-4"
              hx-get="{{ request.get_full_path }}"
+             hx-trigger="reload_table from:body"
```
hemant10yadav: I don’t think this is used anymore. I remember @Charl1996  made some changes, and now we’re redirecting instead of reloading the table.

zandre-eng: Originally removing this was causing an issue where the filter modal would close a few seconds after being opened. After your comment I did some further investigation and discovered that the issue was related to the htmx default trigger behaviour. Removing all the htmx attributes resolves the problem, and should be fine given that we're not reloading the table anymore.

Addressed in [c1677cf](https://github.com/dimagi/commcare-connect/pull/1098/commits/c1677cf245e2fb5bed5be6ea0eb60ec299004366).

commcare_connect/templates/opportunity/tasks.html
```diff
@@ -3,44 +3,6 @@
...
-      <div class="modal">
-        <div class="header text-lg font-semibold text-brand-deep-purple">
-          <h1>{% trans "Filters" %}</h1>
-          <button @click="showFilterModal = false">
-            <i class="fa-solid fa-xmark text-xl"></i>
-          </button>
-        </div>
-        <form
-          id="taskFilterForm"
-          hx-get="."
```
Charl1996, line 25: What does the string dot do?

zandre-eng: The string dot simply issues the GET request to the current URL. "." is resolved as a relative URL, so it targets the currently open page.

commcare_connect/opportunity/views.py
```diff
@@ -3343,16 +3344,33 @@ def visit_export_count(request, org_slug, opp_id):
...
     paginate_by = DEFAULT_PAGE_SIZE
+    filter_class = AssignedTaskFilterSet
+
+    def get_paginate_by(self, table_data):
+        return get_validated_page_size(self.request)
+
+    def get_filter_kwargs(self):
+        opportunity = self.get_opportunity()
+        return {
+            "queryset": AssignedTask.objects.filter(opportunity_access__opportunity=opportunity)
```
Charl1996: Why not use the `get_queryset` method here since it seems like it's creating the same query?

zandre-eng: Excellent point, I agree they are identical and this could be replaced with `get_queryset`.

Addressed in [843eb00](https://github.com/dimagi/commcare-connect/pull/1098/commits/843eb003bb902d30c0489e95a8994d2ae2f4ba8b).

Charl1996 (approved):

hemant10yadav (approved):

PR #1105, opened 2026-04-07
Change Invoice report permission

hemant10yadav:
## Product Description

The Invoice Report now has a dedicated permission: invoice_report_access. Previously, access was limited to users with all_org_access.

Users without invoice_report_access access will now see a scoped view — limited to invoices and opportunities from organizations they belong to.

## Technical Summary
https://dimagi.atlassian.net/browse/CI-541

Introduced a new permission: invoice_report_access.
Replaced the dependency on all_org_access for accessing the Invoice Report.
Applied organization-based scoping for non-admin users:
Invoice data is filtered by the user’s organizations.
Opportunity filter dropdown only shows relevant options.
Ensured the same scoping logic is applied in background export tasks.

### Safety story
New permission holders only see organization-scoped data.
Filtering is enforced at the queryset level, minimizing risk of data leakage.

### QA Plan
Not needed

### Labels & Review


commcare_connect/reports/views.py
```diff
@@ -219,6 +219,14 @@ class InvoiceReportFilter(django_filters.FilterSet):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        request = getattr(self, "request", None)
+        if request and request.user and not request.user.has_perm(ALL_ORG_ACCESS):
+            self.filters["opportunity_name"].queryset = (
+                ManagedOpportunity.objects.filter(program__organization__memberships__user=request.user)
```
mkangia: nit: I am not very clear about how this works so this might be a confusing query,
do we want to give access to just the admins for the organization or all members?

commcare_connect/reports/views.py
```diff
@@ -247,7 +255,7 @@ class InvoiceReportView(
     model = PaymentInvoice
     table_class = InvoiceReportTable
     filterset_class = InvoiceReportFilter
-    permission_required = ALL_ORG_ACCESS
```
mkangia: how is this still backward compatible as highlighted in PR description?

commcare_connect/reports/views.py
hemant10yadav: It’s not about admins and members in this case. We want to grant access to the report based on a specific permission, and all of them will be Dimagi users. However, we want to restrict which invoices they can see—they should only be able to view invoices for opportunities where they are members.

commcare_connect/reports/views.py
```diff
@@ -219,6 +219,14 @@ class InvoiceReportFilter(django_filters.FilterSet):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        request = getattr(self, "request", None)
+        if request and request.user and not request.user.has_perm(ALL_ORG_ACCESS):
```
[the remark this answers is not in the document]
mkangia: > It’s just a null check. It could have been avoided, but I don’t see any harm in it.

I agree its not adding any major performance factor but we are checking for something that should always be present and if it is not, that should be a concern and not something we should bypass.

ajeety4 (approved):

zandre-eng (commented):
Looks good overall! One request: please could you add tests covering the new filtering behavior and permission changes to verify they work as expected.

mkangia (approved):
Was this PR done with Claude or manually?
It seems manually and I wanted to just check if there was a reason to do so, since that deviates from what we are doing by default now.

zandre-eng (approved):

hemant10yadav:
> Was this PR done with Claude or manually?

We both did—Claude started it, I just gave it a human touch 😄

PR #1114, opened 2026-04-09
feat: add Work Area Assignments tab to Connect Workers page

ajeety4:
## Product Description

Adds a new **Work Area Assignments** tab to the Connect Workers page, gated behind the `MICROPLANNING` waffle flag. The tab shows all accepted workers with their work area assignment metrics:

- Worker name and ID
- Last active date
- Assigned buildings, visits, work areas, and work area groups
- Visits done (only counts visits linked to a work area)
- Footer row with column totals

The tab is only visible when the `MICROPLANNING` flag is active for the opportunity.

image

## Technical Summary

**Ticket:** [CCCT-2209](https://dimagi.atlassian.net/browse/CCCT-2209)

### Implementation details:
- **Helper** (`helpers.py`): `get_worker_work_area_table_data()` uses 4 direct JOIN annotations through the `OpportunityAccess → WorkAreaGroup → WorkArea` FK chain, plus 1 correlated subquery for `visits_done` (filtered by `opportunity` to leverage existing DB index)
- **Table** (`tables.py`): `WorkerWorkAreaTable` with `gettext_lazy` column labels and footer totals using `intcomma`
- **View** (`views.py`): `WorkerWorkAreaView` extends `BaseWorkerListView`, adds the tab dynamically in `get()` when `MICROPLANNING` flag is active, and guards access via `dispatch` override
- **Template**: `worker_list_work_areas.html` extends `worker_list_base.html`
- **URL**: `<opp_id>/workers/work-areas/`

### Safety story

- Feature is gated behind the `MICROPLANNING` waffle flag — no impact on existing users
- The tab is added dynamically (not modifying the static `tabs` list), so existing tabs are unaffected
- Query uses JOIN annotations (not N+1) for performance; the `visits_done` subquery filters on `opportunity` to use the existing `(opportunity, status)` index

### Automated test coverage

4 new tests in `test_helpers.py`:
- `test_get_worker_work_area_table_data` — verifies annotation values for a worker with assignments
- `test_get_worker_work_area_table_data_unassigned_worker` — verifies zero values for worker without assignments
- `test_get_worker_work_area_table_data_visit_without_work_area_not_counted` — ensures visits not linked to a work area are excluded from `visits_done`
- `test_get_worker_work_area_table_data_multiple_workers_and_groups` — verifies correct aggregation across multiple workers and groups

### QA Plan

None

### Labels & Review

[CCCT-2209]: https://dimagi.atlassian.net/browse/CCCT-2209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -635,7 +635,11 @@ def get_worker_tasks_base_queryset(opportunity):
 
 def get_worker_work_area_table_data(opportunity):
     visits_done_subquery = (
-        UserVisit.objects.filter(opportunity_access=OuterRef("pk"), work_area__isnull=False)
+        UserVisit.objects.filter(
+            opportunity=opportunity,
+            opportunity_access=OuterRef("pk"),
+            work_area__isnull=False,
+        )
         .values("opportunity_access")
         .annotate(total=Count("id"))
         .values("total")[:1]
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -633,6 +633,27 @@ def get_worker_tasks_base_queryset(opportunity):
     )
 
 
+def get_worker_work_area_table_data(opportunity):
+    visits_done_subquery = (
+        UserVisit.objects.filter(opportunity_access=OuterRef("pk"), work_area__isnull=False)
+        .values("opportunity_access")
+        .annotate(total=Count("id"))
+        .values("total")[:1]
+    )
+
+    return (
+        OpportunityAccess.objects.filter(opportunity=opportunity, accepted=True)
+        .select_related("user")
+        .annotate(
+            assigned_buildings=Coalesce(Sum("workareagroup__workarea__building_count"), Value(0)),
+            assigned_visits=Coalesce(Sum("workareagroup__workarea__expected_visit_count"), Value(0)),
+            assigned_work_areas=Count("workareagroup__workarea"),
+            assigned_work_area_groups=Count("workareagroup", distinct=True),
+            visits_done=Coalesce(Subquery(visits_done_subquery), Value(0), output_field=IntegerField()),
+        )
+    )
+
+
 def get_opportunity_delivery_progress(opp_id):
     today = now().replace(hour=0, minute=0, second=0, microsecond=0)
     three_days_ago = today - timedelta(days=3)
--- a/commcare_connect/opportunity/tests/test_helpers.py
+++ b/commcare_connect/opportunity/tests/test_helpers.py
@@ -635,32 +635,6 @@ def test_filter_worker_tasks_combined_filters(opportunity):
 
 
 @pytest.mark.django_db
-def test_get_worker_work_area_table_data(opportunity):
-    access = OpportunityAccessFactory(opportunity=opportunity, accepted=True)
-    group = WorkAreaGroupFactory(opportunity=opportunity, opportunity_access=access)
-    WorkAreaFactory(opportunity=opportunity, work_area_group=group, building_count=10, expected_visit_count=5)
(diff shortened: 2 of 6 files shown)


commcare_connect/opportunity/helpers.py
```diff
@@ -633,6 +633,31 @@ def get_worker_tasks_base_queryset(opportunity):
...
+    )
+
+    return (
+        OpportunityAccess.objects.filter(opportunity=opportunity, accepted=True)
+        .select_related("user")
+        .annotate(
+            assigned_buildings=Coalesce(Sum("workareagroup__workarea__building_count"), Value(0)),
+            assigned_visits=Coalesce(Sum("workareagroup__workarea__expected_visit_count"), Value(0)),
+            assigned_work_areas=Count("workareagroup__workarea"),
+            assigned_work_area_groups=Count("workareagroup", distinct=True),
```
Charl1996, line 655: Why is `distinct` necessary here?

Charl1996: I would think that the work area group is inherently distinct since there's a FK to OpportunityAccess

Charl1996 (approved):
LGTM - just a small query.

commcare_connect/opportunity/helpers.py
ajeety4: Good question.
This is needed because joining with workarea creates duplicate rows for each workareagroup, so we use distinct to ensure each group is counted only once.

commcare_connect/opportunity/views.py
```diff
@@ -2549,11 +2553,19 @@ def get_tabs(self, org_slug, opportunity):
         workers_count = (
             UserInvite.objects.filter(opportunity=opportunity).exclude(status=UserInviteStatus.not_found).count()
         )
-        tabs_with_urls[0]["label"] = f"Connect Workers ({workers_count})"
+        tabs_with_urls[0]["label"] = _("Connect Workers") + f" ({workers_count})"
         return tabs_with_urls
 
     def get(self, request, org_slug, opp_id):
         opportunity = self.get_opportunity()
+        if is_flag_active(MICROPLANNING, opportunity):
```
zandre-eng, line 2561: Confirming whether we want to have this gated behind a FF? My understanding is that we want to try and push this feature forward without further flag gating.

Fine to ignore this given that we have other parts of the feature behind the flag, but good to keep in consideration.

commcare_connect/opportunity/tests/test_helpers.py
```diff
@@ -630,3 +632,104 @@ def test_filter_worker_tasks_combined_filters(opportunity):
     assert len(result) == 1
     assert result[0].user.name == "Alice"
     assert result[0].task_status == AssignedTaskStatus.COMPLETED
+
+
+@pytest.mark.django_db
+def test_get_worker_work_area_table_data(opportunity):
```
zandre-eng: Nit: This test looks to be doing largely the same thing as `test_get_worker_work_area_table_data_multiple_workers_and_groups()`. The latter seems to be more comprehensive, and so I think this test can be removed entirely.

zandre-eng (approved):
A few non-blocking comments, but otherwise looks good!

commcare_connect/opportunity/tests/test_helpers.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1114/commits/12079b55b00e2471e4fa7e4bfced1c41ab0f0ee7

zandre-eng (approved):

commcare_connect/opportunity/views.py
ajeety4: >  we have other parts of the feature behind the flag

yes kept the FF for this reason, and for consistency purpose

PR #1116, opened 2026-04-09
Assignment Mode UI

pxwxnvermx:
## Product Description
- Adds an "Assignment Mode" for PMs on the microplanning coverage map to bulk-assign work areas to FLWs                                                                                            - PMs can select work areas via map clicks or group dropdown, pick an assignee, and queue multiple assignments before confirming                                                                   
- Includes FLW Summary panel with independent FLW lookup showing assigned buildings/visits/work areas (including pending queue)                                                                    
- Backend endpoints for group work areas, FLW work areas, FLW summary, and a stub save endpoint
- "Enter Assignment Mode" button (PM-only) with mode badge and exit                                                                                                                                
- Multi-select work areas on map (orange highlight) or Select a Work Area group via dropdown
- Assignee dropdown with "Save" button to queue assignments
- Pending Assignments list with remove capability
- Confirmation and discard modals 
- "Show only unassigned" toggle                                             

[Screencast_20260409_183101.webm](https://github.com/user-attachments/assets/36d33e84-f410-4a41-a829-04714dd8cad0)

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2276)

### Safety story
- Tested Locally, Mostly UI changes.

### Labels & Review

### DIFF
--- a/commcare_connect/microplanning/views.py
+++ b/commcare_connect/microplanning/views.py
@@ -34,7 +34,7 @@ from commcare_connect.flags.decorators import require_flag_for_opp
 from commcare_connect.flags.flag_names import MICROPLANNING
 from commcare_connect.microplanning.const import WORK_AREA_STATUS_COLORS
 from commcare_connect.microplanning.filters import UserVisitMapFilterSet, WorkAreaMapFilterSet
-from commcare_connect.microplanning.forms import WorkAreaModelForm
+from commcare_connect.microplanning.forms import AssignmentModeForm, WorkAreaModelForm
 from commcare_connect.microplanning.models import WorkArea, WorkAreaGroup, WorkAreaStatus
 from commcare_connect.opportunity.models import OpportunityAccess, UserVisit
 from commcare_connect.organization.decorators import (
@@ -139,9 +139,7 @@ def microplanning_home(request, *args, **kwargs):
         org_slug = request.org.slug
         opp_id = opportunity.opportunity_id
 
-        context["work_area_groups_json"] = list(
-            WorkAreaGroup.objects.filter(opportunity=opportunity).values("id", "name")
-        )
+        context["assignment_form"] = AssignmentModeForm(opportunity=opportunity)
         context["assignees_json"] = list(
             OpportunityAccess.objects.filter(opportunity=opportunity, accepted=True, suspended=False)
             .select_related("user")
--- a/commcare_connect/microplanning/views.py
+++ b/commcare_connect/microplanning/views.py
@@ -165,7 +165,7 @@ def _get_assignment_mode_context(request, opportunity):
             .values("id", "user__name", "user_id")
         ),
         "group_work_areas_url": reverse(
-            "microplanning:assignment_group_work_areas",
+            "microplanning:get_work_areas_for_assignment",
             args=[org_slug, opp_id, 0],
         ).replace("/0/", "/__group_id__/"),
         "flw_work_areas_url": reverse(
@@ -532,7 +532,7 @@ class ModifyWorkAreaUpdateView(UpdateView):
 @org_program_manager_required
 @opportunity_required
 @require_flag_for_opp(MICROPLANNING)
-def assignment_group_work_areas(request, org_slug, opp_id, group_id):
+def get_work_areas_for_assignment(request, org_slug, opp_id, group_id):
     work_areas = list(
         WorkArea.objects.filter(
             opportunity=request.opportunity,
--- a/commcare_connect/microplanning/views.py
+++ b/commcare_connect/microplanning/views.py
@@ -136,39 +136,7 @@ def microplanning_home(request, *args, **kwargs):
     }
 
     if assignment_mode:
-        org_slug = request.org.slug
-        opp_id = opportunity.opportunity_id
-
-        context["assignment_form"] = AssignmentModeForm(opportunity=opportunity)
-        context["assignees_json"] = list(
-            OpportunityAccess.objects.filter(opportunity=opportunity, accepted=True, suspended=False)
-            .select_related("user")
-            .values("id", "user__name", "user_id")
-        )
-        context["group_work_areas_url"] = reverse(
-            "microplanning:assignment_group_work_areas",
-            args=[org_slug, opp_id, 0],
-        ).replace("/0/", "/__group_id__/")
-        context["flw_work_areas_url"] = reverse(
-            "microplanning:assignment_flw_work_areas",
-            args=[org_slug, opp_id, 0],
-        ).replace("/0/", "/__assignee_id__/")
-        context["flw_summary_url"] = reverse(
-            "microplanning:assignment_flw_summary",
-            kwargs={"org_slug": org_slug, "opp_id": opp_id},
-        )
-        context["assignment_save_url"] = reverse(
-            "microplanning:assignment_save",
-            kwargs={"org_slug": org_slug, "opp_id": opp_id},
-        )
-        context["user_visits_url"] = reverse(
-            "opportunity:user_visits_list",
-            args=[org_slug, opp_id],
-        )
-        context["worker_list_url"] = reverse(
-            "opportunity:worker_list",
-            args=[org_slug, opp_id],
-        )
+        context.update(_get_assignment_mode_context(request, opportunity))
 
     return render(
         request,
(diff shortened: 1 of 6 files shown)


commcare_connect/microplanning/views.py
```diff
@@ -473,3 +523,66 @@ def form_valid(self, form):
             }
         )
         return response
+
+
+@require_GET
+@org_program_manager_required
+@opportunity_required
```
mkangia, line 533: is the ordering of this opposite to what is needed? `opportunity_required` should be before `org_program_manager_required`?

commcare_connect/microplanning/views.py
```diff
@@ -473,3 +523,66 @@ def form_valid(self, form):
...
+    return JsonResponse(
+        {
+            "assigned_buildings": stats["buildings"] or 0,
+            "assigned_visits": stats["visits"] or 0,
+            "assigned_work_areas": qs.count(),
+        }
+    )
+
+
+@require_http_methods(["POST"])
```
mkangia: isn't this same as require_POST?

commcare_connect/microplanning/views.py
```diff
@@ -473,3 +523,66 @@ def form_valid(self, form):
             }
         )
         return response
+
+
+@require_GET
+@org_program_manager_required
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def assignment_group_work_areas(request, org_slug, opp_id, group_id):
```
mkangia, line 535: nit: not sure about the naming here. Why does this have "assignment" in the name?

commcare_connect/microplanning/views.py
```diff
@@ -103,29 +108,74 @@ def microplanning_home(request, *args, **kwargs):
...
+            kwargs={"org_slug": org_slug, "opp_id": opp_id},
+        )
+        context["user_visits_url"] = reverse(
+            "opportunity:user_visits_list",
+            args=[org_slug, opp_id],
+        )
+        context["worker_list_url"] = reverse(
+            "opportunity:worker_list",
+            args=[org_slug, opp_id],
+        )
```
mkangia: Can this whole if block be in a function of its own?

mkangia (approved):
Non blocking feedback.

To be honest, the super long commits made it hard to review this.
Can I see this on staging somewhere?

commcare_connect/microplanning/views.py
pxwxnvermx: We can order these checks anyway we want. The current order of these checks is the same as other views.

mkangia: Well I understand but logically it seems to be off.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -239,23 +254,48 @@
...
+                if (this.selectedWorkAreas.size === 0 || !this.selectedAssigneeId) return;
+                this.assignmentQueue.push({
                     assignee_id: this.selectedAssigneeId,
+                    assignee_name: this.getAssigneeName(),
                     work_area_ids: Array.from(this.selectedWorkAreas),
-                };
+                    buildings: this.getSelectedStats().buildings,
+                    visits: this.getSelectedStats().visits,
+                });
+                this.clearSelection();
```
mkangia, line 274: does this clear the map entirely?
I am confused if user has selected some work areas and then want to queue that selection, should not that selection be retained or shown as assigned and not available when assigning new users?

mkangia (approved):
Looks good so far. Thanks for sharing the preview.

This is mostly plugged together well, kudos on getting this working. The user experience however would need to be refined as we get more work done on this feature to get it ready.

Two things for now:
1. The work area group is expected to be a multi select
2. left a comment about map clearing

commcare_connect/microplanning/views.py
pxwxnvermx: The execution order is: `require_GET` → `org_program_manager_required` → `opportunity_required` → `require_flag_for_opp`. This checks org-level permission first (which doesn't need the opportunity loaded), then loads the opportunity. This is the same pattern used throughout the file (e.g. `org_admin_required` before `opportunity_required` on other views). I'll swap the order if you feel strongly, but wanted to note it's consistent with the existing convention.

commcare_connect/microplanning/views.py
pxwxnvermx: Already fixed in commit 20eb0c8f — switched to `@require_POST`.

commcare_connect/microplanning/views.py
pxwxnvermx: The `assignment_` prefix namespaces the four new endpoints (`assignment_group_work_areas`, `assignment_flw_work_areas`, `assignment_flw_summary`, `assignment_save`) that are specific to the assignment mode workflow. This distinguishes them from the existing work area endpoints used in the default microplanning view. Happy to rename if you have a better suggestion.

commcare_connect/microplanning/views.py
pxwxnvermx: Fixed in c4e3fa76. Extracted the assignment mode context setup into `_get_assignment_mode_context()` helper.

commcare_connect/microplanning/views.py
```diff
@@ -473,3 +521,66 @@ def form_valid(self, form):
...
+    )
+
+
+@require_http_methods(["POST"])
+@org_program_manager_required
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def assignment_save(request, org_slug, opp_id):
+    """Stub endpoint for saving work area assignments. Accepts the payload but does not persist changes yet."""
+    return JsonResponse({"status": "ok"})
```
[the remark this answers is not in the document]
pxwxnvermx: This is an intentional stub for wiring up the UI flow — persistence will be implemented in a follow-up ticket. The docstring documents this explicitly. Returning 501 would break the UI flow we're developing against. Once persistence lands, the stub will be replaced with actual save logic.

commcare_connect/templates/microplanning/home.html
```diff
@@ -127,63 +143,268 @@ <h2 class="text-lg font-semibold">{% translate "Filter Map Work Areas" %}</h2>
...
+                        <h3 class="text-sm font-semibold mb-2">
+                            {% translate "Pending Assignments" %}
+                            <span class="text-gray-400 font-normal" x-text="'(' + assignmentQueue.length + ')'"></span>
+                        </h3>
+                        <div class="space-y-2 max-h-40 overflow-auto">
+                            <template x-for="(entry, index) in assignmentQueue" :key="index">
+                                <div class="flex items-center justify-between bg-gray-50 rounded-lg px-3 py-2 text-xs">
+                                    <div class="min-w-0">
+                                        <span class="font-medium truncate block" x-text="entry.assignee_name"></span>
+                                        <span class="text-gray-500" x-text="entry.work_area_ids.length + ' work areas'"></span>
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed in c4e3fa76. Split the x-text concatenation into a separate `<span x-text>` for the count + a Django `{% translate %}` tag for the label, consistent with how line 233 handles the same pattern.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -66,11 +84,268 @@
...
+                        this.flwFilterWorkAreaIds.add(id);
+                        this.map.setFeatureState(
+                            { source: 'workareas', sourceLayer: 'workareas', id },
+                            { assignment_selected: true }
+                        );
+                    }
+                } catch (e) {
+                    this.showToast("{% translate 'Failed to load FLW work areas' %}", true);
+                }
+                this.updateFlwSummary();
```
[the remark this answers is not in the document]
pxwxnvermx: The partial clear is intentional — it allows additive selections where you pick a group and then manually add/remove individual work areas before queuing. Each selection source (group, FLW filter, manual clicks) is tracked independently so only that source's prior IDs are cleared on change. The workflow is: select a group → optionally tweak → queue → repeat. Using `clearSelection()` would discard any manual adjustments the user made.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -66,11 +84,268 @@
...
+                let buildings = 0;
+                let visits = 0;
+                for (const id of this.selectedWorkAreas) {
+                    const props = this.workAreaProperties.get(id);
+                    if (props) {
+                        buildings += parseInt(props.building_count) || 0;
+                        visits += parseInt(props.expected_visit_count) || 0;
+                    }
+                }
+                return { buildings, visits };
```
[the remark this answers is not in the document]
pxwxnvermx: Good catch. Fixed in c4e3fa76. The group/FLW work area endpoints now return `building_count` and `expected_visit_count` alongside each ID, and the JS loaders populate `workAreaProperties` so `getSelectedStats()` reports correct totals for bulk selections.

commcare_connect/templates/microplanning/map_handler.html
pxwxnvermx: Currently, `clearSelection()` after queueing clears the orange highlights so the user can make fresh selections for the next assignee. Queued entries are shown in the sidebar list. Adding a visual "queued" state on the map (e.g. a different highlight color for already-queued polygons) is a good idea for a future iteration — it would require a new feature-state and map layer. Filed as a follow-up.

commcare_connect/microplanning/views.py
mkangia: > This checks org-level permission first (which doesn't need the opportunity loaded),

Thanks, that was a miss in my understanding.

commcare_connect/microplanning/views.py
mkangia: so should this be "assign_grouped_work_areas"?

I appreciate that we want to distinguish. It's just that it's not intuitive.

mkangia: seeing the definition of that function now, how about `get_work_areas_for_assignment`?

mkangia (approved):
Not blocking on view renaming though that could be taken as a follow up in a different PR if needed.

commcare_connect/microplanning/views.py
pxwxnvermx: https://github.com/dimagi/commcare-connect/pull/1116/commits/1b71730b3b1b2602343a6d4cc628d53376b20e16

pxwxnvermx:
@mkangia I also made an update to change the highlight color from Orange to Black. The black highlighting is much more visible.
image

commcare_connect/microplanning/views.py
```diff
@@ -532,7 +532,7 @@ def form_valid(self, form):
 @org_program_manager_required
 @opportunity_required
 @require_flag_for_opp(MICROPLANNING)
-def assignment_group_work_areas(request, org_slug, opp_id, group_id):
+def get_work_areas_for_assignment(request, org_slug, opp_id, group_id):
```
mkangia, line 535: should we extend the same naming style for other views?

mkangia (approved):
Color change looks great!

commcare_connect/microplanning/views.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1116/commits/332d3c3ad2339d0604f27f7800197f1e58fe54b0

mkangia (approved):

PR #1117, opened 2026-04-09
[Claude design doc] Add design spec for WorkArea status updates via XForm

mkangia:
## Product Description

No user facing change

## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2252
The corresponding implementation can be found for reference here, https://github.com/dimagi/commcare-connect/pull/1118 (in draft right now but can be used if no changes are made to the spec here)

### Labels & Review


calellowitz (approved):
Generally looks good. As mentioned offline, I think it would be good to think through the feedback loop for missing reason, and be explicit about how people are supposed to discover that, but the stuff in here seems right

mkangia:
Closing this as its done and is not to be merged.

PR #1118, opened 2026-04-09
Add ability to request for a work area to be marked as inaccessible

mkangia:
## Product Description

This PR extends connect to accept form submissions to mark a work area as inaccessible.

## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2252
Implements [plan created by Claude](https://github.com/dimagi/commcare-connect/pull/1117/changes#diff-e352a93177f66990c240a0ee9ccf6e70092ad9030d324423212ff608508eb683)

### Safety story

Tests added for new bits

### Labels & Review


commcare_connect/microplanning/migrations/0005_workareaexpectedvisitcountworkareagroupstatusevent_and_more.py
```diff
@@ -0,0 +1,137 @@
...
+                    operation="UPDATE",
+                    pgid="pgtrigger_update_update_c5fd6",
+                    table="microplanning_workarea",
+                    when="AFTER",
+                ),
+            ),
+        ),
+        migrations.DeleteModel(
+            name="WorkAreaExpectedVisitCountWorkAreaGroupEvent",
+        ),
```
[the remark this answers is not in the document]
ajeety4: This is fine as the feature is not currently launched for real opportunities.
I have also checked the data and the existing audit records are for test opportunities.

FYI @mkangia in case of concerns.

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +258,56 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+    if new_status not in allowed:
+        raise ProcessingError(f"Cannot transition work area from {work_area.status} to {new_status}")
+
+    reason = block.get("reason", "")
+    if new_status in WORK_AREA_STATUS_REASON_REQUIRED and not reason:
+        raise ProcessingError(f"Reason is required for status {new_status}")
+
+    work_area.status = new_status
+    if new_status in WORK_AREA_STATUS_REASON_REQUIRED:
+        work_area.inaccessibility_reason = reason
```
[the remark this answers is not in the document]
ajeety4: https://github.com/dimagi/commcare-connect/pull/1118/commits/617732e5f25df24a08b1fd83d22267b55837c3b2

commcare_connect/form_receiver/tests/test_receiver_integration.py
```diff
@@ -986,3 +988,163 @@ def test_receiver_deliver_form_with_invalid_work_area_id(
...
+    events = work_area.expected_visit_count_work_area_group_status_inaccessibility_reason_events
+    assert events.count() == initial_event_count + 1
+    event = events.last()
+    assert event.pgh_context.metadata["username"] == mobile_user_with_connect_link.username
+    assert event.pgh_context.metadata["user_email"] == mobile_user_with_connect_link.email
+    assert event.status == WorkAreaStatus.REQUEST_FOR_INACCESSIBLE
+
+
+@pytest.mark.django_db
+def test_work_area_update_wrong_status(
```
zandre-eng: Nit: A lot of the tests added here have near identical implementation. I think they are good candidates to be parametrized.

commcare_connect/form_receiver/processor.py
```diff
@@ -47,6 +48,13 @@
 TASK_MODULE_JSONPATH = parse("$..task")
 ASSESSMENT_JSONPATH = parse("$..assessment")
 DELIVER_UNIT_JSONPATH = parse("$..deliver")
+WORK_AREA_UPDATE_JSONPATH = parse("$..work_area_update")
+
+ALLOWED_WORK_AREA_STATUS_TRANSITIONS = {
+    WorkAreaStatus.NOT_STARTED: {WorkAreaStatus.REQUEST_FOR_INACCESSIBLE},
+}
+
+WORK_AREA_STATUS_REASON_REQUIRED = {WorkAreaStatus.REQUEST_FOR_INACCESSIBLE}
```
zandre-eng: Do we expect more transitions or reason-required statuses to be added here in the future? If not, (YAGNI) the dict-of-sets and the singleton could be collapsed to a direct check like `work_area.status != NOT_STARTED or new_status != REQUEST_FOR_INACCESSIBLE:`

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +258,56 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+    if new_status not in allowed:
+        raise ProcessingError(f"Cannot transition work area from {work_area.status} to {new_status}")
+
+    reason = block.get("reason", "").strip()
+    if new_status in WORK_AREA_STATUS_REASON_REQUIRED and not reason:
+        raise ProcessingError(f"Reason is required for status {new_status}")
+
+    work_area.status = new_status
+    if new_status in WORK_AREA_STATUS_REASON_REQUIRED:
+        work_area.inaccessibility_reason = reason
```
zandre-eng: Any worker with a valid OAuth token could bloat both the row and the audit event with a massive reason string. Doesn't feel like a big concern, but still worth flagging. Should we cap this? Something like 1024 on the model or processor feels like ample space.

zandre-eng (approved):

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +258,56 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+
+
+def process_work_area_update(user: User, opportunity: Opportunity, block: dict):
+    work_area_case_id = block.get("work_area_id")
+    if not work_area_case_id or not is_a_uuid(work_area_case_id):
+        raise ProcessingError(f"Invalid work area case id specified: {work_area_case_id}")
+
+    try:
+        work_area = WorkArea.objects.select_for_update().get(case_id=work_area_case_id, opportunity=opportunity)
+    except WorkArea.DoesNotExist:
```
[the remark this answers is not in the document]
ajeety4: Not an issue as all requests are atomic

ajeety4:
Added 3816a88c to drop WorkArea.inaccessibility_reason and its required-validation. [CCCT-2210](https://dimagi.atlassian.net/browse/CCCT-2210) will capture the FLW request payload (including reason) on a dedicated WorkAreaInaccessibilityRequest model, so this avoids adding a column we'd immediately migrate away.                   
                                                                                                                                                     
The status transition (NOT_STARTED → REQUEST_FOR_INACCESSIBLE) and its pghistory tracking are unchanged. Migration 0005 was regenerated in place (the prior version was never deployed). Tests updated accordingly.

[CCCT-2210]: https://dimagi.atlassian.net/browse/CCCT-2210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

commcare_connect/form_receiver/processor.py
ajeety4: Good point. Although the form uses a select choice for reason field, it is good to cap it.  Will do it in ticket where I add this field now. See https://github.com/dimagi/commcare-connect/pull/1118#issuecomment-4345968384

commcare_connect/form_receiver/tests/test_receiver_integration.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1118/commits/50c22f465e981ccfef75c7eac40aec78ee89dae9

commcare_connect/form_receiver/processor.py
ajeety4: Good point. 
Addressed in https://github.com/dimagi/commcare-connect/pull/1118/commits/50c22f465e981ccfef75c7eac40aec78ee89dae9

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +252,48 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+        raise ProcessingError("Work area not found")
+
+    try:
+        access = OpportunityAccess.objects.get(opportunity=opportunity, user=user)
```
sravfeyn, line 269: We are already querying access and work_area in `process_deliver_unit`, so it's better to structure the code in a way that makes use of those available objects instead of the extra DB calls.

I did that as part of another status-update ticket [here](https://github.com/dimagi/commcare-connect/pull/1154/changes#diff-0d916b95ee38daa9788eca7b8b3266f406c5c791cab6ff6bda1d5aeefc84d5d2R468). It would be great to PR this off of that PR instead of main as both of them do the work-area status updates. If not, I do see that this is a new xml block, so it would still be nice to avoid the multiple DB calls at least

commcare_connect/microplanning/tests/test_views.py
```diff
@@ -170,7 +170,7 @@ def test_successful_field_updates(self, mock_sync, client, org_user_admin, oppor
         group = WorkAreaGroupFactory(opportunity=opportunity)
         work_area = WorkAreaFactory(opportunity=opportunity, expected_visit_count=10)
```
sravfeyn, line 172: Would be nice to add another fresh test for processing the newly added block and the relevant status updates.

commcare_connect/form_receiver/processor.py
```diff
@@ -47,6 +48,7 @@
 TASK_MODULE_JSONPATH = parse("$..task")
 ASSESSMENT_JSONPATH = parse("$..assessment")
 DELIVER_UNIT_JSONPATH = parse("$..deliver")
+WORK_AREA_UPDATE_JSONPATH = parse("$..work_area_update")
```
sravfeyn, line 51: Currently the deliver block itself has `work_area_id`. If we are adding a new block, then this will require corresponding HQ/Vellum changes as well. In addition, we will need to update existing work_area_id processing in the Connect processor as well. 

@mkangia 
This could impact multiple tickets that relate to work-area processing.

mkangia: @sravfeyn 
Aha,
Claude and I didn't think of that.

We did consider using the existing deliver unit block but went with a different block to keep things cleaner and separate.

In your opinion is it better to use the existing deliver unit block or add a new one?

> update existing work_area_id

Not sure I understand this bit, If we add a new block why would that need changes to existing workflow?

sravfeyn: > We did consider using the existing deliver unit block but went with a different block to keep things cleaner and separate.

> Not sure I understand this bit, If we add a new block why would that need changes to existing workflow?

I don't think it make sense to require app-developers to worry about setting the same `work_area_id` in two places. If you are keen on introducing a work_area block, you should remove the work_area_id from the deliver block and keep it just in the work_area block. HQ/Vellum work needs to happen regardless whether there is a new work_area block or not, since you are adding a new `status` field to the block. And then update the existing work_area update in processor to be based on the new work-area block instead of deliver-block

ajeety4: From my understanding, a user will have two separate forms - one for “deliver unit” and another for “inaccessibility.”
When the user submits the inaccessibility form, they would not submit the deliver unit form.
Therefore, I think the app developer needs to ensure that the work_area_id is set correctly in both flows.

I’m not very well-versed with the form structure at this stage, so please let me know if this understanding makes sense.

Also, I agree that we will likely need Vellum changes in both cases.

mkangia: ^ that is my understanding as well.

The inaccessibility request with use a different block and so it should not conflict with existing worfklows.
And yes they would set the work area id in this one as well which I think is fine.
I agree that this would need a change to add a new unit just like deliver unit. I don't believe there is a different way to identify the inaccessibility request in the form.

fyi @calellowitz

mkangia: I reached out to @calellowitz to help us agree on a way forward.

calellowitz: > ^ that is my understanding as well.

This is also my understanding. Since this is a new question type, in a new form, there is no impact on existing workflows, either in the app or our form processor

commcare_connect/form_receiver/processor.py
ajeety4: Since the inaccessibility update uses a different form block (question type) from the deliver unit, it's processed separately so process_deliver_unit's access and work_area objects are not in scope  when process_work_area_update runs.                                                                                                                                                                      
                                                                                                                                                                                                           
Great point on reducing DB calls. I've refactored process_work_area_update to accept all blocks at once and query access a single time (similar to how process_task_modules handles task blocks). The next  step — querying access once in process_deliver_form and sharing it across deliver, task, and inaccessibility processing — is a worthwhile optimisation but I'd prefer to do that in a separate PR to avoid conflicts with the other in-flight work-area PR.

commcare_connect/form_receiver/processor.py
mkangia: Thanks all.

I will add a new ticket for Vellum update but otherwise I believe you are now unblocked @ajeety4 .

@sravfeyn I want to be sure we don't miss anything so please let me know if you still think we are misunderstanding something here.

sravfeyn: If it's a separate form and question and there won't be a form with both blocks at the same time, then it makes sense. But, perhaps there is no constraint preventing someone to include both the deliver and this new work-area blocks, it would be very unlikely edge case, so probably not worth handling. Then this doesn't impact existing work-area (from deliver block) processing.

Yeah, the Vellum/HQ changes needs separate ticket.

commcare_connect/microplanning/tests/test_views.py
ajeety4: The tests here cover ModifyWorkAreaUpdateView (NM/PM editing expected_visit_count and work_area_group). The status transition and its audit event are covered by [test_work_area_update_inaccessible](https://github.com/dimagi/commcare-connect/pull/1118/files#diff-e9b24363614f8a71a6d5fbf2d28f37b8cf75d4bf62d7d3555589b626f168861bR1014) in test_receiver_integration.py.

Does this address your concern, or could you clarify what specific scenario you’d like the new test to cover?

commcare_connect/form_receiver/processor.py
mkangia: > there won't be a form with both blocks at the same time

for the inaccessible workflow, no.

> there is no constraint preventing someone to include both the deliver and this new work-area blocks

This is correct from my understanding. Connect can process two different blocks in one form so there should be no issue with that.

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +252,52 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
-        match.value for match in DELIVER_UNIT_JSONPATH.find(xform.form) if match.value["@xmlns"] == CCC_LEARN_XMLNS
-    ]
-    for deliver_unit_block in deliver_matches:
+    for deliver_unit_block in _get_matching_blocks(DELIVER_UNIT_JSONPATH, xform):
         process_deliver_unit(user, xform, app, opportunity, deliver_unit_block)
 
-    task_matches = [
-        match.value for match in TASK_MODULE_JSONPATH.find(xform.form) if match.value["@xmlns"] == CCC_LEARN_XMLNS
-    ]
+    task_matches = _get_matching_blocks(TASK_MODULE_JSONPATH, xform)
```
mkangia, line 258: is there a reason for it to not follow the same pattern as above?

`for deliver_unit_block in _get_matching_blocks:`

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +252,52 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+        work_area_case_id = block.get("work_area_id")
+        if not work_area_case_id or not is_a_uuid(work_area_case_id):
+            raise ProcessingError(f"Invalid work area case id specified: {work_area_case_id}")
+
+        try:
+            work_area = WorkArea.objects.select_for_update().get(case_id=work_area_case_id, opportunity=opportunity)
+        except WorkArea.DoesNotExist:
+            raise ProcessingError("Work area not found")
+
+        if not work_area.work_area_group or work_area.work_area_group.opportunity_access_id != access.id:
```
mkangia, line 285: would be good to add a ToDo here and coordinate with @hemant10yadav 

Assignment should exist on work area and not the group and that is being done in this ticket
https://dimagi.atlassian.net/browse/CCCT-2277

so depending on which PR gets merged first, this code would need to be updated.

commcare_connect/form_receiver/processor.py
```diff
@@ -250,18 +252,52 @@ def process_assessments(user, xform: XForm, app: CommCareApp, opportunity: Oppor
...
+        except ValueError:
+            raise ProcessingError(f"Invalid work area status: {requested_status}")
+
+        if not (
+            work_area.status == WorkAreaStatus.NOT_STARTED and new_status == WorkAreaStatus.REQUEST_FOR_INACCESSIBLE
+        ):
+            raise ProcessingError(f"Cannot transition work area from {work_area.status} to {new_status}")
+
+        work_area.status = new_status
+        with pghistory.context(username=user.username, user_email=user.email):
```
mkangia, line 300: very cool 👍

mkangia (commented):
non blocking feedback and I can't approve this since i am a co-author

commcare_connect/form_receiver/processor.py
ajeety4: yes, it allows to query the opportunity access to be queried once and be reused for all matching blocks

commcare_connect/form_receiver/processor.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1118/commits/307c01a3c08c1430b55e61f419f192cf4bac2bc7

zandre-eng (approved):

sravfeyn (approved):

PR #1120, opened 2026-04-14
CCCT-2338: Standardize status indicator across Learn, Deliver, and Payments tabs

mmmary314:
## Summary
- Replaces the simple suspended-only color bar on the **Learn**, **Deliver**, and **Payments** tabs with the same icon-based \`StatusIndicatorColumn\` already used on the **Connect Workers** tab
- Annotates each tab's queryset with the worker's \`UserInvite\` status via subquery so the correct icon renders
- Moves the Status column to the second position (after \`#\`) on all tabs for consistency

**Status icons:**
- ✅ Green check — Invite accepted
- 🕐 Orange clock — Invite pending
- ❌ Red X — User not found / Invite failed  
- ⊟ Black minus — User suspended

Jira: https://dimagi.atlassian.net/browse/CCCT-2338

## Screenshots

Screenshot 2026-04-14 at 5 43 37 PM

## Test plan
- [ ] Navigate to an opportunity with workers in different invite states
- [ ] Verify Connect Workers tab status icons are unchanged
- [ ] Verify Learn tab shows matching status icons
- [ ] Verify Deliver tab shows matching status icons
- [ ] Verify Payments tab shows matching status icons with Status as second column
- [ ] Verify suspended workers show the minus-square icon on all tabs

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -1212,7 +1212,7 @@ class WorkerStatusTable(tables.Table):
 class WorkerPaymentsTable(tables.Table):
     index = IndexColumn()
     user = UserInfoColumn(footer="Total")
-    suspended = SuspendedIndicatorColumn()
+    status = StatusIndicatorColumn(orderable=False)
     last_active = DMYTColumn()
     payment_accrued = tables.Column(
         verbose_name="Accrued", footer=lambda table: intcomma(sum(x.payment_accrued or 0 for x in table.data))
@@ -1243,11 +1243,11 @@ class WorkerPaymentsTable(tables.Table):
 
     class Meta:
         model = OpportunityAccess
-        fields = ("user", "suspended", "payment_accrued", "confirmed_paid")
+        fields = ("user", "payment_accrued", "confirmed_paid")
         sequence = (
             "index",
+            "status",
             "user",
-            "suspended",
             "last_active",
             "payment_accrued",
             "total_paid",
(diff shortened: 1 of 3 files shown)


commcare_connect/opportunity/tables.py
```diff
@@ -1588,11 +1588,6 @@ def render_action(self, record):
         self.run_after_every_row(record)
         return format_html(template, url)
 
-    def render_suspended(self, record, value):
-        if self._is_seen(record):
-            return ""
-        return SuspendedIndicatorColumn().render(value)
```
hemant10yadav, line 1594: I believe the `_is_seen` check is crucial here. On the deliver tab, we can have multiple rows per user, and we intentionally avoid re-rendering worker-level fields (like name and icons) for subsequent rows. Without this guard, the status icon will repeat on each sub-row. We can address this by adding:

                                                                                                           
  ```
def render_status(self, record, value):
      if self._is_seen(record):          
          return ""                                                                                                                                                         
      return StatusIndicatorColumn.render(self.columns["status"].column, record)
```

mmmary314 (commented):
Addressed — added `render_status` with `_is_seen` guard to `WorkerDeliveryTable` so the status icon only renders on the first row per worker, matching the existing behaviour of the `user` and `index` columns.

mkangia:
@hemant10yadav 
can you take another review pass at this?

hemant10yadav (approved):

PR #1125, opened 2026-04-16
Display error message when email already exists for social-login account

sravfeyn:
## Product Description

[Ticket](https://dimagi.atlassian.net/browse/CI-619)

This resolves a bug in sign-in flow where blank onboarding screen is shown when a user with a given email/password tries to login with HQ with same email. Instead of showing a blank screen we now find if the user account already exists and display a relevant error message to login with email/password instead.
Screenshot 2026-04-21 at 1 11 43 PM

### Safety story

Tested locally

### Automated test coverage

Tested locally

### QA Plan

Tested locally
### Labels & Review

### DIFF
--- a/commcare_connect/users/adapters.py
+++ b/commcare_connect/users/adapters.py
@@ -30,8 +30,6 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
             return
         if email_address_exists(email):
             messages.error(
-                request,
-                _("An account with the email %(email)s already exists. Please sign in with your email and password.")
-                % {"email": email},
+                request, _("An account with that email already exists. Please sign in with your email and password.")
             )
             raise ImmediateHttpResponse(redirect("account_login"))
--- a/commcare_connect/users/adapters.py
+++ b/commcare_connect/users/adapters.py
@@ -10,6 +10,7 @@ from django.conf import settings
 from django.contrib import messages
 from django.http import HttpRequest
 from django.shortcuts import redirect
+from django.utils.translation import gettext as _
 
 
 class AccountAdapter(DefaultAccountAdapter):
@@ -30,6 +31,7 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
         if email_address_exists(email):
             messages.error(
                 request,
-                f"An account with the email {email} already exists. " "Please sign in with your email and password.",
+                _("An account with the email %(email)s already exists. Please sign in with your email and password.")
+                % {"email": email},
             )
             raise ImmediateHttpResponse(redirect("account_login"))
--- a/commcare_connect/users/adapters.py
+++ b/commcare_connect/users/adapters.py
@@ -29,7 +29,5 @@ class SocialAccountAdapter(DefaultSocialAccountAdapter):
         if not email:
             return
         if email_address_exists(email):
-            messages.error(
-                request, _("An account with that email already exists. Please sign in with your email and password.")
-            )
+            messages.error(request, _("Unable to sign in with SSO. Please sign in with your email and password."))
             raise ImmediateHttpResponse(redirect("account_login"))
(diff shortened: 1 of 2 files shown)


commcare_connect/users/adapters.py
```diff
@@ -14,3 +20,16 @@ def is_open_for_signup(self, request: HttpRequest):
...
+    def pre_social_login(self, request: HttpRequest, sociallogin: SocialLogin):
+        if sociallogin.is_existing:
+            return
+        email = user_email(sociallogin.user)
+        if not email:
+            return
+        if email_address_exists(email):
+            messages.error(
+                request,
+                f"An account with the email {email} already exists. " "Please sign in with your email and password.",
```
zandre-eng, line 33: Nit: Could you please add translation for this?

zandre-eng (approved):
A minor non-blocking nit.

pxwxnvermx (changes requested):
Specifying that the email exists in the database is a security vulnerability known as [User Enumeration](https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/03-Identity_Management_Testing/04-Testing_for_Account_Enumeration_and_Guessable_User_Account). Attackers use this technique to confirm whether an email has a valid account on a service. An error message that references the specific email guarantees an account exists. A generic response such as "Invalid username or password" is the recommended approach.

sravfeyn:
@pxwxnvermx Thanks. I have updated the text to remove email. Though, in this case the email isn't manually entered and comes from a verified HQ account. But nevertheless, there's no need to disclose the email address

commcare_connect/users/adapters.py
sravfeyn: Added

commcare_connect/users/adapters.py
```diff
@@ -30,8 +30,6 @@ def pre_social_login(self, request: HttpRequest, sociallogin: SocialLogin):
             return
         if email_address_exists(email):
             messages.error(
-                request,
-                _("An account with the email %(email)s already exists. Please sign in with your email and password.")
-                % {"email": email},
+                request, _("An account with that email already exists. Please sign in with your email and password.")
```
zandre-eng, line 33: I feel like this change might not entirely remove the concern Pawan brought up, since the message still essentially indicates that the email used exists. Even though the email isn't manually entered a user can still make a guess based on the email of the verified HQ account.

What if we used a more generic message like:

"Unable to sign in with SSO. Please sign in with your email and password instead."

Charl1996: +1

sravfeyn: Updated

Charl1996 (approved):

mkangia:
Ya, I am in a split too about this. I don't think we want to share that email already exists, though I understand you mentioned that the account is present on HQ.

@calellowitz can you help us align here?

pxwxnvermx (approved):
Thanks. 🙌🏽

mkangia:
@sravfeyn 

Looks like you revised your approach here, is the description out of date?

fyi @calellowitz the original concern I flagged to you might not be there anymore.

sravfeyn:
The approach is same, only the error message has been updated to simply say SSO login failed and to try with email/password instead. @mkangia 

The only alternate approach is to merge both SSO and email/password based accounts for that user and allow signing in with both the modes into same user account. But I think that's more nuanced and need to be thought out a bit more. The current approach of simply showing a better messaging to nudge users to sign in using email/password is much better as this is a case of people mostly forgetting that they have already signed up using email/password and then trying SSO login.

mkangia:
Okay, can you please update the screenshot in the description? I believe the error message is different now.

sravfeyn:
Done, let me know if this g2g are you want to wait for Cal on this @mkangia

mkangia:
@sravfeyn 
Good to go, we can follow up if @calellowitz has any additional concerns. I had pinged him based on the earlier approach that was shared in the description.

sravfeyn:
Thanks!

PR #1127, opened 2026-04-16
Back-end View to Exclude Work Areas

zandre-eng:
## Product Description

Organization admins can now exclude work areas from an opportunity when they are no longer actionable (for example, due to flooding, inaccessibility, or a mapping error). Excluded work areas are removed from the mobile workers' assignment on HQ and marked `EXCLUDED` in Connect with an audit trail of who excluded them and why.

<insert screenshot of the exclude flow on the work areas page>
<insert screenshot showing an excluded work area in its post-exclusion state>

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2279)

Adds a POST endpoint that enqueues an async Celery task to exclude a set of work areas. The task batches HQ case-unassign calls to **50 per PUT request** against HQ's v2 case API (`/a/<domain>/api/case/v2/`) and applies the local DB side via `WorkArea.objects.bulk_update` inside a per-chunk `transaction.atomic() + pghistory.context`. A chunk that fails on HQ logs a warning, counts the chunk as failed, and leaves those rows as `NOT_STARTED` for admin retry — other chunks still commit. HQ's own bulk endpoint is all-or-nothing (verified against [corehq/apps/hqcase/api/updates.py:210-213](https://github.com/dimagi/commcare-hq/blob/master/corehq/apps/hqcase/api/updates.py#L210-L213) and empirically against a local HQ instance), so one bad `case_id` in a chunk cannot cause split-state between HQ and Connect.

- **New view** `exclude_work_areas` at [commcare_connect/microplanning/views.py](commcare_connect/microplanning/views.py): POST-only, org-admin-gated, validates `exclusion_reason` (required, ≤500 chars) and `work_area_ids[]` (required, integers, ≤200 items), then enqueues the task and returns `202`.
- **New async task** `exclude_work_areas_task` at [commcare_connect/microplanning/tasks.py](commcare_connect/microplanning/tasks.py): partitions requested work areas into HQ-needing and DB-only sets, chunks HQ calls by `HQ_BULK_CHUNK_SIZE = 50`, and delegates DB mutation to a private `_bulk_exclude` helper that sets `status=EXCLUDED`, `excluded_by`, `excluded_reason`, and clears `work_area_group` in a single `bulk_update`.
- **New HQ helper** `bulk_update_cases` at [commcare_connect/commcarehq/api.py](commcare_connect/commcarehq/api.py): PUT of a JSON array to `/api/case/v2/`; injects HQ's required `create: False` flag per item and raises `CommCareHQAPIException` on any non-2xx.
- **Unassignment convention**: sends `owner_id="-"` (HQ's `UNOWNED_EXTENSION_OWNER_ID`) rather than `""`. Empty string triggers HQ's fallback at [corehq/form_processor/backends/sql/update_strategy.py:140-141](https://github.com/dimagi/commcare-hq/blob/master/corehq/form_processor/backends/sql/update_strategy.py#L140-L141) that silently re-assigns the case to the submitting API user.
- **Model changes** in [commcare_connect/microplanning/models.py](commcare_connect/microplanning/models.py): `WorkArea` gains `excluded_by` (EmailField) and `excluded_reason` (CharField), and `excluded_reason` is added to the existing `@pghistory.track` field list so the change is captured in the audit-event table.
- **Audit trail**: every chunk's `bulk_update` runs inside `pghistory.context(reason, username, user_email)`; pghistory's DB triggers on `WorkArea` write the context into the event table alongside the `excluded_reason` column. `excluded_by` is stored as an email (not a user FK) so the audit remains readable if the user is later deleted.
- **Design doc**: [docs/plans/2026-04-16-exclude-work-areas-design.md](docs/plans/2026-04-16-exclude-work-areas-design.md) and [docs/plans/2026-04-24-bulk-exclude-work-areas-design.md](docs/plans/2026-04-24-bulk-exclude-work-areas-design.md).

### Safety story

- **Access control**: endpoint requires `org_admin_required` + `opportunity_required` decorators.
- **Input validation**: `exclusion_reason` (required, max 500 chars) and `work_area_ids[]` (required, integers, max 200 per request) validated server-side; malformed input returns 400 before the task is enqueued.
- **Only eligible rows are excluded**: the task filters on `status=NOT_STARTED`. Work areas in any other state (e.g., `VISITED`, `INACCESSIBLE`, already `EXCLUDED`) are counted as `skipped` and left untouched — no silent override of existing state.
- **All-or-nothing per HQ chunk**: HQ bulk-update either applies all rows in the chunk or none. On HQ failure, the whole chunk's Connect DB rows are left as `NOT_STARTED` (retryable), preventing Connect/HQ state divergence. Verified by reading HQ source (`_submit_case_blocks` runs only after all case blocks are validated) and by running the integration script against a local HQ with a deliberately-invalid `case_id` mixed in.
- **Per-chunk atomicity**: each chunk's DB update is its own `transaction.atomic()`, so if the task crashes mid-run, already-committed chunks remain excluded and the rest are retryable.
- **Audit trail survives user deletion**: `excluded_by` is stored as an email string rather than a user FK; pghistory context captures username and email for the event row.
- **Async execution**: the view returns 202 and the task runs under Celery, so a large request cannot block an admin's browser.

### Automated test coverage

**Task-level** (`TestExcludeWorkAreasTask` in [commcare_connect/microplanning/tests/test_views.py](commcare_connect/microplanning/tests/test_views.py), 12 test cases):
- `test_happy_path_excludes_not_started_areas` — 2 areas → 1 HQ call, all excluded.
- `test_mixed_batch_only_not_started_is_excluded` — other statuses in the batch are skipped.
- `test_hq_batch_failure_skips_local_exclusion_for_whole_chunk` — HQ raises → whole chunk stays `NOT_STARTED`, `work_area_group` preserved.
- `test_no_case_id_excludes_locally_without_hq_call` — areas without `case_id` go down the DB-only path and still get excluded.
- `test_work_area_from_other_opportunity_is_ignored` — cross-opportunity IDs in the input are a no-op.
- `test_work_started_statuses_are_not_excluded` — parametrized across `VISITED`, `NOT_VISITED`, `UNASSIGNED`, `REQUEST_FOR_INACCESSIBLE`, `EXPECTED_VISIT_REACHED`.
- `test_work_areas_over_chunk_size_are_split_into_batches` — 125 areas → 3 chunks sized `[50, 50, 25]`; all excluded.
- `test_one_failed_chunk_does_not_block_other_chunks` — 150 areas across 3 chunks, chunk 2 raises; chunks 1 and 3 still commit.

**View-level** (`TestExcludeWorkAreasView`, 4 cases): valid request → 202 + task enqueued; invalid `exclusion_reason` → 400 (parametrized over missing/blank/too-long); invalid `work_area_ids[]` → 400 (missing/non-integer); too-many-items → 400.

**HQ helper** (`test_bulk_update_cases` in [commcare_connect/commcarehq/tests/test_api.py](commcare_connect/commcarehq/tests/test_api.py), parametrized over 200 and 500): asserts correct HTTP verb (PUT), payload shape (including injected `create: False` per item), and that non-2xx raises `CommCareHQAPIException`.

**Manual integration verification** against a local HQ via [scripts/verify_exclude_bulk_hq.py](scripts/verify_exclude_bulk_hq.py): confirmed real `owner_id` transitions on HQ after running the task, and all-or-nothing behaviour when a non-existent `case_id` is mixed into a batch.

### QA Plan

- [ ] As an org admin on an opportunity with existing work areas, select one or more `NOT_STARTED` work areas and submit the exclude request with a non-empty reason. Expect a 202 response and the task to run asynchronously.
- [ ] Refresh the work-areas view and confirm the selected rows now show `EXCLUDED` with the reason and excluder email recorded.
- [ ] Inspect the corresponding cases on HQ and confirm their `owner_id` is now `-` (unassigned), and that a mobile worker logging in sees those cases gone from their assignment.
- [ ] Submit an exclude request with a work area whose status is not `NOT_STARTED` (e.g., `INACCESSIBLE`) and confirm it is silently skipped (no state change).
- [ ] Submit an exclude request with empty `exclusion_reason` and confirm a 400 response.
- [ ] Submit an exclude request with more than 200 IDs and confirm a 400 response.
- [ ] Check the pghistory event table for a new row recording the exclusion with the correct `excluded_reason`, `username`, and `user_email` context.
- [ ] Confirm that excluding a work area with no `case_id` still marks it `EXCLUDED` locally and does not produce an HQ call.

### Labels & Review


commcare_connect/microplanning/views.py
```diff
@@ -409,6 +409,87 @@
...
+                    create_or_update_case(api_key, domain, {"owner_id": ""}, case_id=str(work_area.case_id))
+
+        except CommCareHQAPIException as e:
+            logger.info(f"Failed to unassign HQ case for work area {work_area_id}: {e}")
+            failed.append({"id": work_area_id, "reason": f"HQ sync failed: {e}"})
+            continue
+
+        excluded.append(work_area_id)
+
+    return JsonResponse({"excluded": excluded, "skipped": skipped, "failed": failed})
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [b3e96af](https://github.com/dimagi/commcare-connect/pull/1127/commits/b3e96afbb9dbd2ac796aa21586842beb85fed4f7).

commcare_connect/microplanning/views.py
```diff
@@ -409,6 +409,87 @@ def clustering_status(request, org_slug, opp_id):
...
+        return JsonResponse({"error": "exclusion_reason is required"}, status=400)
+
+    raw_ids = request.POST.getlist("work_area_ids[]")
+    if not raw_ids:
+        return JsonResponse({"error": "work_area_ids[] is required"}, status=400)
+
+    try:
+        work_area_ids = [int(i) for i in raw_ids]
+    except (ValueError, TypeError):
+        return JsonResponse({"error": "work_area_ids[] must be integers"}, status=400)
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [cd5a028](https://github.com/dimagi/commcare-connect/pull/1127/commits/cd5a0284b3407009da509be8feea6ab179012b09)

commcare_connect/microplanning/migrations/0005_workarea_excluded_fields.py
```diff
@@ -0,0 +1,133 @@
...
+                    operation="UPDATE",
+                    pgid="pgtrigger_update_update_c5fd6",
+                    table="microplanning_workarea",
+                    when="AFTER",
+                ),
+            ),
+        ),
+        migrations.DeleteModel(
+            name="WorkAreaExpectedVisitCountWorkAreaGroupEvent",
+        ),
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [cd5a028](https://github.com/dimagi/commcare-connect/pull/1127/commits/cd5a0284b3407009da509be8feea6ab179012b09)

commcare_connect/microplanning/views.py
```diff
@@ -409,6 +409,87 @@ def clustering_status(request, org_slug, opp_id):
     return HttpResponse(headers={"HX-Redirect": redirect_url})
 
 
+@require_POST
+@org_admin_required
+@opportunity_required
+def exclude_work_areas(request, org_slug, opp_id):
```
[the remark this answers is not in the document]
zandre-eng: This is a non-issue. We are pushing towards releasing more parts of this feature without requiring a feature flag.

commcare_connect/microplanning/models.py
```diff
@@ -40,7 +41,7 @@ def building_count(self):
         return self.workarea_set.aggregate(total=Sum("building_count"))["total"] or 0
 
 
-@pghistory.track(fields=["expected_visit_count", "work_area_group"])
+@pghistory.track(fields=["expected_visit_count", "work_area_group", "excluded_reason"])
```
Charl1996, line 44: Why would we want to include the `excluded_reason` in the audit trail?

commcare_connect/microplanning/views.py
```diff
@@ -409,6 +409,89 @@ def clustering_status(request, org_slug, opp_id):
...
+    raw_ids = request.POST.getlist("work_area_ids[]")
+    if not raw_ids:
+        return JsonResponse({"error": "work_area_ids[] is required"}, status=400)
+
+    try:
+        work_area_ids = [int(i) for i in raw_ids]
+    except (ValueError, TypeError):
+        return JsonResponse({"error": "work_area_ids[] must be integers"}, status=400)
+
+    excluded = []
```
Charl1996, line 431: Nit: It feels like this part (up until line 490) could be a function on its own returning `excluded`, `skipped` and `failed`.

commcare_connect/microplanning/views.py
```diff
@@ -409,6 +409,89 @@ def clustering_status(request, org_slug, opp_id):
...
+@org_admin_required
+@opportunity_required
+def exclude_work_areas(request, org_slug, opp_id):
+    exclusion_reason = request.POST.get("exclusion_reason", "").strip()
+    if not exclusion_reason:
+        return JsonResponse({"error": "exclusion_reason is required"}, status=400)
+    if len(exclusion_reason) > 500:
+        return JsonResponse({"error": "exclusion_reason must be at most 500 characters"}, status=400)
+
+    raw_ids = request.POST.getlist("work_area_ids[]")
```
sravfeyn, line 422: The PR description mentions that we expect small amount of work-areas for preferring synchronous processing. Should the count of raw_ids be capped here?

mkangia:
Couple of things:
1. Do we need the model change to track excluded_by if we already have that info via pghistory? I see we want to have audit even in case of user deletion. I am not sure from where AI (claude) got that learning but it could be from a few PRs we have been discussing this point. It's possible we will not be deleting pghistory records.
Additionally, this introduces an additional complexity if the status is updated from EXCLUDED to something else which most probably won't happen since EXCLUDED won't be updated any further.
So, I believe both approaches are okay for now but just wanted to drop this thought here.
2. I believe we would want even the INACCESSIBLE ones to be allowed to go into EXCLUDED. Is this documented somewhere that we won't update INACCESSIBLE TO EXCLUDED?

I am a bit split about doing synchronous update for a smaller batch. I'd vote that we always do it async to avoid the branching.

commcare_connect/microplanning/models.py
zandre-eng: My reasoning was that it allows us to keep an audit in case anything were to clear the `excluded_by` field so that we can still track who initially excluded this work area. In saying that, this doesn't feel like highly valuable auditing and I think I'll do one of two things: (1) remove this from auditing and we rely on the `excluded_by` and `excluded_reason` fields or (b) remove the `excluded_by` field and track the reason in pghistory where we'll store the user that updated the work area.

zandre-eng:
> Do we need the model change to track excluded_by if we already have that info via pghistory? I see we want to have audit even in case of user deletion. I am not sure from where AI (claude) got that learning but it could be from a few PRs we have been discussing this point. It's possible we will not be deleting pghistory records.
Additionally, this introduces an additional complexity if the status is updated from EXCLUDED to something else which most probably won't happen since EXCLUDED won't be updated any further.
So, I believe both approaches are okay for now but just wanted to drop this thought here.

@mkangia My motivation for having a separate `excluded_by` field was to allow for easier querying. I imagine that we might want to eventually include in the UI or a report who excluded a work area. If that's not true, then I'm happy to move this over to `pghistory` if we're keeping it purely for auditing.

> I believe we would want even the INACCESSIBLE ones to be allowed to go into EXCLUDED. Is this documented somewhere that we won't update INACCESSIBLE TO EXCLUDED?

The ticket acceptance criteria lists the following "Work Areas cannot be updated to 'Excluded' if the Work Area has already been marked as 'Inaccessible'". I'm happy to update the code if this is an incorrect statement?

> I am a bit split about doing synchronous update for a smaller batch. I'd vote that we always do it async to avoid the branching.

I did some performance testing and while the function itself runs fine (0.73s for 1k work areas), I can see the synchronous call to HQ becoming a bottleneck for timing out this request. Thinking through this some more I can agree that an async approach might be the better strategy here. Will update.

commcare_connect/microplanning/views.py
zandre-eng: That makes sense, addressed in [8e309c4](https://github.com/dimagi/commcare-connect/pull/1127/commits/8e309c42a2aaf17b838a950e51778169c1fd880c).

commcare_connect/microplanning/views.py
```diff
@@ -462,6 +462,100 @@ def clustering_status(request, org_slug, opp_id):
...
+            with transaction.atomic(), pghistory.context(
+                reason=exclusion_reason,
+                username=user.username,
+                user_email=user.email,
+            ):
+                work_area.status = WorkAreaStatus.EXCLUDED
+                work_area.excluded_by = user
+                work_area.excluded_reason = exclusion_reason
+                work_area.work_area_group = None
+                work_area.save(update_fields=["status", "excluded_by", "excluded_reason", "work_area_group"])
```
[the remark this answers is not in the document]
zandre-eng: I went a different route on this.

Skipping `select_for_update` — the races here are idempotent (concurrent excludes converge on the same `EXCLUDED` state, and the HQ unassign call is idempotent too) or impossible given the model (visits require prior assignment, which transitions out of `NOT_STARTED`). Extra query per row without closing a race that has benign outcomes.

What I did change: reordered HQ → DB (was DB → HQ). If HQ fails now, local state is untouched, so admin retry replays cleanly — no divergent state to reconcile. Also drops external I/O out of the transaction, so the row lock is held for milliseconds instead of across an HQ round-trip.

HQ credential reads still happen before any mutation, as you noted.

Addressed in [a28541d](https://github.com/dimagi/commcare-connect/pull/1127/commits/a28541d22fb60b1678f6fbb2ec4e74d7ccc8da19).

commcare_connect/microplanning/views.py
zandre-eng: That's a fair point, I think it makes sense to have a reasonable cap in the view.

Addressed in [1159026](https://github.com/dimagi/commcare-connect/pull/1127/commits/1159026a74ccfbe1e16ef687b88bfc7feea6c628).

commcare_connect/microplanning/models.py
Charl1996: Oh, I see. I think it's fine as is.

Charl1996 (approved):

commcare_connect/microplanning/tasks.py
```diff
@@ -309,3 +315,69 @@ def cluster_work_areas_task(opp_id):
...
+        domain = None
+        if work_area.work_area_group and work_area.work_area_group.opportunity_access:
+            opp_access = work_area.work_area_group.opportunity_access
+            if opp_access.opportunity.api_key and opp_access.opportunity.deliver_app:
+                api_key = opp_access.opportunity.api_key
+                domain = opp_access.opportunity.deliver_app.cc_domain
+
+        if work_area.case_id and api_key and domain:
+            try:
+                create_or_update_case(api_key, domain, {"owner_id": ""}, case_id=str(work_area.case_id))
```
hemant10yadav: HQ has bulk update user case api right?

commcare_connect/microplanning/tasks.py
```diff
@@ -309,3 +315,69 @@ def cluster_work_areas_task(opp_id):
...
+            skipped += 1
+            continue
+
+        api_key = None
+        domain = None
+        if work_area.work_area_group and work_area.work_area_group.opportunity_access:
+            opp_access = work_area.work_area_group.opportunity_access
+            if opp_access.opportunity.api_key and opp_access.opportunity.deliver_app:
+                api_key = opp_access.opportunity.api_key
+                domain = opp_access.opportunity.deliver_app.cc_domain
```
pxwxnvermx: The ApiKey and Domain can be taken directly from the Opportunity object already fetched in the task definition. This would remove the need to fetch them per Work Area, since all Work Areas belong to a single Opportunity and eliminate the extra joins introduced by the select_related queries above.

commcare_connect/microplanning/tasks.py
```diff
@@ -309,3 +315,69 @@ def cluster_work_areas_task(opp_id):
...
+
+    for work_area_id in work_area_ids:
+        work_area = work_areas_map.get(work_area_id)
+        if work_area is None or work_area.status != WorkAreaStatus.NOT_STARTED:
+            skipped += 1
+            continue
+
+        api_key = None
+        domain = None
+        if work_area.work_area_group and work_area.work_area_group.opportunity_access:
```
pxwxnvermx: The Opportunity Access is moving from WorkAreaGroups to WorkAreas soon. Can you add a TODO comment here to make that change once the migration is done?

commcare_connect/microplanning/models.py
```diff
@@ -65,6 +66,14 @@ class WorkArea(geo_models.Model):
     )
     case_id = geo_models.UUIDField(null=True, blank=True, unique=True)
     case_properties = geo_models.JSONField(default=dict, null=True, blank=True)
+    excluded_by = geo_models.ForeignKey(
```
pxwxnvermx, line 69: We generally store the email for the User that made these changes. How do you feel about switching this field from a User FK to email?

commcare_connect/microplanning/tasks.py
zandre-eng: After some investigation I can confirm HQ does support bulk updates. I've updated my implementation to use bulk updates.

Addressed in [f78c777](https://github.com/dimagi/commcare-connect/pull/1127/commits/f78c77712069cc74554058bd36554dd8dc3e35ba).

commcare_connect/microplanning/tasks.py
zandre-eng: Good point. I refactored the code to use the opp object instead as part of the bulk update changes.

Addressed in [b780a28](https://github.com/dimagi/commcare-connect/pull/1127/commits/b780a2862755794a2511e64b119aae469edaa197).

commcare_connect/microplanning/tasks.py
zandre-eng: This is no longer needed. Work area groups are not referenced at all anymore in the task.

commcare_connect/microplanning/models.py
zandre-eng: I'm happy with that change if it's more consistent with our code.

Addressed in [b2da4d2](https://github.com/dimagi/commcare-connect/pull/1127/commits/b2da4d2b76726d5c4012022987d675eab3be1fe0)

commcare_connect/commcarehq/api.py
```diff
@@ -109,6 +109,26 @@ def create_or_update_case(
...
+
+    All-or-nothing: raises CommCareHQAPIException on any non-2xx response.
+    """
+    base_url = f"{api_key.hq_server.url}/a/{domain}/api/case/v2/"
+    headers = {"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"}
+    payload = [{**update, "create": False} for update in updates]
+
+    try:
+        with httpx.Client(base_url=base_url, headers=headers) as client:
+            response = client.put("", json=payload)
```
pxwxnvermx: The API for bulk updating cases is a POST api.
https://commcare-hq.readthedocs.io/api/cases-v2.html#bulk-create-update-cases

commcare_connect/microplanning/models.py
```diff
@@ -65,6 +65,8 @@ class WorkArea(geo_models.Model):
     )
     case_id = geo_models.UUIDField(null=True, blank=True, unique=True)
     case_properties = geo_models.JSONField(default=dict, null=True, blank=True)
+    excluded_by = geo_models.EmailField(blank=True, default="")
```
pxwxnvermx: Can you please revert this change? I had a similar comment for another PR and Cal provided an explanation that has changed my understanding regarding tracking User using emails vs FK.

Apologies for the back and forth on this. 😄 

https://github.com/dimagi/commcare-connect/pull/1147/changes#r3137612260

commcare_connect/commcarehq/api.py
zandre-eng: Good catch. The docs specify `POST` for the bulk endpoint, but it seems `PUT` works as well because HQ's view routes both methods to the same handler. But given only `POST` is documented, I'll update here.

zandre-eng: Addressed in [f43a648](https://github.com/dimagi/commcare-connect/pull/1127/commits/f43a6485c25f03253ddb9dcb43d5fe74e69891f6). Also tested this switch using a local HQ instance.

commcare_connect/microplanning/models.py
zandre-eng: No worries! That makes sense for tighter auditing.

Reverted in [7fa32ad](https://github.com/dimagi/commcare-connect/pull/1127/commits/7fa32ad449acde16cb569c724432a4621df5683b).

commcare_connect/commcarehq/api.py
```diff
@@ -109,6 +109,26 @@ def create_or_update_case(
     return CommCareCase(**data.get("case", {}))
 
 
+def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, Any]]) -> None:
+    """POST a JSON array of case updates to /api/case/v2/.
+
+    HQ's bulk endpoint requires each row to carry a create flag; this helper
+    only updates existing cases, so create=False is injected per item.
+
+    All-or-nothing: raises CommCareHQAPIException on any non-2xx response.
```
mkangia, line 180: I think the documentation needs to be updated as we are doing a single request. So there is no "any" non-2xx response now.

commcare_connect/microplanning/views.py
```diff
@@ -462,6 +464,39 @@ def clustering_status(request, org_slug, opp_id):
     return HttpResponse(headers={"HX-Redirect": redirect_url})
 
 
+@require_POST
+@org_admin_required
+@opportunity_required
+def exclude_work_areas(request, org_slug, opp_id):
+    exclusion_reason = request.POST.get("exclusion_reason", "").strip()
+    if not exclusion_reason:
+        return JsonResponse({"error": "exclusion_reason is required"}, status=400)
```
mkangia: do we need translation here? Assuming this is user facing.

mkangia: seems to be applicable for other response strings as well.

commcare_connect/microplanning/views.py
```diff
@@ -462,6 +464,39 @@ def clustering_status(request, org_slug, opp_id):
...
+    if len(exclusion_reason) > 500:
```
mkangia, line 477: what is the reason for the 500 limit?

commcare_connect/microplanning/views.py
```diff
@@ -462,6 +464,39 @@ def clustering_status(request, org_slug, opp_id):
...
+        return JsonResponse({"error": "exclusion_reason must be at most 500 characters"}, status=400)
+
+    raw_ids = request.POST.getlist("work_area_ids[]")
+    if not raw_ids:
+        return JsonResponse({"error": "work_area_ids[] is required"}, status=400)
```
mkangia: is that a user facing error? If yes, I wonder if `work_area_ids[]` is the right thing to show.

commcare_connect/microplanning/views.py
```diff
@@ -462,6 +464,39 @@ def clustering_status(request, org_slug, opp_id):
...
+            {"error": f"work_area_ids[] must contain at most {MAX_EXCLUDE_WORK_AREAS} items"},
+            status=400,
+        )
+
+    try:
+        work_area_ids = [int(i) for i in raw_ids]
+    except (ValueError, TypeError):
+        return JsonResponse({"error": "work_area_ids[] must be integers"}, status=400)
+
+    exclude_work_areas_task.delay(
```
mkangia: it seems we capped to 200 work areas so we can make this synchronous but then we are using a celery task?
what did I miss?

mkangia (dismissed):

mkangia:
Just noting that once this is merged, the following PRs can move ahead as well
https://github.com/dimagi/commcare-connect/pull/1138
https://github.com/dimagi/commcare-connect/pull/1143

commcare_connect/microplanning/views.py
zandre-eng: Since we are doing HQ calls in batches of 50, if we get 200 work areas that means 4 API calls will need to be made to HQ. My motivation here was to be cautious and not cause the view to be stuck waiting on external API calls in case HQ might be slow, especially if this is getting called from the UI somewhere.

With the above said, I'm happy to switch over to a completely synchronous call if you think doing multiple calls to HQ should be fine.

commcare_connect/microplanning/views.py
zandre-eng: This is to match the DB constraint on the `exclusion_reason` field which is capped at 500 characters. I added this DB constraint given the assumption that this value will be provided by user input, and so I wanted to ensure users can't create massive multi-mb string entries in the DB.

commcare_connect/commcarehq/api.py
zandre-eng: There could still be a non 2xx response. In the function we have a try/except block for `httpx.HTTPStatusError` and `httpx.RequestError`, so if the HQ request fails for any reason (e.g. HQ is unreachable) then it'll be caught and a `CommCareHQAPIException` will be raised.

commcare_connect/microplanning/views.py
zandre-eng: > is that a user facing error?

It could be. There is no front-end wired up yet, so these errors can either be used directly or the front-end can provide its own strings based on what the back-end sends.

But assuming that these will get used directly, I've updated all the error strings to be more user friendly. Addressed in [c8719b0](https://github.com/dimagi/commcare-connect/pull/1127/commits/c8719b06541622b09286dc4c6100e13fee37b4f1)

commcare_connect/microplanning/views.py
zandre-eng: Based on my response from [this](https://github.com/dimagi/commcare-connect/pull/1127#discussion_r3187130822) comment, I've updated these error strings to be more user friendly and have added translation.

Addressed in [c8719b0](https://github.com/dimagi/commcare-connect/pull/1127/commits/c8719b06541622b09286dc4c6100e13fee37b4f1).

commcare_connect/microplanning/views.py
zandre-eng: Based on how it seems we are handling bulk HQ calls in other places, I've decided to change this over to a synchronous call. Thinking it through, it is only 4 API calls which should still be responsive enough in a view.

Addressed in [c141a45](https://github.com/dimagi/commcare-connect/pull/1127/commits/c141a45d5ec33ea0ea0fc8a75ea85a82952ebbf5).

mkangia: Sorry, I was just checking on the mismatch.
It appeared we adding capping to make synchronous calls but then did the update async.

We just need to ensure the same state between HQ and Connect when it comes to work area records and cases.

zandre-eng: > It appeared we adding capping to make synchronous calls but then did the update async.

That's a fair point, which is why I think making this call synchronous is more consistent and and is a bit more clear for a dev reading through it.

> We just need to ensure the same state between HQ and Connect when it comes to work area records and cases

That should be accounted for here. The synchronous call will do the update in batches and if either the HQ call or DB update fails, the entire batch should be rolled back. I did push a refactor in ([ae48939](https://github.com/dimagi/commcare-connect/pull/1127/commits/ae48939bcf1f7123596df68a0c089bd05254ee65)) so that we can be sure that if HQ fails the Connect DB will still be in sync.

pxwxnvermx (approved):

hemant10yadav (approved):

pxwxnvermx (approved):

PR #1133, opened 2026-04-21
create new release on deploy to production

calellowitz:
## Technical Summary

This updates the deploy workflow to create a release automatically when deploying to production

### Safety story

deploy workflow only

### QA Plan

Tested on staging

### Labels & Review

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,7 +7,7 @@ env:
 
 permissions:
   id-token: write
-  contents: read
+  contents: write
 
 on:
   #  workflow_run:
@@ -112,6 +112,19 @@ jobs:
           set_commits: manual
           repo: dimagi/commcare-connect
           commit: ${{ github.sha }}
+
+      - name: Set release date
+        if: ${{ success() && inputs.environment == 'production' }}
+        run: echo "RELEASE_DATE=$(date -u +'%Y-%m-%d')" >> "$GITHUB_ENV"
+
+      - name: Create GitHub release
+        if: ${{ success() && inputs.environment == 'production' }}
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.RELEASE_DATE }}
+          name: ${{ env.RELEASE_DATE }}
+          target_commitish: ${{ github.sha }}
+          generate_release_notes: true
 
       - name: Notify Deploy End
         uses: slackapi/slack-github-action@v1.24.0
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -113,6 +113,19 @@ jobs:
...
+        run: echo "RELEASE_DATE=$(date -u +'%Y-%m-%d')" >> "$GITHUB_ENV"
+
+      - name: Create GitHub release
+        if: ${{ success() && inputs.environment == 'production' }}
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.RELEASE_DATE }}
+          name: ${{ env.RELEASE_DATE }}
+          target_commitish: ${{ github.sha }}
+          generate_release_notes: true
```
mkangia, line 127: What does this do? List all PRs and descriptions?

.github/workflows/deploy.yml
```diff
@@ -113,6 +113,19 @@ jobs:
...
 
+      - name: Set release date
+        if: ${{ success() && inputs.environment == 'production' }}
+        run: echo "RELEASE_DATE=$(date -u +'%Y-%m-%d')" >> "$GITHUB_ENV"
+
+      - name: Create GitHub release
+        if: ${{ success() && inputs.environment == 'production' }}
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.RELEASE_DATE }}
```
mkangia, line 124: what happens if there are two deploys on the same day?

.github/workflows/deploy.yml
calellowitz: yeah, its the same thing that currently happens when you press the button manually. Its a builtin github option

.github/workflows/deploy.yml
calellowitz: That's a great question, this step would probably fail. I think that's a pretty rare occurrence, but not impossible. HQ uses the time as well. That would be easy to add

pxwxnvermx (approved):

.github/workflows/deploy.yml
mkangia: so you want to add that or use just the date?

calellowitz: sorry took a while, but did update the tag. feel free to review again

pxwxnvermx (approved):

ajeety4 (approved):

Charl1996 (approved):

hemant10yadav (approved):

mkangia (approved):

PR #1135, opened 2026-04-21
opportunity automation apis

calellowitz:
## Technical Summary

<!--
    Provide a link to any tickets, design documents, and/or technical specifications
    associated with this change. Describe the rationale and design decisions.
-->
This adds new APIs to allow automated creation of an opportunity from the point of program creation to inviting users

● Seven new POST endpoints under /api/, split across the program and opportunity apps:                                                                                                        
                                                                                                                                                                                              
  - POST /api/programs/ — Create a program
  - POST /api/programs/{program_id}/applications/ — Invite an organization to a program                                                                                                       
  - POST /api/programs/{program_id}/applications/{application_id}/accept/ — Accept an invited organization (automation shortcut)                                                              
  - POST /api/programs/{program_id}/opportunities/ — Create a managed opportunity (fetches app metadata from CommCare HQ synchronously, includes dates + budget)                              
  - POST /api/opportunities/{opportunity_id}/payment_units/ — Add payment units (accepts a list)                                                                                              
  - POST /api/opportunities/{opportunity_id}/activate/ — Activate the opportunity                                                                                                             
  - POST /api/opportunities/{opportunity_id}/invite_users/ — Invite users by phone number                                                                                                     

Best reviewed commit by commit, each of which handles one view except the final one which handles the end to end tests.

Claude [Design Doc](https://github.com/dimagi/commcare-connect/commit/bfdc2a12c675a7ce8267c82f5d72896129fb1a79) here as well

## Safety Assurance

### Safety story

<!--
Describe how you became confident in this change, such as
local testing, why the change is inherently safe, and/or plans to limit the blast radius of a defect.

In particular consider how existing data may be impacted by this change.
-->
Internal use only and heavily covered by tests

### Automated test coverage

<!-- Identify the related test coverage and the tests it would catch -->
new tests for each individual API and the end to end workflow


### Labels & Review


commcare_connect/program/api/serializers.py
```diff
@@ -99,3 +111,220 @@ class ProgramApplicationResponseSerializer(serializers.ModelSerializer):
...
+                return app["name"]
+        raise serializers.ValidationError(
+            {
+                "non_field_errors": [
+                    _("App '{cc_app_id}' not found in domain '{domain}'.").format(cc_app_id=cc_app_id, domain=domain)
+                ]
+            }
+        )
+
+    def create(self, validated_data):
```
Charl1996, line 210: Nit:
There's a lot of stuff being created during this routine, so I wonder if it makes sense to wrap this as an atomic transaction?

From a functional point of view I think this method is fine without wrapping as an atomic transaction since you're using `get_or_create` where it matters, but my concern is more geared towards making sure there's no half-baked opportunities in the DB if it fails halfway through.

commcare_connect/opportunity/api/serializers/automation.py
```diff
@@ -0,0 +1,129 @@
...
+
+from commcare_connect.opportunity.models import DeliverUnit, PaymentUnit
+
+
+class PaymentUnitItemSerializer(serializers.Serializer):
+    name = serializers.CharField(max_length=255)
+    description = serializers.CharField(required=False, default="")
+    amount = serializers.IntegerField(min_value=0)
+    org_amount = serializers.IntegerField(min_value=0, required=False)
+    max_total = serializers.IntegerField(min_value=1)
```
Charl1996, line 19: We should probably add validation around `amount`, `max_daily` and `max_total`.

commcare_connect/opportunity/api/serializers/automation.py
```diff
@@ -0,0 +1,129 @@
...
+    start_date = serializers.DateField(required=False, allow_null=True, default=None)
+    end_date = serializers.DateField(required=False, allow_null=True, default=None)
+
+    def validate(self, data):
+        opportunity = self.context["opportunity"]
+
+        if opportunity.managed and not data.get("org_amount"):
+            raise serializers.ValidationError({"org_amount": _("org_amount is required for managed opportunities.")})
+
+        if data["start_date"] and data["end_date"] and data["end_date"] < data["start_date"]:
```
Charl1996, line 32: Add validation against the opportunity's start/end dates.

Charl1996:
Out of curiosity, why are these endpoints added? Do we expect some client to use these to set up opportunities in the future?

Charl1996:
Also, why add the `api/automation_views.py` file when there exists a `api/views.py` file? I'm not sure what is expected to go into the one vs the other.

commcare_connect/program/api/serializers.py
```diff
@@ -0,0 +1,330 @@
...
+        """Fetch the app name from HQ. Results cached per (api_key, domain) to avoid duplicate calls."""
+        if not hasattr(self, "_apps_cache"):
+            self._apps_cache = {}
+        cache_key = (api_key.pk, domain)
+        if cache_key not in self._apps_cache:
+            try:
+                self._apps_cache[cache_key] = get_applications_for_user_by_domain(api_key, domain)
+            except CommCareHQAPIException as exc:
+                raise serializers.ValidationError(
+                    {"non_field_errors": [_("Failed to fetch apps from CommCare HQ: {exc}").format(exc=exc)]}
```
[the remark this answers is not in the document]
pxwxnvermx: this looks serious. maybe we can log the stack trace only to sentry and return this error without the exception.

commcare_connect/opportunity/api/views/automation.py
```diff
@@ -0,0 +1,92 @@
...
+    PaymentUnitResponseSerializer,
+    UserInviteSerializer,
+)
+from commcare_connect.opportunity.models import Opportunity, PaymentUnit
+from commcare_connect.opportunity.tasks import add_connect_users
+from commcare_connect.organization.decorators import user_is_org_admin
+from commcare_connect.program.models import ManagedOpportunity
+
+
+class OpportunityMixin:
```
pxwxnvermx: Currently, this mixin calls `permission_denied`, which creates an implicit dependency on APIView. Because it cannot function or be tested independently, we should update the mixin to inherit from APIView directly. This will ensure it is fully self-contained.

commcare_connect/opportunity/api/views/automation.py
```diff
@@ -0,0 +1,92 @@
...
+    """Shared logic for views scoped to an opportunity — resolves opportunity and checks permissions."""
+
+    def get_opportunity(self):
+        opportunity = get_object_or_404(Opportunity, opportunity_id=self.kwargs["opportunity_id"])
+        allowed_orgs = [opportunity.organization]
+        if opportunity.managed:
+            managed = ManagedOpportunity.objects.get(pk=opportunity.pk)
+            allowed_orgs.append(managed.program.organization)
+        if not any(user_is_org_admin(self.request.user, org) for org in allowed_orgs):
+            self.permission_denied(self.request)
```
pxwxnvermx, line 32: this seems safe, but to be extra careful can we also add an empty `return` here to avoid returning any opportunity.

commcare_connect/program/api/views.py
```diff
@@ -0,0 +1,111 @@
...
+    permission_classes = [IsAuthenticated, IsProgramManagerAdmin]
+
+    def post(self, request):
+        serializer = ProgramCreateSerializer(data=request.data, context={"request": request})
+        serializer.is_valid(raise_exception=True)
+        program = serializer.save()
+        return Response(ProgramResponseSerializer(program).data, status=status.HTTP_201_CREATED)
+
+
+class ProgramMixin:
```
pxwxnvermx: this does the same thing as the earlier mixin. Can both of these be updated to inherit the API View?

pxwxnvermx (commented):
Some comments around view Mixins. Rest of the changes look good.

calellowitz:
Thanks for the review @Charl1996 , will reply to the inline comments there, but for your top level questions

> Out of curiosity, why are these endpoints added? Do we expect some client to use these to set up opportunities in the future?

A few people are working on a tool to speed up the process of testing new interventions using AI tools. They are covering everything from once they have an idea to determining success, including iterating on the idea, building the apps, and analyzing the data that comes out of it. This PR is about building hooks for that tool to handle the opportunity creation aspects.

> Also, why add the api/automation_views.py file when there exists a api/views.py file? I'm not sure what is expected to go into the one vs the other.

My mental model was that the existing file was for the mobile APIs, anything FLWs are doing to interact with the opportunity, while this is web automation. The users, web vs mobile, and use case are fairly different, and I think it can be nice to break things up along those lines to avoid massive files, but I would agree the naming isn't ideal.

calellowitz:
I think i got to all the human feedback and most of the coderabbit stuff. It is insatiable and keeps posting a ton of stuff on every push, so I haven't gotten to all of it, but I am ready for a new review @Charl1996 @pxwxnvermx 

I also added [design doc](https://github.com/dimagi/commcare-connect/commit/bfdc2a12c675a7ce8267c82f5d72896129fb1a79) to the description, because I realized I had missed it before.

commcare_connect/opportunity/api/serializers/automation.py
calellowitz: Can you say more about this? I am not sure how we can validate those before the payment units are created, which happens in a later step

commcare_connect/opportunity/api/serializers/automation.py
calellowitz: Offlined about this, but leaving out for now since this is for internal use and that maintains existing behavior

commcare_connect/program/api/views.py
calellowitz: done

commcare_connect/opportunity/api/views/automation.py
calellowitz: done

commcare_connect/opportunity/api/views/automation.py
calellowitz: The point of `permission_denied` is to short circuit the view. If that fails, I feel like we have bigger problems (it would be like putting `return` after `raise`)

commcare_connect/program/api/serializers.py
```diff
@@ -0,0 +1,330 @@
...
+            status=ProgramApplicationStatus.ACCEPTED,
+        ).exists():
+            raise serializers.ValidationError(_("Organization must have an accepted application for this program."))
+        return value
+
+    def validate(self, data):
+        program = self.context["program"]
+
+        if data["learn_app"]["cc_app_id"] == data["deliver_app"]["cc_app_id"]:
+            raise serializers.ValidationError(_("Learn app and deliver app must be different."))
```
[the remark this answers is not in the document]
calellowitz: those IDs are intended to be globally unique so ignoring for now

commcare_connect/program/api/serializers.py
```diff
@@ -0,0 +1,330 @@
...
+            currency=program.currency,
+            country=program.country,
+            delivery_type=program.delivery_type,
+            api_key=learn_api_key,
+            hq_server=learn_data["hq_server"],
+            active=False,
+            created_by=user.email,
+            modified_by=user.email,
+        )
+        return opportunity
```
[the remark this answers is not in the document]
calellowitz: This matches existing behavior where the opp only has one API key. It's a reasonable point but its out of scope for this PR

commcare_connect/opportunity/api/serializers/automation.py
```diff
@@ -0,0 +1,165 @@
...
+                amount=pu_data["amount"],
+                org_amount=pu_data.get("org_amount", 0),
+                max_total=pu_data["max_total"],
+                max_daily=pu_data["max_daily"],
+                start_date=pu_data.get("start_date"),
+                end_date=pu_data.get("end_date"),
+            )
+            self._assign_deliver_units(pu_data.get("required_deliver_units", []), pu, optional=False)
+            self._assign_deliver_units(pu_data.get("optional_deliver_units", []), pu, optional=True)
+            created.append(pu)
```
hemant10yadav, line 106: Don’t we need to create claims for already invited users?

commcare_connect/program/api/serializers.py
```diff
@@ -0,0 +1,329 @@
...
+class ManagedOpportunityCreateSerializer(serializers.Serializer):
+    name = serializers.CharField(max_length=255)
+    description = serializers.CharField()
+    short_description = serializers.CharField(max_length=255)
+    organization = serializers.SlugRelatedField(slug_field="slug", queryset=Organization.objects.all())
+    start_date = serializers.DateField()
+    end_date = serializers.DateField()
+    total_budget = serializers.IntegerField(min_value=1)
+    learn_app = LearnAppInputSerializer()
+    deliver_app = CommCareAppInputSerializer()
```
hemant10yadav, line 138: Just wanted to confirm—leaving the is_test field as-is is intentional, since it defaults to True.

commcare_connect/program/api/views.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class ProgramApplicationCreateView(ProgramAPIView):
+    def post(self, request, program_id):
+        program = self.get_program()
+        serializer = ProgramApplicationCreateSerializer(
+            data=request.data, context={"request": request, "program": program}
+        )
+        serializer.is_valid(raise_exception=True)
+        application = serializer.save()
```
hemant10yadav, line 59: send_program_invite_email ?

commcare_connect/program/api/views.py
```diff
@@ -0,0 +1,115 @@
...
+            httpx.RequestError,
+            httpx.TimeoutException,
+            httpx.ConnectError,
+        ):
+            logger.exception("Failed to fetch app metadata from HQ while creating managed opportunity")
+            return Response(
+                {"non_field_errors": [_("Failed to fetch app metadata from CommCare HQ.")]},
+                status=status.HTTP_502_BAD_GATEWAY,
+            )
+
```
hemant10yadav, line 114: send_opportunity_created_email ?

hemant10yadav (commented):
Looks good overall—just had a few questions and suggestions.

commcare_connect/opportunity/api/serializers/automation.py
calellowitz: there are no already invited users at this time. unles im missing something

commcare_connect/program/api/views.py
calellowitz: good point

commcare_connect/program/api/views.py
calellowitz: good idea

commcare_connect/program/api/serializers.py
calellowitz: this should probably be configurable in the API, let me update that

commcare_connect/opportunity/api/serializers/automation.py
hemant10yadav: I think an admin can add additional payment units even after users have been invited. In that case, we need to ensure that claims are created for already invited users.
Our current UI workflow already accounts for this scenario and handles the creation of claims for existing users when new payment units are added.

calellowitz: I see. I think thats unlikely for the use cases of this API, but added to be safe.

hemant10yadav (approved):

PR #1144, opened 2026-04-23
Visually Group User Visits with Work Areas on Microplanning Map

zandre-eng:
## Product Description

When a work area is selected on the microplanning map, its linked user visits are now highlighted as green diamonds rendered on top of the existing red dots. All unrelated visits simultaneously fade to 20% opacity so the selected work area's visits visually pop out. Clearing the selection returns visits to their normal red-dot appearance.

An example of a work area selected, showing green diamonds overlaid on its visits with other visits dimmed:
Screenshot_20260424_151449

An example of the default state (no selection) for comparison:
Screenshot_20260423_131606

This only applies in normal mode — assignment mode does not support showing user visits and so highlighting of user visits does not apply to that mode.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2340)

Implementation is frontend-first with one small backend change. The existing `selectedFeature` in the `mapController` Alpine component remains the single source of truth; a derived `highlightedWorkAreaId` (synced via a `$watch` on `selectedFeature`) drives styling on two Mapbox layers. A `symbol` layer type is required for the diamond shape because Mapbox's `circle` layer can only render circles.

- **Backend — expose `work_area_id` on visit tiles:** `UserVisitVectorLayer` now includes `work_area_id` in `tile_fields` and in the queryset values so the client can filter features without re-fetching tiles.
- **Client — derived highlight state:** new Alpine field `highlightedWorkAreaId`, updated by `updateHighlightedWorkAreaId()` which returns `null` in assignment mode or when nothing is selected, otherwise the selected feature's id. Wired via `$watch('selectedFeature', ...)`.
- **Client — hidden symbol layer:** new `user-visits-highlight` layer (symbol type, `◆` text-field, green fill, white halo) added alongside `user-visits-circle`. Added via `addVisitHighlightLayer()` when visits are first shown; starts hidden.
- **Client — styling driver:** new `refreshVisitStyling()`, wired via `$watch('highlightedWorkAreaId', ...)`, sets `circle-opacity` on the base circles (0.2 dimmed / 0.7 default) and toggles + filters the highlight layer (`["==", ["get", "work_area_id"], id]`).
- **Client — paint builder extraction:** circle paint config was extracted into `buildVisitCirclePaint()` so layer creation and styling refresh stay in sync.

### Safety story

This change is purely additive:

- Backend change is a no-op for existing consumers: `tile_fields` and `.values()` additions ship one extra column (`work_area_id`) in tiles. Existing clients that don't reference the property ignore it.
- No migrations, no data mutations, no new endpoints, no permission changes.
- New layer and state are gated on `showVisits` and start hidden; in the default (no-selection) state the map looks and behaves exactly as before.
- Assignment mode is explicitly excluded from the highlight logic via `updateHighlightedWorkAreaId`, so the assignment-mode UX is unaffected.
- Reversible by revert — no persisted state.

### Automated test coverage

- `TestUserVisitVectorLayer.test_queryset_annotates_location_point` in [commcare_connect/microplanning/tests/test_views.py](commcare_connect/microplanning/tests/test_views.py) was extended to assert `visit["work_area_id"] == visit_data.work_area.id`. Catches regressions in `tile_fields` or the queryset `.values()` that would strip `work_area_id` from emitted tiles.

Frontend interaction changes are verified manually (see QA Plan).

### QA Plan

- [ ] Open the microplanning map for an opportunity that has user visits and confirm the baseline state: red dots at 0.7 opacity, no diamonds visible.
- [ ] Toggle "Show Visits" off and on; confirm visits appear/disappear without errors and styling state is preserved.
- [ ] Click a work area that has linked visits; confirm:
  - Green diamonds appear on exactly the visits whose `work_area_id` matches the selected work area.
  - All other visit circles dim to ~0.2 opacity.
  - Diamonds are visibly larger than the base circles at typical zoom levels (zoom 8 → 14).
- [ ] Click the same work area again to deselect; confirm diamonds disappear and circles return to 0.7 opacity.
- [ ] Click a different work area; confirm the diamond set updates to match the new selection (no stale diamonds from the previous selection).
- [ ] Click a work area with **no** linked visits; confirm circles dim but no diamonds appear.
- [ ] Change filters while a work area is selected; confirm the tile reload preserves correct diamond/dim state for the new feature set.
- [ ] Enter assignment mode, multi-select several work areas; confirm **no** diamond highlighting and **no** circle dimming occurs (assignment mode is excluded).
- [ ] Exit assignment mode back to normal mode; confirm highlight behavior resumes on next single-select.
- [ ] Zoom in and out across the visit zoom range; confirm diamonds remain legibly sized and aligned over their circles.

### Labels & Review

### DIFF
--- a/commcare_connect/templates/microplanning/map_handler.html
+++ b/commcare_connect/templates/microplanning/map_handler.html
@@ -5,6 +5,7 @@
         Alpine.data("mapController", () => ({
             map: null,
             selectedFeature: null,
+            highlightedWorkAreaId: null,
             showWorkAreaEditModal: false,
             showVisits: false,
             filterQuery: '',
@@ -33,6 +34,12 @@
                 const url = editWorkAreaUrl + workAreaId + '/';
                 htmx.ajax('GET', url, { target: '#work-area-form', swap: 'innerHTML' })
                 .then(() => this.showWorkAreaEditModal = true)
+            },
+
+            updateHighlightedWorkAreaId() {
+                this.highlightedWorkAreaId = (this.assignmentMode || !this.selectedFeature)
+                    ? null
+                    : this.selectedFeature._id;
             },
 
             getDownloadUrl() {
@@ -338,6 +345,8 @@
                 this.$nextTick(() => {
                     this.initializeMap();
                 });
+
+                this.$watch('selectedFeature', () => this.updateHighlightedWorkAreaId());
 
                 if (this.assignmentMode) {
                     const assigneesEl = document.getElementById('assignees-data');
--- a/commcare_connect/templates/microplanning/map_handler.html
+++ b/commcare_connect/templates/microplanning/map_handler.html
@@ -79,8 +79,8 @@
                     },
                     paint: {
                         'text-color': '#22c55e',
-                        'text-halo-color': '#ffffff',
-                        'text-halo-width': 1.5,
+                        'text-halo-color': '#000000',
+                        'text-halo-width': 2.5,
                     },
                 });
             },
--- a/commcare_connect/templates/microplanning/map_handler.html
+++ b/commcare_connect/templates/microplanning/map_handler.html
@@ -58,6 +58,35 @@
                 };
             },
 
+            refreshVisitStyling() {
+                if (!this.map) return;
+
+                if (this.map.getLayer('user-visits-circle')) {
+                    this.map.setPaintProperty(
+                        'user-visits-circle',
+                        'circle-opacity',
+                        this.highlightedWorkAreaId !== null ? 0.2 : 0.7,
+                    );
+                }
+
+                if (this.map.getLayer('user-visits-highlight')) {
+                    const id = this.highlightedWorkAreaId;
+                    if (id === null) {
+                        this.map.setLayoutProperty('user-visits-highlight', 'visibility', 'none');
+                    } else {
+                        this.map.setFilter(
+                            'user-visits-highlight',
+                            ['==', ['get', 'work_area_id'], id],
+                        );
+                        this.map.setLayoutProperty(
+                            'user-visits-highlight',
+                            'visibility',
+                            this.showVisits ? 'visible' : 'none',
+                        );
+                    }
+                }
+            },
+
             getDownloadUrl() {
                 const baseUrl = '{{ download_url|escapejs }}';
                 return this.filterQuery ? `${baseUrl}?${this.filterQuery}` : baseUrl;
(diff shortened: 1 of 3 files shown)


commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -35,6 +36,80 @@
...
+                        this.map.setLayoutProperty('user-visits-highlight', 'visibility', 'none');
+                    } else {
+                        this.map.setFilter(
+                            'user-visits-highlight',
+                            ['==', ['get', 'work_area_id'], id],
+                        );
+                        this.map.setLayoutProperty('user-visits-highlight', 'visibility', 'visible');
+                    }
+                }
+            },
```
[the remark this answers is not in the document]
zandre-eng: Did a force push and addressed in commit [4d89126](https://github.com/dimagi/commcare-connect/pull/1144/commits/4d89126869957b8dfb440698c43926d33ae6afd6).

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -35,6 +36,84 @@
...
+                    source: 'user-visits',
+                    'source-layer': 'user-visits',
+                    layout: {
+                        'text-field': '◆',
+                        'text-size': [
+                            'interpolate', ['linear'], ['zoom'],
+                            8, 13,
+                            14, 22,
+                        ],
+                        'text-font': ['DIN Pro Medium', 'Arial Unicode MS Regular'],
```
pxwxnvermx, line 75: nit: Can these be replaced with commonly available fonts or browser generics like "Sans", "Serif" etc?

pxwxnvermx (approved):
Just a small nit. Rest of the changes look good to me.

commcare_connect/templates/microplanning/map_handler.html
zandre-eng: I got the following response after investigating it with Claude:

`text-font` here isn't a CSS font — Mapbox GL renders text from SDF glyphs served by the style's `glyphs` endpoint, so the value has to match a font the active style actually ships. Generics like `Sans`/`Serif` aren't resolved and the layer would fail to render glyphs. `DIN Pro Medium` + `Arial Unicode MS Regular` are the fonts bundled with the `mapbox://styles/mapbox/standard` style we're using (Arial Unicode is the standard CJK fallback). Happy to swap to another Mapbox-hosted stack if you'd prefer, but browser generics won't work here.

Given the above, I'm opted to keep the current font as it ships with the Mapbox style. For verification I did switch it to `Sans` font and received a 404 error in the browser's dev console which confirms that generic fonts won't work.

mkangia:
:diamond_shape_with_a_dot_inside:

hemant10yadav (approved):

pxwxnvermx (approved):

PR #1106, opened 2026-04-07
add tech spec for weekly performance report

calellowitz:
As per our recent discussion about using Claude design docs in place of tech specs, and using github to review them, I am opening this draft PR to facilitate discussion of my spec for the weekly performance report as described in [this ticket](https://dimagi.atlassian.net/browse/CCCT-2273). Please leave any comments on this PR, and I can continue to iterate on the underlying doc.

The doc should be fairly self explanatory and is designed to be read in order, but as with any spec, all questions welcome.

PR #1147, opened 2026-04-24
Audit backend

calellowitz:
## Technical Summary

This is the first part of the audit work outlined in [this spec](https://github.com/dimagi/commcare-connect/pull/1106). It lays the groundwork by adding the relevant models and setting up the audit calculation system. It does not include any of the UI elements or implement any of the calculations. Those will come in future PRs.

### Safety story

This change modifies no behavior, just adds unused models and a base class

### Automated test coverage

The models are and calculation class are tested.

### QA Plan

QA will happen once the feature is complete.

### Labels & Review


commcare_connect/audit/calculations.py
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    name: ClassVar[str]
+    label: ClassVar[str]
+    min_sample_size: ClassVar[int] = 1
+    lower_bound: ClassVar[float | None] = None
+    upper_bound: ClassVar[float | None] = None
+
+    @abstractmethod
+    def compute(self, opportunity_access, period_start, period_end) -> tuple[Any, int]:
```
pxwxnvermx, line 52: nit: please add a `raise NotImplementedError` for this method.

commcare_connect/audit/models.py
```diff
@@ -0,0 +1,67 @@
...
+    opportunity_access = models.ForeignKey(
+        OpportunityAccess, on_delete=models.CASCADE, related_name="audit_report_entries"
+    )
+    results = models.JSONField(default=dict)
+    flagged = models.BooleanField(default=False)
+    reviewed = models.BooleanField(default=False)
+    review_action = models.CharField(max_length=32, choices=ReviewAction.choices, null=True, blank=True)
+
+    class Meta:
+        ordering = ["opportunity_access__user__name", "date_created"]
```
pxwxnvermx: nit: can we remove this `opportunity_access__user__name` ordering here? I would prefer to do this explicitly in querysets due to the joins.

commcare_connect/audit/models.py
```diff
@@ -0,0 +1,67 @@
...
+    class Status(models.TextChoices):
+        PENDING = "pending", "Pending"
+        COMPLETED = "completed", "Completed"
+
+    audit_report_id = models.UUIDField(default=uuid.uuid4, unique=True, editable=False)
+    opportunity = models.ForeignKey(Opportunity, on_delete=models.CASCADE, related_name="audit_reports")
+    period_start = models.DateField()
+    period_end = models.DateField()
+    status = models.CharField(max_length=16, choices=Status.choices, default=Status.PENDING)
+    completed_by = models.ForeignKey(
```
pxwxnvermx, line 20: We use email for tracking changes made by users to objects. How do you feel about updating this to be an email instead of the User?

pxwxnvermx (commented):
just some minor comments and queries.

commcare_connect/audit/models.py
```diff
@@ -0,0 +1,67 @@
...
+
+    audit_report_entry_id = models.UUIDField(default=uuid.uuid4, unique=True, editable=False)
+    audit_report = models.ForeignKey(AuditReport, on_delete=models.CASCADE, related_name="entries")
+    opportunity_access = models.ForeignKey(
+        OpportunityAccess, on_delete=models.CASCADE, related_name="audit_report_entries"
+    )
+    results = models.JSONField(default=dict)
+    flagged = models.BooleanField(default=False)
+    reviewed = models.BooleanField(default=False)
+    review_action = models.CharField(max_length=32, choices=ReviewAction.choices, null=True, blank=True)
```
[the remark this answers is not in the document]
calellowitz: I don't believe this is correct. null means the row has not been reviewed, `NONE` means the review is complete and no further action is required

commcare_connect/audit/calculations.py
calellowitz: The `@abstractmethod` [decorator](https://docs.python.org/3/library/abc.html#abc.abstractmethod) implies that. It doesn't seem to be standard practice to mix both

commcare_connect/audit/models.py
calellowitz: Since this will be displayed in the UI, having access to the name of the user, rather than an email (and all the guarantees of a true foreign key rather than an implicit one) make this a bit nicer to have. As we try to develop more robust auditing workflows, I imagine we will want more of this and fewer free text fields like we used to use.

commcare_connect/audit/models.py
calellowitz: done

commcare_connect/audit/models.py
pxwxnvermx: Thanks for the context. Using the user fields sounds good to me.

pxwxnvermx (approved):
Changes look good. The test failure looks related to the migrations sequence.

pxwxnvermx (approved):

hemant10yadav (approved):

PR #1148, opened 2026-04-24
Push Notification on Task Assignment

pxwxnvermx:
## Product Description
- New Celery task `send_task_assignment_notification` in `commcare_connect/opportunity/tasks.py` that builds the notification payload and dispatches push notification.
- create_task view schedules the task via transaction.on_commit(...)  required because ATOMIC_REQUESTS = True, so the Celery worker must not race ahead of the DB commit.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2373)

### Safety story
- Tested on staging.

### Automated test coverage
- Added automated tests.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tests/test_tasks.py
+++ b/commcare_connect/opportunity/tests/test_tasks.py
@@ -494,8 +494,8 @@ def test_send_task_assignment_notification(send_message_patch):
             usernames=[access.user.username],
             data={
                 "action": "ccc_generic_opportunity",
-                "title": "Notification Title",
-                "body": "Notification description",
+                "title": "New Task Assigned",
+                "body": "A task has been assigned to you. You must complete it before continuing Delivery activities.",
                 "opportunity_uuid": str(opportunity.opportunity_id),
                 "opportunity_status": "delivery",
                 "key": "task_assignment",
--- a/commcare_connect/opportunity/tests/test_tasks.py
+++ b/commcare_connect/opportunity/tests/test_tasks.py
@@ -33,9 +33,11 @@ from commcare_connect.opportunity.tasks import (
     generate_work_status_export,
     notify_user_for_scored_assessment,
     save_export,
+    send_task_assignment_notification,
 )
 from commcare_connect.opportunity.tests.factories import (
     AssessmentFactory,
+    AssignedTaskFactory,
     CompletedModuleFactory,
     CompletedWorkFactory,
     LearnModuleFactory,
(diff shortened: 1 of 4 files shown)


ajeety4 (approved):
LGTM

commcare_connect/opportunity/tests/test_tasks.py
```diff
@@ -473,3 +476,30 @@ def test_generate_catchment_area_export(self, mock_export_fn, mock_save, opportu
...
+        Message(
+            usernames=[access.user.username],
+            data={
+                "action": "ccc_generic_opportunity",
+                "title": "New Task Assigned",
+                "body": "A task has been assigned to you. You must complete it before continuing Delivery activities.",
+                "opportunity_uuid": str(opportunity.opportunity_id),
+                "opportunity_status": "delivery",
+                "key": "task_assignment",
+                "session_endpoint_id": "cc_app_home",
```
mkangia, line 502: Ideally, `cc_app_home` will be a constant.
I understand it was not obvious here but it will be used at one more place (work area assignment notification)

mkangia (approved):

commcare_connect/opportunity/views.py
```diff
@@ -3505,13 +3506,14 @@ def create_task(request, org_slug, opp_id):
...
 
-    AssignedTask.objects.create(
+    assigned_task = AssignedTask.objects.create(
         task_type=task,
         opportunity_access=access,
         due_date=due_date,
         status=AssignedTaskStatus.ASSIGNED,
         assigned_by=request.user,
     )
+    transaction.on_commit(lambda: send_task_assignment_notification.delay(assigned_task.pk))
```
mkangia, line 3516: is this expected to work without an atomic block?

mkangia: this PR conflicted with a different PR, so I resolved the conflict but then moved the notification call to be within an atomic block
https://github.com/dimagi/commcare-connect/pull/1150/changes/b9a6dd912bc22ca76c39f055f524ba74293883fe

pxwxnvermx: > is this expected to work without an atomic block?

The Push Notification should be sent when the transaction block is commited, otherwise the task will not find the task as it would not be created until the commit happens. I think it leads to an error if we place the push notification function call in the atomic block as it tries to access a table that has a write lock.

mkangia: So my query was that if this would work as expected without an atomic block, I see there is no atomic block here.

hence the change I made in the PR this code got added during merge conflict
https://github.com/dimagi/commcare-connect/pull/1150/changes/b9a6dd912bc22ca76c39f055f524ba74293883fe

pxwxnvermx: yeah, that looks good. Thanks for making the change.

PR #1149, opened 2026-04-27
Work Area Assigned Notification

pxwxnvermx:
## Product Description
- Adds `send_work_area_assignment_notification` Celery task that sends a single push notification to an FLW with the opportunity UUID and a `work_area_assignment` key (CCCT-2374).
- Wires the task into the `save_assignment` view: dedupes assignees from the payload, validates they belong to the opportunity, and schedules one notification per FLW via `transaction.on_commit`, so multi-area assignments produce a single consolidated notification.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2374) 

### Automated test coverage
- Added Automated Tests

### Labels & Review

### DIFF
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -315,7 +315,7 @@ def send_work_area_assignment_notification(opportunity_access_id: int):
         data={
             "action": "ccc_generic_opportunity",
             "title": "New Work Areas Assigned",
-            "body": "You have been assigned new work areas. Open the app to start.",
+            "body": "You have been assigned new work areas. Click here to begin.",
             "opportunity_uuid": str(access.opportunity.opportunity_id),
             "opportunity_status": "delivery",
             "key": "work_area_assignment",
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -9,6 +9,9 @@ from django.core.files.storage import default_storage
 from django.utils.html import strip_tags
 from django.utils.translation import gettext as _
 
+from commcare_connect.connect_id_client import send_message
+from commcare_connect.connect_id_client.models import Message
+from commcare_connect.opportunity.models import OpportunityAccess
 from config import celery_app
 
 from .clustering import WorkAreaGrouper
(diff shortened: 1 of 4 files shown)


commcare_connect/microplanning/views.py
```diff
@@ -588,4 +589,22 @@ def get_flw_summary_for_assignment(request, org_slug, opp_id):
...
+        if isinstance(a, dict) and a.get("assignee_id") is not None
+    }
+    valid_assignee_ids = list(
+        OpportunityAccess.objects.filter(pk__in=assignee_ids, opportunity=request.opportunity).values_list(
+            "pk", flat=True
+        )
+    )
+    for access_id in valid_assignee_ids:
+        transaction.on_commit(lambda aid=access_id: send_work_area_assignment_notification.delay(aid))
+
```
pxwxnvermx, line 609: This is just generated stub code as we don't have the save_assignment code yet for Work Area Assignments. Will be changed in future PRs. I will add a TODO comment here for more clarity.

commcare_connect/microplanning/tasks.py
```diff
@@ -304,6 +307,24 @@ def rows(cls, queryset):
...
 
+@celery_app.task()
+def send_work_area_assignment_notification(opportunity_access_id: int):
+    access = OpportunityAccess.objects.select_related("user", "opportunity").get(pk=opportunity_access_id)
+    message = Message(
+        usernames=[access.user.username],
+        data={
+            "action": "ccc_generic_opportunity",
+            "title": "New Work Areas Assigned",
+            "body": "You have been assigned new work areas. Open the app to start.",
```
mkangia: just confirming that this message body was checked with Product about?

Confused about "Open the app to start". Clicking on the notification redirects user to the app so they should not need to do anything?

commcare_connect/microplanning/views.py
```diff
@@ -588,4 +589,22 @@ def get_flw_summary_for_assignment(request, org_slug, opp_id):
 @require_flag_for_opp(MICROPLANNING)
 def save_assignment(request, org_slug, opp_id):
     """Stub endpoint for saving work area assignments. Accepts the payload but does not persist changes yet."""
+    try:
+        payload = json.loads(request.body or b"{}")
```
mkangia, line 593: in what scenario won't there be a request.body?

mkangia (commented):
Just couple of queries but this seems like on the right path.

mkangia (approved):

commcare_connect/microplanning/views.py
pxwxnvermx: I think that is the stub generated code. I looked at a recent PR for adding the code for this stub. I think that PR does a similar thing so we can keep this and it will be updated by that PR.

commcare_connect/microplanning/tasks.py
pxwxnvermx: Updated. https://github.com/dimagi/commcare-connect/commit/9abb4d89f76c257880604d6336334d24c1a6414c

mkangia (approved):

PR #1150, opened 2026-04-27
Update HQ Usercase on Task Assignment

zandre-eng:
## Product Description

When an admin assigns a task to a worker in Connect, the worker's CommCare HQ mobile-worker usercase is automatically updated with a property matching the `TaskType.case_property` set to `"1"`. This lets HQ apps gate forms, modules, or display logic on whether a Connect task has been assigned. No UI changes — the effect is server-side only.

## Technical Summary

[Link to ticket](https://dimagi.atlassian.net/browse/CCCT-2381)

The change wires up a one-way sync from a newly created `AssignedTask` to the worker's HQ usercase. Rather than plug the side-effect into the single existing creation site, the `AssignedTask` model gets a central `assign` classmethod that handles both row creation and the post-commit Celery enqueue, so any future creation site is covered too. The HQ call is async via Celery (with retries) so HQ outages don't block task assignment, and `transaction.on_commit` ensures the worker never sees a row that hasn't been committed.

- **New classmethod `AssignedTask.assign(*, task_type, opportunity_access, due_date, assigned_by=None)`**: single entry point that creates the row inside the request transaction and registers a `transaction.on_commit` hook to enqueue the sync task. Function-local import of the Celery task avoids a circular import (`tasks.py` already imports `AssignedTask`).
- **New Celery task `sync_assigned_task_to_usercase(assigned_task_id)`**: loads the `AssignedTask`, reads `task_type.case_property`, and calls the existing `update_usercase` helper with `{"properties": {<case_property>: "1"}}`. Skips silently when `case_property` is null/empty.
- **Retry policy** mirrors the existing `download_user_visit_attachments` precedent: `autoretry_for=(CommCareHQAPIException,)`, `max_retries=5`, exponential backoff with jitter capped at 5 minutes. Any HQ failure (timeout, 4xx, 5xx) retries; `ConnectIDUserLink.DoesNotExist` (worker not yet linked to HQ) intentionally fails fast since retrying won't help.
- **`create_task` view migrated**: the only production creation site now calls `AssignedTask.assign(...)` instead of `AssignedTask.objects.create(...)`. Test factories and `form_receiver` test setup still use `objects.create` directly — bypassing the side effect is the desired behavior for unit tests.

### Safety story

- The HQ call is asynchronous and isolated from the request path, so a misbehaving HQ cannot block or fail task assignment from the admin's perspective.
- `transaction.on_commit` guarantees the Celery task is only enqueued after the `AssignedTask` row is durably committed — no race where the worker fetches a not-yet-visible row.
- `update_usercase` is idempotent (PUT on a known case ID), so retries cannot corrupt state.
- After 5 exhausted retries the task lands in Celery's failed-task logs / Sentry; the `AssignedTask` row remains regardless. The case property simply won't be set on HQ — recoverable manually or by re-assigning.
- A live integration test was run against a local HQ: an `AssignedTask.assign(...)` call wrote `connect_assigned_test=1` to a real usercase via the full code path, was verified by reading the case back, then cleaned up.
- Existing data is not affected — no migrations, no backfill of the property for already-existing `AssignedTask` rows. The behavior only applies to rows created after this lands.

### Automated test coverage

- `commcare_connect/opportunity/tests/test_tasks.py::TestSyncAssignedTaskToUsercase` — covers the happy path (correct payload `{"properties": {<slug>: "1"}}`), the skip-when-no-`case_property` cases (parametrized over `None` and `""`), and that `CommCareHQAPIException` propagates so the Celery decorator can retry (would catch a regression where the function silently swallows HQ errors).
- `commcare_connect/opportunity/tests/test_models.py::TestAssignedTaskAssign::test_creates_row_and_enqueues_sync` — covers row fields and confirms the Celery task is enqueued only after commit (uses `django_capture_on_commit_callbacks`).
- `commcare_connect/opportunity/tests/test_views.py::TestCreateTask::test_create_task_success` — updated to assert the view path actually triggers the after-commit enqueue.

### QA Plan

- [ ] As an admin on a managed opportunity, assign a task to a worker that has a linked HQ user — confirm the worker's HQ usercase shows the new `<task_type.case_property>=1` property.
- [ ] Create a `TaskType` with no `case_property` set, assign it — confirm the row is created and no HQ call is made.
- [ ] Stop HQ (or simulate an HQ outage) and assign a task — confirm task assignment still succeeds in Connect; the Celery task retries; no admin-facing error.
- [ ] Assign a task to a worker that has no `ConnectIDUserLink` (not yet linked to HQ) — confirm the row is created and the failure surfaces in Sentry / Celery logs without blocking.
- [ ] Confirm existing `AssignedTask` creation paths in tests/factories continue to work without the side effect.

### Labels & Review


commcare_connect/opportunity/models.py
```diff
@@ -453,6 +453,20 @@ class Meta:
...
+
+        task = cls.objects.create(
+            task_type=task_type,
+            opportunity_access=opportunity_access,
+            due_date=due_date,
+            status=AssignedTaskStatus.ASSIGNED,
+            assigned_by=assigned_by,
+        )
+        transaction.on_commit(lambda: sync_assigned_task_to_usercase.delay(task.id))
+        return task
```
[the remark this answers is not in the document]
zandre-eng: Not applying either change.

**BaseModel inheritance:** Every other `XFormBaseModel` subclass in this file (`CompletedModule`, `Assessment`, `UserVisit`) follows the same pattern of not mixing in `BaseModel`. `AssignedTask` already has its own `date_created` (would clash with `BaseModel.date_created`), and audit trail is covered by the existing `@pghistory.track` decorator plus the explicit `assigned_by` FK. Adding BaseModel here just for one model would create asymmetry and require a migration for fields nothing consumes.

**Unprotected on_commit enqueue:** This matches the established convention — three other callsites (`views.py:1868`, `processor.py:249`, `processor.py:472`) use the same unguarded `transaction.on_commit(... .delay(...))` pattern. Failures aren't silent: Django propagates exceptions from `on_commit` callbacks, and the task itself has retry semantics (`autoretry_for`, `max_retries=5`, backoff). A transactional outbox / `QueuedJob` table doesn't exist anywhere in the codebase; introducing one for a single enqueue site is out of scope. If at-least-once celery handoff is a real requirement, it should be a project-wide decision applied uniformly, not bolted onto this PR.

commcare_connect/opportunity/models.py
```diff
@@ -453,6 +453,20 @@ class Meta:
...
+        from commcare_connect.opportunity.tasks import sync_assigned_task_to_usercase
+
+        task = cls.objects.create(
+            task_type=task_type,
+            opportunity_access=opportunity_access,
+            due_date=due_date,
+            status=AssignedTaskStatus.ASSIGNED,
+            assigned_by=assigned_by,
+        )
+        transaction.on_commit(lambda: sync_assigned_task_to_usercase.delay(task.id))
```
mkangia: can we do this as a part of the assignment itself and fail if things go wrong?
I know @calellowitz has been pushing to have both connect and HQ on same page with status of things.

commcare_connect/opportunity/views.py
```diff
@@ -3546,11 +3546,10 @@ def create_task(request, org_slug, opp_id):
     access = form.cleaned_data["access"]
     due_date = form.cleaned_data["due_date"]
 
-    AssignedTask.objects.create(
```
mkangia: 👍

commcare_connect/opportunity/models.py
zandre-eng: The task itself does have a retry mechanism so if HQ fails it will retry with fall off up to 5 times. If you think we need something more rigid I could add creating the task instance itself to the Celery task as well and then have it only be created once the HQ update has succeeded.

Let me know if you would prefer the alternate approach suggested above.

mkangia: The ask is that either both things happen or neither does.

> add creating the task instance itself to the Celery task

The only challenge with this would be that the user won't have an insight into what happened with the task till it eventually shows up on the UI after N seconds/minutes. So the user experience would be tricky.

I would think since its a single quick API call, we do it as a part of the assigned task creation itself unless there is a strong reason to keep the API call separate.

P.S: we are doing quite a few API calls in the Microplanning and Re-Learn workflow and so these discussions are happening on different tickets but the narrative we have is to have either all actions happen or neither.
We might end up doing something different to support scale but unless you are dealing with a bulk update, doing the API call with the record update itself seems like the right thing to do for now.

zandre-eng: Thanks for the helpful response. I'm happy to update this so that everything happens synchronously. My main motivation for putting the HQ call into a Celery task was to have access to the retry mechanism Celery supports. 

You make a good point regarding this being a single API call, so I agree it should be relatively quick. I'll refactor the logic to match the alternative I proposed where we do the HQ call first and, only if it's successful, do we commit the changes to DB. If the HQ call fails, we can simply alert the user and they can retry again in the UI.

zandre-eng: Addressed in [88363be](https://github.com/dimagi/commcare-connect/pull/1150/commits/88363be9f6ffe215a76831ee706698d87597b026).

commcare_connect/opportunity/models.py
```diff
@@ -453,6 +453,21 @@ class Meta:
             )
         ]
 
+    @classmethod
+    def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) -> "AssignedTask":
+        from commcare_connect.commcarehq.api import update_usercase
+
+        case_property = task_type.case_property
+        if case_property:
```
mkangia: I am thinking if the opposite happens, HQ succeeds but saving the record on Connect fails?

What if we do
```
with @transaction.atomic:
    # create the object
    # do the api call
```

I assume that any issue with api call will also roll back the db operation

zandre-eng: That could work. `update_usercase()` doesn't swallow exceptions so any issues with the API call will get raised and the transaction will be rolled back.

Addressed in [8521599](https://github.com/dimagi/commcare-connect/pull/1150/commits/85215995748d19fe44ac2d427beaf6e5f5b940bb).

commcare_connect/opportunity/tests/test_views.py
```diff
@@ -2658,21 +2659,45 @@ def _url(self, opportunity):
...
         assert any("successfully" in str(m) for m in msgs)
 
+    def test_create_task_hq_failure_shows_error_and_no_row(self, client, org_user_member, opportunity, access):
+        client.force_login(org_user_member)
+        task = TaskTypeFactory(app=opportunity.deliver_app, case_property="some_prop")
+        due_date = date.today() + timedelta(days=7)
+
+        with mock.patch(
+            "commcare_connect.commcarehq.api.update_usercase",
+            side_effect=CommCareHQAPIException("boom"),
```
mkangia, line 2688: 🧨

mkangia (approved):

mkangia:
Going to merge this since I just resolved merge conflict mainly and made a minor improvement for code that got added due to merge conflict, that is not related to the changes in this PR  but I am tagging @pxwxnvermx so he can ensure that the improvement I made works as he expects it to.

PR #1158, opened 2026-05-05
Ce/weekly perf report

calellowitz:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-2382

This is the UI for the audit detail report. It includes both the report and the actions tied to the report (completing the audit and assigning tasks). I will add screenshots tomorrow, but wanted to get the code up for review

## Technical Summary

This renders the Weekly performance report with a dynamic django table using the audit calculation definitions to determine the columns and adds views for the two actions from that page. It builds on the previous two PRs, including only the new views and templates, and takes advantage of the new `AssignedTask.assign` method.

This should be reviewable commit by commit

### Safety story

New behavior only and behind a feature flag.

### Automated test coverage

New views are tested

### QA Plan

Will be tested as part of broader audit QA.

### Labels & Review


commcare_connect/audit/views.py
```diff
@@ -68,6 +69,80 @@ def audit_report_list(request, org_slug, opportunity_id):
...
+    RequestConfig(request, paginate={"per_page": DEFAULT_PAGE_SIZE}).configure(no_action_table)
+
+    path = [
+        {"title": "Opportunities", "url": reverse("opportunity:list", args=(org_slug,))},
+        {
+            "title": opportunity.name,
+            "url": reverse("opportunity:detail", args=(org_slug, opportunity.opportunity_id)),
+        },
+        {
+            "title": "Audits",
```
zandre-eng: Nit: Translation for strings in this breadcrumb.

commcare_connect/audit/tables.py
```diff
@@ -75,3 +75,77 @@ def render_view(self, record):
...
+        self.calc_name = calc_name
+        super().__init__(empty_values=(), orderable=False, **kw)
+
+    def render(self, record):
+        r = record.results.get(self.calc_name, {})
+        if not r.get("has_sufficient_data"):
+            return format_html('<span class="text-gray-400">{}</span>', _("N/A"))
+        value = r.get("value")
+        if not r.get("in_range"):
+            return format_html('<span class="badge badge-md negative-dark">{}</span>', value)
```
zandre-eng, line 97: For the default return we only return `value` if it is not `None` but here we simply return `value`. If it is `None` in this conditional then I believe `format_html` will transform it into a "None" string that will be shown on the UI. Confirming whether we want a check here as well?

commcare_connect/templates/audit/audit_report_task_modal.html
```diff
@@ -0,0 +1,97 @@
...
+        {{ report.period_start }} – {{ report.period_end }}
+      </p>
+    </div>
+
+    <form method="post"
+          hx-post="{% url 'opportunity:audit:audit_report_task_action' org_slug=opportunity.organization.slug opportunity_id=opportunity.opportunity_id audit_report_id=report.audit_report_id entry_id=entry.audit_report_entry_id %}"
+          hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
+          hx-swap="none"
+          hx-on:htmx:after-request="if (event.detail.successful) { document.getElementById('modal-root').innerHTML = ''; htmx.trigger(document.body, 'refreshDetail'); } else { error = event.detail.xhr.responseText || '{% trans "Something went wrong. Please try again." %}'; confirm = false; }">
+      {% csrf_token %}
```
zandre-eng: I don't think this CSRF token is needed here since it is already being passed via the HX headers.

commcare_connect/audit/tests/test_views.py
```diff
@@ -66,3 +67,249 @@ def test_list_view_404_when_flag_disabled(client, program_manager_org_user_admin
     )
     response = client.get(url)
     assert response.status_code == 404
+
+
+# ---------------------------------------------------------------------------
+# Detail view tests
+# ---------------------------------------------------------------------------
```
zandre-eng, line 74: For the detail view tests it could be nice to have a test that ensures workers from another opp cannot be viewed in the table.

zandre-eng (approved):
Implementation looks good overall. All non-blocking feedback.

hemant10yadav (approved):

commcare_connect/templates/audit/audit_report_task_modal.html
calellowitz: thanks, removed

commcare_connect/audit/tables.py
calellowitz: added an empty value to render there so it won't be an english word, but I think either is probably fine

calellowitz:
@zandre-eng @hemant10yadav tried to get to all the feedback, let me know what you think

hemant10yadav (approved):

pxwxnvermx (approved):

zandre-eng (approved):
LGTM.

PR #1161, opened 2026-05-06
CCCT-2210: NM/PM review flow for inaccessible work area requests

ajeety4:
## Product Description

Adds a review flow for NMs/PMs to act on inaccessible work area requests submitted by CHWs via the mobile app.

When a CHW submits a form flagging a work area as inaccessible, the system records the request (reason, date of visit, GPS location, photos). NMs/PMs can then open a review modal from the microplanning map, view the supporting evidence, and approve (marks the area `INACCESSIBLE`) or deny (returns it to `NOT_VISITED` and notifies the CHW).

Screemshots:

1. Review Inaccessible Button
image

2. Review Inaccessible Modal
image

## Technical Summary

JIRA: [CCCT-2210](https://dimagi.atlassian.net/browse/CCCT-2210)

- **`WorkAreaInaccessibilityRequest` model + migration** — stores xform ID, date of visit, GPS location, reason, additional details, estimated duration
- **Form receiver** (`processor.py`) — on `request_for_inaccessible` block, creates the request record; location from `xform.metadata.location`, attachments from `xform.raw_form["attachments"]`; duplicate form replay handled idempotently via IntegrityError catch; TODO comments reference CCCT-2397 in case xform structure changes
- **Celery task** — `download_inaccessibility_request_attachments` downloads photo attachments; refactored alongside `download_user_visit_attachments` to share a common `_download_attachments` helper
- **Attachment serving** — `fetch_attachment` extended to serve inaccessibility request photo blobs scoped to the opportunity
- **Views** — `review_inaccessibility_request` (GET: renders htmx modal with photos, Mapbox boundary + GPS marker, reason, date) and `act_on_inaccessibility_request` (POST: approve/deny with pghistory audit trail)
- **Frontend** — "Review Inaccessible" button on the map, htmx-powered modal, Alpine.js state, `modal-lg` CSS class added to Tailwind

### Safety story

- All new write paths are guarded by the `MICROPLANNING` waffle flag and org-admin/PM role checks
- `select_for_update()` on `WorkArea` in the review action prevents double-approval races under `ATOMIC_REQUESTS = True`
- Duplicate form replay (CommCare HQ can resend forms) is handled: the `unique_xform_work_area_inaccessibility` constraint violation is caught and logged, not propagated as a 500
- Attachment blob access is scoped to the opportunity before serving — no cross-opportunity data leak

### Automated test coverage

- `test_receiver_integration.py` — full form submission cycle creating a `WorkAreaInaccessibilityRequest`, parametrised edge cases for missing/invalid `reason` and `date_of_visit`, on-commit attachment task queuing
- `test_tasks.py` — attachment download and blob creation
- `test_views.py` (`TestReviewInaccessibilityModal`) — GET modal renders, 404 for non-pending statuses, photo filtering by xform ID, approve/deny status transitions, audit trail, push notification, flag requirement
- `test_views.py` (opportunity) — `fetch_attachment` serving inaccessibility request blobs

### QA Plan

None

### Labels & Review

[CCCT-2210]: https://dimagi.atlassian.net/browse/CCCT-2210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/microplanning/tests/test_views.py
```diff
@@ -815,6 +820,173 @@ def test_ignores_assignees_from_other_opportunity(
...
+            opportunity=opportunity,
+            work_area_group=group,
+            opportunity_access=access,
+            status=WorkAreaStatus.REQUEST_FOR_INACCESSIBLE,
+        )
+        inacc_request = WorkAreaInaccessibilityRequestFactory(
+            work_area=work_area,
+            opportunity_access=access,
+        )
+        return work_area, inacc_request
```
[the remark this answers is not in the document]
ajeety4: Fixed in f0c47e09. Split `pending_wa` so a separate `field_worker_access` (backed by `mobile_user`) owns the inaccessibility request while `org_user_admin` remains the work-area assignee/reviewer. The deny-notification assertion now checks that the notified user ID matches the field worker's, explicitly asserting it differs from `org_user_admin.id`.

commcare_connect/opportunity/tests/test_views.py
```diff
@@ -1000,6 +1001,43 @@ def test_user_can_fetch_managed_opp(self, storage_handler_getitem_mock, org_user
...
+    ):
+        inacc_request = WorkAreaInaccessibilityRequestFactory(
+            work_area__opportunity=opportunity,
+        )
+        blob_meta = BlobMetaFactory(parent_id=inacc_request.xform_id)
+
+        url = reverse(
+            "opportunity:fetch_attachment",
+            args=(organization.slug, opportunity.opportunity_id, blob_meta.blob_id),
+        )
```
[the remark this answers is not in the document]
ajeety4: Fixed in f0c47e09. Both `fetch_attachment` URL constructions now use `opportunity.id` (integer PK) to match the established pattern in the rest of the test class.

commcare_connect/microplanning/models.py
```diff
@@ -66,17 +65,30 @@ class WorkArea(geo_models.Model):
...
+
+class WorkAreaInaccessibilityRequest(geo_models.Model):
+    work_area = geo_models.ForeignKey(WorkArea, on_delete=geo_models.CASCADE)
+    opportunity_access = geo_models.ForeignKey(OpportunityAccess, on_delete=geo_models.CASCADE)
+    xform_id = geo_models.CharField(max_length=50)
+    date_of_visit = geo_models.DateField()
+    location = geo_models.PointField(null=True, blank=True, srid=SRID)
+    reason = geo_models.CharField(max_length=256)
+    additional_details = geo_models.TextField(blank=True, default="")
+    estimated_duration = geo_models.CharField(max_length=256, blank=True, default="")
```
Charl1996: Why do we need to store `estimated_duration` here?

commcare_connect/form_receiver/processor.py
```diff
@@ -294,6 +312,32 @@ def process_work_area_update(user: User, opportunity: Opportunity, blocks: list[
...
+        raw_date = block.get("date_of_visit", "")
+        date_of_visit = parse_date(raw_date) if raw_date else None
+        if date_of_visit is None:
+            raise ProcessingError("date_of_visit is required and must be a valid ISO date")
+
+        additional_details = block.get("additional_details", "")
+        estimated_duration = block.get("estimated_duration", "")
+        location = _parse_xform_location(xform.metadata.location)
+
+        WorkAreaInaccessibilityRequest.objects.create(
```
Charl1996, line 325: Is it OK if this fails here in the event of a unique constraint violation, or do we just need to catch and skip any such exceptions?

commcare_connect/microplanning/views.py
```diff
@@ -134,6 +141,10 @@ def microplanning_home(request, *args, **kwargs):
         "workarea_min_zoom": WORKAREA_MIN_ZOOM,
         "edit_work_area_url": edit_work_area_url,
         "download_url": download_url,
+        "review_inaccessibility_url": reverse(
+            "microplanning:review_inaccessibility_request",
+            args=[request.org.slug, opportunity.opportunity_id, 0],
+        ).replace("/0/", "/"),
```
Charl1996, line 153: Can you explain why this is needed?

commcare_connect/microplanning/views.py
```diff
@@ -654,3 +665,90 @@ def save_assignment(request, org_slug, opp_id):
...
+        )
+    except WorkAreaInaccessibilityRequest.DoesNotExist:
+        raise Http404
+
+    work_area.status = new_status
+    with pghistory.context(username=request.user.username, user_email=request.user.email):
+        work_area.save(update_fields=["status"])
+
+    if action == InaccessibilityReviewAction.DENY:
+        transaction.on_commit(
```
Charl1996, line 867: Why do you use `transaction.on_commit` here instead of just calling the task directly?

commcare_connect/microplanning/models.py
ajeety4: This is not currently displayed in the UI, but is part of the form.  We (discussed with product) decided to store it in case we decide to show it in the UI in future (avoids backfilling too). 

In case you are asking its meaning, I am guessing it is the duration for how much the area would be inacessible.

commcare_connect/opportunity/views.py
```diff
@@ -1123,10 +1124,12 @@ def reject_visits(request, org_slug=None, opp_id=None):
 def fetch_attachment(request, org_slug, opp_id, blob_id):
     blob_meta = get_object_or_404(BlobMeta, blob_id=blob_id)
 
-    if not UserVisit.objects.filter(
-        opportunity=request.opportunity,
-        xform_id=blob_meta.parent_id,
-    ).exists():
+    if not (
+        UserVisit.objects.filter(opportunity=request.opportunity, xform_id=blob_meta.parent_id).exists()
+        or WorkAreaInaccessibilityRequest.objects.filter(
```
Charl1996, line 1133: I assume we don't create a UserVisit record in the event of a "request for inaccessible area" form submission?

commcare_connect/form_receiver/processor.py
ajeety4: Yes, it is done [here](https://github.com/dimagi/commcare-connect/blob/ay/CCCT-2210-review-inaccessible-work-areas/commcare_connect/form_receiver/views.py#L45)

commcare_connect/microplanning/views.py
ajeety4: The URL pattern requires <int:work_area_id> as a required positional argument:
 `path("<slug:opp_id>/review_inaccessibility/<int:work_area_id>/", ...)`
  
reverse() can't build the URL without supplying something for work_area_id. But the actual ID isn't known server-side as it's chosen client-side in Alpine.js at map_handler.html, where `openReviewInaccessibilityModal(workAreaId)` appends the id and trailing slash:

So the view passes 0 as a placeholder just to satisfy reverse(), then strips /0/ → / to produce a base URL like `/a/<org>/microplanning/<opp>/review_inaccessibility/` that the JS can concatenate <workAreaId>/ onto.
The pattern mirrors `edit_work_area_url` right above it (same approach used elsewhere in this file).

commcare_connect/microplanning/views.py
```diff
@@ -688,3 +699,99 @@ def save_assignment(request, org_slug, opp_id):
...
+    work_area = get_object_or_404(
+        WorkArea,
+        id=work_area_id,
+        opportunity=request.opportunity,
+        status=WorkAreaStatus.REQUEST_FOR_INACCESSIBLE,
+    )
+    try:
+        inacc_request = WorkAreaInaccessibilityRequest.objects.filter(work_area=work_area).latest("pk")
+    except WorkAreaInaccessibilityRequest.DoesNotExist:
+        raise Http404
```
zandre-eng: Would there ever be more than one inaccessibility request for a work area? If not, instead of doing a filter within a try block here, you could replicate how we fetch the work area:

```suggestion
    inacc_request = get_object_or_404(WorkAreaInaccessibilityRequest, work_area=work_area)
```

commcare_connect/microplanning/views.py
```diff
@@ -688,3 +699,99 @@ def save_assignment(request, org_slug, opp_id):
...
+@org_admin_required
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def act_on_inaccessibility_request(request, org_slug, opp_id, work_area_id):
+    try:
+        action = InaccessibilityReviewAction(request.POST.get("action", ""))
+    except ValueError:
+        return HttpResponseBadRequest("Invalid action")
+
+    new_status = _ACTION_TO_NEW_STATUS[action]
```
zandre-eng, line 842: Nit: This can be wrapped in a try/except block so that we can return a more graceful response if the action is not in the allowed list. This doesn't look to be possible normally given that we control what is sent from the front-end, but could be good to add since this view can technically be called outside of the UI.

commcare_connect/microplanning/views.py
ajeety4: Since we have atomic requests, this is to ensure that push notification is only send after the db makes the commit. Without it, the task can get queued and sent even in the case of rollback.

commcare_connect/opportunity/views.py
ajeety4: Correct

commcare_connect/microplanning/tests/test_views.py
```diff
@@ -813,6 +816,17 @@ def test_ignores_assignees_from_other_opportunity(
         delay_patch.assert_not_called()
 
 
+@pytest.mark.django_db
+def test_work_area_inaccessibility_request_factory_creates_instance():
```
Charl1996: This test feels redundant.

commcare_connect/opportunity/tasks.py
```diff
@@ -370,28 +370,17 @@ def send_push_notification_task(
...
-    retry_jitter=True,
-)
-def download_user_visit_attachments(self, user_visit_id: int):
-    user_visit = UserVisit.objects.get(id=user_visit_id)
-    api_key = user_visit.opportunity.api_key
-    blobs = user_visit.form_json.get("attachments", {})
-    domain = user_visit.opportunity.deliver_app.cc_domain
-    form_id = user_visit.xform_id
-    for name, blob in blobs.items():
+def _download_attachments(api_key, domain: str, xform_id: str, attachments: dict):
```
Charl1996, line 373: 👍

commcare_connect/microplanning/views.py
Charl1996: I see.

Charl1996:
Generally looks good. Will wait till you solve the conflicts, then I shall do a final pass review.

commcare_connect/microplanning/views.py
```diff
@@ -688,3 +699,99 @@ def save_assignment(request, org_slug, opp_id):
...
+        logger.info(f"Failed to sync work area {work_area.id} to HQ after review action. Error: {e}")
+        return HttpResponse(status=500, content=_("Failed to sync work area status. Please try again."))
+
+    if action == InaccessibilityReviewAction.DENY:
+        transaction.on_commit(
+            partial(
+                send_push_notification_task.delay,
+                [inacc_request.opportunity_access.user_id],
+                _("Inaccessibility Request Denied"),
+                _("Your request to mark a work area inaccessible has been declined."),
```
zandre-eng: Nit: I don't know if translating these will be helpful to the user. These will be translated based on the web request, and so won't necessarily be the same language that the worker uses on their mobile device.

commcare_connect/templates/microplanning/review_inaccessibility_modal.html
```diff
@@ -0,0 +1,88 @@
...
+    <div class="mb-6">
+        <p class="text-sm font-medium text-brand-deep-purple mb-2">{% translate "Location" %}</p>
+        <div id="review-inaccessibility-map" class="w-full rounded-lg h-56"></div>
+    </div>
+
+    <div class="mb-6">
+        <p class="text-sm font-medium text-brand-deep-purple mb-2">{% translate "Photos" %}</p>
+        {% if photos %}
+        <div class="grid grid-cols-2 gap-3">
+            {% for photo in photos %}
```
zandre-eng: Nit: If there are a large number of photos (e.g. 5-6) does the modal correctly display these without issues (e.g. creating a horizontal scrollbar on page)?

zandre-eng (commented):
Took a first pass. The implementation looks good, with a few minor comments.

commcare_connect/microplanning/views.py
ajeety4: I think the check just above this should account for invalid action and _ACTION_TO_NEW_STATUS should cover all the actions in `InaccessibilityReviewAction`.
```

    try:
        action = InaccessibilityReviewAction(request.POST.get("action", ""))
    except ValueError:
        return HttpResponseBadRequest("Invalid action")
```

commcare_connect/templates/microplanning/review_inaccessibility_modal.html
ajeety4: Yes it displays in grid of two's and uses a vertical scrollbar.

commcare_connect/microplanning/views.py
ajeety4: Good point. This was a defensive approach, and looking at the processor logic it should block any multiple inacessibility requests. 
https://github.com/dimagi/commcare-connect/pull/1161/commits/3248040f308f45a197cd846fda28b9e2246cc951

commcare_connect/microplanning/views.py
ajeety4: Great catch. 
Addressed in https://github.com/dimagi/commcare-connect/pull/1161/commits/3248040f308f45a197cd846fda28b9e2246cc951

commcare_connect/microplanning/tests/test_views.py
ajeety4: Oh man, I remember removing this and claude keeps adding this test case :)
Thanks for catching
Addressed in https://github.com/dimagi/commcare-connect/pull/1161/commits/1cf1f13995738c7910df6f485dbc86bb8f168dbb

commcare_connect/microplanning/models.py
```diff
@@ -108,3 +108,24 @@ def update_status(self, user):
             self.status = new_status
             with pghistory.context(username=user.username, user_email=user.email):
                 self.save(update_fields=["status"])
+
+
+class WorkAreaInaccessibilityRequest(geo_models.Model):
```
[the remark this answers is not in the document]
ajeety4: This is consistent with the rest of the models defined in micro planning which use geo_models when there is an associated geo field

zandre-eng (approved):
Thanks for addressing my feedback, LGTM!

zandre-eng (approved):

ajeety4:
Removed the `estimated_duration` field as it is no longer needed by Delivery

zandre-eng (approved):

PR #1164, opened 2026-05-07
Work Area Clustering Feedback

pxwxnvermx:
## Product Description

- Caps each Work Area Group at 60 work areas in addition to the existing 300-building cap. Whichever cap trips first stops cluster growth.
- Tighten buffer_distance default to 10m was 100m. The old default let BFS bridge clusters across unrelated regions whenever any pair of WAs from different regions came within 100 m. 10 m is large enough to absorb minor polygonization slivers, small enough not to bridge realistic inter-region gaps.
- Filter over-the-top adjacency edges. A WA-A ↔ WA-B distance edge used to form even when WA-X physically sat between them. If X was assigned to a different group, BFS would then bridge A↔B via that over-the-top edge. Now: for each distance-based candidate pair, the connecting line segment between their nearest points is STRtree-queried for any intervening WA. If one is found, the edge is dropped. Also fixes a latent bug where empty-intersection pairs were misclassified as shared-edge (Shapely returns dim=2 on an empty Polygon).
- Disallow diagonal/corner-only adjacency. Two WAs touching only at a single corner point used to count as adjacent because dist == 0 passed the buffer check. That isn't really contiguous, so the distance branch now requires 0 < dist <=  buffer_distance. Shared-edge pairs are unaffected.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2392)

### Safety story
Tested Locally

### Automated test coverage
- Added automated tests

### Labels & Review


sravfeyn (approved):
LGTM, but would wait for another review.

hemant10yadav (approved):

PR #1165, opened 2026-05-07
Front-end Linting

zandre-eng:
## Product Description

This PR adds automated linting for JavaScript and Django HTML templates to the pre-commit pipeline. There are no user-facing changes.

## Technical Summary

[CCCT-2388](https://dimagi.atlassian.net/browse/CCCT-2388)

Adds two linters as pre-commit hooks alongside the existing Python linting:

- **ESLint (`eslint:recommended`)** — lints `commcare_connect/static/js/**/*.js`. JS-in-HTML templates is intentionally excluded: Django template tags (`{% %}`, `{{ }}`) interpolated into `<script>` blocks cause parse errors. Existing JS files are grandfathered via an `overrides` block (rules downgraded to `warn`); new files get `error`-level rules, blocking commits.
- **djlint (Django profile)** — lints and reformats Django HTML templates in `commcare_connect/templates/`. Runs as two hooks: `djlint-reformat-django` (auto-fixes indentation) then `djlint-django` (blocks on lint violations). A number of rules are suppressed — see `[tool.djlint]` in `pyproject.toml` for the full list. Notable suppressions: Two deferred cleanup items (H026: empty class attributes, H029: uppercase method="POST") that will be addressed in a [bulk reformat PR](https://github.com/dimagi/commcare-connect/pull/1166).

As mentioned a [separate PR](https://github.com/dimagi/commcare-connect/pull/1166) here has been opened to do a bulk reformat of all our HTML code to bring it in line with the new linters.

For `.eslintrc.js` here is a summary breakdown of what each of the settings do:
- `env`: Declares global variables that are pre-defined in those environments.
  - `browser: true`: Makes browser globals like `window`, `document`, and `fetch` available. Would otherwise flag these as `no-undef` errors.
  - `es2020: true`: Makes ES2020 globals like `Promise` and `BigInt` available. Would otherwise flag these as `no-undef` errors.
- `parserOptions`: Controls how ESLint parses the JS.
  - `ecmaVersion: 2020`: Allows modern syntax like optional chaining and nullish coalescing.
  - `sourceType: 'module'`: Enables ES module syntax (`import`/`export`).
- `extends: ['eslint:recommended']`: Enables a curated set of rules that catch common bugs (e.g. no duplicate keys, no unreachable code).
- `rules`: Project-wide overrides on top of `recommended`.
  - `no-console: 'error'`: Disallows `console.log` etc.
  - `no-unused-vars`: Disallows declared but unused variables
- `overrides`: Applies different rules to specific files. The grandfathered existing JS files all get rules set to `warn` instead of `error`, so that existing files won't break CI but will still surface issues as warnings.
- `ignorePatterns`: Files ESLint skips entirely.    

### Safety story

Config-only change. No application code, models, views, or migrations are modified. The hooks only run at commit time and have no effect on the running application. The ESLint grandfathering strategy ensures existing code produces warnings only, so no commits are blocked on legacy violations.

This change was also tested locally by creating an HTML and JS file with linting errors and ensuring that commits were blocked for those files.

### Automated test coverage

No automated tests — this change adds developer tooling with no runtime behaviour to test.

### QA Plan

- [ ] Run `pre-commit run --all-files` and confirm all hooks pass
- [ ] Create a new JS file in `commcare_connect/static/js/` with a `console.log` and an unused variable, stage it, and confirm ESLint blocks the commit with errors
- [ ] Confirm that staging an existing grandfathered JS file (e.g. `dashboard.js`) produces warnings but does not block the commit
- [ ] Create a Django template with a missing `alt` attribute on an ``, stage it, and confirm djlint blocks the commit
- [ ] Confirm that djlint-reformat auto-fixes indentation in staged templates before the lint check runs

### Labels & Review

[CCCT-2388]: https://dimagi.atlassian.net/browse/CCCT-2388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,30 +12,10 @@ module.exports = {
     'no-console': 'error',
     'no-unused-vars': 'error',
   },
-  overrides: [
-    {
-      // Grandfather existing files — all rules are warnings only
-      files: [
-        'commcare_connect/static/js/alpine.js',
-        'commcare_connect/static/js/dashboard.js',
-        'commcare_connect/static/js/flatpickr.js',
-        'commcare_connect/static/js/gtx.js',
-        'commcare_connect/static/js/htmx.js',
-        'commcare_connect/static/js/livesession.js',
-        'commcare_connect/static/js/mapbox.js',
-        'commcare_connect/static/js/tomselect.js',
-        'commcare_connect/static/js/vendors.js',
-      ],
-      rules: {
-        'no-console': 'warn',
-        'no-unused-vars': 'warn',
-        'no-undef': 'warn',
-      },
-    },
-  ],
   ignorePatterns: [
-    '.eslintrc.js',
-    'commcare_connect/static/bundles/',
     'node_modules/',
+    'commcare_connect/static/bundles/',
+    // Vendor tracking snippet (LiveSession) — minified third-party code, not ours to lint
+    'commcare_connect/static/js/livesession.js',
   ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,5 +33,9 @@ module.exports = {
       },
     },
   ],
-  ignorePatterns: ['commcare_connect/static/bundles/', 'node_modules/'],
+  ignorePatterns: [
+    '.eslintrc.js',
+    'commcare_connect/static/bundles/',
+    'node_modules/',
+  ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,37 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2020: true,
+  },
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+  },
+  extends: ['eslint:recommended'],
+  rules: {
+    'no-console': 'error',
+    'no-unused-vars': 'error',
+  },
+  overrides: [
+    {
+      // Grandfather existing files — all rules are warnings only
+      files: [
+        'commcare_connect/static/js/alpine.js',
+        'commcare_connect/static/js/dashboard.js',
+        'commcare_connect/static/js/flatpickr.js',
+        'commcare_connect/static/js/gtx.js',
+        'commcare_connect/static/js/htmx.js',
+        'commcare_connect/static/js/livesession.js',
+        'commcare_connect/static/js/mapbox.js',
+        'commcare_connect/static/js/tomselect.js',
+        'commcare_connect/static/js/vendors.js',
+      ],
+      rules: {
+        'no-console': 'warn',
+        'no-unused-vars': 'warn',
+        'no-undef': 'warn',
+      },
+    },
+  ],
+  ignorePatterns: ['commcare_connect/static/bundles/', 'node_modules/'],
+};
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "concurrently": "^8.0.1",
     "css-loader": "^6.5.1",
     "dotenv-webpack": "^8.0.1",
+    "eslint": "^8.57.1",
     "flatpickr": "^4.6.13",
     "gulp-concat": "^2.6.1",
     "htmx.org": "^1.9.4",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -116,6 +116,7 @@ dev = [
     "black>=23.7.0",
     "coverage>=7.2.7",
     "django-debug-toolbar>=4.1.0",
+    "djlint>=1.36.4",
     "factory-boy>=3.3.0",
     "flake8>=6.1.0",
     "flake8-isort>=6.0.0",
@@ -140,3 +141,9 @@ production = [
     "gunicorn[gevent]>=21.2.0",
     "psycopg2>=2.9.6",
 ]
+
+# ==== djlint ====
+[tool.djlint]
+profile = "django"
+indent = 2
+ignore = "H006,H008,H014,H016,H020,H021,H023,H025,H030,H031,H037,T001,T002,T003"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,4 +146,4 @@ production = [
 [tool.djlint]
 profile = "django"
 indent = 2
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -15,6 +15,7 @@
     "concurrently": "^8.0.1",
     "css-loader": "^6.5.1",
     "dotenv-webpack": "^8.0.1",
+    "eslint": "^8.57.1",
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [41a568d](https://github.com/dimagi/commcare-connect/pull/1165/commits/41a568d7e584bcf882913ad898f0f082d9c027d3).

.eslintrc.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [41a568d](https://github.com/dimagi/commcare-connect/pull/1165/commits/41a568d7e584bcf882913ad898f0f082d9c027d3).

.eslintrc.js
mkangia, line 1: Can you help understand the different settings added in this file, what they mean?

mkangia (commented):
I don't have enough knowledge to review this but if you can share what each setting/file is specifically for, that would be helpful.

.eslintrc.js
zandre-eng: I have updated the PR description to give a breakdown summary of what each of the settings in `eslintrc.js` mean.

ajeety4 (dismissed):

pxwxnvermx (dismissed):

hemant10yadav (dismissed):

pyproject.toml
```diff
@@ -141,3 +142,9 @@ production = [
     "gunicorn[gevent]>=21.2.0",
     "psycopg2>=2.9.6",
 ]
+
+# ==== djlint ====
+[tool.djlint]
+profile = "django"
+indent = 2
+ignore = "H006,H008,H014,H016,H020,H021,H023,H025,H030,H031,H037,T001,T002,T003"
```
Charl1996: Claude defines the tags as follows:
```
  │ Code │                                         Meaning                                         │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H006 │ <img> tag should have height and width attributes (prevents layout shift).              │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H008 │ Attributes should be double-quoted (class="x", not class='x').                          │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H014 │ Found extra/consecutive blank lines that should be removed.                             │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H016 │ Missing <title> tag in the HTML.                                                        │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H020 │ Empty tag pair found (e.g. <span></span>) — consider removing it.                       │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H021 │ Inline styles should be avoided (use CSS classes instead).                              │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H023 │ Don't use HTML entity references (e.g. prefer the Unicode character over &nbsp;/&amp;). │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H025 │ Tag appears to be orphaned / unbalanced (unclosed tag).                                 │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H030 │ Consider adding a <meta name="description"> tag.                                        │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H031 │ Consider adding <meta name="keywords"> tags.                                            │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ H037 │ Duplicate attribute found on a tag.                                                     │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ T001 │ Variables should be wrapped in a single whitespace — {{ var }}, not {{var}}.            │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ T002 │ Double quotes should be used inside template tags — {% extends "base.html" %}.          │
  ├──────┼─────────────────────────────────────────────────────────────────────────────────────────┤
  │ T003 │ {% endblock %} should include its name — {% endblock content %}.    
```

Why are we ignoring these ones?
  - H025 - This seems like a good check to keep; an unclosed tag can break rendering.
  - H037 - Defining a duplicate `class` attribute could lead to unintended consequences, since the second silently wins.

zandre-eng: Initially, it seems Claude put these on the ignore list as they would lead to a few supressions being required in the codebase. But after looking through the list again, and what those supressions would be, I agree these could be enabled.

I'll be re-enabling the following:
- H008
- H014
- H025
- H037
- T001

Addressed in [3f043ec](https://github.com/dimagi/commcare-connect/pull/1165/commits/3f043ec8c97deff6094e5b9b7b969ff4823513e6).

mkangia:
> overrides: Applies different rules to specific files. The grandfathered existing JS files all get rules set to warn instead of error, so that existing files won't break CI but will still surface issues as warnings.

Can you say more about this just so I understand this well? Is this going to lead to things getting ignored endlessly because they are just warnings?

zandre-eng:
I did a force push to cleanly resolve conflicts and changes with upstream branch.

zandre-eng:
> Can you say more about this just so I understand this well? Is this going to lead to things getting ignored endlessly because they are just warnings?

@mkangia This is a fair point. This was initially introduced by Claude to make adding the new linters as frictionless as possible, but after looking into this some more there's no real threat in linting the existing JS files. I've updated this branch (and the upstream reformat branch) to no longer ignore these files.

Charl1996 (approved):

pyproject.toml
mkangia: Thanks for the review here @Charl1996

PR #1167, opened 2026-05-07
feat(prelogin): public marketing site at /

jjackson:
## Product Description

Visitors to `/` now land on a public marketing page (currently a placeholder mockup from Gillian, sourced manually from `dimagi-internal/connect-prelogin`) instead of the existing `pages/home.html` template. A header Login CTA takes signed-in users to `/accounts/login/` (which then bounces them onward via `LOGIN_REDIRECT_URL = users:redirect`); anonymous visitors see the marketing page. Sign Up CTA links out to `connect.dimagi.com/accounts/signup/`.

This is the public-facing companion to a parallel landing site running on `labs.connect.dimagi.com/` ([connect-labs#169](https://github.com/jjackson/connect-labs/pull/169) — already merged). The new `prelogin` app is intentionally self-contained (no labs-specific or connect-specific imports beyond the one configurable setting) so the two deployments can stay in sync.

## Technical Summary

- New Django app `commcare_connect/prelogin/` with one `TemplateView` for `/`.
- The mockup HTML lives in `commcare_connect/templates/prelogin/home.html` as a self-contained document (own DOCTYPE + inline `<style>`, no `base.html` extension — the mockup ships its own typography and palette).
- Header Login CTA's URL is read from `PRELOGIN_APP_LOGIN_URL` setting; default `/accounts/login/`.
- Replaces the existing `TemplateView(template_name="pages/home.html")` at `/`.
- Two prep edits applied at re-import time, documented in a Django comment at the top of `home.html`: strip the `.mockup-bar` element + its CSS rule (the top "MOCKUP" banner used in Gillian's preview), and template the login link's `href`.
- `name="home"` is now namespaced as `prelogin:home`; the one inbound `reverse("home")` call in `commcare_connect/users/views.py` is updated.

Layout follows the project's house-style conventions (per `CLAUDE.md`):
- Tests in `commcare_connect/prelogin/tests/test_views.py` (directory layout, `pytest.mark`-style fixture-based tests).
- Template in top-level `commcare_connect/templates/prelogin/` (not app-local).
- No app-local `static/`; future static assets go in top-level `commcare_connect/static/prelogin/`.

### Safety story

Static page, no models, no migrations, no data access, no auth changes. The only behavioral change is that `/` no longer renders `pages/home.html` — anonymous and authenticated visitors both see the marketing page; logged-in users navigate into the app via the header Login CTA (which falls through to allauth, which then forwards them via `LOGIN_REDIRECT_URL`). Blast radius: anyone with `/` bookmarked who expected the old `pages/home.html` content gets the new marketing page instead.

### Automated test coverage

Three tests in `commcare_connect/prelogin/tests/test_views.py`:
- `test_renders_with_brand` — `GET /` returns 200 with expected branding text.
- `test_login_url_defaults_to_accounts_login` — verifies the default for the login CTA.
- `test_login_url_respects_setting_override` — locks the contract that the `PRELOGIN_APP_LOGIN_URL` override flows through.

### QA Plan

- Visit `/` on the deployed environment and confirm the marketing page renders without the "MOCKUP · Connect by Dimagi · v11 reshape …" top banner.
- Confirm header Login CTA navigates to `/accounts/login/` (and from there forwards correctly for already-signed-in users).
- Confirm Sign Up CTA still points to `connect.dimagi.com/accounts/signup/`.


jjackson:
Pushed `c59f5f74` — restructures the app to be static-first per Cal's recommendation, and pulls in Gillian's latest content from upstream.

**Architecture shift:** the marketing site is now plain static HTML under `commcare_connect/static/prelogin/index.html` (no Django template syntax). The view shrinks to a ~10-line file-read + one-placeholder substitution:

```python
class HomeView(View):
    def get(self, request):
        login_url = getattr(settings, "PRELOGIN_APP_LOGIN_URL", "/accounts/login/")
        html = _INDEX_HTML.read_text(encoding="utf-8").replace(_LOGIN_PLACEHOLDER, login_url)
        return HttpResponse(html)
```

Designers can drop new upstream content in without touching Django code. The static dir is byte-comparable to `dimagi-internal/connect-prelogin` modulo four documented import-time edits (each tagged with an `import-time edit` HTML comment in `index.html`):
1. `<base href="/static/prelogin/">` so upstream's relative `images/` + `gifs/` paths resolve under Django staticfiles unchanged.
2. `.mockup-bar` CSS rule stripped (preview-only banner).
3. `.mockup-bar` element stripped.
4. Login link `href` → `__APP_LOGIN_URL__` placeholder.

**New content from upstream:** 10 program-area SVGs, 8 field photos (CHC/KMC/Readers + join-hero), 3 product GIFs (learn/deliver/pay), and the homepage hero photo, copy refinements, and `/sign-up` → `/join` route rename.

**Asset weight:** the 8 upstream JPGs were raw camera shots totaling **41.9 MB**. Resized to 1600px wide at JPEG q=82 progressive — drops the set to **2.95 MB** with no visible loss at typical viewport sizes. Total `static/prelogin/` is now ~8 MB.

**Tests:** updated to assert against rendered HTML directly (the view returns `HttpResponse`, not a `TemplateResponse` with context). Added `test_no_unsubstituted_placeholder` to catch regressions where the placeholder leaks to clients.

Verified locally: `/` returns 200, all 22 assets serve via Django staticfiles, all 4 tests pass, pre-commit clean.

jjackson:
Re-aligned to the project's existing layout convention (`84ebf449`):

- HTML moved back to `commcare_connect/templates/prelogin/home.html` (every other `.html` in this repo lives in `templates/`; there are zero `.html` files under `static/`).
- Binary assets stay in `commcare_connect/static/prelogin/{images,gifs}/`.
- HTML is a real Django template again, but with a small directive count: one `{% load static %}`, 18 `{% static 'prelogin/...' %}` on asset references, one `{{ app_login_url }}`. No conditional/loop/inheritance logic.
- View simplifies back to a stock `TemplateView` with one context var; tests revert to `resp.context["app_login_url"]` assertions.
- All four import-time edits to upstream HTML still tagged with `import-time edit` comments and documented at the top of `home.html`.

Net effect: the static surface (CSS/JPG/SVG/GIF) is in the conventional place, and the HTML is in the conventional place. The Django piece is still small (~10 LOC view, ~20 directives in 5111-line HTML). Resolves the "first .html under static/" oddity from the previous commit.

Verified locally: `/` returns 200, all 18 assets serve at `/static/prelogin/...`, all 3 tests pass, pre-commit clean.

jjackson:
**AI feedback audit — all addressed.**

Only one piece of actionable AI feedback was posted on this PR (CodeRabbit, on `c59f5f74`):

> Avoid reading `index.html` from disk on every request — cache at import time.

That finding was already resolved by the very next commit, `84ebf449` (CodeRabbit acknowledged this in-thread with `✅ Addressed in commit 84ebf44`). The resolution went further than CodeRabbit's suggested fix: instead of caching a module-level string, we moved the HTML back into `commcare_connect/templates/prelogin/home.html` as a real Django template. Django's template engine caches compiled templates internally, so there's no per-request disk I/O on the hot path, and the view simplified to a stock `TemplateView`. No subsequent AI feedback has been posted on the more recent commits (`7095fdea` rebase, `ed0d6c6d` content sync).

commcare_connect/templates/prelogin/home.html
```diff
@@ -0,0 +1,8341 @@
...
+        <a href="#/insights" class="btn ghost">See all Campaign Insights →</a>
+      </div>
+    </div>
+  </div>
+
+  <div class="closing-cta">
+    <div class="closing-inner">
+      <span class="eyebrow on-dark">Support Kangaroo Care</span>
+      <h2>$36 funds one newborn. <em>Verified delivery.</em></h2>
+      <p class="lead">$36 covers the full Connect KMC cycle for one small and vulnerable newborn
```
calellowitz: This JS can move to the static directory and be loaded in with a `{% static %}` tag

commcare_connect/templates/prelogin/home.html
```diff
@@ -0,0 +1,8341 @@
...
+{# its player surfaces error 153). #}
+<html lang="en">
+<head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<title>Connect by Dimagi, Mockup</title>
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link href="https://fonts.googleapis.com/css2?family=Work+Sans:wght@200;300;400;500;600;700&family=JetBrains+Mono:wght@400;500&display=swap" rel="stylesheet">
+<style>
```
calellowitz: all this CSS can be moved to a static file and loaded in via `{% static %}`

.pre-commit-config.yaml
```diff
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: prettier
         args: ['--tab-width', '2', '--single-quote']
-        exclude: 'commcare_connect/templates/'
+        exclude: '(commcare_connect/templates/|commcare_connect/static/prelogin/)'
```
calellowitz: I think this is fine in a pinch, but ideally we wouldn't exclude the directory and instead just let the formater do its thing on this code too. It won't affect the function at all, but keeps the code consistent on the off chance someone needs to go read it.

calellowitz (commented):
A few small things. I reviewed the python pieces to a typical standard, the html/css/js I just scanned for things that could cause issues on the backend and made a few small suggestions to get more of it into static.

The main html file is not very maintainable, but I am assuming there are enough changes that you won't really care. Ideally the inline styling and repeated components would be pulled out into static classes and components. It is also inconsistent about SVGs (some are inline some are in static). None of those are blocking.

One suggestion, assuming we want to translate this page in the future, is that all text should be marked with a `{% translate %}` tag. You can do that later, but I imagine it will be necessary at some point.

commcare_connect/templates/prelogin/home.html
jjackson: Done in `71b01114` — the inline `<style>` (~4940 lines) is now in `commcare_connect/static/prelogin/styles.css` and the template loads it via `<link rel="stylesheet" href="{% static 'prelogin/styles.css' %}">`. All the `{% static 'prelogin/images/…' %}` refs inside the CSS were rewritten to relative `images/…` paths so the file is a plain CSS asset.

commcare_connect/templates/prelogin/home.html
jjackson: Done in `71b01114` — the inline `<script>` (~150 lines) is now in `commcare_connect/static/prelogin/app.js` and the template loads it via `<script src="{% static 'prelogin/app.js' %}"></script>`. No `{% static %}` refs lived in the JS, so it was a clean lift.

.pre-commit-config.yaml
jjackson: Done in `71b01114` — the exclude is now just `commcare_connect/templates/` (Django templates still need to be excluded so prettier doesn't mangle `{% %}` tags). The newly-extracted `commcare_connect/static/prelogin/` is no longer excluded, so prettier formats the CSS/JS on every commit — that's why `styles.css` lands here already reflowed (rule blocks per line) vs the tight inline form it had as `<style>` content.

jjackson:
@calellowitz — all three review comments are addressed in `71b01114`, and there are three additional fixes piggybacking on the same push:

### Your review

| # | Comment | Resolution |
|---|---------|-----------|
| 1 | "all this CSS can be moved to a static file and loaded in via `{% static %}`" | Inline `<style>` (~4940 lines) → `commcare_connect/static/prelogin/styles.css`. `{% static 'prelogin/images/…' %}` references rewritten as relative `images/…` paths so the CSS file stands alone. |
| 2 | "This JS can move to the static directory" | Inline `<script>` (~150 lines) → `commcare_connect/static/prelogin/app.js`. No template tags in the JS, clean lift. |
| 3 | "ideally we wouldn't exclude the directory" (`.pre-commit-config.yaml:25`) | Exclude reduced to `commcare_connect/templates/` only — the new `static/prelogin/` is included so prettier formats the CSS/JS on commit. That's why `styles.css` lands here already reflowed. |

`home.html` shrank from 8341 → 3266 lines.

### Three mobile fixes that came along

These were live regressions on `labs.connect.dimagi.com` that I caught while iterating on the static-file extraction; same code is portable to `connect.dimagi.com`:

4. **Two-Sided Marketplace stacks on mobile.** `.market-pair` had `grid-template-columns: 1fr 1fr` with no mobile override. The existing 980px rules tried to fix it but targeted `.two-sided-market > .market-side-funder` — a relationship that never exists in the DOM (the sides live inside `.market-pair`, not `.two-sided-market`). At 375px the Funder card was 175px wide and the Frontline card overflowed to 518px (past the viewport). Now both stack 1-up.
5. **Field-story photo block is full-bleed on mobile.** The "Closing the Post-Discharge Gap" KMC hero had 20px white gutters on each side at ≤600px because the surrounding `.wrap` pads horizontally. Targeted `a.photo-block` only — program tiles (`div.photo-block`) are unaffected, desktop view keeps its rounded card.
6. **Mobile nav is reachable.** The previous mobile CSS hid every link except Login with no replacement, so How It Works / Programs / Insights / Connect Network / Sign Up were unreachable from a phone. Added a ☰ button (visible only at ≤980px) that toggles a slide-down panel with all seven targets; closes on link tap, outside tap, Escape, or hashchange.

### Verification

All six changes were built and verified at `labs.connect.dimagi.com` first ([connect-labs#192](https://github.com/jjackson/connect-labs/pull/192) + [connect-labs#193](https://github.com/jjackson/connect-labs/pull/193)) — Playwright at 375px and 1280px viewports, no console errors, SPA routing intact, static assets serve 200. The diff that landed here is the same diff (verbatim file copies plus the surgical exclude edit).

calellowitz (approved):
Thanks for the followup. Looks good to go.

calellowitz:
I am not sure why the codeql bits didn't fire, but merging this since the build and linter passed.

PR #1174, opened 2026-05-11
HQ Reset Case Property on Assigned Task Delete

zandre-eng:
## Product Description

When assigned tasks are deleted via the task management UI, CommCare HQ is now kept in sync: the case property set on the worker's user case at assignment time is cleared on deletion. If another assigned task for the same worker still references that same case property, the reset is skipped to avoid disrupting the active task.

This change has no visible UI difference for the happy path. On HQ API failure, the delete is now rolled back and an error banner is shown instead of silently succeeding.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2401)

Previously, deleting assigned tasks left stale case properties on the worker's HQ user case. This PR closes that gap with `AssignedTask.delete_and_reset_hq()`, which wraps the DB delete and HQ reset in a single `transaction.atomic()` block — a HQ failure rolls back the delete, ensuring tasks are never silently lost. HQ updates are sent as a single bulk POST via `bulk_update_usercases`, which delegates to the existing `bulk_create_or_update_cases` infrastructure (chunked at 100). A post-delete query + Python set subtraction determines which `(access, case_property)` pairs actually need resetting, avoiding unnecessary HQ calls when other tasks still hold the property.

As part of this work, `bulk_update_cases` was removed as a duplicate of `bulk_create_or_update_cases` — callers now inject `create: False` directly, giving all bulk case writes a single code path.

- **New classmethod — `AssignedTask.delete_and_reset_hq(task_ids, opportunity)`** (`opportunity/models.py`): Deduplicates `(access_id, case_property)` pairs, deletes within a transaction, runs one query to find still-active pairs, and resets only those that are now orphaned via `bulk_update_usercases`.
- **New API helper — `bulk_update_usercases(updates)`** (`commcarehq/api.py`): Accepts `list[tuple[OpportunityAccess, dict]]`, resolves `hq_case_id` per access (fetching from HQ and caching if missing), injects `create: False`, and calls `bulk_create_or_update_cases` for a single chunked POST.
- **`delete_tasks` view updated** (`opportunity/views.py`): Delegates to `delete_and_reset_hq()` and catches `CommCareHQAPIException`, showing a user-facing error and preserving tasks on HQ failure.
- **`bulk_update_cases` removed** (`commcarehq/api.py`): Was a thin wrapper around `bulk_create_or_update_cases` with `create: False` injected. Callers (`bulk_update_usercases`, `microplanning/helpers.py`) now call `bulk_create_or_update_cases` directly.

### Safety story

The DB delete and HQ reset are wrapped in `transaction.atomic()` — a HQ failure rolls back the delete so tasks are never silently lost. The "still assigned" check runs inside the transaction after the delete, on the committed state, so it's correct under concurrent deletions. The function is scoped to the current opportunity and excludes `COMPLETED` tasks, matching existing delete behaviour. Manually integration-tested against a local HQ instance: assignment correctly set the case property to `"1"` via a bulk POST, and deletion cleared it to `""` in a second bulk POST.

### Automated test coverage

- **`TestAssignedTaskDeleteAndResetHQ`** (`test_models.py`):
  - `test_deletes_rows_and_resets_hq` — happy path: task deleted, HQ reset called with correct payload
  - `test_skips_hq_when_no_case_property` — no HQ call when task type has no `case_property`
  - `test_does_not_delete_when_hq_fails` — DB delete rolled back on `CommCareHQAPIException`
  - `test_skips_completed_tasks` — completed tasks excluded from deletion and reset
  - `test_skips_hq_when_other_assigned_task_remains` — reset skipped when another task still uses the property
  - `test_deduplicates_hq_calls` — multiple tasks sharing a property produce one `bulk_update_usercases` call
- **`TestAssignedTaskAssign`** (`test_models.py`): updated to assert `bulk_update_usercases` is called on assignment
- **`TestDeleteTasks`** (`test_views.py`):
  - `test_delete_tasks_resets_hq_case_property` — view triggers HQ reset on delete
  - `test_delete_tasks_hq_failure_shows_error_and_keeps_tasks` — error message shown and tasks preserved on HQ failure
- **`TestExcludeWorkAreas`** (`test_helpers.py`): updated mock patches for the `bulk_update_cases` → `bulk_create_or_update_cases` rename

### QA Plan

- [ ] Assign a task to a worker where the task type has a `case_property` configured — verify the property is set to `"1"` on the worker's HQ user case
- [ ] Delete that task — verify the property is cleared (`""`) on the HQ user case and a success message appears
- [ ] Assign two tasks with the same `case_property` to the same worker, then delete only one — verify the HQ case property is **not** reset (the remaining task still references it)
- [ ] Delete the remaining task — verify the HQ case property is now cleared
- [ ] Simulate an HQ API failure (e.g. wrong API key) and attempt deletion — verify no tasks are deleted and an error message is shown
- [ ] Delete a task whose task type has no `case_property` — verify no HQ call is made and the task is deleted successfully
- [ ] Attempt to delete a `COMPLETED` task — verify it is excluded and no changes occur

### Labels & Review

### DIFF
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -191,18 +191,25 @@ def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, An
         raise CommCareHQAPIException(f"Failed to bulk-update {len(updates)} cases for {domain}. HQ Error: {e}") from e
 
 
-def update_usercase(opportunity_access: OpportunityAccess, data: dict[str, Any]) -> CommCareCase:
-    domain = opportunity_access.opportunity.deliver_app.cc_domain
-    api_key = opportunity_access.opportunity.api_key
+def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]]) -> None:
+    if not updates:
+        return
+
+    first_access = updates[0][0]
+    domain = first_access.opportunity.deliver_app.cc_domain
+    api_key = first_access.opportunity.api_key
     hq_server = api_key.hq_server
 
-    link = ConnectIDUserLink.objects.get(user=opportunity_access.user, domain=domain, hq_server=hq_server)
-    if link.hq_case_id is None:
-        usercase = get_usercase(opportunity_access)
-        link.hq_case_id = usercase.case_id
-        link.save()
+    cases_data = []
+    for access, data in updates:
+        link = ConnectIDUserLink.objects.get(user=access.user, domain=domain, hq_server=hq_server)
+        if link.hq_case_id is None:
+            usercase = get_usercase(access)
+            link.hq_case_id = usercase.case_id
+            link.save()
+        cases_data.append({"case_id": link.hq_case_id, **data})
 
-    return create_or_update_case(api_key, domain, data, case_id=link.hq_case_id)
+    bulk_update_cases(api_key, domain, cases_data)
 
 
 def get_usercase(opportunity_access: OpportunityAccess) -> CommCareCase:
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -172,6 +172,11 @@ def create_or_update_case(
 
 
 def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> None:
+    """Update usercase properties on CommCare HQ for multiple users in a single bulk request.
+
+    All entries in `updates` must belong to the same opportunity. The domain, API key, and
+    HQ server are derived from the first entry and applied to the entire batch.
+    """
     if not updates:
         return
 
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -171,26 +171,6 @@ def create_or_update_case(
     return CommCareCase(**data.get("case", {}))
 
 
-def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, Any]]) -> None:
-    """POST a JSON array of case updates to /api/case/v2/.
-
-    HQ's bulk endpoint requires each row to carry a create flag; this helper
-    only updates existing cases, so create=False is injected per item.
-
-    All-or-nothing: raises CommCareHQAPIException on any non-2xx response.
-    """
-    base_url = f"{api_key.hq_server.url}/a/{domain}/api/case/v2/"
-    headers = {"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"}
-    payload = [{**update, "create": False} for update in updates]
-
-    try:
-        with httpx.Client(base_url=base_url, headers=headers) as client:
-            response = client.post("", json=payload)
-        response.raise_for_status()
-    except (httpx.HTTPStatusError, httpx.RequestError) as e:
-        raise CommCareHQAPIException(f"Failed to bulk-update {len(updates)} cases for {domain}. HQ Error: {e}") from e
-
-
 def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]]) -> None:
     if not updates:
         return
@@ -207,9 +187,9 @@ def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]
             usercase = get_usercase(access)
             link.hq_case_id = usercase.case_id
             link.save()
-        cases_data.append({"case_id": link.hq_case_id, **data})
+        cases_data.append({"case_id": link.hq_case_id, "create": False, **data})
 
-    bulk_update_cases(api_key, domain, cases_data)
+    bulk_create_or_update_cases(api_key, domain, cases_data)
 
 
 def get_usercase(opportunity_access: OpportunityAccess) -> CommCareCase:
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -171,21 +171,21 @@ def create_or_update_case(
     return CommCareCase(**data.get("case", {}))
 
 
-def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]]) -> None:
+def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> None:
     if not updates:
         return
 
-    first_access = updates[0][0]
+    first_access = next(iter(updates))
     domain = first_access.opportunity.deliver_app.cc_domain
     api_key = first_access.opportunity.api_key
     hq_server = api_key.hq_server
 
-    users = [access.user for access, _ in updates]
+    users = [access.user for access in updates]
     links = ConnectIDUserLink.objects.filter(user__in=users, domain=domain, hq_server=hq_server)
     links_by_user = {link.user_id: link for link in links}
 
     cases_data = []
-    for access, data in updates:
+    for access, data in updates.items():
         link = links_by_user[access.user_id]
         if link.hq_case_id is None:
             usercase = get_usercase(access)
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -191,12 +191,16 @@ def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> N
 
     cases_data = []
     for access, data in updates.items():
-        link = links_by_user[access.user_id]
-        if link.hq_case_id is None:
-            usercase = get_usercase(access)
-            link.hq_case_id = usercase.case_id
+        link = links_by_user.get(access.user_id)
+        if link is None:
+            hq_case_id = get_usercase(access).case_id
+        elif link.hq_case_id is None:
+            hq_case_id = get_usercase(access).case_id
+            link.hq_case_id = hq_case_id
             link.save()
-        cases_data.append({"case_id": link.hq_case_id, "create": False, **data})
+        else:
+            hq_case_id = link.hq_case_id
+        cases_data.append({"case_id": hq_case_id, "create": False, **data})
 
     bulk_create_or_update_cases(api_key, domain, cases_data)
 
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -180,9 +180,13 @@ def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]
     api_key = first_access.opportunity.api_key
     hq_server = api_key.hq_server
 
+    users = [access.user for access, _ in updates]
+    links = ConnectIDUserLink.objects.filter(user__in=users, domain=domain, hq_server=hq_server)
+    links_by_user = {link.user_id: link for link in links}
+
     cases_data = []
     for access, data in updates:
-        link = ConnectIDUserLink.objects.get(user=access.user, domain=domain, hq_server=hq_server)
+        link = links_by_user[access.user_id]
         if link.hq_case_id is None:
             usercase = get_usercase(access)
             link.hq_case_id = usercase.case_id
--- a/commcare_connect/microplanning/helpers.py
+++ b/commcare_connect/microplanning/helpers.py
@@ -3,7 +3,7 @@ import logging
 import pghistory
 from django.db import transaction
 
-from commcare_connect.commcarehq.api import bulk_update_cases
+from commcare_connect.commcarehq.api import bulk_create_or_update_cases
 from commcare_connect.microplanning.const import HQ_BULK_CHUNK_SIZE
 from commcare_connect.microplanning.models import WorkArea, WorkAreaStatus
 from commcare_connect.utils.commcarehq_api import CommCareHQAPIException
@@ -44,11 +44,11 @@ def exclude_work_areas_for_opportunity(opportunity, work_area_ids, user, exclusi
     for i in range(0, len(needs_hq), HQ_BULK_CHUNK_SIZE):
         chunk = needs_hq[i : i + HQ_BULK_CHUNK_SIZE]  # noqa: E203
         # HQ's "unassigned" convention is "-"; empty string falls back to the submitting user.
-        updates = [{"case_id": str(wa.case_id), "owner_id": "-"} for wa in chunk]
+        updates = [{"case_id": str(wa.case_id), "owner_id": "-", "create": False} for wa in chunk]
         try:
             with transaction.atomic(), pghistory.context(**pghistory_ctx):
                 _bulk_exclude(chunk, user, exclusion_reason)
-                bulk_update_cases(api_key, domain, updates)
+                bulk_create_or_update_cases(api_key, domain, updates)
         except CommCareHQAPIException as e:
             logger.warning("Failed to unassign HQ case chunk (size=%d): %s", len(chunk), e)
             failed += len(chunk)
--- a/commcare_connect/opportunity/exceptions.py
+++ b/commcare_connect/opportunity/exceptions.py
@@ -1,0 +1,2 @@
+class ListTooLongError(Exception):
+    """Raised when a list exceeds the maximum allowed size for a bulk operation."""
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -496,7 +496,9 @@ class AssignedTask(XFormBaseModel):
                 hq_updates.setdefault((task.opportunity_access_id, prop), task.opportunity_access)
 
         with transaction.atomic():
-            deleted_count, _ = cls.objects.filter(pk__in=[t.pk for t in tasks]).delete()
+            deleted_count, _ = (
+                cls.objects.filter(pk__in=[t.pk for t in tasks]).exclude(status=AssignedTaskStatus.COMPLETED).delete()
+            )
             if hq_updates:
                 still_assigned = set(
                     cls.objects.filter(
@@ -505,10 +507,10 @@ class AssignedTask(XFormBaseModel):
                         task_type__case_property__in={p for _, p in hq_updates},
                     ).values_list("opportunity_access_id", "task_type__case_property")
                 )
-                to_reset = {
-                    hq_updates[access_id, prop]: {"properties": {prop: ""}}
-                    for access_id, prop in hq_updates.keys() - still_assigned
-                }
+                to_reset: dict[OpportunityAccess, dict] = {}
+                for access_id, prop in hq_updates.keys() - still_assigned:
+                    access = hq_updates[access_id, prop]
+                    to_reset.setdefault(access, {"properties": {}})["properties"][prop] = ""
                 if len(to_reset) > HQ_CASE_BULK_CHUNK_SIZE:
                     raise ListTooLongError(
                         f"Too many HQ case property resets ({len(to_reset)}); "
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -472,6 +472,35 @@ class AssignedTask(XFormBaseModel):
             transaction.on_commit(lambda: send_task_assignment_notification.delay(assigned_task.pk))
         return assigned_task
 
+    @classmethod
+    def delete_and_reset_hq(cls, task_ids: list[int], opportunity: "Opportunity") -> int:
+        from commcare_connect.commcarehq.api import update_usercase
+
+        tasks = list(
+            cls.objects.filter(
+                pk__in=task_ids,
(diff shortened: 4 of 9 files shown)


commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,25 @@ def create_or_update_case(
...
+    cases_data = []
+    for access, data in updates:
+        link = ConnectIDUserLink.objects.get(user=access.user, domain=domain, hq_server=hq_server)
+        if link.hq_case_id is None:
+            usercase = get_usercase(access)
+            link.hq_case_id = usercase.case_id
+            link.save()
+        cases_data.append({"case_id": link.hq_case_id, "create": False, **data})
+
+    bulk_create_or_update_cases(api_key, domain, cases_data)
```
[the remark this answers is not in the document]
zandre-eng: The chunking concern is addressed at the call site in `delete_and_reset_hq`, which raises a `ValueError` before any HQ call is made if the number of resets would exceed a single chunk. This makes the single-request guarantee explicit and enforceable without adding compensating write complexity. If the limit is ever legitimately hit, that's the point to revisit with a proper compensating strategy.

Addressed in [f92e6a7](https://github.com/dimagi/commcare-connect/pull/1174/commits/f92e6a793373c2b627abf614a80ad3a9458feaf0).

commcare_connect/opportunity/views.py
```diff
@@ -3574,17 +3574,13 @@ def delete_tasks(request, org_slug, opp_id):
...
-    )
-
-    if deleted_count:
-        messages.success(request, _("Successfully deleted %(count)d task(s).") % {"count": deleted_count})
-
     redirect_url = _task_redirect_url(request, org_slug, opp_id)
+
+    try:
+        deleted_count = AssignedTask.delete_and_reset_hq(task_ids, request.opportunity)
+    except CommCareHQAPIException:
```
hemant10yadav, line 3585: Missed `ValueError`?

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,25 @@ def create_or_update_case(
...
-    link = ConnectIDUserLink.objects.get(user=opportunity_access.user, domain=domain, hq_server=hq_server)
-    if link.hq_case_id is None:
-        usercase = get_usercase(opportunity_access)
-        link.hq_case_id = usercase.case_id
-        link.save()
-
-    return create_or_update_case(api_key, domain, data, case_id=link.hq_case_id)
+    cases_data = []
+    for access, data in updates:
+        link = ConnectIDUserLink.objects.get(user=access.user, domain=domain, hq_server=hq_server)
```
hemant10yadav: nit: Batch the lookup before the loop.

hemant10yadav (commented):
Minor suggestion, otherwise looks good to me.

commcare_connect/opportunity/views.py
pxwxnvermx, line 3585: can we log the error message here? it becomes hard to debug issues without the information from the error.

zandre-eng: I was contemplating whether we want to fail hard here. I'm assuming we would control how many tasks are sent to the view, so if a `ValueError` does pop up it's an unexpected coding error rather than a user error. The error message might therefore be misleading as a user trying again might not be able to resolve the issue.

With that said, I'm happy to handle this more gracefully so that the request doesn't fail hard.

commcare_connect/commcarehq/api.py
zandre-eng: Addressed in [8bbc697](https://github.com/dimagi/commcare-connect/pull/1174/commits/8bbc697adca06ed82743b42d8229664d345baf0b).

commcare_connect/opportunity/views.py
zandre-eng: I added a separate catch for `ValueError` so that we can give a more applicable error message to the user.

Addressed in [8bbc697](https://github.com/dimagi/commcare-connect/pull/1174/commits/8bbc697adca06ed82743b42d8229664d345baf0b).

Charl1996 (approved):
LGTM

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,29 @@ def create_or_update_case(
     return CommCareCase(**data.get("case", {}))
 
 
-def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, Any]]) -> None:
-    """POST a JSON array of case updates to /api/case/v2/.
+def bulk_update_usercases(updates: list[tuple[OpportunityAccess, dict[str, Any]]]) -> None:
```
pxwxnvermx: nit: Can the `updates` param be changed to a dictionary with keys for OpportunityAccess and `data` instead? That will make the code a bit more easier to read.

commcare_connect/opportunity/models.py
```diff
@@ -469,10 +468,54 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
             )
             case_property = task_type.case_property
             if case_property:
-                update_usercase(opportunity_access, data={"properties": {case_property: "1"}})
+                bulk_update_usercases([(opportunity_access, {"properties": {case_property: "1"}})])
             transaction.on_commit(lambda: send_task_assignment_notification.delay(assigned_task.pk))
         return assigned_task
 
+    @classmethod
+    def delete_and_reset_hq(cls, task_ids: list[int], opportunity: "Opportunity") -> int:
```
pxwxnvermx: nit: The "Opportunity" type hint could be replaced with the Opportunity class.

commcare_connect/opportunity/models.py
```diff
@@ -469,10 +468,54 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
...
+                pk__in=task_ids,
+                opportunity_access__opportunity=opportunity,
+            )
+            .exclude(status=AssignedTaskStatus.COMPLETED)
+            .select_related("task_type", "opportunity_access")
+        )
+        if not tasks:
+            return 0
+
+        hq_updates: dict[tuple[int, str], "OpportunityAccess"] = {}
```
pxwxnvermx: nit: The "OpportunityAccess" type hint could be replaced with the OpportunityAccess class.

pxwxnvermx (commented):
Some minor comments. Rest of the changes looks good to me.

commcare_connect/opportunity/views.py
zandre-eng: Good idea, addressed in [d632738](https://github.com/dimagi/commcare-connect/pull/1174/commits/d63273863195d52529c27c7db96696411fc5f55e).

commcare_connect/commcarehq/api.py
zandre-eng: Addressed in [bc214bd](https://github.com/dimagi/commcare-connect/pull/1174/commits/bc214bdb10a940e28736194ccc7355eab78624ff).

commcare_connect/opportunity/models.py
zandre-eng: Addressed in [30d24cf](https://github.com/dimagi/commcare-connect/pull/1174/commits/30d24cf5fb55ef7d5469f6c12d9c7101ae1d42bf).

commcare_connect/opportunity/models.py
zandre-eng: Addressed in [30d24cf](https://github.com/dimagi/commcare-connect/pull/1174/commits/30d24cf5fb55ef7d5469f6c12d9c7101ae1d42bf).

pxwxnvermx (approved):

commcare_connect/opportunity/models.py
```diff
@@ -473,6 +472,35 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
             transaction.on_commit(lambda: send_task_assignment_notification.delay(assigned_task.pk))
         return assigned_task
 
+    @classmethod
+    def delete_and_reset_hq(cls, task_ids: list[int], opportunity: "Opportunity") -> int:
```
mkangia: nit: I appreciate the details in naming. Here however, I think just delete would be fine, same as we have above for "assign". Resetting HQ is an internal detail or a side effect that I believe is not needed in the name.

commcare_connect/opportunity/models.py
```diff
@@ -473,6 +472,35 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
...
+        from commcare_connect.commcarehq.api import update_usercase
+
+        tasks = list(
+            cls.objects.filter(
+                pk__in=task_ids,
+                opportunity_access__opportunity=opportunity,
+            )
+            .exclude(status=AssignedTaskStatus.COMPLETED)
```
mkangia, line 487: should we filter for only the assigned status instead of excluding completed?

commcare_connect/commcarehq/api.py
```diff
@@ -191,18 +191,25 @@ def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, An
...
+    if not updates:
+        return
 
-    link = ConnectIDUserLink.objects.get(user=opportunity_access.user, domain=domain, hq_server=hq_server)
-    if link.hq_case_id is None:
-        usercase = get_usercase(opportunity_access)
-        link.hq_case_id = usercase.case_id
-        link.save()
+    first_access = updates[0][0]
```
mkangia: we are assuming that all updates are for the same opportunity -> domain -> hq_server.
Is it a safe assumption?

hemant10yadav (approved):

commcare_connect/opportunity/views.py
```diff
@@ -3580,6 +3580,8 @@ def delete_tasks(request, org_slug, opp_id):
         deleted_count = AssignedTask.delete_and_reset_hq(task_ids, request.opportunity)
     except CommCareHQAPIException:
         messages.error(request, _("Task deletion failed: could not update CommCare HQ. Please try again."))
+    except ValueError:
```
mkangia: I believe this exception does not seem the right fit for a "ValueError"

mkangia (changes requested):
Hey @zandre-eng 

I appreciate the improvement made to the api call though I will be honest that this made the PR's scope bigger and hence harder to review. Keeping the api improvement separate from the core implementation in a different PR would have been ideal.

Took a pass and shared some queries for now.

commcare_connect/commcarehq/api.py
zandre-eng: This should be a safe assumption given that we are filtering tasks by opportunity, so all tasks in the list should belong to the same opportunity/domain/hq_server. But since this is a public function that can be called elsewhere let me add a docstring to make this explicit.

commcare_connect/opportunity/models.py
zandre-eng: This is pre-existing logic that I copied over to the model function. I'm opted to leave it as is, in case we might eventually add an additional pre-complete status (e.g. unassigned) that we would like to include here as well. To my understanding the only status we don't want to delete are those that have been completed, hence the explicit `.exclude()` here.

commcare_connect/opportunity/models.py
zandre-eng: That's a fair point. I went with `bulk_delete` to more accurately reflect that this function takes a list.

Addressed in [349fa71](https://github.com/dimagi/commcare-connect/pull/1174/commits/349fa717865d93e579dfac515a6f96508bf76602).

commcare_connect/opportunity/views.py
zandre-eng: I ended up creating a custom `ListTooLongError` exception class as I couldn't think of what else would fit better here than `ValueError`.

Addressed in [349fa71](https://github.com/dimagi/commcare-connect/pull/1174/commits/349fa717865d93e579dfac515a6f96508bf76602).

zandre-eng:
> I appreciate the improvement made to the api call though I will be honest that this made the PR's scope bigger and hence harder to review. Keeping the api improvement separate from the core implementation in a different PR would have been ideal.

@mkangia That's a fair point. At the time it felt a little challenging to separate as the API improvements are specifically for the core implementation. It allows us to do a bulk DB and HQ update together, which prevents a single failed HQ update from rolling back all the DB changes and leaving the two out of sync. With that said, in hindsight this could've gone in a follow-up PR.

commcare_connect/commcarehq/api.py
zandre-eng: Docstring added in [349fa71](https://github.com/dimagi/commcare-connect/pull/1174/commits/349fa717865d93e579dfac515a6f96508bf76602).

commcare_connect/opportunity/models.py
```diff
@@ -468,10 +471,54 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
...
+                    hq_updates[access_id, prop]: {"properties": {prop: ""}}
+                    for access_id, prop in hq_updates.keys() - still_assigned
+                }
+                if len(to_reset) > HQ_CASE_BULK_CHUNK_SIZE:
+                    raise ListTooLongError(
+                        f"Too many HQ case property resets ({len(to_reset)}); "
+                        "split the delete into smaller batches."
+                    )
+                if to_reset:
+                    bulk_update_usercases(to_reset)
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [44d70c5](https://github.com/dimagi/commcare-connect/pull/1174/commits/44d70c52da6f599085ce7e7b33de92203930b59f).

commcare_connect/opportunity/models.py
```diff
@@ -468,10 +471,54 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
...
+        if not tasks:
+            return 0
+
+        hq_updates: dict[tuple[int, str], OpportunityAccess] = {}
+        for task in tasks:
+            if prop := task.task_type.case_property:
+                hq_updates.setdefault((task.opportunity_access_id, prop), task.opportunity_access)
+
+        with transaction.atomic():
+            deleted_count, _ = cls.objects.filter(pk__in=[t.pk for t in tasks]).delete()
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [44d70c5](https://github.com/dimagi/commcare-connect/pull/1174/commits/44d70c52da6f599085ce7e7b33de92203930b59f).

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,34 @@ def create_or_update_case(
...
 
-    return create_or_update_case(api_key, domain, data, case_id=link.hq_case_id)
+    cases_data = []
+    for access, data in updates.items():
+        link = links_by_user[access.user_id]
+        if link.hq_case_id is None:
+            usercase = get_usercase(access)
+            link.hq_case_id = usercase.case_id
+            link.save()
+        cases_data.append({"case_id": link.hq_case_id, "create": False, **data})
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [44d70c5](https://github.com/dimagi/commcare-connect/pull/1174/commits/44d70c52da6f599085ce7e7b33de92203930b59f).

commcare_connect/opportunity/models.py
mkangia: > To my understanding the only status we don't want to delete are those that have been completed

that makes sense, thank you.

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,38 @@ def create_or_update_case(
...
-def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, Any]]) -> None:
-    """POST a JSON array of case updates to /api/case/v2/.
+def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> None:
+    """Update usercase properties on CommCare HQ for multiple users in a single bulk request.
 
-    HQ's bulk endpoint requires each row to carry a create flag; this helper
-    only updates existing cases, so create=False is injected per item.
-
-    All-or-nothing: raises CommCareHQAPIException on any non-2xx response.
+    All entries in `updates` must belong to the same opportunity. The domain, API key, and
```
mkangia, line 177: nit: the documentation seems a fair place to explain though a code validation to ensure same opportunity would be the real check that would prevent any wrong update attempts.

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,38 @@ def create_or_update_case(
     return CommCareCase(**data.get("case", {}))
 
 
-def bulk_update_cases(api_key: HQApiKey, domain: str, updates: list[dict[str, Any]]) -> None:
-    """POST a JSON array of case updates to /api/case/v2/.
+def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> None:
+    """Update usercase properties on CommCare HQ for multiple users in a single bulk request.
```
mkangia, line 175: nit: would be nice to add here what structure you expect of the updates dictionary.

commcare_connect/commcarehq/api.py
```diff
@@ -171,38 +171,38 @@ def create_or_update_case(
...
-        link.save()
+    users = [access.user for access in updates]
+    links = ConnectIDUserLink.objects.filter(user__in=users, domain=domain, hq_server=hq_server)
+    links_by_user = {link.user_id: link for link in links}
 
-    return create_or_update_case(api_key, domain, data, case_id=link.hq_case_id)
+    cases_data = []
+    for access, data in updates.items():
+        link = links_by_user.get(access.user_id)
+        if link is None:
```
mkangia, line 195: nit: just checking if this is something we really expect to happen? can a user not have a link?

commcare_connect/microplanning/helpers.py
```diff
@@ -44,11 +44,11 @@ def exclude_work_areas_for_opportunity(opportunity, work_area_ids, user, exclusi
     for i in range(0, len(needs_hq), HQ_BULK_CHUNK_SIZE):
         chunk = needs_hq[i : i + HQ_BULK_CHUNK_SIZE]  # noqa: E203
         # HQ's "unassigned" convention is "-"; empty string falls back to the submitting user.
-        updates = [{"case_id": str(wa.case_id), "owner_id": "-"} for wa in chunk]
+        updates = [{"case_id": str(wa.case_id), "owner_id": "-", "create": False} for wa in chunk]
```
mkangia, line 47: nit: Ideally if this is done by the function doing the api call, so, `bulk_update_cases` then it makes it simpler for calls to that function to not worry about setting this

mkangia: and I see we are now using the same function for create and update, so in this case, one would have a param called "create" that would be a boolean to tell if it needs to prepare data for create or update

commcare_connect/opportunity/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class ListTooLongError(Exception):
```
mkangia, line 1: 👍

commcare_connect/opportunity/models.py
```diff
@@ -469,10 +471,56 @@ def assign(cls, *, task_type, opportunity_access, due_date, assigned_by=None) ->
             )
             case_property = task_type.case_property
             if case_property:
-                update_usercase(opportunity_access, data={"properties": {case_property: "1"}})
+                bulk_update_usercases({opportunity_access: {"properties": {case_property: "1"}}})
             transaction.on_commit(lambda: send_task_assignment_notification.delay(assigned_task.pk))
         return assigned_task
 
+    @classmethod
+    def bulk_delete(cls, task_ids: list[int], opportunity: Opportunity) -> int:
```
mkangia, line 479: nit: would be nice to refactor this method.

mkangia (approved):
non blocking feedback, feel free to take in a follow up PR/ticket if needed

mkangia:
> With that said, in hindsight this could've gone in a follow-up PR.

Just adding that at times doing the change in the same PR seems fine when its a small tweak but this one here was a big enough change to warrant its own PR. It's the delay in review that punts the benefit one gets from keeping things at one place.

commcare_connect/commcarehq/api.py
zandre-eng: My understanding is that this should be unusual to happen, but regardless I decided to add some defensive code here so that we can gracefully handle potential edge cases. Happy to remove this if you think this is just adding defensive noise to the function.

zandre-eng:
> non blocking feedback, feel free to take in a follow up PR/ticket if needed

@mkangia Thanks for the review. Since it's non-blocking I'm going to go ahead and merge for now and will address the comments in a follow-up PR.

PR #1176, opened 2026-05-12
UI changes for automatic verification Opportunities

pxwxnvermx:
## Product Description
When an opportunity has automatic_visit_verification=True, manual NM/PM review surfaces are no longer meaningful. This branch hides those UI elements and guards the backend endpoints, while leaving manual-mode opportunities untouched.

- Backend guard
  -  block_when_automatic_verification decorator returns 403 + HX-Trigger: reload_table for direct posts to: user_visit_review, approve_visits, reject_visits, update_visit_status_import, review_visit_export, review_visit_import.

- Verification Rules Configuration page, when auto-verify is on:
  - Page title, breadcrumb, and section heading conditionally show "Rules" wording ("Verification Rules Configuration", "Deliver Unit Rules"); manual mode keeps the original "Flags" wording.
  - Form hides Check Duplicates, Check GPS, Catchment Area, Location Distance, and Require Attachments. 
  - Save force-sets those fields to falsy so stored values can't keep flagging logic alive.

- User Visits page, when auto-verify is on:
  - PM sees only the All tab; NM sees Approved / Rejected / All (no Pending NM Review, Pending PM Review, Agree, Disagree, or Revalidate).
  - Bulk Approve/Reject buttons and per-visit Agree/Disagree/Approve/Reject buttons (with modals) are hidden.
  
- Worker Delivery page, when auto-verify is on:
  - Import button + modal hidden (export retained).
  - "PM Review Sheet" radio dropped from the export modal.
  - "Pending" column excluded from the workers table.
  - review_pending and has_duplicates filters dropped from the filter modal.
  - "Pending" status choice stripped from the User Visits export form.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2286)

### Safety story
- Tested locally and on staging.

### Automated test coverage
- Added automated tests.

### Labels & Review


commcare_connect/opportunity/views.py
```diff
@@ -2298,8 +2312,28 @@ def get_table(self, **kwargs):
...
+                        "label": "Approved",
+                        "count": user_visit_counts.get("approved", 0),
+                    },
+                    {
+                        "name": "rejected",
+                        "label": "Rejected",
+                        "count": user_visit_counts.get("rejected", 0),
+                    },
+                    {"name": "all", "label": "All", "count": user_visit_counts.get("total", 0)},
+                ]
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/1b93daa08158752a1c0a33efebc029f997be3202

commcare_connect/opportunity/views.py
```diff
@@ -620,6 +623,7 @@ def update_visit_status_import(request, org_slug=None, opp_id=None):
 
 
 @opportunity_required
+@block_when_automatic_verification
 def review_visit_import(request, org_slug=None, opp_id=None):
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/3d8137a2deba1c22bcd7451bba404f0f861a3e15

commcare_connect/opportunity/forms.py
```diff
@@ -1386,36 +1386,62 @@ class Meta:
...
+                self.fields.pop(hidden, None)
+            self.helper.layout = Layout(
+                Fieldset(
+                    "Form Submission Hours",
+                    Row(
+                        Field("form_submission_start"),
+                        Field("form_submission_end"),
+                        css_class="grid grid-cols-2 gap-2",
+                    ),
+                ),
```
zandre-eng: Nit: This is also inserted in the `else` conditional and can be pulled out into a shared variable.

commcare_connect/opportunity/forms.py
```diff
@@ -1386,36 +1386,62 @@ class Meta:
...
-            Row(Field("location")),
-            Fieldset(
-                "Form Submission Hours",
+        auto_verify = bool(self.opportunity and self.opportunity.automatic_visit_verification)
+        if auto_verify:
+            for hidden in ("duplicate", "gps", "catchment_areas", "location"):
+                self.fields.pop(hidden, None)
+            self.helper.layout = Layout(
+                Fieldset(
+                    "Form Submission Hours",
```
zandre-eng: Nit: Translation.

commcare_connect/opportunity/views.py
```diff
@@ -1196,7 +1206,12 @@ def verification_flags_config(request, org_slug=None, opp_id=None):
             "title": request.opportunity.name,
             "url": reverse("opportunity:detail", args=(org_slug, request.opportunity.opportunity_id)),
         },
-        {"title": "Verification Flags Config", "url": request.path},
+        {
+            "title": "Verification Rules Configuration"
```
zandre-eng: Nit: Translation for titles.

commcare_connect/templates/opportunity/verification_flags_config.html
```diff
@@ -8,7 +8,7 @@
 {% include 'components/breadcrumbs.html' with path=path %}
 
 <div class="container md:max-w-screen-lg mx-auto flex flex-col gap-6">
-  <h2 class="title">Verification Flags Configuration</h2>
+  <h2 class="title">{% if opportunity.automatic_visit_verification %}Verification Rules Configuration{% else %}Verification Flags Configuration{% endif %}</h2>
```
zandre-eng: Nit: Translation.

commcare_connect/opportunity/tests/test_automatic_verification.py
```diff
@@ -0,0 +1,299 @@
...
+            (a for a in response.context["recent_activities"] if a["title"] == "Pending Review"), None
+        )
+        assert pending_review_card is not None
+        opp_names = {row["opportunity__name"] for row in pending_review_card["rows"]}
+        assert manual_opp.name in opp_names
+        assert auto_opp.name not in opp_names
+
+
+@pytest.mark.django_db
+class TestAutomaticVerificationBackendGuards:
```
zandre-eng, line 200: Nit: The tests in this class all share a very similar structure and can be parametrized.

commcare_connect/opportunity/forms.py
```diff
@@ -1386,36 +1392,62 @@ class Meta:
...
-            Row(
-                Field("duplicate", css_class=f"{CHECKBOX_CLASS} block"),
-                Field("gps", css_class=f"{CHECKBOX_CLASS} block"),
-                Field("catchment_areas", css_class=f"{CHECKBOX_CLASS} block"),
-                css_class="grid grid-cols-3 gap-2",
-            ),
-            Row(Field("location")),
-            Fieldset(
-                "Form Submission Hours",
+        auto_verify = bool(self.opportunity and self.opportunity.automatic_visit_verification)
```
zandre-eng: Nit: This could be saved as an instance variable so that it can also be used in the `save()` method.

commcare_connect/opportunity/forms.py
```diff
@@ -1428,20 +1460,41 @@ def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity")
         super().__init__(*args, **kwargs)
 
+        auto_verify = self.opportunity.automatic_visit_verification
```
zandre-eng: Should we be checking that `self.opportunity` exists first here (similar to what we do in `OpportunityVerificationFlagsConfigForm`)?

commcare_connect/opportunity/views.py
```diff
@@ -2298,8 +2314,28 @@ def get_table(self, **kwargs):
 
     def get_context_data(self, **kwargs):
         user_visit_counts = get_user_visit_counts(self.opportunity, self.filter_queryset)
+        opportunity = self.get_opportunity()
 
-        if self.request.is_opportunity_pm:
+        if opportunity.automatic_visit_verification:
+            if self.request.is_opportunity_pm:
+                tabs = [
+                    {"name": "all", "label": "All", "count": user_visit_counts.get("total", 0)},
```
zandre-eng: Nit: Should all the labels here get translations?

commcare_connect/program/views.py
```diff
@@ -310,6 +310,7 @@ def program_manager_home(request, org):
             opportunity__managed=True,
             opportunity__managedopportunity__program__in=programs_qs,
         )
+        .exclude(opportunity__automatic_visit_verification=True)
```
zandre-eng, line 313: With automatic verification there shouldn't ever be any pending visits, however if a bug/issue in the logic causes a visit to be pending would we still want to exclude it here? Just thinking whether valid visits might go completely unnoticed due to a logic error.

zandre-eng (commented):
Implementation looks good overall. Happy to approve after resolving non-nit questions. Otherwise, mostly nits about translation.

commcare_connect/opportunity/decorators.py
```diff
@@ -0,0 +1,23 @@
+from functools import wraps
+
+from django.http import HttpResponseForbidden
+
+
+def block_when_automatic_verification(view_func):
```
sravfeyn: nit: Would rename this to something like `require_manual_visit_verification`

commcare_connect/opportunity/views.py
```diff
@@ -2298,8 +2314,28 @@ def get_table(self, **kwargs):
 
     def get_context_data(self, **kwargs):
         user_visit_counts = get_user_visit_counts(self.opportunity, self.filter_queryset)
+        opportunity = self.get_opportunity()
 
-        if self.request.is_opportunity_pm:
+        if opportunity.automatic_visit_verification:
```
sravfeyn, line 2342: Instead of branching inside the views, it is cleaner to use inheritance and conditionally switch the views. 

This PR is doing something similar to that of https://github.com/dimagi/commcare-connect/pull/1094/ which did a bunch of ground-work for this PR. Routing to simpler view without all tabs when that switch was active.

Below was introduced in that PR that can be used here by passing `or request.opportunity.automatic_visit_verification` condition. Or re-spec which condition to use.

```
@org_viewer_required
@opportunity_required
def visit_verification_table_view(request, org_slug, opp_id):
    if switch_is_active(WORKER_VISITS_TASKS):
        return WorkerVisitTableView.as_view()(request, org_slug=org_slug, opp_id=opp_id)
    return VisitVerificationTableView.as_view()(request, org_slug=org_slug, opp_id=opp_id)
```

commcare_connect/templates/opportunity/deliver.html
```diff
@@ -5,17 +5,21 @@
 {% block tab_options %}
   {% include "components/filter_modal.html" with form_id="filterForm" hx_indicator="#loadingIndicator" %}
   <div x-cloak x-data="{showVisitImportModal: false, showVisitExportModal: false}">
+    {% if not opportunity.automatic_visit_verification %}
```
sravfeyn, line 8: It's better to define something like `show_worker_tasks_tabs = switch_is_active(WORKER_VISITS_TASKS)` (what the other PR does) and use that template var inside this template.

commcare_connect/opportunity/forms.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/48b83c0dddbb927d07359625a322699062a65433

commcare_connect/opportunity/forms.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/01f69df03f658e02287c3863f31b5422ba72835c

commcare_connect/opportunity/views.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/5b18d75119e4070100fb559846d8226daf06bf16

commcare_connect/templates/opportunity/verification_flags_config.html
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/5b18d75119e4070100fb559846d8226daf06bf16

commcare_connect/opportunity/tests/test_automatic_verification.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/c0a320e39ae9a8b21d5e314c2ea505c166fb1e73

commcare_connect/opportunity/forms.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/53cf8ec704dafebf713adf3b4ed073a3f5347238

commcare_connect/opportunity/forms.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/53cf8ec704dafebf713adf3b4ed073a3f5347238

commcare_connect/opportunity/views.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/22d16edca263f67294500c0a3f59c75747cdfd7b

commcare_connect/opportunity/decorators.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/8d5ebfd3657e0b2633e9d04d1a3625a9873c501b

commcare_connect/templates/opportunity/deliver.html
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1176/commits/ced411b4ab51c484848f7fb00dae61aac6dcf469

commcare_connect/opportunity/views.py
pxwxnvermx: This view had conditional switching for tabs, the view remained the same only the tabs being shown were changed. I have created changes to make the code for tabs more concise and readable the conditional switching now happens in the `get_tabs` method. 
https://github.com/dimagi/commcare-connect/pull/1176/commits/fe29e5eee172355c144311635010542273edc560

commcare_connect/program/views.py
pxwxnvermx: The Opportunity managers can still use the visit_export to see a full list of visits and raise tickets for any anomalies i.e Pending visits that they spot.

pxwxnvermx:
@sravfeyn @zandre-eng Thanks for the review feedback. I have addressed it. Can you please do another review whenever possible? 😄 

PS: Earlier I left this comment on the wrong PR please ignore that one.

zandre-eng (approved):

commcare_connect/opportunity/decorators.py
```diff
@@ -0,0 +1,27 @@
...
+    The endpoint is only meaningful for manual-review opportunities. Must be
+    applied after @opportunity_required so request.opportunity is set.
+    Returns 403 with an HX-Trigger: reload_table header so HTMX clients
+    refresh after a guard hit instead of showing a stale UI.
+    """
+
+    @wraps(view_func)
+    def _inner(request, *args, **kwargs):
+        assert (
+            getattr(request, "opportunity", None) is not None
```
mkangia, line 19: 👍

commcare_connect/opportunity/filters.py
```diff
@@ -125,6 +125,13 @@ class DeliverFilterSet(django_filters.FilterSet):
     class Meta:
         form = CSRFExemptForm
 
+    def __init__(self, *args, **kwargs):
+        opportunity = kwargs.pop("opportunity", None)
+        super().__init__(*args, **kwargs)
+        if opportunity and opportunity.automatic_visit_verification:
+            self.filters.pop("review_pending", None)
```
mkangia: nit: Do you need to the default None here?

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
             ),
         )
 
-        self.fields["duplicate"].required = False
-        self.fields["gps"].required = False
-        self.fields["catchment_areas"].required = False
+        if self.auto_verify:
+            for hidden in ("duplicate", "gps", "catchment_areas", "location"):
```
mkangia: nit: took me a moment to figure this out,
I didn't realize "hidden" was for a "field". Since the iteration is of the fields, it should still be called "field" or "field_name".

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
...
+                self.fields.pop(hidden, None)
+            self.helper.layout = Layout(form_submission_hour_fields)
```
mkangia, line 1417: nit: just confirming that since you are manually setting the entire Layout, do you need to remove the hidden/redundant fields?

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
...
             self.fields["form_submission_start"].initial = self.instance.form_submission_start
             self.fields["form_submission_end"].initial = self.instance.form_submission_end
 
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        if self.auto_verify:
+            instance.duplicate = False
+            instance.gps = False
+            instance.catchment_areas = False
+            instance.location = 0
```
mkangia, line 1442: (non blocking) I wonder if this should be done?
we are removing some values without letting the user know but it is a reasonable reset anyway.
Since an opportunity should have this flag set from starting and not switch mid way, this would probably never reset any value. So it should be safe as well.

mkangia: my suggested alternate would be to either return an error if these values are set (for an update)

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
...
+
 
 class DeliverUnitFlagsForm(forms.ModelForm):
     class Meta:
         model = DeliverUnitFlagRules
         fields = ("deliver_unit", "check_attachments", "duration")
-        help_texts = {"duration": "Minimum time to complete form (minutes)"}
-        labels = {"check_attachments": "Require Attachments"}
+        help_texts = {"duration": _("Minimum time to complete form (minutes)")}
+        labels = {"check_attachments": _("Require Attachments")}
```
mkangia, line 1453: thanks for changing these to be translated 👍

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
...
 
     def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity")
         super().__init__(*args, **kwargs)
 
+        self.auto_verify = bool(self.opportunity and self.opportunity.automatic_visit_verification)
```
mkangia: do we expect opportunity not to be passed?

I believe that would be unexpected so we should just check for the flag instead of checking first that opportunity is present.

commcare_connect/opportunity/forms.py
```diff
@@ -1409,39 +1411,86 @@ def __init__(self, *args, **kwargs):
...
+                ),
+            )
         self.fields["deliver_unit"] = forms.ModelChoiceField(
             queryset=DeliverUnit.objects.filter(app=self.opportunity.deliver_app), disabled=True, empty_label=None
         )
 
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+        if self.auto_verify:
+            instance.check_attachments = False
```
mkangia, line 1489: same as https://github.com/dimagi/commcare-connect/pull/1176/changes#r3257571186

commcare_connect/opportunity/views.py
```diff
@@ -2296,78 +2312,42 @@ def get_table(self, **kwargs):
...
-            ]
-        else:
-            tabs = [
-                {
-                    "name": "pending",
-                    "label": "Pending NM Review",
-                    "count": user_visit_counts.get("pending", 0),
-                }
-            ]
+    def get_tabs(self, **kwargs):
```
mkangia: Nice clean up.
Though it seems "rejected" and "all" were available for all branches but now it is not?

commcare_connect/opportunity/views.py
```diff
@@ -2296,78 +2312,42 @@ def get_table(self, **kwargs):
...
+        tabs_to_labels = {
+            "all": _("All"),
+            "approved": _("Approved"),
+            "rejected": _("Rejected"),
+            "pending": _("Pending NM Review"),
+            "pending_review": _("Pending PM Review"),
+            "agree": _("Agree") if self.request.is_opportunity_pm else _("Approved"),
+            "disagree": _("Disagree") if self.request.is_opportunity_pm else _("Revalidate"),
+        }
+        tabs_to_display = self.get_tabs()
```
mkangia: nit: looks like this function should be above `get_tabs` for a better ordering of functions.

commcare_connect/opportunity/views.py
```diff
@@ -2383,7 +2363,9 @@ def get_queryset(self):
         self.exclude_columns = []
         self.filter_queryset = super().get_queryset()
 
-        self.filter_status = self.request.GET.get("filter_status")
+        requested_status = self.request.GET.get("filter_status")
+        allowed = set(self.get_tabs())
```
mkangia: nit: do we want to memoize this function since its called multiple times?

mkangia (commented):
couple of minor tweaks for you to consider but others non blocking.

commcare_connect/opportunity/views.py
pxwxnvermx: They are available for the workflows that they were available for before.

Program Managers had access to these tabs "Pending PM Review", "Agree", "Disagree" and "All".
Network Managers had access to "Pending NM Review", "pending PM Review", "Accepted", "Rejected" and "All".

Non Managed Opportunities had access to "Pending NM Review", "Accepted", "Rejected" and "All".

commcare_connect/opportunity/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/731ae02d9a483c23f714f5a98eba4fe9c60b15c2

commcare_connect/opportunity/views.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/95387d48fc05696d620b45343a80201188de7720

commcare_connect/opportunity/forms.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/a76bdeb113b9646668cf03ed484b468e211efab3

opportunity is a required variable on the form, the form would throw an error if it is not present.

commcare_connect/opportunity/forms.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1176/commits/59ac1e57c4ed36a07f1583bbb3c0a13585c312db

commcare_connect/opportunity/forms.py
pxwxnvermx: Removing the hidden fields makes it impossible to set them through a form submission as the Django code would not process them.

commcare_connect/opportunity/filters.py
pxwxnvermx: Nope that is not required. Removed it.

commcare_connect/opportunity/views.py
mkangia: I am looking at the last few lines in the older code that adds rejected and all to tabs.

```
tabs.extend(
                [
                    {
                        "name": "rejected",
                        "label": "Rejected",
                        "count": user_visit_counts.get("rejected", 0),
                    },
                    {"name": "all", "label": "All", "count": user_visit_counts.get("total", 0)},
                ]
            )
```

Now when `automatic_visit_verification` is set and `if self.request.is_opportunity_pm` is true we are showing just all and not rejected. Rejected would still be available for new workflow right?

mkangia: I am trying my best to figure out what was present and what has been added and compare it but it's quite hard to map.

pxwxnvermx: In the older code the if statement separated PM and NM tabs. The "Rejected" and "All" tabs were always shown for the NMs only.

[PM Tabs](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/views.py#L2309-L2327)

[NM Tabs](https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/opportunity/views.py#L2328-L2374)

mkangia: okay, I trust your understanding of this.

commcare_connect/opportunity/forms.py
pxwxnvermx: The defaults for some of these values are non-falsy. So we need to set these to false when the instance is saved to avoid setting these to true which would be incorrect for the auto_verify workflows.

```
    gps = models.BooleanField(default=True)
    duplicate = models.BooleanField(default=True)
    location = models.PositiveIntegerField(default=10)
```

commcare_connect/opportunity/forms.py
pxwxnvermx: added a clarification on the linked comment.

commcare_connect/opportunity/forms.py
mkangia: okay :+1:

mkangia (approved):

mkangia (approved):

PR #1179, opened 2026-05-12
Add .md file for replication model addition options

Charl1996:
For review and input only!

Ticket: https://dimagi.atlassian.net/browse/CCCT-2094

Please review and leave comments / thoughts / suggestions.

### DIFF
--- a/docs/superset-sync-options.md
+++ b/docs/superset-sync-options.md
@@ -1,0 +1,87 @@
+# Superset Table Sync: Options for Updating Replicated Tables Without a Deploy
+
+## Background
+
+Tables are synced to Superset via PostgreSQL logical replication from the primary DB to a read-only secondary DB.
+The list of replicated tables is currently defined as `REPLICATION_ALLOWED_MODELS` in
+`commcare_connect/multidb/constants.py`.
+
+Adding a table today requires a code change, PR, and a full GitHub Actions deploy (image build → push to ECR →
+rolling restart). This couples an infrastructure concern to the release cycle.
+
+---
+
+## Option 1 (Current): Hardcoded Python list
+
+`REPLICATION_ALLOWED_MODELS` is a Python list of model classes checked into the codebase.
+The `setup_logical_replication` management command reads this list and issues
+`ALTER PUBLICATION ... SET TABLE ...` on the primary and `ALTER SUBSCRIPTION ... REFRESH PUBLICATION`
+on the secondary.
+
+**Pros**
+- Zero complexity — nothing extra to build or operate
+- Full git audit trail; changes go through code review
+- Code review acts as an intentional gate on what gets exposed to Superset
+
+**Cons**
+- Every table addition is a release event, bundling an infra change with feature code
+- Slow — hours to days depending on review and deploy queue
+- Whatever is on `main` at the time of the deploy ships alongside the change
+
+---
+
+## Option 2: Environment variable (1Password → Ansible → restart)
+
+Replace the hardcoded list with a `SUPERSET_REPLICATED_TABLES` env var (comma-separated table names).
+The management command reads from the env at runtime instead of from the Python constant.
+
+**One-time setup**: a small PR to add the var to `deploy/roles/connect/templates/docker.env.j2`.
+
+**To add a table after that**:
+1. Update the value in 1Password (Connect Tech vault)
+2. Run `ansible-playbook play.yml -t django_settings` to regenerate `docker.env` on servers
+3. `kamal app restart` to reload the env
+
+No image rebuild. No GitHub Actions pipeline. Estimated time: ~10–20 min with infra access.
+
+**Pros**
+- Fast after one-time setup
+- Decoupled from code releases; table list changes are purely an ops concern
+- Low implementation effort — small, low-risk one-time PR
+- Uses existing tooling (1Password, Ansible) with no new infrastructure
+
+**Cons**
+- Container restart still required — brief interruption or rolling restart risk
+- Requires infra access; not self-serve for non-engineers
+- Table list lives outside version control after the initial PR
+
+---
+
+## Option 3: DB-backed config with Django admin
+
+Store replicated table names in a Django model (e.g. `SupersetReplicatedTable`).
+Override `save()`/`delete()` or use a post-save signal to immediately issue
+`ALTER PUBLICATION ... SET TABLE ...` on the primary and
+`ALTER SUBSCRIPTION ... REFRESH PUBLICATION` on the secondary.
+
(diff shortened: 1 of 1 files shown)


mkangia (commented):
I am aligned with removing the requirement of a deploy from this process.

3a with 3b (running management command manually) as a backup seems okay to me.

docs/superset-sync-options.md
```diff
@@ -0,0 +1,109 @@
...
+2. Run `./manage.py setup_logical_replication` (or trigger an admin action) to apply the change
+
+**Pros**
+- Simpler to implement than 3a — no signal hooks, no DDL credentials in app settings
+- DDL is only issued on explicit operator intent, not as a side effect of a model save
+- Easier to reason about failures — the sync step is explicit and observable
+- Still decoupled from code deploys and the release cycle
+
+**Cons**
+- Not fully self-serve — applying the change still requires someone to run a command or trigger an action
```
zandre-eng, line 100: I don't think this is necessarily a bad thing. It allows us to be more aware when the sync happens such that we can immediately see if anything went wrong. The other benefit is we can continue to restrict this ability to devs only if we don't want non-devs to be able to sync tables.

docs/superset-sync-options.md
```diff
@@ -0,0 +1,109 @@
...
+- Two-step process introduces a window where the admin shows a table that isn't yet replicated
```
zandre-eng, line 101: Perhaps we can have a boolean field on the new model that reflects whether the table has been synced. Once the management command is run for that row we simply flip `is_synced` to true to help give an accurate reflection of which tables are currently synced to Superset.

docs/superset-sync-options.md
```diff
@@ -0,0 +1,109 @@
...
+- `setup_logical_replication` is currently interactive (prompts for DB credentials); would need a
```
zandre-eng, line 102: This should probably be fine to keep as is. Since a dev will need to run the management command anyways, we can continue using the interactive command and that way keep from having to store anything as application secrets.

zandre-eng (approved):
I'm in favour of option 3b. It removes any need for a restart, and allows us to more easily add entries instead of having it as a potentially long list in our env vars. The other benefit with a model is that we can provide more information regarding the sync status of each table.

docs/superset-sync-options.md
```diff
@@ -0,0 +1,109 @@
...
+Tables are synced to Superset via PostgreSQL logical replication from the primary DB to a read-only secondary DB.
+The list of replicated tables is currently defined as `REPLICATION_ALLOWED_MODELS` in
+`commcare_connect/multidb/constants.py`.
+
+Adding a table today requires a code change, PR, and a full GitHub Actions deploy (image build → push to ECR →
+rolling restart). This couples an infrastructure concern to the release cycle.
+
+---
+
+## Option 1 (Current): Hardcoded Python list
```
calellowitz, line 14: I think we generally want every table to be in superset, one option here is to just add tests that make sure all new models are in this list (and maybe add an exempted list so that we exclude tables if necessary). That would create a build failure when a table is missed and mean that new models would always be properly configured on their initial deploy (even if we still need to run the command).

docs/superset-sync-options.md
calellowitz: Made a similar comment above, but I think the ideal is actually for all this to happen in an automated way. We want basically every table immediately available in superset, and the motivation for this ticket was our team repeatedly missing adding these tables, and then needing dev action to get them in.

calellowitz (commented):
Is there a way to automate the running of the management command? I think that is the big thing here, which is that right now new models don't get added unless someone remembers to add it to the list and run the command. This doesn't address the root cause of users discovering that a table they are looking for is missing, and the need for manual dev action that someone needs to remember to do or do in an interrupt request.

I think the suggestions in the spec are nice improvements to make it easier, but all still rely on manual dev behavior.

docs/superset-sync-options.md
```diff
@@ -0,0 +1,109 @@
...
+**Cons**
+- Highest implementation effort — new model, admin registration, signal hook, error handling
+- DDL credentials (superuser on secondary DB) must be stored as application secrets, increasing security surface
+- A failed `ALTER PUBLICATION` silently diverges what admin shows from what Postgres actually replicates;
+  needs careful error handling and monitoring
+- Coupling a Django admin save to a DDL statement is unusual and harder to reason about under failure
+
+---
+
+## Option 3b: DB-backed config with Django admin + manual sync
```
sravfeyn, line 83: Since the goal is simply to avoid a code-review/deploy cycle, this option seems good enough to me without the additional complexity of 3a and giving enough visibility to a dev. 

Though, I am not opposed to 3a as well, if we get sufficient alerts/monitoring of the automated process.

docs/superset-sync-options.md
sravfeyn: If we are actually fine syncing almost every table (with probably an Exclusion-list) then yeah, this is fine. But I thought, we wanted to be selective each time as syncing tables without someone reviewing which table it is can result in unintended data exposure on superset side. (For e.g. some sort of Payment/Credit-Card model being added in future).

May be we can maintain two separate lists `EXCLUDE/INCLUDE` and have the CI fail if a model is not listed in either of places. This will let the dev make an explicit decision each time a model is added to the codebase.

When the `EXCLUDE/INCLUDE` lists are modified we can create a DB migration that calls `setup_logical_replication`

sravfeyn:
(Regardless of which automated option we are going with) It might be good to include some sort of check to make sure table is getting synced to superset after the command run succeeds.

I usually run `logical_replication_status`  command that prints table counts on primary vs superset to make sure the table I added is actually getting synced correctly.

docs/superset-sync-options.md
Charl1996: @calellowitz 

> I think we generally want every table to be in superset

I was not aware that this is the real problem to be solved and thought it's only about reducing friction when adding new models manually. I was under the impression that we explicitly _don't_ want every model synced to superset, lest (to @sravfeyn's point) we have unintended data exposure.

But now that we're on the same page, it sounds like we want to have a proactive way of pushing tables to superset instead of having to rely on a developer to add it whenever we realize a certain table is not there. 

I like @sravfeyn 's idea of having to explicitly think about superset whenever a new model is added to the codebase and triggering the sync as part of a migration, presumably the same migration adding the table to the connect DB. I'll do a short writeup of what this could look like.

Charl1996:
> I think we generally want every table to be in superset

See [this new spec](https://github.com/dimagi/commcare-connect/pull/1185/changes) addressing the larger requirement (making this one redundant).

FYI @sravfeyn @calellowitz

PR #1183, opened 2026-05-14
Sr/target population

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CCCT-2399

Adds support for setting `target_population` field on WorkArea during csv upload

## Technical Summary

Simply adds a field to existing upload code

### Safety story

Tested locally

### Automated test coverage

Tests modified
### QA Plan

NA

### Labels & Review


commcare_connect/microplanning/tasks.py
```diff
@@ -157,12 +159,15 @@ def get_slug(self, row):
         slug = (row.get(self.HEADERS.get("slug")) or "").strip()
         return strip_tags(slug)
 
+    def _get_int(self, row, key):
+        raw = row.get(self.HEADERS.get(key))
+        return int(raw) if raw else 0
```
ajeety4, line 164: will this cause 500 if the value is not integer and has text ?

ajeety4 (approved):
LGTM

Just a non blocking query

zandre-eng (approved):

commcare_connect/microplanning/tasks.py
sravfeyn: This gets validated inside `_validate_numbers`

sravfeyn:
@zandre-eng or @ajeety4 Thanks for the review, looking for a re-review. Should be super quick.

commcare_connect/microplanning/tasks.py
```diff
@@ -157,12 +159,15 @@ def get_slug(self, row):
...
+
     def get_building_and_visit(self, row):
-        building_raw = row.get(self.HEADERS.get("building_count"))
-        visit_raw = row.get(self.HEADERS.get("visit_count"))
-        building = int(building_raw) if building_raw else 0
-        visit = int(visit_raw) if visit_raw else 0
-        return building, visit
+        return self._get_int(row, "building_count"), self._get_int(row, "visit_count")
+
+    def get_target_population(self, row):
```
mkangia, line 169: nit: it is quite strange that there is a function that checks for both building and visit and then there is a separate function that checks for target population, and then all three are the same checks. But then the error message has all three mentioned even if one is off.

mkangia (approved):

PR #1184, opened 2026-05-14
Fix UI glitch in tasks page

sravfeyn:
## Product Description

Fixes a UI glitch. 
https://dimagi.atlassian.net/browse/CCCT-2393

(While the table is being fetched)

Screenshot 2026-05-14 at 3 11 58 PM

## Technical Summary

Solves the UI glitch using a table-loading skeleton.

### Safety story

Tested locally using slow connection simulator in Chrome Dev Tools 

### Automated test coverage

NA

### Labels & Review

### DIFF
--- a/commcare_connect/templates/opportunity/user_tasks.html
+++ b/commcare_connect/templates/opportunity/user_tasks.html
@@ -66,14 +66,36 @@
 
   <div class="w-full grid grid-cols-10 gap-2">
     <div class="col-span-6" hx-indicator="#loading-spinner">
-      <div class="w-full relative shadow-sm rounded-lg flex items-center justify-center"
-        style="min-height: 70vh"
+      <div class="w-full"
         hx-get="{% url 'opportunity:user_tasks_table' request.org.slug opportunity.opportunity_id %}{% querystring %}"
-        hx-trigger="load" hx-swap="innerHTML"
-        hx-on::after-swap="this.style.minHeight = ''; this.classList.remove('flex', 'items-center', 'justify-center')">
-        <div role="status">
-          <div class="animate-spin rounded-full h-12 w-12 border-4 border-brand-mango border-t-transparent"></div>
-          <span class="sr-only">{% trans "Loading..." %}</span>
+        hx-trigger="load" hx-swap="innerHTML">
+        <div class="table-container flex flex-col gap-2 mb-1">
+          <div class="w-full overflow-x-auto block">
+            <table class="base-table animate-pulse">
+              <thead>
+                <tr>
+                  <th><div class="h-3 w-4 rounded bg-gray-300"></div></th>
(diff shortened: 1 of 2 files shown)


commcare_connect/templates/opportunity/user_tasks.html
```diff
@@ -66,9 +66,38 @@ <h1>{% trans "Filters" %}</h1>
...
+                  <th><div class="h-3 w-4 rounded bg-gray-300"></div></th>
+                  <th><div class="h-3 w-28 rounded bg-gray-300"></div></th>
+                  <th><div class="h-3 w-24 rounded bg-gray-300"></div></th>
+                  <th><div class="h-3 w-24 rounded bg-gray-300"></div></th>
+                  <th><div class="h-3 w-20 rounded bg-gray-300"></div></th>
+                  <th><div class="h-3 w-16 rounded bg-gray-300"></div></th>
+                </tr>
+              </thead>
+              <tbody>
+                {% for _ in "xxx" %}
```
mkangia, line 86: what is "xxx"?

mkangia (commented):
(non blocking) I would suggest if the commit messages could be more specific about what is being changed from code perspective, for example: "Fix UI glitch in tasks page" could be instead be "swap innerHTML instead of outerHTML" or a detailed commit message that has the title "improve loading behavior" and then with specific details in the commit message listed below.

zandre-eng (approved):
Overall looks good. Does this also address the issue where the "Create Task" and "Delete Tasks" buttons are visible for a split second before disappearing if the user doesn't have permission? I think that might have also been part of the user confusion. Also, could you please add the ticket link to the PR description?

sravfeyn:
@zandre-eng 
Thanks, I have linked the ticket.

Yes this addresses the described issue by clearly displaying a table-loading animation and with buttons right below it that move when the table expands. Previously there was no loading indicator, and even after the table loaded there was a large empty space between table and the buttons, making buttons appear far below which made users think they disappeared. 

@mkangia 
The first commit (used a different approach), and if you observe it gets removed in the second commit. The second commit simply adds a standard tailwind table loading animation, which I mentioned in the commit. The `innerHtml` and `outerHtml` are a pretty foundational htmx concepts that span many commits, so that is not the focus to be described in the commit.

commcare_connect/templates/opportunity/user_tasks.html
sravfeyn: This is a simple way to iterate 3 times.

sravfeyn:
I will wait for @mkangia's approval as well

mkangia:
Thanks for the reasoning @sravfeyn 

> The second commit simply adds a standard tailwind table loading animation, which I mentioned in the commit

My suggestion was for the first commit message, the second one was good.

mkangia (approved):

PR #1185, opened 2026-05-14
Auto-discovered logical replication: CI-enforced table set + split bootstrap/refresh lifecycle

Charl1996:
Ticket: https://dimagi.atlassian.net/browse/CCCT-2094
Supersedes the approach in https://github.com/dimagi/commcare-connect/pull/1179.

## Product Description

N/A — internal replication plumbing for the Superset secondary DB. No user-facing change.

## Technical Summary

Replaces the hand-maintained `REPLICATION_ALLOWED_MODELS` allowlist (a developer had to edit it and re-run an interactive command on every model change) with auto-discovery + a split lifecycle. Design and rationale: `docs/replication-redesign.md` and `docs/replication-redesign-recommendation.md` (full decision write-up, D1/D2/D3); implementation plan in `docs/superpowers/plans/2026-06-12-auto-discovered-logical-replication.md`.

Three pieces:
- **Twin lists** (`REPLICATION_INCLUDED_MODELS` / `REPLICATION_EXCLUDED_MODELS` in `multidb/constants.py`). Every local app model must be in exactly one — a pytest (`test_no_unclassified_local_models`) fails in CI otherwise, so adding a model forces a deliberate replicate/don't-replicate decision. pghistory event models and m2m through tables are auto-excluded; third-party tables are never replicated.
- **`refresh_replication`** management command (non-interactive), wired into `docker/start_migrate` after `migrate_multi` — the only race-safe trigger, since `migrate_multi` migrates both DBs in parallel and the publication SQL must run after both schemas exist. It reconciles the publication on the primary and `REFRESH`es the subscription on the secondary using only the connections Django already has — no superuser.
- **Bootstrap** (`setup_logical_replication`) now transfers subscription ownership to the app secondary role, which is what lets the recurring refresh run without a superuser. `logical_replication_status` likewise drops its superuser prompt and uses Django's connections.

### Safety story

Gated behind `REPLICATION_ENABLED` (default **False**), so this is a no-op in every environment until the flag is flipped — merging changes nothing in production today. `refresh_replication` exits cleanly (0) when replication is inactive or when the publication doesn't yet exist, so it cannot break the deploy's `set -o errexit` in `start_migrate`.

**Out of scope / deploy ordering:** the no-superuser refresh path is only *legal* on **Postgres 16+** (non-superuser subscription ownership). The production secondary currently runs PG15, so `REPLICATION_ENABLED` must stay False until the secondary is upgraded via blue/green (see the "Postgres version prerequisite" section in the recommendation doc). `ALTER SUBSCRIPTION ... REFRESH PUBLICATION` must run outside a transaction — relies on management commands running in autocommit and the secondary alias having no `ATOMIC_REQUESTS`; noted in a code comment on `refresh_subscription`.

### Automated test coverage

19 tests in `commcare_connect/multidb/tests/`: the classification guard (proves the lists are complete and exempts events/m2m), gating (`replication_is_active`), bootstrap-state and in-sync detection, the publication/subscription refresh SQL, and the bootstrap ownership transfer.

### QA Plan

No QA needed for merge (feature disabled). Real validation happens against the PG16 secondary during the blue/green upgrade: bootstrap once, confirm `refresh_replication` reconciles the table set and Superset sees new/removed tables.

### Labels & Review

---

Note: branch is well behind `main` (based on an older commit) — likely wants a rebase before merge.

### DIFF
--- a/docs/replication-redesign.md
+++ b/docs/replication-redesign.md
@@ -329,6 +329,41 @@ anywhere in the recurring path.
   and runs bootstrap; documenting "you must run bootstrap once" is
   a process risk. 
 
+#### ⚠️ Hard prerequisite — Postgres ≥ 16 on the secondary
+
+D3.B's "no superuser in the recurring path" property depends entirely on
+the secondary's subscription being **owned by an ordinary (non-superuser)
+role** — the app's secondary role that `migrate_multi` already connects
+as. However, that ownership is only legal on **Postgres 16+**; earlier
+versions do not allow non-superusers to own subscriptions. The PG15
+[`ALTER SUBSCRIPTION`](https://www.postgresql.org/docs/15/sql-altersubscription.html)
+reference says so directly in its Description:
+
+> The new owner has to be a superuser. (Currently, all subscription owners
+> must be superusers, so the owner checks will be bypassed in practice. But
+> this might change in the future.)
+
+So `ALTER SUBSCRIPTION ... OWNER TO <app_role>` cannot hand the subscription
+to an ordinary role on PG15, and `REFRESH PUBLICATION` (which requires
+ownership) stays superuser-only. PG16 is the "might change in the future"
+above: it added the
+[`pg_create_subscription`](https://www.postgresql.org/docs/16/predefined-roles.html)
+predefined role and non-superuser subscription ownership.
+
+The primary/publication leg (`ALTER PUBLICATION ... SET TABLE` via
+ownership transfer) works on any version; only the subscriber leg is
+gated by this.
+
+**Current state:** our **production secondary DB runs Postgres 15**, so
+D3.B is *not* implementable as-is today.
+
+**Plan to unblock:** upgrade the secondary to Postgres 16+ via a
+**blue/green** approach — stand up the PG16 instance in parallel,
+replicate/validate against it, cut over once it's confirmed working, and
+decommission the old PG15 instance. D3.B can only be rolled out *after*
+that cutover completes; until then the recurring subscriber refresh still
+requires a superuser (D3.A behaviour).
+
 ---
 
 ## Recommendation
@@ -350,7 +385,10 @@ The following options are recommended:
   recurring path**. The high-privilege secondary superuser credential is
   typed by an operator once at bootstrap and never persisted. The cost is
   two code paths to maintain and a documented "run bootstrap once per
-  environment" step.
+  environment" step. **Prerequisite:** the secondary must run Postgres
+  16+; our production secondary is currently Postgres 15, so this requires
+  a blue/green upgrade of the secondary first (see the D3.B prerequisite
+  note above).
 
 ### Out of scope
 
--- a/docs/replication-redesign.md
+++ b/docs/replication-redesign.md
@@ -234,6 +234,14 @@ def is_publication_in_sync() -> bool:
     return desired == actual
 ```
 
+#### Gating: an env flag for whether replication runs on deploy
+
+Since not all environments have replication set up we should gate when
+this happens. We can thus add a `REPLICATION_ENABLED = env.bool("REPLICATION_ENABLED",
+default=False)` in `base.py` — set to `True` only on the environment(s)
+that should replicate (via the per-environment `docker.env`). The refresh
+command checks it first and skips as a no-op when it's `False`.
+
 ---
 
 ### D3. How are credentials handled in a non-interactive context?
@@ -348,6 +356,7 @@ The following options are recommended:
 1. Developer adds a new model in the code.
 2. A Django system check (run by `manage.py check` locally and in CI) fails unless the new model is specified in either `REPLICATION_INCLUDED_MODELS` or `REPLICATION_EXCLUDED_MODELS`.
 3. On deploy, after `migrate_multi` has executed, a new command will be run which
+   - skips as a no-op if `REPLICATION_ENABLED` is `False`
    - checks to see if the publication is in sync with the codebase
    - if so: exit (noop)
    - if not: invoke `setup_logical_replication` command (which obtains credentials from django settings)
--- a/docs/replication-redesign.md
+++ b/docs/replication-redesign.md
@@ -237,10 +237,8 @@ def is_publication_in_sync() -> bool:
 #### Gating: an env flag for whether replication runs on deploy
 
 Since not all environments have replication set up we should gate when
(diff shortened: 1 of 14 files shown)


docs/replication-redesign.md
```diff
@@ -0,0 +1,355 @@
...
+REPLICATION_EXCLUDED_MODELS = [
+    SomePIIModel,       # PII
+    SomeSecretsModel,   # secrets
+]
+
+# Enforced via a Django system check (runs on `manage.py check`, in CI,
+# and on `runserver`):
+unclassified = set(apps.get_models()) - set(INCLUDED) - set(EXCLUDED)
+if unclassified:
+    raise SystemCheckError(unclassified)
```
calellowitz, line 77: what are the pros and cons of this approach vs a test like
```
def test_superset_replication():
    unclassified = set(apps.get_models()) - set(INCLUDED) - set(EXCLUDED)
    assert len(unclassified) == 0
```

docs/replication-redesign.md
```diff
@@ -0,0 +1,355 @@
...
+---
+
+## Recommendation
+The following options are recommended:
+
+- D1.B: Maintaining two lists makes it trivial to enforce explicit thinking about which models should be replicated and which ones not.
+- D2: deploy-pipeline command, sequenced after `migrate_multi` returns
+  (the only safe trigger — see D2 background on the ordering constraint
+  imposed by logical replication)
+- D3.A: Since we already have the command written, swapping the inputs for reading the necessary credentials should be minimum effort while accepting the trade-off that the secondary superuser password (and primary replication user password) stay in env on every deploy host, rather than doing the ownership-transfer work D3.B would require.
```
calellowitz: I agree up to this point, I think if there are decent ways to keep postgres superuser credentials out of the env, we should strive to do that, and I am ok with bootstrap requiring a one time extra option. Curious if you think that is being overly cautious

calellowitz (commented):
generally looks good, left a few questions/comments

docs/replication-redesign.md
Charl1996: The only benefit of doing the check as part of django check (and thus when the developer runs `runserver`) is that the fact that a model is not listed will be surfaced way earlier. The developer is forced to think about replication as soon as adding the model instead of only when running the tests. 

Practically I don't think it matters where we do this check, as long as it happens before the PR is merged. A test could work just as good.

docs/replication-redesign.md
Charl1996: > Curious if you think that is being overly cautious

No, I think generally extra caution is preferred. My main motivation for D3.A was its simplicity, but truth be told, after a night's sleep I'm starting to lean back towards D3.B again as a long term solution. 

If we decide to go with D3.B:
The bootstrap is technically already done if we decide to not transfer ownership of the `tables_for_superset_pub` and `tables_for_superset_sub` objects over to the primary DB role / secondary DB roles (the one used for running `migrate_multi`). If we decide it's fine to transfer ownership of these pub/sub object to these primary/secondary roles it's a once-off developer action (and it would allow us to not have to add any new credentials anywhere).

sravfeyn: I also like if there is a way that they can be stored outside of env, but I doubt if the separate command works without those credentials. If you have tested locally and it works, then great! Also need to consider that all queries will continue to work correctly on superset as well.

docs/replication-redesign.md
```diff
@@ -0,0 +1,355 @@
...
+table set already matches the desired set, and exit early when it does.
+Postgres exposes the live publication membership via
+`pg_publication_tables`:
+
+```python
+from django.apps import apps
+from django.db import connection
+from commcare_connect.multidb.constants import PUBLICATION_NAME
+
+def is_publication_in_sync() -> bool:
```
zandre-eng, line 226: Would we also need a check here to validate we're on the correct env to allow publishing (e.g. we don't want this to run for staging)?

zandre-eng (approved):
I'm in favour of D1.B + D2 + D3.B. For D1.B the exhaustive classification will prevent silent inclusions and I think the risk of merge conflicts on the file where this const resides should be relatively low. I don't think model additions are frequent enough to make this a real pain point in practice.

For D3 I have to agree with the other comments in that persisting the secondary superuser password on every deploy host can widen the risk range for attacks, and D3.B sounds like a better approach if we're happy with transferring ownership.

docs/replication-redesign.md
```diff
@@ -0,0 +1,355 @@
...
+  default DB; the secondary is populated solely via Postgres logical
+  replication and consumed by Superset.
+
+## Decision dimensions
+
+The following 3 decision dimensions will be discussed:
+1. Where we define which models should be included/excluded from replication
+2. When we trigger the update to the replication object in postgres to add/remove tables
+3. Where we read the necessary credentials from   
+
```
hemant10yadav, line 33: Is there a reason we haven't considered the option of introducing a `ReplicatedModel` model in the database and managing replication membership there?

We could store references to models (potentially via `ContentType`) and manage replication membership through Django Admin. When a new table needs to be replicated, a dev could simply add the corresponding model to `ReplicatedModel` in Django Admin and run the `setup_logical_replication` command, without requiring an additional code change or deployment solely to update the replication configuration.

Am I missing any significant drawbacks with this approach compared to the proposed above?

docs/replication-redesign.md
Charl1996: Good point. I'll add it

docs/replication-redesign.md
Charl1996: I did initially think about maintaining a DB model for storing this info, but the issue is that it's still a manual task to 
1. add the record after deploy
2. trigger the sync after the deploy

Both of the above carry the risk of the developer forgetting to do this and it will eventually land on the interrupt queue. 

That said, both of the items above can be automated which would look similar to the solution proposed in the spec. The real difference then is just where/how we define the models that should be replicated.

docs/replication-redesign.md
Charl1996: https://github.com/dimagi/commcare-connect/pull/1185/commits/8f42bbbd6f28b0e17952617ce88bee3d8d2a2eef

calellowitz (approved):
I only read the high level doc, but agree with the suggestions there, including the upgrade. I have a slight preference for doing the CI check in a test vs the SystemCheck, beacuse that is the way my brain works, but am fine with either if you have a preference in the other direction.

Charl1996:
>  CI check in a test vs the SystemCheck

I don't have a strong preference, so let's do CI test

PR #1186, opened 2026-05-14
Add name search to Opportunities list

mmmary314:
## Summary

- Adds an inline search bar to the Opportunities list page header
- Filters opportunities by name (case-insensitive `icontains`) on form submit
- Search input shares the same `<form>` as the filter modal so search and filters compose together
- Search is excluded from the filter badge count (the sliders icon)
Screenshot 2026-05-14 at 3 35 58 PM

Jira: [CCCT-2420](https://dimagi.atlassian.net/browse/CCCT-2420)

## Test plan

- [ ] Search for a partial opportunity name — matching rows appear, others disappear
- [ ] Applying a filter while a search is active preserves the search term
- [ ] Searching while filters are active preserves the filter values
- [ ] Clearing the search field and submitting restores the full list
- [ ] `filters_applied_count` badge does not increment when only search is set

[CCCT-2420]: https://dimagi.atlassian.net/browse/CCCT-2420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -157,6 +157,7 @@ class OpportunityListFilterSet(django_filters.FilterSet):
             else:
                 del self.filters["program"]
 
+        self.form.helper.form_tag = False
         self.form.helper.layout = Layout(*modal_fields)
 
 
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -127,6 +127,7 @@ class DeliverFilterSet(django_filters.FilterSet):
 
 
 class OpportunityListFilterSet(django_filters.FilterSet):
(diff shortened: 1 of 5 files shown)


calellowitz (commented):
how does this search get triggered? The code looks like it relies on a button press but I don't see a button anywhere in the screenshots

mmmary314:
> how does this search get triggered? The code looks like it relies on a button press but I don't see a button anywhere in the screenshots

It should work the same as on the Connect Workers page, which searches in real time as the user enters info.

calellowitz:
Offlined with Mary. This PR does not do what it was intended to do, so closing

PR #1187, opened 2026-05-14
Fix work area assignment

mkangia:
## Product Description

Work areas should be assigned to the mobile worker and not its usercase. This PR fixes that.

## Technical Summary

We didn't have the mobile workers ID available so that was added to the ConnectIDUserLink.
Then a utility was added to get the ID from HQ. However, HQ currently does not support filtering by username so an alternate was to get all mobile workers and then filter here in Connect. This is not the ideal solution but for two reasons this was considered for now. Firstly, it is needed only if ConnectIDUserLink does not already have it. A link should always be present and we will back fill the new column added. Secondly the number of mobile workers is expected to be low in project spaces so this should still have a fast turnaround.
And then simply the work area assignment was updated.

### Safety story

Tests added.

### Labels & Review


pxwxnvermx (approved):
Changes look good to me. Just a couple of queries.

1. What is the reason for going with the User API instead of the Case Data API to get the user `id` / `owner_id`?
2. How do you feel about a pre-fill command that runs for every active opportunity and fills the hq_user_uuid from HQ? This will allow us to backfill already existing users and avoid doing `fetch_hq_user_uuid`.

zandre-eng (approved):

mkangia:
Hey @pxwxnvermx 

> What is the reason for going with the User API instead of the Case Data API to get the user id / owner_id?

we need the user ID of the commcare-user. So we would get that from the User API right?

> How do you feel about a pre-fill command that runs for every active opportunity and fills the hq_user_uuid from HQ? This will allow us to backfill already existing users and avoid doing fetch_hq_user_uuid.

Absolutely, we will have that as well. It will be done in a follow up PR. I will confirm now that its mentioned on the ticket

pxwxnvermx:
@mkangia 

> we need the user ID of the commcare-user. So we would get that from the User API right?

The user ID is also available on the usercase under the `external_id`, `hq_user_id` or `owner_id` properties. Are there any concerns regarding using the data present on the usercase?

mkangia:
Thanks @pxwxnvermx 
we can surely consider that once the usercase lookup is fixed, though I do feel using the user's API is more ideal way to do it. We can check with Cal for his guidance on this.
The fix we have [for now](https://github.com/dimagi/commcare-connect/pull/1189) seems like a solid fix.

zandre-eng (approved):

commcare_connect/commcarehq/api.py
```diff
@@ -90,6 +86,24 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
     return case
 
 
+def _get_hq_user_uuid_for_owner(user, domain, api_key):
```
sravfeyn: Is it not possible to avoid this extra method that is pretty much same as `_resolve_hq_user_uuid`?

commcare_connect/commcarehq/api.py
```diff
@@ -90,6 +86,24 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
...
+    link = ConnectIDUserLink.objects.filter(commcare_username=user.username.lower()).first()
+    return _resolve_hq_user_uuid(user, domain, api_key, link)
+
+
+def _resolve_hq_user_uuid(user, domain, api_key, link):
```
sravfeyn: Is this being called multiple times for same user? If so, might be worth memoizing it?

pxwxnvermx: This would be called only once per user per opportunity and the return is getting saved on the Link so it is getting memoized to the database.

commcare_connect/commcarehq/api.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1187/commits/260b82115f3583a758ea9a1009ac65443cc4bcec

sravfeyn (approved):

zandre-eng (approved):

commcare_connect/users/helpers.py
```diff
@@ -24,14 +24,42 @@ def create_hq_user_and_link(user, domain, opportunity):
...
+            hq_server=hq_server,
+            hq_user_uuid=hq_user_uuid,
+        )
     return True
 
 
+def fetch_hq_user_uuid(commcare_username, domain, api_key):
+    hq_username = f"{commcare_username.lower()}@{domain}.commcarehq.org"
+    headers = {"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"}
+    next_url = f"{api_key.hq_server.url}/a/{domain}/api/v0.5/user/?limit=1000"
```
mkangia: nit: I didn't think this part through, but do we want to keep this limit to a smaller number like 200 although we are assuming a small user base so a 1000 would actually help get all users at once.

commcare_connect/commcarehq/api.py
```diff
@@ -87,7 +87,7 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
 
 
 def _resolve_hq_user_uuid(user, domain, api_key):
-    link = ConnectIDUserLink.objects.filter(commcare_username=user.username.lower()).first()
+    link = ConnectIDUserLink.objects.filter(commcare_username__startswith=user.username.lower()).first()
```
mkangia: is this okay? won't this have unwanted matches?

pxwxnvermx: It would not have unwanted matches as we use the same usernames for HQ and Connect, so startswith matches with username part, I can make it more strict using the domain and hq_server to construct the hq username.

commcare_connect/users/helpers.py
pxwxnvermx: yeah, 200 sounds much more reasonable. I will make that change.

commcare_connect/commcarehq/api.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1187/commits/ba0dc48e38034ed2862d1d8a00c70e0c4c7aa03e

commcare_connect/users/helpers.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1187/commits/cc32b4e887567fbed2ccbb64e0eab7224d9b88f8

commcare_connect/commcarehq/api.py
```diff
@@ -90,6 +86,20 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
     return case
 
 
+def _resolve_hq_user_uuid(user, domain, api_key):
+    commcare_username = f"{user.username.lower()}@{domain}.commcarehq.org"
+    link = ConnectIDUserLink.objects.filter(commcare_username=commcare_username).first()
+    if link and link.hq_user_uuid:
+        return link.hq_user_uuid
+    hq_user_uuid = fetch_hq_user_uuid(user.username, domain, api_key)
```
mkangia: nit: looks like we are creating the full username again in fetch_hq_user_uuid which can be avoided.

mkangia (commented):
thanks for the improvements. Just one non blocking refactor but I can't approve this anyway since I co-authored.

commcare_connect/commcarehq/api.py
```diff
@@ -90,6 +86,20 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
     return case
 
 
+def _resolve_hq_user_uuid(user, domain, api_key):
+    commcare_username = f"{user.username.lower()}@{domain}.commcarehq.org"
+    link = ConnectIDUserLink.objects.filter(commcare_username=commcare_username).first()
```
calellowitz: This can just fetch by the `user` and `domain` which will uniquely identify a user link and not require recreating the username. It should also use `.get` since you are expecting exactly one result, and `.first()` will result in potentially incorrect behavior if there are zero or multiple returned.

commcare_connect/commcarehq/api.py
```diff
@@ -90,6 +86,20 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
...
+    if link and link.hq_user_uuid:
+        return link.hq_user_uuid
+    hq_user_uuid = fetch_hq_user_uuid(user.username, domain, api_key)
+    if hq_user_uuid is None:
+        raise CommCareHQAPIException(f"Failed to find HQ user for {user.username.lower()} on {domain} HQ domain.")
+    if link:
```
calellowitz: it seems like a substantial issue if there isn't a link here because we are attaching the cases to a possibly unrelated mobile user. If there isn't a link I think we need to fail hard because it means we never made a mobile user for that person in commcare.

commcare_connect/users/helpers.py
```diff
@@ -24,14 +24,42 @@ def create_hq_user_and_link(user, domain, opportunity):
...
+            user=user,
+            domain=domain,
+            hq_server=hq_server,
+            hq_user_uuid=hq_user_uuid,
+        )
     return True
 
 
+def fetch_hq_user_uuid(commcare_username, domain, api_key):
+    hq_username = f"{commcare_username.lower()}@{domain}.commcarehq.org"
```
calellowitz: we should pass in the user link here and get the username from there

commcare_connect/users/helpers.py
```diff
@@ -24,14 +24,42 @@ def create_hq_user_and_link(user, domain, opportunity):
...
+    headers = {"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"}
+    next_url = f"{api_key.hq_server.url}/a/{domain}/api/v0.5/user/?limit=200"
```
calellowitz, line 47: I think this chunk can be a lot bigger (at least 1k), to speed this up.

calellowitz (approved):
We spoke offline about this, but I think given the small number of users we have, it would be much better to do a onetime migration of the data to populate this field all at once, rather than lazily migrating when we run into empty ones, and then we can make the field not nullable and guarantee we always have it. Otherwise our data ends up in an indeterminate state, we can't enforce things at the db level, and random actions may sometimes take far longer than usual because the system needs to make an API call to populate a missing field.

I am tentatively approving because I know there is a deadline to release this, but I think it is important to follow up on some of the feedback and I would otherwise leave it as blocking.

commcare_connect/users/helpers.py
pxwxnvermx: I made the chunk smaller yesterday after a small discussion on another comment on this PR regarding the older 1K limit being too big.

@mkangia How do you feel about making this 1k again?

sravfeyn (approved):

commcare_connect/commcarehq/api.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1187/commits/cc4a8f1e93b9d53113a88c7f80e42705a6b63b73

commcare_connect/commcarehq/api.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1187/commits/cc4a8f1e93b9d53113a88c7f80e42705a6b63b73

commcare_connect/users/helpers.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1187/commits/fd38cd448e6366b1a0f6b35defab7e97c9bea458

commcare_connect/users/helpers.py
mkangia: fine with 1k. 
Like I said in https://github.com/dimagi/commcare-connect/pull/1187#discussion_r3317832711, 1k was also a good number.

PR #1188, opened 2026-05-14
Fix usercase filter

mkangia:
## Product Description

Just a bugfix for filter used to fetch usercase from HQ for a mobile worker.

## Technical Summary

`case_name` stopped working after Connect push the user's username as the first name for the mobile worker when sending additional fields during user creation. `case_name` also seemed unreliable.
I found that HQ sets a case property called username on the usercase and keeps that in sync with the user's username, so that was more reliable to fetch the usercase.

https://github.com/dimagi/commcare-connect/pull/1187 now adds the user's HQ ID on the link, so that can also be used to fetch the commcare-user case assigned to the mobile worker but the change in this PR seems solid as well.

### Safety story

Testing on staging for integration but otherwise tests added.

### Labels & Review


mkangia:
Accidentally got staging here.

PR #1191, opened 2026-05-15
Fix Worker Tasks Filters

zandre-eng:
## Product Description

Several bugs in the worker tasks tab have been fixed. The **Task Type** filter dropdown now correctly populates with the task types for the opportunity. The **Date Assigned** filter inputs no longer accept future dates, since assignment dates are always in the past. When closing the filter modal without applying, any in-progress filter changes are discarded and the filter fields revert to their last applied state — so unapplied edits never persist across modal opens.

Screenshot_20260515_144907

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/QA-8509)

Three bugs in the tasks feature are fixed in this PR. The root cause of the task type filter being empty was that `TaskType.objects.filter(opportunity=...)` was used to build filter choices, but `TaskType.opportunity` is a nullable FK that is never set — task types are linked to an opportunity via `app = opportunity.deliver_app`. The filter queries are corrected to use `app`. `AddTaskTypeForm.save()` is also updated to populate the `opportunity` FK going forward so both paths are consistent. The modal close behaviour is handled client-side: on close, each filter field reads the current URL params and restores its applied value (or clears if absent), so Tom Select and date inputs stay in sync with what's actually active.

- **Filter choices fix** — `TasksFilterSet` and `UserTasksFilterSet` both queried `TaskType` via the nullable `opportunity` FK; corrected to `app=opportunity.deliver_app`, matching how every other view looks up task types for an opportunity.
- **`AddTaskTypeForm.save()`** — now sets `task_type.opportunity = self.opportunity` so newly created task types have the FK populated.
- **Date inputs capped at today** — both filter sets set `max = date.today().isoformat()` on `date_assigned_from` / `date_assigned_to` widgets in `__init__`, preventing future date selection in the browser picker. `date_assigned` is an annotation of `auto_now_add` so it can never be in the future.
- **Filter modal close** — replaces the previous "reset everything if nothing applied" logic with a per-field restore: on close, `URLSearchParams` is read and each date input / Tom Select is set to its applied value, discarding any unapplied edits while preserving active filters.

### Safety story

All changes are read-path or form-save additions with no destructive data operations. The `opportunity` FK addition to `AddTaskTypeForm.save()` is additive and backward-compatible — existing task types with `opportunity=null` continue to work and are still found via the `app`-based filter. The date `max` constraint is browser-side only and does not add server-side validation (the filter accepts any valid date regardless). The modal close reset is entirely client-side with no server interaction.

Local testing was also performed on the changes in this PR.

### Automated test coverage

- **`test_tasks_filterset_task_type`** — verifies filter choices are populated and filtering works when task types are linked only via `app` (no `opportunity` FK set)
- **`test_tasks_filterset_task_type_excludes_inactive`** — verifies inactive task types are excluded from choices
- **`TestUserTasksFilterSet::test_task_type`** — same coverage for `UserTasksFilterSet`; factories updated to omit `opportunity=` to reflect real data
- **`test_filter_worker_tasks_by_task_type`** — helper-level test updated to use app-only task types
- **`TestAddTaskTypeForm::test_save_sets_slug_and_app`** — extended with `assert task_type.opportunity == opportunity`

### QA Plan

- [ ] Navigate to an opportunity's worker tasks tab
- [ ] Open the filter modal — confirm the **Task Type** dropdown is populated with the opportunity's task types
- [ ] Select a task type and confirm the table filters correctly
- [ ] Open the filter modal — confirm the **Date Assigned From / To** pickers do not allow selecting a future date
- [ ] Enter a filter value without clicking Apply, then close the modal — confirm the field reverts to the previously applied value (or clears if nothing was applied)
- [ ] Apply a filter, reopen the modal, change a value, then close without applying — confirm the applied filter remains active and the changed value is discarded

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -209,7 +209,7 @@ class TasksFilterSet(django_filters.FilterSet):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
         if self.opportunity:
-            active_tasks = TaskType.objects.filter(opportunity=self.opportunity, is_active=True)
+            active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
             self.filters["task_type"].extra["choices"] = [(str(t.pk), t.name) for t in active_tasks]
 
             worker_queryset = (
@@ -270,7 +270,7 @@ class UserTasksFilterSet(django_filters.FilterSet):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
         if self.opportunity:
-            active_tasks = TaskType.objects.filter(opportunity=self.opportunity, is_active=True)
+            active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
             self.filters["task_type"].extra["choices"] = [(str(t.pk), t.name) for t in active_tasks]
         self.form.helper.layout = Layout(
             "task_status",
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -373,6 +373,9 @@ class AssignedTaskFilterSet(django_filters.FilterSet):
     def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
+        today = date.today().isoformat()
+        self.filters["date_assigned_after"].extra["widget"].attrs["max"] = today
+        self.filters["date_assigned_before"].extra["widget"].attrs["max"] = today
         if self.opportunity:
             self.filters["task_type"].extra["choices"] = self._get_task_type_choices()
             self.filters["worker_name"].extra["choices"] = self._get_worker_name_choices()
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -214,7 +214,7 @@ class TasksFilterSet(django_filters.FilterSet):
         self.filters["date_assigned_from"].extra["widget"].attrs["max"] = today
         self.filters["date_assigned_to"].extra["widget"].attrs["max"] = today
         if self.opportunity:
-            active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
+            active_tasks = TaskType.objects.filter(opportunity=self.opportunity, is_active=True)
             self.filters["task_type"].extra["choices"] = [(str(t.pk), t.name) for t in active_tasks]
 
             worker_queryset = (
@@ -278,7 +278,7 @@ class UserTasksFilterSet(django_filters.FilterSet):
         self.filters["date_assigned_from"].extra["widget"].attrs["max"] = today
         self.filters["date_assigned_to"].extra["widget"].attrs["max"] = today
         if self.opportunity:
-            active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
+            active_tasks = TaskType.objects.filter(opportunity=self.opportunity, is_active=True)
             self.filters["task_type"].extra["choices"] = [(str(t.pk), t.name) for t in active_tasks]
         self.form.helper.layout = Layout(
             "task_status",
--- a/commcare_connect/opportunity/filters.py
+++ b/commcare_connect/opportunity/filters.py
@@ -1,3 +1,5 @@
+from datetime import date
+
 import django_filters
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import HTML, Column, Div, Layout, Row
@@ -208,6 +210,9 @@ class TasksFilterSet(django_filters.FilterSet):
     def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
+        today = date.today().isoformat()
+        self.filters["date_assigned_from"].extra["widget"].attrs["max"] = today
+        self.filters["date_assigned_to"].extra["widget"].attrs["max"] = today
         if self.opportunity:
             active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
             self.filters["task_type"].extra["choices"] = [(str(t.pk), t.name) for t in active_tasks]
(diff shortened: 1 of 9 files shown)


commcare_connect/opportunity/forms.py
```diff
@@ -1980,6 +1980,7 @@ def _populate_task_unit_choices(self):
     def save(self, commit=True):
         task_type = super().save(commit=False)
         task_type.app = self.opportunity.deliver_app
+        task_type.opportunity = self.opportunity
```
Charl1996, line 1983: What issue does this solve? Its odd that the opp wasn't persisted before this PR.

commcare_connect/opportunity/filters.py
```diff
@@ -269,8 +274,11 @@ class Meta:
     def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
+        today = date.today().isoformat()
+        self.filters["date_assigned_from"].extra["widget"].attrs["max"] = today
+        self.filters["date_assigned_to"].extra["widget"].attrs["max"] = today
         if self.opportunity:
-            active_tasks = TaskType.objects.filter(opportunity=self.opportunity, is_active=True)
+            active_tasks = TaskType.objects.filter(app=self.opportunity.deliver_app, is_active=True)
```
Charl1996: Why was this change made? Should we also keep the opportunity filter here in addition to the app clause?

Charl1996:
Looks good - just [one blocking](https://github.com/dimagi/commcare-connect/pull/1191#discussion_r3257759795) item.

commcare_connect/opportunity/filters.py
mkangia: ya, I am also not sure why opportunity won't be present. I believe all task types should have an opportunity, no?

commcare_connect/opportunity/filters.py
```diff
@@ -269,6 +274,9 @@ class Meta:
     def __init__(self, *args, **kwargs):
         self.opportunity = kwargs.pop("opportunity", None)
         super().__init__(*args, **kwargs)
+        today = date.today().isoformat()
+        self.filters["date_assigned_from"].extra["widget"].attrs["max"] = today
+        self.filters["date_assigned_to"].extra["widget"].attrs["max"] = today
```
mkangia, line 279: 👍

commcare_connect/opportunity/forms.py
mkangia, line 1983: 😿

commcare_connect/templates/opportunity/user_tasks.html
```diff
@@ -12,6 +12,20 @@
        showEditTaskModal: false,
        showFilterModal: false,
        editTaskError: false,
+       closeFilter() {
```
mkangia: nit: is this something we have at multiple places? I wonder if this should exist as a common utility.

mkangia (commented):
Thanks for this @zandre-eng 

If I am understanding this correctly, the opportunity was not being saved and hence we had to change the filter to deliver app since that was present.
Can we instead retain the opportunity filter but backfill the values? I believe this is only needed for test data on staging.
The reason to use deliver app is missing data, which does not seem like the right reason for the change.

All other changes look good.

commcare_connect/opportunity/forms.py
zandre-eng: This solves the issue where we never saved the opportunity to the `TaskType` instance. It is a bit odd that it was never persisted, but I suspect it was perhaps a timing issue with multiple folks working on this feature.

commcare_connect/opportunity/filters.py
zandre-eng: I made this change considering that `TaskType.opportunity` is a nullable field. I switched to app since it's a non-nullable field and is linked to the opportunity as well.
With the changes in this PR we should now always have an opportunity link, so perhaps I was being too defensive here.

Will revert the above to filter on opp instead.

zandre-eng: Reverted in [72894b3](https://github.com/dimagi/commcare-connect/pull/1191/commits/72894b33de9298a04a9c1fa07a91af0795e526f3).

zandre-eng:
> If I am understanding this correctly, the opportunity was not being saved and hence we had to change the filter to deliver app since that was present.
Can we instead retain the opportunity filter but backfill the values? I believe this is only needed for test data on staging.
The reason to use deliver app is missing data, which does not seem like the right reason for the change.

@mkangia That is partially true. I did change the filter to accommodate the existing data we have, but I think I was also trying to be too defensive here. `TaskType.opportunity` is a nullable field on the model, so I switched to filtering on the app because it is non-null and we will always have that. But with the changes in this PR we should always have an opp link. I've reverted to use the opp filter instead and will backfill the missing data.

commcare_connect/opportunity/tests/test_filters.py
```diff
@@ -56,8 +56,8 @@ def test_tasks_filterset_task_status_single():
 def test_tasks_filterset_task_type():
     opp = OpportunityFactory()
     access = OpportunityAccessFactory(opportunity=opp, accepted=True)
-    task_a = TaskTypeFactory(opportunity=opp, app=opp.deliver_app, is_active=True, name="Survey")
-    task_b = TaskTypeFactory(opportunity=opp, app=opp.deliver_app, is_active=True, name="Follow-up")
+    task_a = TaskTypeFactory(app=opp.deliver_app, is_active=True, name="Survey")
+    task_b = TaskTypeFactory(app=opp.deliver_app, is_active=True, name="Follow-up")
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [e5d6d26](https://github.com/dimagi/commcare-connect/pull/1191/commits/e5d6d2618e34d9d00b24d17cbf77e9cb5c93e760).

commcare_connect/templates/opportunity/user_tasks.html
zandre-eng: Looking at the code this seems to be something we can reuse for the other filter modals we have. I've extracted this out into a utility function that can be reused. I won't implement using this util in other filter modals in this PR to keep it in scope, but this can be followed up in a separate PR.

Addressed in [56b3c71](https://github.com/dimagi/commcare-connect/pull/1191/commits/56b3c71fc01364a18ca5a7d33b3aa8be25aeee00).

Charl1996 (approved):
Fine from my side.

pxwxnvermx (approved):

hemant10yadav (approved):

PR #1202, opened 2026-05-20
Group task

hemant10yadav:
## Product Description

<video src="https://github.com/user-attachments/assets/5732378b-35d8-4a6c-8c06-21332c358bab"></video>

## Technical Summary

**https://dimagi.atlassian.net/browse/CCCT-2406 - **RP-3****
[Connect-tech-demo link](https://dimagi.slack.com/archives/C09H7D1V1NK/p1779202353685859)

A new reusable `GroupedTable` base class and companion `GroupedTableData` were added to `commcare_connect/utils/tables.py`. The new implementation groups queryset rows by `pk`, paginates based on distinct group headers instead of individual rows, and builds each page into grouped head rows with their corresponding sub-rows. A new `grouped_table.html` template renders these groups as Alpine.js-powered accordion rows.

`WorkerTasksTable` was migrated from the old `GroupedByWorkerMixin` to the new stateless `GroupedTable`. This removes the previous fragile ordering dependency, where tables had to manually call `run_after_every_row` in the final render method and rely on stateful tracking of seen workers.

### Safety story
Tested locally

### Automated test coverage
 All existing tests continue to pass. One test was removed because it validated behaviour specific to the old implementation (`run_after_every_row`), which no longer exists in the new stateless `GroupedTable` implementation.

### QA Plan
Not needed

### Labels & Review

### DIFF
--- a/commcare_connect/templates/grouped_table.html
+++ b/commcare_connect/templates/grouped_table.html
@@ -1,0 +1,56 @@
+{% extends "base_table.html" %}
+{% load l10n %}
+
+{% block table.tbody %}
+  {% for row in table.paginated_rows %}
+  <tbody x-data="{open: false}">
+
+    <tr @click="open = !open" class="cursor-pointer select-none">
+      {% block group_header_row %}
+        {% for column, cell in row.items %}
+          {% if column.name in table.header_columns %}
+            <td {{ column.attrs.td.as_html }}>
+              {% if column.localize == None %}{{ cell }}
+              {% elif column.localize %}{{ cell|localize }}
+              {% else %}{{ cell|unlocalize }}{% endif %}
+            </td>
+          {% endif %}
+        {% endfor %}
+      {% endblock group_header_row %}
+      {% block group_header_extra %}
+        <td colspan="{{ table.non_header_column_count }}" class="pr-4">
+          <div class="flex items-center justify-end gap-2">
+            {% if table.group_item_label %}
+              {% with count=row.record.sub_rows|length %}
+                <span class="text-xs text-slate-400">{{ count }} {% if count == 1 %}{{ table.group_item_label.0 }}{% else %}{{ table.group_item_label.1 }}{% endif %}</span>
+              {% endwith %}
+            {% endif %}
+            <i class="fa-solid fa-chevron-down text-slate-400 transition-transform duration-200" :class="{'rotate-180': open}"></i>
+          </div>
+        </td>
+      {% endblock group_header_extra %}
+    </tr>
+
+    {% for sub_row in row.record.sub_rows %}
+    <tr x-show="open" x-cloak>
+      {% block group_item_row %}
+        {% for column, cell in sub_row.items %}
+          <td {{ column.attrs.td.as_html }}>
+            {% if column.localize == None %}{{ cell }}
+            {% elif column.localize %}{{ cell|localize }}
+            {% else %}{{ cell|unlocalize }}{% endif %}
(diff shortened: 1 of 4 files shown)


commcare_connect/templates/grouped_table.html
```diff
@@ -0,0 +1,56 @@
+{% extends "base_table.html" %}
+{% load l10n %}
+
+{% block table.tbody %}
+  {% for row in table.paginated_rows %}
+  <tbody x-data="{open: false}">
+
+    <tr @click="open = !open" class="cursor-pointer select-none">
```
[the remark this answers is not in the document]
hemant10yadav: I think it fine to have this inline.

commcare_connect/utils/tables.py
```diff
@@ -113,6 +116,80 @@ def get_validated_page_size(request):
...
+                groups[row.pk] = row
+            groups[row.pk].sub_rows.append(table.SubBoundRow(record=row, table=table))
+        return [groups[pk] for pk in pks if pk in groups]
+
+
+class GroupedTable(tables.Table):
+    """A table where rows are grouped by `pk` into collapsible sections.
+    Each group renders as one clickable header row with sub-rows hidden beneath
+    it. Pagination counts groups, not individual rows — so a page size of 20
+    shows 20 headers, regardless of how many sub-rows they expand to.
```
zandre-eng, line 161: Looking long-term I wonder if this will become a performance concern as task lists grow for workers. This will largely depend on how many tasks we expect an average worker to have, but it might be worth at least adding a comment here that we could make a potential performance improvement (e.g. paginate sub-rows or fetch on demand) if needed in the future.

zandre-eng (approved):
One question about performance, but otherwise looks good.

commcare_connect/utils/tables.py
hemant10yadav: Agree, although implementing pagination for sub-rows or fetching them on demand would be quite complex within the table structure.

Instead, a simpler and more convenient approach would be to reduce the page size specifically for this table. That should help manage the data volume effectively while keeping the implementation straightforward.

Also, I believe the number of tasks per worker is unlikely to be very high, so the chances of this becoming a significant issue are relatively low.

commcare_connect/utils/tables.py
```diff
@@ -113,6 +116,80 @@ def get_validated_page_size(request):
...
 
 
+class GroupedTableData(TableQuerysetData):
+    """Data source for `GroupedTable`. Paginates by distinct `pk` count, then
+    groups rows for each page into one head row per pk with a `sub_rows` list.
+    """
+
+    @cached_property
+    def _pks(self):
+        return list(dict.fromkeys(self.data.values_list("pk", flat=True)))
```
ajeety4, line 126: So we build the full ordered PK list first and perform group pagination in memory, before fetching rows for the current page. I assume this is acceptable performance-wise since the number of workers per opportunity is expected to remain relatively low.

ajeety4 (approved):
Looks great.

commcare_connect/utils/tables.py
hemant10yadav: Yes — pks only contains integers, not full objects, so even with a few hundred workers it uses only a small amount of memory. The actual row data is fetched only for the current page via pk__in=pks.

We need the full PK list upfront because there’s no DB-level way to support group-aware pagination otherwise — the paginator requires both __len__ and ordered __getitem__ on the same sequence.

PR #1203, opened 2026-05-20
CI-670: Fix feature flag leaking across orgs

Charl1996:
## Product Description

For users who belong to multiple workspaces, feature flags enabled on one workspace were leaking into other workspaces the user belongs to. The two user-visible symptoms today are the chat widget (`open_chat_studio_widget`) and session tracking (`session_tracking`) — both rendered on workspaces that hadn't opted into the flag. Other request-scoped uses of `Flag.is_flag_active_for_request` would have had the same problem.

### Important note
Given the nature of the issue I think we need to deploy this to production as soon as it's merged.

## Technical Summary

[CI-670](https://dimagi.atlassian.net/browse/CI-670). Two related bugs in `commcare_connect/flags/`:

1. **`is_flag_active_for_request` ignored the request's workspace scope.** It delegated to `active_flags_for_user`, which returns flags active via *any* of the user's org/opportunity/program memberships — `request.org` was never consulted. Reworked it to short-circuit on request-independent activations (`everyone`, `staff`, `superusers`, direct `users`), then check `request.opportunity` → `request.org` via the already-existing scope-correct `flag.is_active_for()`.

2. **Custom m2m relations didn't invalidate the per-flag relation cache.** waffle registers `m2m_changed` handlers for its built-in `users` / `groups` m2m, but our custom `organizations` / `opportunities` / `programs` relations had no equivalent. Any `.add()` / `.remove()` / `.set()` / `.clear()` left `_get_relation_ids` returning stale (or `CACHE_EMPTY`) sets until the Flag itself was saved. Added `m2m_changed` handlers wired to waffle's `flag_membership_changed`.

### Safety story

Scoped change to the toggle plumbing. The new request-scope check is additive after the existing user/role checks, so flags previously activated for a user directly (via `flag.users`) or via role (`everyone` / `staff` / `superusers`) continue to activate exactly as before. The m2m signal handlers only fire on relation changes and delegate to waffle's own flush logic.

### Automated test coverage

- `test_flag_does_not_leak_across_orgs_for_multi_org_user` — user is a member of two workspaces, flag enabled on workspace A only, request scoped to workspace B → expects `chat_widget_enabled` is False. Fails on `main`, passes with the fix.
- `test_cache_is_invalidated_when_relation_membership_changes` — verifies `flag.is_active_for(org)` reflects subsequent `.add()` / `.remove()` calls within the same process (i.e. the relation cache is flushed). Fails if the `m2m_changed` handler is disconnected.

### QA Plan
- As a user who is admin of two workspaces, enable `open_chat_studio_widget` on workspace A only. Confirm the chat widget appears on workspace A's pages and does **not** appear on workspace B's pages.
- In the internal Toggles UI, add a workspace to a flag's `organizations`, then immediately reload a page that gates on that flag — the new state should be reflected without a server restart.

### Labels & Review

[CI-670]: https://dimagi.atlassian.net/browse/CI-670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/flags/models.py
+++ b/commcare_connect/flags/models.py
@@ -66,8 +66,28 @@ class Flag(AbstractUserFlag):
         user = getattr(request, "user", None)
         if not (user and user.is_authenticated):
             return False
-        active_flags_query = cls.active_flags_for_user(user, include_role_flags=True).filter(name=flag_name)
-        return active_flags_query.exists()
+        try:
+            flag = cls.objects.get(name=flag_name)
+        except cls.DoesNotExist:
+            return False
+
+        if flag.everyone:
+            return True
+        if flag.staff and user.is_staff:
+            return True
+        if flag.superusers and user.is_superuser:
+            return True
+        if flag.users.filter(pk=user.pk).exists():
+            return True
+
+        opportunity = getattr(request, "opportunity", None)
+        if opportunity and flag.is_active_for(opportunity):
+            return True
+        org = getattr(request, "org", None)
+        if org and flag.is_active_for(org):
+            return True
+
+        return False
 
     def is_active_for(self, obj: Organization | Opportunity | Program):
         if isinstance(obj, Organization):
--- a/commcare_connect/flags/models.py
+++ b/commcare_connect/flags/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.db.models.signals import m2m_changed
 from django.utils.translation import gettext_lazy
 from waffle.models import CACHE_EMPTY, AbstractUserFlag
 from waffle.utils import get_cache, get_setting, keyfmt
@@ -126,3 +127,19 @@ class Flag(AbstractUserFlag):
 
         cache.add(cache_key, ids)
         return ids
+
+
+def _flush_flag_relation_cache(sender, instance, action, **kwargs):
+    from waffle.signals import flag_membership_changed
+
+    flag_membership_changed(sender, instance, action, **kwargs)
+
+
+# Ties the custom flag relationships to the m2m_changed signal
+# to ensure flag's cache is cleared after editing
+for _relation in ("organizations", "opportunities", "programs"):
+    m2m_changed.connect(
+        _flush_flag_relation_cache,
+        sender=getattr(Flag, _relation).through,
+        dispatch_uid=f"commcare_connect.flag.{_relation}.through",
+    )
--- a/commcare_connect/flags/models.py
+++ b/commcare_connect/flags/models.py
@@ -84,6 +84,9 @@ class Flag(AbstractUserFlag):
         opportunity = getattr(request, "opportunity", None)
         if opportunity and flag.is_active_for(opportunity):
             return True
+        program = _get_program_for_opportunity(opportunity)
+        if program and flag.is_active_for(program):
+            return True
         org = getattr(request, "org", None)
         if org and flag.is_active_for(org):
             return True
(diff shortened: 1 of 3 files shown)


Charl1996:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

zandre-eng (approved):

Charl1996:
### Code review (revised)

Correction to my earlier "no issues" comment — on a closer look there is one real issue:

**Program-scoped flag regression.** On `main`, `is_flag_active_for_request` delegates to `active_flags_for_user`, whose filter set includes `Q(programs__organization__members=user)` — so a flag configured only via `flag.programs` is picked up (leaky, but functional). The PR's rewrite checks `flag.is_active_for(opportunity)` and `flag.is_active_for(org)` only, neither of which traverses to programs, so program-only flags now silently return `False` from this code path.

https://github.com/dimagi/commcare-connect/blob/b87d4a0b3a0a0a351851e56236dab5bf052eb677/commcare_connect/flags/models.py#L83-L91

CodeRabbit raised this earlier and suggested adding `getattr(request, "program", None)`, but that won't actually fix it — `request.program` is not set anywhere in the codebase (only `request.opportunity` is, via `opportunity_required` in `commcare_connect/organization/decorators.py`). The functional fix is to resolve the program through the opportunity (as in the previously-tried `request.opportunity.managedopportunity.program` traversal) and call `flag.is_active_for(program)`.

Practical impact today is low because there are no `programs`-scoped flag rows in production, but the capability silently breaks for any future use.

zandre-eng (approved):

commcare_connect/flags/models.py
```diff
@@ -66,8 +67,31 @@ def is_flag_active_for_request(cls, request, flag_name: str):
         user = getattr(request, "user", None)
         if not (user and user.is_authenticated):
             return False
-        active_flags_query = cls.active_flags_for_user(user, include_role_flags=True).filter(name=flag_name)
-        return active_flags_query.exists()
+        try:
+            flag = cls.objects.get(name=flag_name)
```
hemant10yadav: Any good reason not to use `cls.get(name=flag_name)`? It checks the cache before hitting the DB.

commcare_connect/flags/models.py
```diff
@@ -66,8 +67,31 @@ def is_flag_active_for_request(cls, request, flag_name: str):
...
+        except cls.DoesNotExist:
+            return False
+
+        if flag.everyone:
+            return True
+        if flag.staff and user.is_staff:
+            return True
+        if flag.superusers and user.is_superuser:
+            return True
+        if flag.users.filter(pk=user.pk).exists():
```
hemant10yadav: same here flag._get_user_ids() it also uses cache.

ajeety4 (approved):

commcare_connect/flags/models.py
Charl1996: Good point. I'll updated

commcare_connect/flags/models.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1203/commits/2ddd8f0eaf7ad829e435a35224dc71dbafc24ace

commcare_connect/flags/models.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1203/commits/2ddd8f0eaf7ad829e435a35224dc71dbafc24ace

hemant10yadav (approved):

PR #1205, opened 2026-05-21
Fix new-user onboarding flow after prelogin launch

Charl1996:
This change as not prompted by a ticket - I noticed on my local env the two things this ticket adress:
1. A signed in user is redirected to the "home" page (i.e. marketing page)
2. The "Sign Up" button from the home page redirects to the production URL and does not respect the env.

## Product Description

After the prelogin marketing site replaced `pages/home.html`, users logging in without any organization memberships were dropped on the public marketing page with no indication of what to do next — and the Sign Up link on the prelogin page hard-coded the connect.dimagi.com URL so it didn't work in other environments.

This PR does 2 things
### 1 - Restore "no memberships" page
Redirects post-login, membershipless users to a new `/users/no_memberships/` page. It shows a **Create a Workspace** CTA for users with `workspace_entity_management_access`, otherwise the original "ask your administrator for an invite" message.

image

### 2 - Update signup link
Points the prelogin **Sign Up** link at `{% url 'account_signup' %}`.

## Technical Summary

- `UserRedirectView` now sends membershipless users to `users:no_memberships` (a new `TemplateView`) instead of `prelogin:home` (the marketing page).
- New template `templates/users/no_memberships.html` branches on `perms.users.workspace_entity_management_access`.

### Safety story

Change is confined to the post-login redirect destination and a static template. No data, schema, or permission changes. The previous target (`prelogin:home`) remains reachable directly at `/`.

### Automated test coverage

No existing test asserts where `UserRedirectView` sends an orgless user; worth adding one that confirms the new destination.

### QA Plan

- Log in as a fresh user with no memberships and **no** `workspace_entity_management_access` — lands on the "ask your administrator" page.
- Log in as a fresh user **with** that perm — lands on the "Create a Workspace" CTA, click through, create an org, end up on `opportunity:list`.
- Log in as a user with an existing membership — lands on `opportunity:list` as before.
- Visit `/` and verify the **Sign Up** CTA routes to `/accounts/signup/`.

### Labels & Review

### DIFF
--- a/commcare_connect/templates/prelogin/home.html
+++ b/commcare_connect/templates/prelogin/home.html
@@ -50,7 +50,7 @@
       <a href="#/programs" data-route="/programs">Programs</a>
       <a href="#/insights" data-route="/insights">Insights</a>
       <a href="#/join" data-route="/join">Connect Network</a>
-      <a href="{% url 'account_signup' %}" class="cta cta-ghost" target="_blank" rel="noopener">Sign Up</a>
+      <a href="{% url 'account_signup' %}" class="cta cta-ghost">Sign Up</a>
       <a href="{{ app_login_url }}" class="cta">Login</a>{# import-time edit: login href templated #}
     </nav>
     <button class="nav-toggle" id="nav-toggle" aria-label="Open menu" aria-expanded="false" aria-controls="mobile-nav">
@@ -65,7 +65,7 @@
     <a href="#/programs" data-route="/programs">Programs</a>
     <a href="#/insights" data-route="/insights">Insights</a>
     <a href="#/join" data-route="/join">Connect Network</a>
-    <a href="{% url 'account_signup' %}" class="mobile-nav-cta-secondary" target="_blank" rel="noopener">Sign Up</a>
+    <a href="{% url 'account_signup' %}" class="mobile-nav-cta-secondary">Sign Up</a>
     <a href="{{ app_login_url }}" class="mobile-nav-cta">Login</a>
   </nav>
 </header>
--- a/commcare_connect/templates/prelogin/home.html
+++ b/commcare_connect/templates/prelogin/home.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 {% load static %}
+{% load i18n %}
 {# Imported from dimagi-internal/connect-prelogin. Three import-time edits #}
 {# (search "import-time edit"): login link template var, asset paths wrapped #}
 {# in static template tags, and referrerpolicy on YouTube iframes (Django #}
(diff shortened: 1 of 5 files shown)


ajeety4 (approved):
Thanks for catching this

commcare_connect/users/views.py
```diff
@@ -80,6 +80,13 @@ def get_redirect_url(self):
 user_redirect_view = UserRedirectView.as_view()
 
 
+class NoMembershipsView(LoginRequiredMixin, TemplateView):
+    template_name = "users/no_memberships.html"
+
+
+no_memberships_view = NoMembershipsView.as_view()
```
zandre-eng: Any specific reason for putting this into a variable? Normally we simply call `.as_view()` inline in the `urls.py` file.

commcare_connect/templates/prelogin/home.html
```diff
@@ -65,7 +65,7 @@
     <a href="#/programs" data-route="/programs">Programs</a>
     <a href="#/insights" data-route="/insights">Insights</a>
     <a href="#/join" data-route="/join">Connect Network</a>
-    <a href="{% url 'account_signup' %}" class="mobile-nav-cta-secondary" target="_blank" rel="noopener">Sign Up</a>
+    <a href="{% url 'account_signup' %}" class="mobile-nav-cta-secondary">Sign Up</a>
```
zandre-eng: Minor nit: This isn't part of your changes so feel free to ignore, but it would be nice to have translations for the strings here.

zandre-eng (approved):
Looks good. I'm curious what prompted this change, is there a ticket associated with this?

commcare_connect/users/views.py
Charl1996: It's a pattern in the file. I was also finding that strange. Maybe the right course is to update the whole file instead of perpetuating this pattern if we're not doing it in the rest of the codebase.

commcare_connect/templates/prelogin/home.html
Charl1996: Translations - yes. Let me update.

Charl1996:
> Looks good. I'm curious what prompted this change, is there a ticket associated with this?

No ticket associated - only some odd behaviour I noticed while on my local env. I've updated the PR description to reflect this.

commcare_connect/templates/prelogin/home.html
Charl1996: https://github.com/dimagi/commcare-connect/pull/1205/commits/f14f75bc7a4c633c7d8571a3c0e4138fa4c3388c

commcare_connect/users/views.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1205/changes/532b1471dcce680c21288dff9a3074d475992edc

zandre-eng (approved):

PR #1213, opened 2026-05-25
update slack channel

calellowitz:
## Technical Summary

This updates our deploy workflow to post in the web pod channel for production deploys, so that the team is kept aware of when deploys are happening, and keeps staging deploys in a separate channel to avoid being too noisy.

### Safety story

slack notification only

### Labels & Review

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -61,7 +61,7 @@ jobs:
         name: Notify Deploy Start
         uses: slackapi/slack-github-action@v1.24.0
         with:
-          channel-id: 'C09TVKX4N2C'
+          channel-id: ${{ inputs.environment == 'production' && 'C09HWCJSVJ6' || 'C09TVKX4N2C' }}
           payload: |
             {
               "blocks": [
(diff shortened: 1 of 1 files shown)


mkangia (approved):
web pod channel seems like a fine choice.
Do we want to consider the tech channel instead?

calellowitz:
> Do we want to consider the tech channel instead?

mobile devs got annoyed when we did that, so we figured this time we would restrict it to a channel for web devs and EMs, but I am open to the connect tech channel if we prefer that

hemant10yadav (approved):

Charl1996:
@calellowitz  So only staging deploys will be posted to #connect-deploy channel from now? Can we rename that channel then after this change to make it clearer it only pertains to staging, otherwise it would be confusing for new folks down the line?

ajeety4 (approved):

calellowitz:
> Can we rename that channel then after this change to make it clearer

Yep! that's the plan

pxwxnvermx (approved):

PR #1224, opened 2026-05-28
Unassign Work Areas

pxwxnvermx:
## Product Description
- New unassign_work_areas_for_opportunity() function. Mirrors the existing exclude helper: clears the assignee, resets status to UNASSIGNED, preserves the work area group, and sets the HQ case owner_id to "-". Skips already-unassigned and EXCLUDED work areas. Chunked HQ calls with per-chunk atomicity so a partial HQ failure rolls back only that chunk's DB writes. Plus 9 helper tests.

 - New PM-only POST endpoint that accepts a JSON list of work area IDs, calls the helper, and returns counts for unassigned, skipped, and failed. Returns 502 only when every HQ chunk failed and nothing was unassigned. The endpoint URL is also added to the assignment-mode template context so the frontend can reach it. Plus 8 view tests.

- Replaces the previously-disabled "Unassign Selected Areas" button with one that opens a confirmation modal showing eligible vs ineligible counts. Confirming POSTs eligible IDs to the new endpoint, repaints affected work areas as UNASSIGNED on the map, clears the selection, and refreshes the FLW summary.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2451)

### Safety story
- Tested Locally

### Automated test coverage
Automated tests added.

### QA Plan
[QA Ticket](https://dimagi.atlassian.net/browse/QA-8532)

### Labels & Review


commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,70 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+    work_areas_map = {wa.id: wa for wa in WorkArea.objects.filter(id__in=work_area_ids, opportunity=opportunity)}
+
+    api_key = opportunity.api_key
+    domain = opportunity.deliver_app.cc_domain if opportunity.deliver_app else None
+
+    needs_hq = []
+    db_only = []
+
+    for work_area_id in work_area_ids:
+        work_area = work_areas_map.get(work_area_id)
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed in c5e252bd. Deduped `work_area_ids` with `dict.fromkeys()` (preserves order) before building the work-areas map, the classify loop, and the logging count. Added a helper test `test_duplicate_ids_are_deduped`.

commcare_connect/microplanning/views.py
```diff
@@ -786,6 +793,39 @@ def save_assignment(request, org_slug, opp_id):
...
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def unassign_work_areas(request, org_slug, opp_id):
+    try:
+        data = json.loads(request.body)
+        raw_ids = data["work_area_ids"]
+        if not isinstance(raw_ids, list) or not raw_ids:
+            raise ValueError
+        work_area_ids = [int(i) for i in raw_ids]
+    except (json.JSONDecodeError, KeyError, TypeError, ValueError):
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed in c5e252bd. View now rejects `type(i) is not int` (catches bool/float/str) and `len(set(raw_ids)) != len(raw_ids)` (catches duplicates) before dispatching. Added parametrized cases for `float`, `bool`, and duplicate IDs.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -419,6 +426,55 @@
...
+                            this.map.setFeatureState(
+                                { source: 'workareas', sourceLayer: 'workareas', id },
+                                { status: 'UNASSIGNED' }
+                            );
+                            const props = this.workAreaProperties.get(id);
+                            if (props) {
+                                props.status = 'UNASSIGNED';
+                                props.assignee_name = null;
+                            }
+                        }
```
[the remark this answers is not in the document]
pxwxnvermx: Fixed in c5e252bd. Took option 1: the view now returns `unassigned_ids` (the actual list from the helper) instead of just a count, and the frontend iterates that list to set feature state. Work areas whose HQ sync failed stay assigned and keep their existing status on the map.

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,74 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+
+        if work_area.case_id and api_key and domain:
+            needs_hq.append(work_area)
+        else:
+            db_only.append(work_area)
+
+    pghistory_ctx = dict(reason="unassigned", username=user.username, user_email=user.email)
+
+    for i in range(0, len(needs_hq), HQ_BULK_CHUNK_SIZE):
+        chunk = needs_hq[i : i + HQ_BULK_CHUNK_SIZE]  # noqa: E203
```
calellowitz: This chunk size is really small. Especially for such a lightweight operation. I would expect at minimum 100, and we could probably do much more.

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,74 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+    needs_hq = []
+    db_only = []
+
+    for work_area_id in unique_work_area_ids:
+        work_area = work_areas_map.get(work_area_id)
+        if work_area is None or work_area.opportunity_access_id is None or work_area.status == WorkAreaStatus.EXCLUDED:
+            skipped += 1
+            continue
+
+        if work_area.case_id and api_key and domain:
```
calellowitz: Under what circumstance will these be false?

commcare_connect/microplanning/views.py
```diff
@@ -786,6 +793,44 @@ def save_assignment(request, org_slug, opp_id):
...
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def unassign_work_areas(request, org_slug, opp_id):
+    try:
+        data = json.loads(request.body)
+        raw_ids = data["work_area_ids"]
+        if not isinstance(raw_ids, list) or not raw_ids:
+            raise ValueError
+        # Reject bool/float/str — JSON ints arrive as `int`, anything else is a client bug.
+        if any(type(i) is not int for i in raw_ids):
```
calellowitz, line 811: Since we control the client here (and all of these result in error back to the frontend), what benefit do we get from checking these?

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,74 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+            db_only.append(work_area)
+
+    pghistory_ctx = dict(reason="unassigned", username=user.username, user_email=user.email)
+
+    for i in range(0, len(needs_hq), HQ_BULK_CHUNK_SIZE):
+        chunk = needs_hq[i : i + HQ_BULK_CHUNK_SIZE]  # noqa: E203
+        # HQ's "unassigned" convention is "-"; empty string falls back to the submitting user.
+        updates = [{"case_id": str(wa.case_id), "owner_id": "-", "create": False} for wa in chunk]
+        try:
+            with transaction.atomic(), pghistory.context(**pghistory_ctx):
```
calellowitz, line 132: I think doing the transactions at the chunk level is likely to be very confusing to the user if one transaction fails, since they will have very limited ability to figure out which ones failed and reselect only those, vs just trying the whole thing again. It also may end up in a weird undesired intermediate state where some but not all were unassigned

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,74 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+
+    api_key = opportunity.api_key
+    domain = opportunity.deliver_app.cc_domain if opportunity.deliver_app else None
+
+    needs_hq = []
+    db_only = []
+
+    for work_area_id in unique_work_area_ids:
+        work_area = work_areas_map.get(work_area_id)
+        if work_area is None or work_area.opportunity_access_id is None or work_area.status == WorkAreaStatus.EXCLUDED:
```
calellowitz: do we need to prevent people from unassigning work areas that have already been started

calellowitz (approved):
Similar to the other PR. Approving due to the upcoming deadline, but I don't love the approach

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,70 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+def unassign_work_areas_for_opportunity(opportunity, work_area_ids, user):
+    """Unassign work areas and set their HQ case owner_id to '-'.
+    HQ calls are batched by HQ_BULK_CHUNK_SIZE; a batch failure skips DB unassignment
+    for the whole chunk.
+    """
+    unassigned_ids = []
+    skipped = 0
+    failed = 0
+
+    work_areas_map = {wa.id: wa for wa in WorkArea.objects.filter(id__in=work_area_ids, opportunity=opportunity)}
```
ajeety4: Should we have a limit to how much work areas can be unassigned in a single request as this is a sync operation ?

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +81,74 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
...
+            continue
+
+        if work_area.case_id and api_key and domain:
+            needs_hq.append(work_area)
+        else:
+            db_only.append(work_area)
+
+    pghistory_ctx = dict(reason="unassigned", username=user.username, user_email=user.email)
+
+    for i in range(0, len(needs_hq), HQ_BULK_CHUNK_SIZE):
```
zandre-eng: Nit: Noting that we do have a `batched` util which handles the chunking for us:
`for chunk in batched(needs_hq, HQ_BULK_CHUNK_SIZE):`

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -419,6 +426,58 @@
                 }
             },
 
+            async unassignWorkAreas() {
+                const eligibleIds = Array.from(this.selectedWorkAreas).filter(id => !this.cannotUnassign(id));
+                if (eligibleIds.length === 0) {
+                    this.showUnassignModal = false;
```
zandre-eng, line 453: Nit: Does the user get any sort of alert or message when this happens? I know we show the update/skip count on the modal, but a confirmation toast could help users understand what the system actually did on submit (or didn't do in this case).

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -419,6 +426,58 @@
...
+                                { status: 'UNASSIGNED' }
+                            );
+                            const props = this.workAreaProperties.get(id);
+                            if (props) {
+                                props.status = 'UNASSIGNED';
+                                props.assignee_name = null;
+                            }
+                        }
+                        this.showUnassignModal = false;
+                        this.clearSelection();
```
zandre-eng, line 487: I'm wondering if it might be helpful to keep failed work areas selected. This can be accompanied with a user message that all work areas that failed to update have remained selected for further inspection. This can at least assist the user to see which ones failed instead of trying to guess. What do you think?

zandre-eng (approved):
Just some minor comments from my side. Approving as it sounds like this PR needs to be merged soon.

commcare_connect/microplanning/helpers.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1224/commits/205509c9e7f219791e2312cb63b986960aa3f6ca

commcare_connect/microplanning/helpers.py
pxwxnvermx: The API_KEY and domain should not be false at all. claude is really paranoid about things being null.
Fixed. https://github.com/dimagi/commcare-connect/pull/1224/commits/205509c9e7f219791e2312cb63b986960aa3f6ca

commcare_connect/microplanning/views.py
pxwxnvermx: I think its mostly for validating the input we have received. Its a low effort operation so keeping it just gurantees a bit more defense against bad input.

commcare_connect/microplanning/helpers.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1224/commits/205509c9e7f219791e2312cb63b986960aa3f6ca

commcare_connect/microplanning/helpers.py
pxwxnvermx: I talked to product regarding this and we do need to prevent people from unassigning work areas that have been started. I have added/removed code to make this happen.

commcare_connect/microplanning/helpers.py
pxwxnvermx: Fixed. https://github.com/dimagi/commcare-connect/pull/1224/commits/205509c9e7f219791e2312cb63b986960aa3f6ca

commcare_connect/templates/microplanning/map_handler.html
pxwxnvermx: This has been added as a part of another change where the transaction fails if any of the batches have a failed update. Due to that now the whole batch remains selected after a failed update.

commcare_connect/templates/microplanning/map_handler.html
pxwxnvermx: Added a "No eligible work areas to unassign." toast instead of silently closing the modal

commcare_connect/microplanning/helpers.py
pxwxnvermx: I think we can allow doing these updates on a bulk level for now. The save assignment code also does a similar thing where we are currently not forcing a limit on the number of work areas that can be updated at once in a bulk operation.

commcare_connect/microplanning/helpers.py
```diff
@@ -81,3 +82,77 @@ def _bulk_exclude(work_areas, user, exclusion_reason):
         work_areas,
         fields=["status", "excluded_by", "excluded_reason", "work_area_group"],
     )
+
+
+def unassign_work_areas_for_opportunity(opportunity, work_area_ids, user):
+    """Unassign work areas and set their HQ case owner_id to '-'.
+    The whole operation runs in a single transaction: HQ calls are batched by
+    HQ_UNASSIGN_BULK_CHUNK_SIZE, and if any batch fails the DB changes roll back so
+    the caller never sees a partial (some-but-not-all) unassignment.
```
zandre-eng: We might need to consider the sync state between HQ and Connect for this. Let's say there are 2 batches, and batch A is successful but batch B fails. The DB will rollback for both A+B but the HQ changes for A will still persist and cause an unsynced state between Connect and HQ.

zandre-eng (approved):
New changes look good, although just one query about keeping HQ and Connect in sync.

commcare_connect/microplanning/helpers.py
pxwxnvermx: Thanks. I fixed this issue and added highlights for failed Work Areas on the map. https://github.com/dimagi/commcare-connect/pull/1224/commits/3b72429a04608175fe5a67badbadf3f10a2db3d7

zandre-eng (approved):

calellowitz (approved):
not sold on the batching changes but the rest looks good and i think its fine for now

PR #1225, opened 2026-05-28
Make Inaccessibility Requests Unique Per Work Area

zandre-eng:
## Product Description

This change tightens the uniqueness constraint on `WorkAreaInaccessibilityRequest` records so that only one inaccessibility request can exist per work area (previously, uniqueness was enforced per xform + work area combination, allowing multiple requests per work area from different form submissions). The processor is updated to use `get_or_create` so re-processed forms do not create duplicate records.

## Technical Summary

[Link to Sentry error here](https://dimagi.sentry.io/issues/7510972759/?alert_rule_id=14580978&alert_timestamp=1779954756381&alert_type=email&environment=staging&notification_uuid=a07e12e0-b280-4d7e-a8de-060a411521e8&project=4505635339829248)

Previously, `WorkAreaInaccessibilityRequest` records were constrained to be unique on `(xform_id, work_area)`, meaning the same work area could accumulate multiple inaccessibility requests from different form submissions. The uniqueness constraint has been tightened to `work_area` alone, so a work area can only have one active inaccessibility request at a time. The processor's `process_work_area_update` is updated to use `get_or_create` (keyed on `work_area`) to safely handle re-processing of the same form without raising an integrity error.

- **Model constraint change:** `WorkAreaInaccessibilityRequest.Meta.constraints` updated from `unique_xform_work_area_inaccessibility` (`xform_id`, `work_area`) to `unique_work_area_inaccessibility` (`work_area` only) in [microplanning/models.py](commcare_connect/microplanning/models.py)
- **Migration:** [0011_remove_workareainaccessibilityrequest_unique_xform_work_area_inaccessibility_and_more.py](commcare_connect/microplanning/migrations/0011_remove_workareainaccessibilityrequest_unique_xform_work_area_inaccessibility_and_more.py) removes the old constraint and adds the new one
- **Processor idempotency:** `processor.py` switches from `.create()` to `.get_or_create(work_area=work_area, defaults={...})` so repeated form processing for the same work area is a no-op rather than a duplicate insert

### Safety story

The migration removes a uniqueness constraint and adds a narrower one. Before deploying, any existing duplicate `work_area` values (i.e., multiple `WorkAreaInaccessibilityRequest` rows with the same `work_area`) would need to be resolved — the new constraint will fail to apply if duplicates exist. The feature is early-stage, so data volume is expected to be low. The `get_or_create` change makes the processor idempotent with respect to work area, reducing the risk of integrity errors from replay or retries.

### Automated test coverage

No new tests were added in this change. Existing tests for `process_work_area_update` in `commcare_connect/form_receiver/` cover the happy path; the `get_or_create` change preserves that behavior.

### QA Plan

- [ ] Submit an inaccessibility form for a work area and confirm a `WorkAreaInaccessibilityRequest` record is created
- [ ] Re-submit the same form (or replay the xform via the form receiver) for the same work area and confirm no duplicate record is created
- [ ] Submit a second distinct form for the same work area and confirm no duplicate record is created (only one record per work area exists)
- [ ] Confirm the migration applies cleanly on a database that has no duplicate `work_area` entries

### Labels & Review


Charl1996 (approved):

commcare_connect/form_receiver/processor.py
```diff
@@ -322,14 +322,16 @@ def process_work_area_update(user: User, opportunity: Opportunity, xform: XForm,
...
             work_area=work_area,
-            opportunity_access=access,
-            xform_id=xform.id,
-            date_of_visit=xform.metadata.timeStart.date(),
-            location=location,
-            reason=reason,
-            additional_details=additional_details,
+            defaults={
+                "opportunity_access": access,
+                "xform_id": xform.id,
```
ajeety4, line 329: Just double checking that we are good with losing the previous form submitted for inaccessibility request ?

commcare_connect/form_receiver/processor.py
```diff
@@ -322,14 +322,16 @@ def process_work_area_update(user: User, opportunity: Opportunity, xform: XForm,
...
-        WorkAreaInaccessibilityRequest.objects.create(
+        WorkAreaInaccessibilityRequest.objects.get_or_create(
             work_area=work_area,
-            opportunity_access=access,
-            xform_id=xform.id,
-            date_of_visit=xform.metadata.timeStart.date(),
-            location=location,
-            reason=reason,
-            additional_details=additional_details,
+            defaults={
```
ajeety4, line 327: Does this means the details from the second form won't be updated in the request ?

ajeety4 (commented):
Apolgies for late review (I see that it has been merged). 
Just a couple of queries.

commcare_connect/form_receiver/processor.py
zandre-eng: I believe this should be fine. Workers should only need to submit a single inacessibility request for a work area. If a worker submits more than one, these requests feel like they would be identical since it is for the same work area.

For more context, I also had a chat with Cal about this [here](https://dimagi.slack.com/archives/C09HWCJSVJ6/p1779971496275429).

commcare_connect/form_receiver/processor.py
zandre-eng: That is correct. I'm happy to change this if we feel we should rather update the details, but my thinking was that the original request should already contain all the details of the request form.

commcare_connect/form_receiver/processor.py
ajeety4: Cool, that makes sense.  Thanks for the context.
Agreed that we expect only a single request for now.

commcare_connect/form_receiver/processor.py
ajeety4: Agreed with the reasoning, as we expect only a single request in practice.

However, in the interest of handling this more safely, I think we should deny the second request. In the current scenario, the photo evidence for the second form would still be downloaded but never displayed in the UI.

zandre-eng: That sounds reasonable. Let me open up a follow-up PR so that we can reject duplicate requests.

zandre-eng: A PR has been created [here](https://github.com/dimagi/commcare-connect/pull/1230).

PR #1232, opened 2026-05-29
Fix Work Area Update code

pxwxnvermx:
## Product Description
This PR fixes an issue with the way Connect was storing case_ids received from HQ.

HQ send the case_ids in 2 different format UUID (with dashes) and UUID Hex. HQ also only accepts the case_id formatted in the same way it was initially sent. Connect stored these ids in a UUIDField column which updated all IDS to use the UUID (with dashes) format. HQ stores these IDs in a CharField that mean all ids keep the format they were generated in.
This PR adds a change for the case_id column to be similar to HQ.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/QA-8527)

### Safety story
- Tested Locally and on staging.

### Labels & Review


commcare_connect/commcarehq/api.py
```diff
@@ -75,7 +75,21 @@ def create_or_update_case_by_work_area(work_area: WorkArea) -> CommCareCase:
...
+            case = create_or_update_case(api_key, domain, case_data, case_id=locked_work_area.case_id)
+        except CommCareHQAPIException as e:
+            # This code only gets triggered when the case_id that we stored was in the wrong UUID Format.
+            # HQ sends case_ids in 2 formats UUID (with dashes) and UUID Hex. HQ only accepts the format
+            # that it sent initially. Connect was saving all case_ids as UUID (with dashes) and that caused
+            # the case update to fail, so this code updates the UUID format to HEX and tries to refetch
+            # the case, it also sets case_id to None so it gets updated.
+            if f"Failed to update case data for {domain} with {locked_work_area.case_id}." in str(e):
+                case_id = locked_work_area.case_id.replace("-", "")
+                case = create_or_update_case(api_key, domain, case_data, case_id=case_id)
```
sravfeyn, line 88: We should update the case_id on connect side when this happens?

pxwxnvermx: That would happen in a data migration to update all case_ids at once.

calellowitz (approved):

mkangia:
Wow! Kudos on finding that out @pxwxnvermx

PR #1235, opened 2026-06-01
Sanitize Sentry release name to remove slashes

ajeety4:
## Summary

Sanitize the deploy release identifier by replacing `/` with `-` before it's used as the Sentry release version.

Sentry release versions cannot contain slashes, so deploys from hotfix branches like `mk/hotfix-3-3-2026` failed to create a Sentry release:

```
error: invalid value 'mk/hotfix-3-3-2026.<sha>' for '<VERSION>': Invalid release version. Slashes and certain whitespace characters are not permitted.
```

The deploy itself was unaffected (the Sentry step is non-fatal and runs after `kamal deploy`); only the release tagging was skipped.

## Notes

- No-op for slash-free branches like `main`.
- `APP_RELEASE` is the single source for both creating the Sentry release and tagging the running app's error events, so sanitizing at this one point keeps them consistent.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,7 +32,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set release identifier
-        run: echo "APP_RELEASE=${GITHUB_REF_NAME}.${GITHUB_SHA}" >> "$GITHUB_ENV"
+        run: |
+          # Sentry release versions cannot contain slashes, so sanitize the ref name
+          SAFE_REF_NAME="${GITHUB_REF_NAME//\//-}"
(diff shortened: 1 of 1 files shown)


mkangia (approved):
The reason behind the change seems good to me.

What all does it impact though? Just confirming it does not break any other process

ajeety4:
> The reason behind the change seems good to me.
> 
> What all does it impact though? Just confirming it does not break any other process

Only the Sentry release name is affected. There is no impact on regular staging or production deployments. Name sanitization (replacing slashes with dashes) will only occur when a branch containing a slash is used for a hotfix deployment.

zandre-eng (approved):

mkangia:
Thanks @ajeety4 
Looks good to go.

PR #1240, opened 2026-06-03
add connectid hosts to ansible setup

calellowitz:
## Technical Summary

This adds our connectid servers to our ansible inventory, and updates the tasks to correctly set them up. It also updates the ansible playbook to account for recent changes in ubuntu and docker about the correct way to set up docker in ubuntu VMs. This should finally allow us to automate setup of all servers not just connect ones.

I chose to include these in the same repo rather than replicating the scripts to connectid, because it will be easier to maintain in the long run and protect against drift as all the setup is shared except the env files, which are correctly handled in these updates.

I also added a new limit option to our run ansible task to allow targeted setup and for testing.

### Safety story

I tested these on our new celery machine and was able to set it up successfully, including diffing the manually added env file to the one that the new playbook uses.

### Labels & Review


Charl1996 (approved):
It feels slightly weird to maintain the infra configs for connectid in this repo, but I understand the reasoning.

pxwxnvermx (approved):
Same as Charl. I get the reason for storing the config together, but still feels weird to store ConnectID/PersonalID infra here. 😄

PR #1245, opened 2026-06-03
CCCT-2274: Coverage Progress report metrics backend

ajeety4:
## Product Description

Backend (data layer + live endpoint) for the **Coverage Progress** report (CCCT-2274). The
`coverage_progress` view is wired up and computes the full report context, but for now renders an
**empty page shell** — the two tables that consume this data are added in
[CCCT-2268](https://dimagi.atlassian.net/browse/CCCT-2268). No user-facing UI change yet.

## Technical Summary

Ticket: [CCCT-2274](https://dimagi.atlassian.net/browse/CCCT-2274)

Adds `commcare_connect/microplanning/coverage_progress.py`:
- `CoverageDateFilter` — page-level date window (overall vs. a from/to range).
- Aggregate queries grouped by **ward** and **work-area-group**: targets/denominators, WA-status
  counts + building sums (windowed via pghistory status-transition timestamps), and approved-visit
  counts. EXCLUDED work areas are dropped throughout.
- A **3-slot cache** (static / pinned-last-week / filtered) with a 15-minute TTL, fronted by the
  `CoverageProgressReport` facade exposing `header()` / `ward_rows()` / `wag_rows()`.
- Row builders that merge the aggregates into the top (per-ward, ~30 columns) and bottom (per-WAG)
  table rows, with derived pct/ratio columns driven by declarative tables.

Supporting changes:
- New `0014_workareaevent_status_transition_idx` migration — `CREATE INDEX CONCURRENTLY` on the
  pghistory event table `(pgh_obj_id, status, pgh_created_at)` to back the windowed status queries.
- `coverage_progress` view sets a `SET LOCAL statement_timeout` (30s) for two reasons:
  - **Stop long-running queries:** if a query exceeds 30s, Postgres cancels it (pgcode `57014`) and
    the view returns a graceful, query-free 503 instead of a 500 — first rolling back the txn so
    `ATOMIC_REQUESTS`' end-of-request COMMIT doesn't fail.
  - **Flag it in Sentry:** the timeout is logged at ERROR, which Sentry captures as an event — so a
    slow query gets surfaced for investigation rather than failing silently.

The commit series is structured one logical step per commit (skeleton → aggregates → row builders →
cache → view → rename), so it reads well commit-by-commit.

### Performance

Benchmarked **locally** via `CoverageProgressReport` (full page: header + ward_rows + wag_rows), on a
Docker Postgres capped to a **db.t3.medium-equivalent** box (2 vCPU) with migration 0014's index in
place. These are local numbers, not measured against prod:

| Opp size | Cold (all slots fresh) | Warm (cached) | Custom range (uncached, live) |
|---|---|---|---|
| **250k WAs / 100 wards / 2,500 WAGs** | ~5.5–5.8 s | 30–140 ms | ~5.7–6.0 s |
| 12.5k WAs / 25 wards / 250 WAGs | ~1.0–1.3 s | 3–7 ms | ~1.0–1.5 s |
| 47 WAs | ~40 ms | ~2 ms | ~20 ms |

- **The warm path is the common case** — slots cache 15 min behind a single-flight lock, so almost
  all loads are tens of ms.
- **Cold / custom-date on the largest opp ≈ 6 s**, comfortably under the 30 s `statement_timeout`.
  The cost is dominated by the windowed status aggregate (a correlated `WorkAreaEvent` subquery),
  run once per grouping (ward + WAG).
- **Future optimization:** if a large opp's custom-range path ever proves heavy in practice, the
  windowed status query could be rewritten as a CTE. Deferred for now — the path is cached, and a
  CTE would introduce raw SQL this codebase doesn't currently use.

### Safety story

- **Read-only**: the report issues only `SELECT`s — no writes, no data migration of existing rows.
- **Access-scoped**: gated by `@org_admin_required` + `@opportunity_required` + the `MICROPLANNING`
  flag; every query and every cache key is scoped by `opportunity.id`, so one org cannot read
  another's data (verified in review).
- **Bounded cost**: `statement_timeout` (30s) caps runaway queries — the view returns a graceful
  503 instead of a 500, and the timeout is flagged in Sentry for investigation.
- The migration uses `CREATE INDEX CONCURRENTLY` (non-blocking) and is reversible.

### Automated test coverage

New ones added

### QA Plan

- No user-visible UI in this PR (empty shell); functional QA of the rendered tables will accompany
  CCCT-2268. Suggest a smoke check that `/a/<org>/.../coverage-progress/` returns 200 for an
  opportunity with work areas and that the endpoint is gated by the MICROPLANNING flag.

### Labels & Review

[CCCT-2268]: https://dimagi.atlassian.net/browse/CCCT-2268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CCCT-2274]: https://dimagi.atlassian.net/browse/CCCT-2274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/microplanning/migrations/0014_workareaevent_status_transition_idx.py
```diff
@@ -0,0 +1,27 @@
+from django.db import migrations
+
+EVENT_MODEL = "WorkAreaExpectedVisitCountWorkAreaGroupStatusOpportunityAccessExcludedReasonEvent"
+INDEX_NAME = "wae_obj_status_created_idx"
+
+
+def create_index(apps, schema_editor):
```
zandre-eng: It might be better to define this on the model itself and have it be automatically generated when creating the migration file:
```
@pghistory.track(
    fields=["a", "b", "c"],
    meta={"indexes": [models.Index(fields=["d"], name="wa_evt_obj_idx")]},
)
```
This makes the model definition the source of truth (don't have to guess DB state by looking through migration files). Additionally if we ever add/remove fields to the pghistory table in the future, creating an auto-migration results in the pghistory table being dropped and rebuilt, which will cause us to lose this index. Defining the index in the model will prevent this.

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def window(self) -> tuple[datetime.date, datetime.date] | None:
+        if self.is_overall:
+            return None
+        return (self.start, self.end)
+
+
+def last_week_window() -> tuple[datetime.date, datetime.date]:
+    today = localdate()
```
mkangia: is this the date according to the server or the user?

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -0,0 +1,433 @@
...
+    return WorkArea.pgh_event_model
+
+
+def _earliest_transition_subquery(status):
+    events = status_event_model().objects
+    return Subquery(
+        events.filter(pgh_obj_id=OuterRef("pk"), status=status)
+        .order_by()
+        .values("pgh_obj_id")
+        .annotate(earliest=Min("pgh_created_at"))
```
zandre-eng: Is there a specific reason for fetching the earliest through annotation instead of ordering the results by created date?

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -145,3 +152,23 @@ def status_aggregates(opportunity, group_field, window) -> dict[GroupKey, Status
...
+        opportunity=opportunity,
+        status=VisitValidationStatus.approved,
+        work_area__isnull=False,
+    ).exclude(work_area__status=WorkAreaStatus.EXCLUDED)
+    if window is not None:
+        start_dt, end_dt = _window_datetime_bounds(window)
+        qs = qs.filter(visit_date__gte=start_dt, visit_date__lt=end_dt)
+
+    group_expr = f"work_area__{group_field}"
+    rows = qs.values(group_expr).annotate(visits_approved=Count("id"))
```
mkangia, line 168: is the join and group implicit?

commcare_connect/microplanning/views.py
```diff
@@ -871,3 +877,56 @@ def act_on_inaccessibility_request(request, org_slug, opp_id, work_area_id):
...
+def coverage_progress(request, *args, **kwargs):
+    # Renders an empty page shell for now, but fetches the report data (and exercises the safeguard)
+    # so the endpoint is live. CCCT-2268 (https://dimagi.atlassian.net/browse/CCCT-2268) adds
+    # the two tables to the template, consuming the context built below.
+    opportunity = request.opportunity
+    date_filter = _get_coverage_date_filter(request)
+    template_name = "microplanning/coverage_progress.html"
+    try:
+        with connection.cursor() as cursor:
+            cursor.execute("SET LOCAL statement_timeout = %s", [STATEMENT_TIMEOUT])
```
mkangia, line 952: nit: this is interesting. No concerns against it though also not sure if this should be done here.

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -0,0 +1,433 @@
...
+        evc_reached_at=_earliest_transition_subquery(WorkAreaStatus.EXPECTED_VISIT_REACHED),
+    )
+
+
+def _window_datetime_bounds(window):
+    """Convert an inclusive (start, end) *date* window to a half-open [start, end-exclusive) datetime range.
+
+    Callers use ``__gte=start_dt`` AND ``__lt=end_dt``. Because visit_date / pgh_created_at are
+    timestamps with sub-second precision, the upper bound is midnight of the day AFTER ``end`` (not
+    ``end 23:59:59``) so every instant on the ``end`` day is included.
```
mkangia: can you help understand this?

mkangia (commented):
non blocking queries

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -0,0 +1,433 @@
...
+        "pct_buildings_covered_in_WAs_visited_last_week",
+        "last_week",
+    ),
+)
+
+# How long a computed coverage slot stays cached.
+COVERAGE_CACHE_TTL_SECONDS = 15 * 60
+
+
+def _get_or_compute(key, compute):
```
zandre-eng: Nit: This function could be removed entirely and replaced inline with a `cache.get_or_set(key, default, timeout=COVERAGE_CACHE_TTL_SECONDS)` where `default` can be a callable

commcare_connect/microplanning/views.py
```diff
@@ -876,3 +877,56 @@ def act_on_inaccessibility_request(request, org_slug, opp_id, work_area_id):
...
+@org_admin_required
+@opportunity_required
+@require_flag_for_opp(MICROPLANNING)
+def coverage_progress(request, *args, **kwargs):
+    # Renders an empty page shell for now, but fetches the report data (and exercises the safeguard)
+    # so the endpoint is live. CCCT-2268 (https://dimagi.atlassian.net/browse/CCCT-2268) adds
+    # the two tables to the template, consuming the context built below.
+    opportunity = request.opportunity
+    date_filter = _get_coverage_date_filter(request)
+    template_name = "microplanning/coverage_progress.html"
```
zandre-eng: Minor nit: This doesn't seem to be used anywhere else except at the end in `render` so this variable could be completely dropped and the string just passed inline.

commcare_connect/microplanning/views.py
```diff
@@ -876,3 +877,56 @@ def act_on_inaccessibility_request(request, org_slug, opp_id, work_area_id):
...
+        with connection.cursor() as cursor:
+            cursor.execute("SET LOCAL statement_timeout = %s", [STATEMENT_TIMEOUT])
+        report = CoverageProgressReport(opportunity, date_filter)
+        context = {
+            "opportunity": opportunity,
+            "header": report.header(),
+            "ward_rows": report.ward_rows(),
+            "wag_rows": report.wag_rows(),
+        }
+    except OperationalError as exc:
```
zandre-eng, line 960: Do we want to log this to Sentry as well?

commcare_connect/microplanning/coverage_progress.py
ajeety4: This is according to the server (UTC) and is consistent with how we use dates for rest of reports.

commcare_connect/microplanning/views.py
ajeety4: This is not necessary and is mostly good to have.  Has two advantages
- Stops any query that takes more than STATEMENT_TIMEOUT (30s for this one)
- Logs it into sentry so that we can be aware of it and take any needed action.

I am okay with removing if team thinks it is not worth it.

commcare_connect/microplanning/coverage_progress.py
ajeety4: Yes, both are implicit 
 
 - JOIN: work_area__ward traverses the UserVisit.work_area FK, so the ORM emits the join to WorkArea for us. 
 - GROUP BY: .values(group_expr).annotate(visits_approved=Count("id")) is Django's grouping idiom — .values() before .annotate() produces GROUP BY <group_expr> with a per-group count.

zandre-eng (commented):
Changes look good overall. One question regarding the migration, but rest non-blocking.

commcare_connect/microplanning/coverage_progress.py
ajeety4: For example, Jun 1–Jun 7 becomes >= Jun 1 00:00 and < Jun 8 00:00, ensuring all timestamps on Jun 7 are included. This avoids edge cases where records such as Jun 7 23:59:59.999 could be missed when using end-of-day boundaries like <= Jun 7 23:59:59.

(FYI - This was suggested by Claiude and felt reasonable to me)

We could equivalently use __lte with time.max (23:59:59.999999), since that matches the column's microsecond precision, however the existing approach fellt much cleaner.

commcare_connect/microplanning/views.py
ajeety4: Yes we want to.
The line `logger.exception("Coverage progress query timed out for opportunity %s", opportunity.id)` should send it as per my understanding as sentry is [configured](https://github.com/dimagi/commcare-connect/blob/main/config/settings/production.py#L113) to capture events of error and above.

commcare_connect/microplanning/coverage_progress.py
ajeety4: Good point, No specific reason. `order_by("pgh_created_at").values("pgh_created_at")[:1]` would be equivalent in terms of results returned.Switched as it is bit more readable. 
https://github.com/dimagi/commcare-connect/pull/1245/commits/73a1b4e745e8f0f0992f09b8a6272ff29bd4f9d9

commcare_connect/microplanning/coverage_progress.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1245/commits/73a1b4e745e8f0f0992f09b8a6272ff29bd4f9d9

commcare_connect/microplanning/views.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1245/commits/73a1b4e745e8f0f0992f09b8a6272ff29bd4f9d9

commcare_connect/microplanning/migrations/0014_workareaevent_status_transition_idx.py
ajeety4: Great catch. Thanks.
Addressed in https://github.com/dimagi/commcare-connect/pull/1245/commits/a99854447e2734dfa18f9d8dc204ab678036e348.

We do not need concurrent index creation as the rows in prod is quite low at this point.

zandre-eng (approved):
Thanks for addressing my questions, the changes look good.

commcare_connect/microplanning/coverage_progress.py
mkangia: and is this okay for a UI?
Say I am viewing the report in IST timezone, then the last week would be based on UTC and not IST?
I am okay if that is what we are agreeing to.

commcare_connect/microplanning/views.py
mkangia: In my opinion 30s is a long time anyway. But it is good to keep a time limit, though I am thinking if this should be at the platform level or code level.
No strong opinions

commcare_connect/microplanning/migrations/0014_workareaevent_status_transition_idx.py
mkangia: Very nice :+1:

commcare_connect/microplanning/coverage_progress.py
```diff
@@ -0,0 +1,433 @@
...
+from django.db.models import Count, Min, OuterRef, Q, Subquery, Sum
+from django.db.models.functions import Coalesce
+from django.utils import timezone
+from django.utils.timezone import localdate
+
+from commcare_connect.microplanning.helpers import pct
+from commcare_connect.microplanning.models import WorkArea, WorkAreaGroup, WorkAreaStatus
+from commcare_connect.opportunity.models import UserVisit, VisitValidationStatus
+
+LAST_WEEK_DAYS = 7
```
mkangia: any reason this is called "last" week days?

commcare_connect/microplanning/coverage_progress.py
mkangia: hmm,

I think my concern starts with this function doing things because "callers" use something.
Anyhow, can't last_week_window itself give datetime instead of giving dates that are then converted to time?
And so that function itself has the boundaries as needed and this one is not needed only.

commcare_connect/microplanning/migrations/0014_workareaevent_coverage_idx.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("microplanning", "0013_alter_workarea_case_id"),
+    ]
+
+    operations = [
+        migrations.AddIndex(
+            model_name="workareaexpectedvisitcountworkareagroupstatusopportunityaccessexcludedreasonevent",
+            index=models.Index(fields=["pgh_obj", "status", "pgh_created_at"], name="wae_obj_status_created_idx"),
+        ),
```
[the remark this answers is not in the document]
ajeety4: This is not needed as number of rows in Prod quite low

commcare_connect/microplanning/coverage_progress.py
ajeety4: I am going to double check with product on this.

commcare_connect/microplanning/views.py
ajeety4: I guess we can check the elb logs to see the request time duration. Not sure if there is way to stop the query from execution at the platform level.
Going to keep this for now.

commcare_connect/microplanning/coverage_progress.py
ajeety4: > this function doing things because "callers" use something.

Fair point. I meant the comment to document how the range is consumed, but I agree it reads as if the helper and its callers are coupled, and that's a smell. Good catch.
 
I've reworked it: CoverageDateFilter.window now returns the half-open [start, end) datetime range directly, so the window owns its own boundaries and the separate  `_window_datetime_bounds` converter is gone. last_week is now a CoverageDateFilter.last_week() factory, so it goes through the same single conversion instead of producing bare dates. Callers just unpack start_dt, end_dt = window.

https://github.com/dimagi/commcare-connect/pull/1245/commits/bdfa0a5e694cae93df9baca07357bac4f2831188

commcare_connect/microplanning/coverage_progress.py
ajeety4: Mostly comes from product column terminology (last_week_...), but agreed it is redundant.
https://github.com/dimagi/commcare-connect/pull/1245/commits/bdfa0a5e694cae93df9baca07357bac4f2831188

commcare_connect/microplanning/coverage_progress.py
ajeety4: This has been agreed.
For future, we may move to a timezone respected date filters for all the reports, and a ticket is being logged for it.

zandre-eng (approved):

mkangia (approved):

mkangia:
Thanks for the follow up.

A small note about the commit message though, I'd push that we avoid using generic code message like "address feedback".

zandre-eng (approved):

PR #1252, opened 2026-06-05
Add command

Charl1996:
## Summary

Adds a `create_microplanning_visits` management command that generates sample `UserVisit` records (with locations) for an opportunity, so they can be displayed via the **Show Visits** toggle on the microplanning map.

### Usage

```bash
./manage.py create_microplanning_visits <opportunity_id> <count> [--work-areas N]
```

- `opportunity_id` — the opportunity's UUID (`opportunity_id` field)
- `count` — number of visits to create
- `--work-areas` — how many work areas to create if the opportunity has none (default: 4)

### What it does

- Reuses the opportunity's existing work areas, or creates a grid of small, non-overlapping demo work areas in a default region (around Delhi) if it has none.
- Reuses existing `OpportunityAccess` records, or creates sample mobile workers and assigns one to each work area.
- Reuses or creates a `DeliverUnit` for the opportunity's deliver app.
- Creates the requested number of approved visits, scattering each at a random point inside a work area's boundary with a random visit date within the last two weeks.

Runs in a single atomic transaction.

### DIFF
--- a/commcare_connect/microplanning/management/commands/create_microplanning_visits.py
+++ b/commcare_connect/microplanning/management/commands/create_microplanning_visits.py
@@ -1,0 +1,153 @@
+import random
+from datetime import timedelta
+
+from django.contrib.gis.geos import Point, Polygon
+from django.core.exceptions import ValidationError
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.utils.timezone import now
+
+from commcare_connect.conftest import MobileUserFactory
+from commcare_connect.microplanning.models import SRID, WorkArea, WorkAreaStatus
+from commcare_connect.opportunity.models import DeliverUnit, Opportunity, OpportunityAccess, VisitValidationStatus
+from commcare_connect.opportunity.tests.factories import DeliverUnitFactory, OpportunityAccessFactory, UserVisitFactory
+
+# Default region to drop generated work areas into when the opportunity has
+# none of its own (lon/lat around Delhi, matching WorkAreaFactory).
+DEFAULT_REGION = (77.0, 28.0, 78.0, 29.0)
+
+
+class Command(BaseCommand):
+    help = (
+        "Create sample UserVisits (with locations) for an opportunity so they can be shown "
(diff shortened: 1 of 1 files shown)


commcare_connect/microplanning/management/commands/create_microplanning_visits.py
```diff
@@ -0,0 +1,153 @@
...
+                work_area.opportunity_access = accesses[i % len(accesses)]
+                work_area.save(update_fields=["opportunity_access"])
+        return accesses
+
+    def _create_visit(self, opportunity, work_area, access, deliver_unit):
+        point = self._random_point_in(work_area.boundary)
+        location = f"{point.y:.7f} {point.x:.7f} 0.0 5.0"  # "<lat> <lon> <altitude> <accuracy>"
+        visit_date = now() - timedelta(days=random.randint(0, 14), minutes=random.randint(0, 1440))
+
+        UserVisitFactory(
```
zandre-eng: It might be better to try and bulk this so that we don't do a separate DB write for each user visit. Since each visit needs a different point, location, and visit_date I don't think `create_batch()` would work here, so you might have to create the instance with `UserVisit()` and then do a bulk save at the end.

commcare_connect/microplanning/management/commands/create_microplanning_visits.py
```diff
@@ -0,0 +1,153 @@
...
+        bx_min = x_min + col * cell_w + cell_w * 0.1
+        by_min = y_min + row * cell_h + cell_h * 0.1
+        bx_max = bx_min + cell_w * 0.8
+        by_max = by_min + cell_h * 0.8
+
+        boundary = Polygon(
+            ((bx_min, by_min), (bx_max, by_min), (bx_max, by_max), (bx_min, by_max), (bx_min, by_min)),
+            srid=SRID,
+        )
+        return WorkArea.objects.create(
```
zandre-eng, line 102: You could just return the `WorkArea` instance here instead and then keep them all in a list for a bulk save at the end.

zandre-eng (commented):
Just a few comments around bulk saving. Fine to skip if needed, given that this will be run on staging, but good to have given the size of visits we plan on creating.

PR #1244, opened 2026-06-03
Refresh prelogin marketing site from connect-prelogin (clean URLs)

jjackson:
## Product Description

Refreshes the public marketing site at `connect.dimagi.com/` to the latest design and switches it to clean, shareable URLs — `/platform`, `/portfolio/<program>`, `/the-opportunity`, `/insights`, `/release-notes`, `/frontline-network` — each openable/refreshable directly (no 404). `robots.txt` and `sitemap.xml` are now served at the site root so the pages are search-indexable. "Sign In" links point at the Connect login; "Sign Up" links are unchanged.

Already shipped and verified on `labs.connect.dimagi.com`.

## How to review this PR

The diff is large but **almost all of it is the marketing site content itself** (a wholesale asset replacement), not logic. The real review surface is three small files:

| Review carefully | What to check |
|---|---|
| `commcare_connect/prelogin/urls.py` (+33) | Routes listed explicitly (no catch-all that could shadow `/accounts`, `/a/<org>`, dashboards); all go through `HomeView` |
| `config/urls.py` (+16) | `robots.txt` / `sitemap.xml` served at root via `TemplateView`; doesn't collide with existing routes |
| `commcare_connect/prelogin/tests/test_views.py` (+25) | Route coverage; `django_db` marker |

**Skim, don't line-review** (vendored marketing assets — the site, authored in `jjackson/connect-labs` and copied here):
- `templates/prelogin/{home.html, robots.txt, sitemap.xml}` — the single-file SPA + root files. `home.html` is one large authored document.
- `static/prelogin/{styles.css, app.js, favicon.svg, images/…}` — css/js/photos. These are linted (prettier) like normal source, so this first import also **reformats** the JS/CSS — that's why `styles.css`/`app.js` show large diffs; it's whitespace + wholesale content, not logic.

## Technical Summary

- **Clean URLs:** the SPA moved from hash routes to History-API clean URLs and renamed routes (`/programs` → `/portfolio`, `/how-it-works` → `/platform`, …). `prelogin/urls.py` enumerates each marketing route + `/portfolio/<slug>`, all via `HomeView` so `{{ app_login_url }}` resolves on every page.
- **robots/sitemap:** wired at the site root in `config/urls.py` (`Allow: /`, canonical origin in the sitemap).
- **Tests:** parametrized from `MARKETING_ROUTES` so they can't drift; `django_db` because requests pass through `CustomPGHistoryMiddleware`.

[...]

PR #1257, opened 2026-06-09
Fix Pre-Commit Hook Failure

zandre-eng:
## Product Description

No user-facing changes. This PR fixes pre-commit hook failures introduced when the prelogin marketing site was updated in #1244.

## Technical Summary

[CCCT-2388](https://dimagi.atlassian.net/browse/CCCT-2388)

Three linting issues surfaced after #1244 merged new marketing site assets that had never been linted under the hooks added in #1165:

- **Exclude `home.html` from djlint hooks (`.pre-commit-config.yaml`):** `prelogin/home.html` is a ~450KB / 3,550-line vendored SPA template — djlint was taking ~10 minutes to process it on every run. Since it's authored upstream and not a standard Django template, there's no value in linting or reformatting it. Excluded via pre-commit's `exclude` on both `djlint-reformat-django` and `djlint-django` hooks.
- **ESLint `no-empty` configured to allow empty catches (`.eslintrc.js`):** `app.js` has five `catch (_) {}` blocks that intentionally swallow errors from `history.replaceState` and `sessionStorage` calls, which throw in restricted environments (file://, private browsing). `allowEmptyCatch: true` is the standard ESLint escape hatch for this pattern.
- **Removed write-only `currentFlw` variable (`app.js`):** The variable was declared and assigned inside `activateFlw()` but never read — a genuine unused-variable bug caught by `no-unused-vars`.

### Safety story

Config-only and dead-code changes — no logic altered. The djlint and eslint rule changes are scoped to specific patterns that are verified intentional. The `currentFlw` removal is safe: the variable was write-only and had no effect on behaviour.

### Automated test coverage

No new tests required — no logic changed. Existing prelogin view tests (`commcare_connect/prelogin/tests/test_views.py`) continue to cover all marketing routes.

### QA Plan

- [ ] Run `prek run -a` and confirm all hooks pass with exit code 0
- [ ] Load the prelogin marketing site locally and verify navigation between SPA routes (`/`, `/platform`, `/portfolio`, `/insights`, etc.) works correctly
- [ ] Confirm no JS console errors on page load

### Labels & Review

[CCCT-2388]: https://dimagi.atlassian.net/browse/CCCT-2388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,9 @@ repos:
     rev: v1.36.4
     hooks:
       - id: djlint-reformat-django
+        exclude: commcare_connect/templates/prelogin/home.html
       - id: djlint-django
+        exclude: commcare_connect/templates/prelogin/home.html
 
   - repo: https://github.com/adamchainz/django-upgrade
     rev: '1.13.0'
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,7 @@ module.exports = {
   rules: {
     'no-console': 'error',
     'no-unused-vars': 'error',
+    'no-empty': ['error', { allowEmptyCatch: true }],
   },
   ignorePatterns: [
     'node_modules/',
--- a/commcare_connect/static/prelogin/app.js
+++ b/commcare_connect/static/prelogin/app.js
@@ -913,10 +913,7 @@ document.addEventListener('click', (e) => {
 
     if (!flwTabs.length) return;
 
-    var currentFlw = 0;
-
     function activateFlw(idx) {
-      currentFlw = idx;
       flwTabs.forEach(function (t, i) {
         t.setAttribute('aria-selected', i === idx ? 'true' : 'false');
       });


hemant10yadav (approved):

ajeety4 (approved):

.pre-commit-config.yaml
```diff
@@ -30,7 +30,9 @@ repos:
     rev: v1.36.4
     hooks:
       - id: djlint-reformat-django
+        exclude: commcare_connect/templates/prelogin/home.html
       - id: djlint-django
+        exclude: commcare_connect/templates/prelogin/home.html
```
hemant10yadav, line 35: Is there a reason we can't do this in `pyproject.toml` instead?

zandre-eng: I tried this approach and it didn't seem to work (djlint still reformats the file when in `pyproject.toml`. When pre-commit runs djlint it passes the staged files as explicit CLI arguments. Adding `exclude` for djlint in `pyproject.toml` only applies when djlint scans a directory, and is ignored when files are passed directly.

PR #1258, opened 2026-06-09
Redirect SMS invite link to Play Store with opp referrer

mkangia:
## Product Description

The CommCare Connect job invite SMS no longer drops users on the in-browser invite-accept page. The SMS link now redirects to the Google Play Store listing for the CommCare app, with a `referrer=opp=<opportunity_uuid>` parameter so the Play Store install carries the opportunity context through to first launch on mobile.

The existing in-browser `accept_invite` flow is untouched — only the URL placed in the outgoing SMS changes.

  ## Technical Summary

  Ticket: https://dimagi.atlassian.net/browse/CCCT-2403
  - A defensive `uuid.UUID(str(...))` round-trip was added in the view to satisfy
    CodeQL `py/url-redirection`. The `<uuid:>` URL converter already rejects
    non-UUIDs with a 404, so the try/except is unreachable in practice — an
    inline comment documents that it's purely to silence the static analyzer.

  ## Safety Assurance

  ### Safety story

  - Existing `AcceptInviteView` and its URL pattern are unchanged; the old
    flow continues to work for any in-flight invites with the old link.
  - The redirect target is a hard-coded Play Store URL. The only piece of the
    URL derived from request input is the `referrer` query value, which is
    bound to be a UUID by Django's `<uuid:>` converter (and re-validated by
    the view), so attacker-controlled hosts are not reachable.

  ### Automated test coverage

new tests added

  ### QA Plan

  ### Labels & Review

### DIFF
--- a/commcare_connect/users/views.py
+++ b/commcare_connect/users/views.py
@@ -140,16 +140,18 @@ def start_learn_app(request):
     return Response()
 
 
-class InviteRedirectView(View):
-    def get(self, request, opportunity_uuid):
-        # The <uuid:> URL converter already rejects non-UUIDs with a 404, so this
-        # try/except is unreachable in practice. It's kept to satisfy CodeQL's
-        # py/url-redirection rule, which doesn't model Django URL converters as
-        # sanitizers.
+class InviteRedirectView(RedirectView):
+    permanent = False
+
+    def get_redirect_url(self, *args, **kwargs):
+        # The <uuid:> URL converter already rejects non-UUIDs with a 404, so
+        # this try/except is unreachable in practice. It's kept to satisfy
+        # CodeQL's py/url-redirection rule, which doesn't model Django URL
+        # converters as sanitizers.
         try:
-            opportunity_uuid = str(uuid.UUID(str(opportunity_uuid)))
+            opportunity_uuid = str(uuid.UUID(str(kwargs["opportunity_uuid"])))
         except (ValueError, TypeError):
-            return HttpResponse("Invalid opportunity id", status=400)
+            return None
         params = urlencode(
             {
                 "id": "org.commcare.dalvik",
@@ -157,7 +159,7 @@ class InviteRedirectView(View):
                 "referrer": f"opp={opportunity_uuid}",
             }
         )
-        return redirect(f"https://play.google.com/store/apps/details?{params}")
+        return f"https://play.google.com/store/apps/details?{params}"
 
 
 class AcceptInviteView(View):
--- a/commcare_connect/users/views.py
+++ b/commcare_connect/users/views.py
@@ -1,3 +1,4 @@
+import uuid
 from urllib.parse import urlencode
 
 from allauth.account.models import transaction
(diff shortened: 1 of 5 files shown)


commcare_connect/users/views.py
```diff
@@ -137,6 +139,18 @@
...
+class InviteRedirectView(View):
+    def get(self, request, opportunity_uuid):
+        params = urlencode(
+            {
+                "id": "org.commcare.dalvik",
+                "hl": "en_US",
+                "referrer": f"opp={opportunity_uuid}",
+            }
+        )
+        return redirect(f"https://play.google.com/store/apps/details?{params}")
```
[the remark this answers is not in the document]
mkangia: With <uuid:...> used for the URL, Django's router rejects bad UUIDs with 404 before the view runs, so this should be safe.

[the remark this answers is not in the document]
mkangia: https://github.com/dimagi/commcare-connect/pull/1258/commits/37431458ffb00a600f7f932bc6618b7f0a0b6867

commcare_connect/users/views.py
```diff
@@ -137,6 +140,26 @@ def start_learn_app(request):
...
 
 
+class InviteRedirectView(View):
+    def get(self, request, opportunity_uuid):
+        # The <uuid:> URL converter already rejects non-UUIDs with a 404, so this
+        # try/except is unreachable in practice. It's kept to satisfy CodeQL's
+        # py/url-redirection rule, which doesn't model Django URL converters as
+        # sanitizers.
+        try:
+            opportunity_uuid = str(uuid.UUID(str(opportunity_uuid)))
```
zandre-eng: Do we want to validate that `opportunity_uuid` refers to an actual opportunity? What will the user experience if we return the play store link with a non-existent opportunity ID?

zandre-eng (approved):
Just one non-blocking question, otherwise looks good.

commcare_connect/users/views.py
```diff
@@ -137,6 +140,26 @@ def start_learn_app(request):
     return Response()
 
 
+class InviteRedirectView(View):
```
hemant10yadav: Django's RedirectView is also pretty nice.

commcare_connect/users/views.py
```diff
@@ -137,6 +140,26 @@ def start_learn_app(request):
...
+    def get(self, request, opportunity_uuid):
+        # The <uuid:> URL converter already rejects non-UUIDs with a 404, so this
+        # try/except is unreachable in practice. It's kept to satisfy CodeQL's
+        # py/url-redirection rule, which doesn't model Django URL converters as
+        # sanitizers.
+        try:
+            opportunity_uuid = str(uuid.UUID(str(opportunity_uuid)))
+        except (ValueError, TypeError):
+            return HttpResponse("Invalid opportunity id", status=400)
```
pxwxnvermx: nit: Can this be changed to 404 error instead with a opportunity not found message?

pxwxnvermx (commented):
How do you feel about removing the Old Accept Invite view now, since the only usage it had has been removed?

commcare_connect/users/views.py
mkangia: that is a good point.

The ID in the link is also populated by Connect when we send the SMS. We don't expect someone to hit this request manually and get a wrong ID here. So I chose to avoid that db look up.

commcare_connect/users/views.py
mkangia: I like nice

https://github.com/dimagi/commcare-connect/pull/1258/commits/c195772eff076c3defb3c207d258d101b0022835

commcare_connect/users/views.py
mkangia: Hmm, possibly.

After the change made in https://github.com/dimagi/commcare-connect/pull/1258/commits/c195772eff076c3defb3c207d258d101b0022835, it seems the response would be "410 Gone"

Just adding that we don't really expect this except block to fire and is added just to satisfy the alert requirement. An invalid UUID would get rejected during URL matching itself.

hemant10yadav (approved):
Looks Good Thanks!

zandre-eng (approved):

PR #1259, opened 2026-06-09
Tech spec: invoice line-item locking (CCCT-2401)

ajeety4:
Design doc only (no code changes) for fixing post-invoice drift in invoice line items.

**Problem:** line items recompute from `CompletedWork.saved_*` on every view, so post-invoice approvals/rejections drift from the frozen `PaymentInvoice.amount`.

**Chosen approach (forward-delta):** snapshot line items into a new `CompletedWorkInvoice` table + an `invoiced_approved_count` watermark; late approvals bill forward as a delta on the next invoice; rejections on billed work are blocked.

See the spec for the rejected alternative, lifecycle/edge cases, migration plan, and one open question (managed-opp agreement gating).

### DIFF
--- a/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
+++ b/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
@@ -302,10 +302,11 @@ Populate `CompletedWorkInvoice` rows for every existing invoice, so legacy invoi
 new table once the FK is gone:
 
 - For each existing invoice with linked works (read via the still-present `CompletedWork.invoice`
-  FK), create one row per work from the work's **current** `saved_*` (best available snapshot) and
-  set `invoiced_approved_count = saved_approved_count`. Set `month = TruncMonth(invoice.end_date)`
-  (original billing period) and `exchange_rate = PaymentInvoice.exchange_rate` (the frozen rate, not
-  a re-fetched current one).
+  FK), create one row per work from the work's **current** `saved_*` and set
+  `invoiced_approved_count = saved_approved_count`. Use `month = TruncMonth(status_modified_date)` and
+  `exchange_rate = get_exchange_rate(currency, month)` — matching how `get_invoice_items`
+  (`utils/completed_work.py:330,352`) groups and prices line items today, so backfilled rows reproduce
+  the current display exactly. (This is §1.4's first-billing arm; only *late deltas* use `end_date`.)
 - **Pre-existing drift is frozen as-is, not corrected.** For an already-drifted legacy invoice
   `Σ(rows)` equals the *current* line items, not necessarily the original `PaymentInvoice.amount`.
   The design stops *future* drift; reconciling history is a non-goal and is generally impossible (the
--- a/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
+++ b/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
@@ -100,7 +100,7 @@ recording how much of that work was billed on that invoice:
 
 ```
 CompletedWorkInvoice
-  invoice         FK → PaymentInvoice  (related_name="work_items", on_delete=CASCADE)
+  invoice         FK → PaymentInvoice  (related_name="work_items", on_delete=PROTECT)
   completed_work  FK → CompletedWork   (on_delete=PROTECT)
   payment_unit    FK → PaymentUnit     (snapshot, for display grouping)
   month           DateField            (period attribution, §1.4)
@@ -129,14 +129,13 @@ CompletedWorkInvoice
 - Maintained on create ([§1.3](#13-creation--snapshot--bump-watermark-atomic)) and cancel/reject
   ([§1.6](#16-lifecycle--cancel--reject--pay)).
 
-**Watermark integrity** — it's a cached copy, so guard it:
-
-- **Reconciliation check** (management command or data-integrity test): assert
-  `invoiced_approved_count == SUM(billed_count over live invoices)` per work.
-- **Hard-delete caveat:** `invoice on_delete=CASCADE` means hard-deleting a `PaymentInvoice` drops its
-  rows but leaves the watermark stale. No such path exists today (invoices are cancelled/archived,
-  never deleted), so the precondition is "invoices are never hard-deleted"; switch to `PROTECT` if
-  that changes.
+**Watermark integrity** — it's a cached copy, kept correct by enforcing the invariant in code at the
+call site: the value is written solely by the atomic, `select_for_update`-locked create/rollback code
+(§1.3/§1.6), which is the single writer and where correctness comes from.
+- **Delete guard:** `invoice on_delete=PROTECT` blocks hard-deleting a `PaymentInvoice` while it has
+  rows — the only sanctioned teardown is the §1.6 cancel/reject rollback, which decrements the
+  watermark and deletes the rows atomically. PROTECT forces any other delete path (Django admin,
+  shell) through that rollback, so the watermark can never be left stale.
 
 ### 1.2 Selection — delta-based
 
--- a/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
+++ b/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
@@ -355,6 +355,12 @@ bills only what the PM has agreed to — first billing and later duplicates alik
 the value the watermark and delta compare against ([§1.2](#12-selection--delta-based)), so the
 selection query needs no other change.
 
+### 1.11 Display — late-delta (catch-up) count on the invoice
+
+A read-only form field on the invoice showing how many of the billed units are **late deltas** — units
+billed against a delivery that was already billed on an earlier invoice — with a tooltip giving a short
+description (e.g. *"Follow-up units for deliveries first billed in an earlier period. "*).
+
 ## Shared — Why no rejection guard is needed
 
 An earlier draft added a guard blocking rejection of a billed work's visits. With the
--- a/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
+++ b/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
@@ -99,8 +99,7 @@ recording how much of that work was billed on that invoice:
 ```
 CompletedWorkInvoice
   invoice         FK → PaymentInvoice  (related_name="work_items", on_delete=PROTECT)
-  completed_work  FK → CompletedWork   (on_delete=PROTECT)
-  payment_unit    FK → PaymentUnit     (snapshot, for display grouping)
+  completed_work  FK → CompletedWork   (on_delete=PROTECT)   # payment_unit reached via this
   month           DateField            (period attribution, §1.4)
   billed_count    IntegerField         (delta billed on THIS invoice; > 0)
   amount_local    DecimalField         (frozen)
@@ -110,8 +109,8 @@ CompletedWorkInvoice
 ```
 
 - Each row is **one work (one user, one entity)**.
-- Displayed line items = `GROUP BY payment_unit, month` over these rows — today's `get_invoice_items`
-  shape, but reading frozen rows instead of recomputing.
+- Displayed line items = `GROUP BY completed_work__payment_unit, month` over these rows — today's
+  `get_invoice_items` shape, but reading frozen rows instead of recomputing. 
 
 **Why a join table, not the existing `CompletedWork.invoice` FK?**
 
@@ -224,12 +223,12 @@ On invoice creation (`InvoiceForm.save`, `tasks.py` auto-generate), in a transac
   service-delivery `amount` is a **read-only, auto-computed** field (`forms.py:1747`), not
   user-entered; computing it from the rows at submit also closes the render-to-submit window. (Custom
   invoices are user-entered but link no works → no snapshot rows.)
-- **Per-row amount.** `amount_local = billed_count × payment_unit.amount`; `amount_usd = amount_local
-  / rate`, where `rate` is for the row's `month` ([§1.4](#14-when-a-late-approved-visit-gets-billed-and-under-which-month):
+- **Per-row amount.** `amount_local = billed_count × completed_work.payment_unit.amount`; `amount_usd =
+  amount_local / rate`, where `rate` is for the row's `month` ([§1.4](#14-when-a-late-approved-visit-gets-billed-and-under-which-month):
   billing-month rate for a late delta, approval-month rate for a first billing). `exchange_rate`
   stores that `ExchangeRate` row.
-- **Display.** `get_invoiced_visit_items` becomes a `GROUP BY payment_unit, month` aggregation over
-  `invoice.work_items` — same shape as today but frozen → **drift eliminated**.
+- **Display.** `get_invoiced_visit_items` becomes a `GROUP BY completed_work__payment_unit, month`
+  aggregation over `invoice.work_items` — same shape as today but frozen → **drift eliminated**.
 
 ### 1.4 When a late-approved visit gets billed, and under which month
 
--- a/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
+++ b/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
@@ -17,10 +17,9 @@ to completed work.
 - **Late approvals** bill on the **next** invoice, never retro-added to the issued one. On an
   already-billed work the delta is always a duplicate (the only way a work's approved count exceeds 1
   — see [Background](#background--how-a-works-approved-count-exceeds-1)).
-- **Late rejection is blocked** once a work is billed, so an issued invoice can't be silently lowered.
 
 **Decision:** chosen over the "drop late approvals" alternative (appendix) because product confirmed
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+  ([§1.6](#16-lifecycle--cancel--reject--pay)).
+
+**Watermark integrity** — it's a cached copy, so guard it:
+
+- **Reconciliation check** (management command or data-integrity test): assert
+  `invoiced_approved_count == SUM(billed_count over live invoices)` per work.
+- **Hard-delete caveat:** `invoice on_delete=CASCADE` means hard-deleting a `PaymentInvoice` drops its
+  rows but leaves the watermark stale. No such path exists today (invoices are cancelled/archived,
+  never deleted), so the precondition is "invoices are never hard-deleted"; switch to `PROTECT` if
+  that changes.
```
zandre-eng: Is there any reason not to use `PROTECT` now to future-proof against this? This will make the relationship clearer for devs and guard against other potential forms of delete (through Django admin or shell).

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+
+**New — `CompletedWork.invoiced_approved_count`** (`IntegerField(default=0)`): the watermark.
+
+- A **count only** — money lives solely in `CompletedWorkInvoice`.
+- Maintained on create ([§1.3](#13-creation--snapshot--bump-watermark-atomic)) and cancel/reject
+  ([§1.6](#16-lifecycle--cancel--reject--pay)).
+
+**Watermark integrity** — it's a cached copy, so guard it:
+
+- **Reconciliation check** (management command or data-integrity test): assert
```
zandre-eng: I would be in favour of a data integrity check on the model/DB level so that we can confidently ensure this constraint.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+
+### 1.9 Releases — expand/contract rollout
+
+The FK drop ships in a *separate, later release*, not with the cutover — safe under rolling deploys
+(old pods still reference the FK mid-deploy), preserves clean rollback (a dropped column can't be
+restored), and keeps a replicated-column drop (`multidb`) as its own deliberate step.
+
+**Release 1 — expand (cutover, keep the FK):**
+
+1. **Schema add** — add `CompletedWorkInvoice`; add `CompletedWork.invoiced_approved_count`
```
zandre-eng, line 324: Curious to hear your thoughts on querying this instead of saving it in a counter. Something like:

```
completed_work_instance.completedworkinvoice_set.count()
```
This will not be as performant at scale (could be mitigated through things like `annotate` for loops), but it will mean we are guaranteed accuracy and don't have to worry about this value drifting.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+Populate `CompletedWorkInvoice` rows for every existing invoice, so legacy invoices render from the
+new table once the FK is gone:
+
+- For each existing invoice with linked works (read via the still-present `CompletedWork.invoice`
+  FK), create one row per work from the work's **current** `saved_*` (best available snapshot) and
+  set `invoiced_approved_count = saved_approved_count`. Set `month = TruncMonth(invoice.end_date)`
+  (original billing period) and `exchange_rate = PaymentInvoice.exchange_rate` (the frozen rate, not
+  a re-fetched current one).
+- **Pre-existing drift is frozen as-is, not corrected.** For an already-drifted legacy invoice
+  `Σ(rows)` equals the *current* line items, not necessarily the original `PaymentInvoice.amount`.
```
zandre-eng, line 306: Would we need to update `PaymentInvoice.amount` for any open backfilled invoices (that aren't already marked as `PAID`)? After the backfill, the sum of the already drifted line items will then permanently differ from what's reflected in the invoice amount, which could cause confusion.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Shared — Rejection-guard enforcement
+
+**Why the guard is needed.** A billed work can also hold an approved-but-not-yet-agreed visit, and
+the count includes *all* approved visits. That visit is still rejectable (`reject_visits` only blocks
+`review_status=agree`, `views.py:1123`), so rejecting it drops `saved_approved_count` and **drifts
+the already-billed invoice**. The guard stops this: once a work is billed, none of its approved
+visits can be rejected.
+
+**Enforcement.** A single `completed_work_is_locked(work)` predicate (Approach 1 checks the
```
zandre-eng: Did the product team have any input on this? This is a change in our existing workflow, and so just want to make sure we are fine with this change.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,421 @@
...
+
+## Good to have — legacy drift report
+
+Optional, not required for correctness. The backfill ([§1.8](#18-backfill)) freezes legacy rows from
+current `saved_*`, so an already-drifted invoice is frozen *with* its drift. To avoid blessing that
+silently, add a read-only report (management command / query, not a migration step):
+
+- For each existing invoice, compare `Σ(CompletedWorkInvoice.amount)` against `PaymentInvoice.amount`.
+- Where they differ, emit invoice number, opportunity, and delta. **Report only — no auto-correct**
+  (history is a non-goal; the original per-work split can't be reconstructed).
```
zandre-eng: Would it be possible to determine drift if we do the following:

- Take the invoice date which is set at invoice creation
- Find all user visits approved after the invoice date that are  linked to the work
- Subtract their payment unit value from saved_payment_accrued to determine the pre-drift amounts.

I haven't tested the above theory, so just an idea I'm raising.

zandre-eng (commented):
The spec generally looks good. Just a few queries from my side.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: I have shared the proposed workflow, and will follow up on this.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Good point, I was thinking we shouldn't rewrite amount in the backfill itself — non-PAID still includes READY_TO_PAY and PENDING_PM_REVIEW invoices that have already been signed off, so we'd be changing approved figures silently. Instead I'm considering a management command we can run for the affected invoices (reconciling the amount to match the line items) and inform the admins before/after running it. Does that sound good?

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: .count() would undercount as there's one row per invoice, but a row can bill more than one unit (billed_count > 1, e.g. a delta billing 2 at once). 
So the real number is Sum(billed_count), not the row count. We also use this value directly in DB queries (the billable-works filter and the rejection lock), which is simplest with a stored column, and it's only written by the locked code in §1.3/§1.6 so it can't drift.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Enforcing this at the call site instead - the watermark has a single writer so the invariant is maintained there. A true DB/model constraint can't express it anyway (it's a cross-table aggregate, which CHECK/CheckConstraint can't reference). Updated the "Watermark integrity" section to state this.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Good call. Switched to CompletedWorkInvoice.invoice to on_delete=PROTECT.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,420 @@
...
+### 1.1 Model changes
+
+**New — `CompletedWorkInvoice` (per-work billing record).** One row per `(invoice, completed_work)`
+recording how much of that work was billed on that invoice:
+
+```
+CompletedWorkInvoice
+  invoice         FK → PaymentInvoice  (related_name="work_items", on_delete=PROTECT)
+  completed_work  FK → CompletedWork   (on_delete=PROTECT)
+  payment_unit    FK → PaymentUnit     (snapshot, for display grouping)
```
calellowitz: Is this necessary if it can be fetched from the completed work?

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,420 @@
...
+
+### 1.1 Model changes
+
+**New — `CompletedWorkInvoice` (per-work billing record).** One row per `(invoice, completed_work)`
+recording how much of that work was billed on that invoice:
+
+```
+CompletedWorkInvoice
+  invoice         FK → PaymentInvoice  (related_name="work_items", on_delete=PROTECT)
+  completed_work  FK → CompletedWork   (on_delete=PROTECT)
```
calellowitz: This blocks deletion of the completed work, right? I think thats correct behavior, just want to make sure

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,420 @@
...
+  payment_unit    FK → PaymentUnit     (snapshot, for display grouping)
+  month           DateField            (period attribution, §1.4)
```
calellowitz, line 103: what is this for? I think we can just used dates from the invoice

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
```diff
@@ -0,0 +1,420 @@
...
+
+- Billable delta per work = `saved_approved_count − invoiced_approved_count` (≥ 1).
+- Brand-new works bill their full count via the same path.
+
+> **Depends on [Open question 2](#open-questions-pending-product):** this query (and the
+> [§1.4](#14-when-a-late-approved-visit-gets-billed-and-under-which-month) `invoiced_approved_count ==
+> 0` test) use `saved_approved_count`. If managed billing must require PM agreement, the billable
+> count changes — resolve before finalizing.
+
+**Why late deltas bypass the date window.** A late duplicate keeps the work at `approved`, so its
```
calellowitz, line 175: This seems pretty inefficient if it has to go over every existing completed work. I think we may want to find a way to notate on the completed work itself that it has been updated (or query based on latest approved user visit)

calellowitz (approved):
Left a couple questions but the overall approach fine to me. That said, I think this was really tough to read, likely because of the terminology claude used and its overall verbosity. I think this is a great example of times where claude goes overboard in details in ways that detract from readability. In the future it would be helpful to try to make this a bit more concise (even if that just means telling claude to do that).

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
zandre-eng: Thanks for the clarification!

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
zandre-eng: That sounds good to me. :+1:

zandre-eng (approved):
Thanks for addressing my questions, overall looks good from my side.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Noting that I have removed the locking and rejection guard path as we are going to fix the scenario where duplicate approved and not yet agreed visits are counted towards billing.
Once we fix this, we do not need locking and rejection guard as agreed visits are not allowed to be disagreed by the system,

zandre-eng: Thanks for following up on this, that sounds good.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Followed up on this. Short version: we can compute the total drift per invoice (frozen amount − current recomputed line items), but not the per-work split needed to correct individual amounts.

The reason is the multiple deliver units per work:

-   A work's count isn't a per-visit sum — it's a min across its deliver units, built from several visits that were agreed at different times. To rebuild the count as of the invoice date, we'd need each visit's agreed state on that date.
-  But nothing records when a visit was agreed — agreement only sets review_status (no timestamp; status_modified_date tracks status, not agreement). So we can't tell which visits were already agreed before the invoice date.

Thinking to keep this report-only that surfaces which invoices drifted and by how much, no itemized auto-correction.

Happy to hear any thoughts on this.

zandre-eng: Thanks for investigating more into this. It's unfortunate that we won't be able to calculate the per-item amounts as I think this could've been quite useful. But the plan for keeping this report-only sounds good.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Yeah that's right

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: This indicates the month to which the work is being billed to. An invoice can span across several months.
Also it used to group invoice line items shown in UI which is by month.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Agreed this scans the opp's completed works.

However I don't think it is an issue considering below points:
  - This is a cold path - only runs when an invoice is created.
  - A flag (or visit-based prefilter) will still keep the same scan; it just swaps the filter. And the delta filtering itself has no additional cost (tested on local — ~200 ms for 1M works).
  - A partial index on the delta predicate can significantly help. Not sure it's worth adding at this point, so thinking of deferring it.
  
Happy to hear any thoughts on this

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
ajeety4: Good catch. removed in https://github.com/dimagi/commcare-connect/pull/1259/commits/07327c907804a6ad150e9d13141ebbacadba85f5

ajeety4:
>  That said, I think this was really tough to read, likely because of the terminology claude used and its overall verbosity. 

Agreed and thanks for raising this. 
I did intend to keep it a bit more detailed due to complexity of the feature involved.  Perhaps I should have used a better prompt while trying to make it concise.

docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md
calellowitz: > query based on latest approved user visit

do you have thoughts about this approach? Something like the following pseudo code (I am making up the field names, and am not sure if it makes more sense to start with the Visit or completed work) `UserVisit.objects.filter(status_change_date__gte=last_invoice).selected_related('completed_work')` should efficiently filter for only the relevant completed works and is easy to index the status change column to make it even more efficent if we don't already have that

ajeety4: Yes, adding an index on the status-change date would be more efficient than the current approach, since it limits the scan to only the recent changes.

That said, I still think the count-based approach is the cleaner source of truth for a couple of reasons:

1. **Handling cancellations**. The date-scan approach depends on a "last invoice date" cursor that only moves forward. If an earlier invoice is cancelled after a later invoice already exists (for example, Jan is cancelled but Feb has already been invoiced), the re-opened work from Jan falls behind the cursor and won't be picked up in subsequent runs. This remains true even if we define the cursor as the latest non-cancelled invoice, since that would still be Feb. With the count-based approach, ordering doesn't matter—cancelling an invoice simply reduces the invoiced count, so saved > invoiced again and the next run automatically picks it up.

2. **Single source of truth**. Even with a date-scan, we'd still need the count comparison to calculate the delta and prevent double billing. So the date-scan becomes an additional filtering step before the same count-based validation, rather than a true replacement for it.

zandre-eng (approved):

calellowitz (approved):
Still a bit concerned about performance, but happy to try it like this and see what happens

PR #1262, opened 2026-06-10
Refactor Microplanning Flag Decorator

pxwxnvermx:
## Product Description
This PR fixes the microplanning flag only getting applied on the Opportunity Level.
- All usages of the `require_flag_for_opp` decorator were replaced with `waffle_flag` decorator from the django-waffle library.
- The `is_active` method was overriden on the flag model to allow usage of the `waffle_flag` decorator.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CI-683)

### Safety story
Tested Locally

### Labels & Review


zandre-eng (approved):

commcare_connect/flags/models.py
```diff
@@ -131,6 +131,10 @@ def _get_relation_ids(self, relation_name, cache_key_template):
         return ids
 
     def is_active(self, request, read_only: bool = False) -> bool:
+        user = getattr(request, "user", None)
+        if not (user and user.is_authenticated):
```
ajeety4: Wondering how this interacts with the other Waffle flag settings. For example, if the flag is configured in the admin panel to be active for everyone (including unauthenticated users), the early auth check would prevent super().is_active() from ever evaluating those conditions.

I don't think it matters much for our current use case, but it might be cleaner to move this check below super().is_active() so both cases continue to work as expected.

ajeety4 (commented):
Looks great. Just a single query

commcare_connect/flags/models.py
```diff
@@ -130,6 +130,26 @@ def _get_relation_ids(self, relation_name, cache_key_template):
         cache.add(cache_key, ids)
         return ids
 
+    def is_active(self, request, read_only: bool = False) -> bool:
```
hemant10yadav, line 103: Could we avoid this `is_active()` override entirely? It seems the root issue is that `require_flag_for_opp` was calling `is_flag_active(flag_name, opportunity)` instead of `is_flag_active_for_request(request, flag_name)`, which evaluates the full scope. Updating the existing decorator to use `is_flag_active_for_request` would fix the bug without duplicating the scoping logic, right?

pxwxnvermx: > Could we avoid this is_active() override entirely?

Overriding `is_active` allows us to use the `waffle_flag` decorator from the library, without needing to create a custom decorator for the same use case. Also `is_active` method has code to allow using the same decorators inside tests.

>  It seems the root issue is that require_flag_for_opp was calling is_flag_active(flag_name, opportunity) instead of is_flag_active_for_request(request, flag_name), which evaluates the full scope. Updating the existing decorator to use is_flag_active_for_request would fix the bug without duplicating the scoping logic, right?

Yes, I have made change to refactor the usage to be more library focused. I think it much nicer to not declare too much logic ourselves and defer to using the library logic wherever applicable.

hemant10yadav: Cool, happy if that's what it gives us. In that case, should we get rid of `is_flag_active_for_request`?

pxwxnvermx: Yeah, i was thinking about that. I think we can remove that completely. I am going to make that change. Thanks for the suggestion.

commcare_connect/flags/models.py
pxwxnvermx: I checked the internal code for django-waffle again. It already does the user authentication check after the everyone check is done so we dont need this guard here.

https://github.com/dimagi/commcare-connect/pull/1262/commits/7559f795c06aef96837b34cad9962097500ace7b

commcare_connect/flags/models.py
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1262/commits/ade6d4be1e1b1e2b4e82f67b7678303370e25339

ajeety4 (approved):
Thanks for the changes

hemant10yadav (approved):
Thanks!

PR #1263, opened 2026-06-10
Solicitations module: design + ticket breakdown specs

Charl1996:
## Product Description

No user-facing changes. This PR adds design documentation for the upcoming **Solicitations** module (a public marketplace where Program Managers post RFPs/EOIs, organizations apply, reviewers score, and PMs shortlist and award). No code or schema changes.

## Technical Summary
- Requirements doc: https://docs.google.com/document/d/1QkbWoT0Ebqy4zZjlCfgvJra2mfR9SJo9RysRKapD_4U/edit?tab=t.0
- Ticket: [CCCT-2450](https://dimagi.atlassian.net/browse/CCCT-2450)
- **Release path: 3** — the module will sit behind a global `solicitations` feature switch (`flags/switch_names.py`).

Adds two specs under `docs/superpowers/specs/`:

- **`2026-06-09-solicitations-design.md`** — the design doc: overview, the flow by persona, and the technical design (module/URLs, key decisions, data model, lifecycles, permissions, notifications, Phase 2).
- **`2026-06-10-solicitations-ticket-breakdown.md`** — a recommended ticket breakdown grouping the work into epics E0–E6 with per-ticket scope, acceptance criteria, dependencies, and design-section references.

### Safety story

Documentation only — no code, migrations, or runtime behavior changes. Inherently safe; existing data is unaffected.

### Automated test coverage

N/A — docs only.

### QA Plan

No QA required (documentation-only change).

### Labels & Review

[CCCT-2450]: https://dimagi.atlassian.net/browse/CCCT-2450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+natural to the applicant *and* resolvable to an org at award. *Decision:* after the standard
+login, the applicant either **picks an `LLOEntity` they're affiliated with** (the LLO of an
+org they already belong to) or **creates one inline** (`name`, `short_name`). Picking an
+existing LLO reuses the org they reached it through; creating a new LLO inline also creates a
+Connect `Organization` linked to it (`org.llo_entity = the new LLO`), with the user as a
+member. The **`Application` is keyed on the `LLOEntity`** — one application per LLO per
+solicitation — and also stores the resolved `Organization` so the award has a definite org.
+*Benefit:* the applicant identity matches how LLOs think of themselves, and a real
+`Organization` always exists by submit time, so there's nothing to create lazily at award —
+the award step simply creates/updates the `ProgramApplication` for that org.
```
calellowitz: I thought we decided the ProgramApplication piece was not true

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+*Affiliation* is defined through org membership: the LLOs a user can pick are the
+`llo_entity` values of the orgs they belong to (a user with no orgs simply creates one
+inline).
+
+> **Open question (revisit before the apply-flow tickets):** `Organization.llo_entity` is a
+> plain FK (`organization/models.py`), and the existing org forms already let two orgs point at
+> the same `LLOEntity` — there is no 1:1 enforcement today. That breaks two assumptions here:
```
calellowitz: Im not sure how much this matters, since I think in the medium term we are doing away with the LLOEntity model, but I had the same concern as is raised here

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+**Decision 4 — Reviewer visibility is scoped per-solicitation via a ReviewerAssignment record.**
+*Problem it solves:* the requirement is "a reviewer sees only the solicitations they've been
+added to," but Connect's access model is all-or-nothing per organization — it can't express
+"these two solicitations but not the others this org runs." *Decision:* a
+`ReviewerAssignment(user, solicitation, role)` record *is* the authorization; review screens
+check it rather than relying on org membership alone. Reviewers must still be members of the
+funding org that posted the solicitation (cross-org reviewers — assigning someone outside the
+posting org — are out of scope for v1). But because a single user can belong to *several*
+funding orgs and be assigned in each, the review surface is **non-org-scoped**: it lives on
+its own top-level path so the user's "My assigned solicitations" list shows every assignment
```
calellowitz: >  "My assigned solicitations" list shows every assignment

Can you explain more about this? This is different than every other UI and solicitations are otherwise within an org context (they are issued by an org/llo and responded to by an org/llo). Pulling things out of the org context seems quite confusing

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    solicitation_id = models.UUIDField(editable=False, default=uuid4, unique=True)
+    title = models.CharField(max_length=255)
+    type = models.CharField(max_length=10, choices=Type.choices)
+    status = models.CharField(max_length=20, choices=Status.choices, default=Status.DRAFT)
+    public = models.BooleanField(default=False)
+
+    description = models.TextField()  # the scope of work; UI may label this "Scope of work" or "Description"
+
+    budget_min = models.PositiveBigIntegerField(null=True, blank=True)
+    budget_max = models.PositiveBigIntegerField(null=True, blank=True)
```
calellowitz: what do these mean?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+        UNDER_REVIEW = "under_review", _("Under Review")
+        SHORTLISTED = "shortlisted", _("Shortlisted")
+        AWARDED = "awarded", _("Awarded")
+        REJECTED = "rejected", _("Rejected")
+        WITHDRAWN = "withdrawn", _("Withdrawn")
+
+    application_id = models.UUIDField(editable=False, default=uuid4, unique=True)
+    solicitation = models.ForeignKey(Solicitation, on_delete=models.CASCADE, related_name="applications")
+    # The applicant identity: an LLOEntity, picked or created inline at apply.
+    llo_entity = models.ForeignKey(LLOEntity, on_delete=models.PROTECT, related_name="solicitation_applications")
```
calellowitz: This is implied by org if we include org

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    submitted_by = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True)
+    submitter_name = models.CharField(max_length=255)   # snapshot, kept even if user changes
+    submitter_email = models.EmailField()               # snapshot
+
+    status = models.CharField(max_length=20, choices=Status.choices, default=Status.DRAFT)
+    submitted_date = models.DateTimeField(null=True, blank=True)
+    withdrawn_date = models.DateTimeField(null=True, blank=True)
+
+    # Link into the existing onboarding flow, set on award (program-linked solicitations only).
+    program_application = models.ForeignKey(ProgramApplication, on_delete=models.SET_NULL, null=True, blank=True)
```
calellowitz, line 452: How is this column used? Are we replacing parts of the old program flow?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+**Decision 7 — No AI features in v1; contracts limited to a manual upload placeholder.**
+The prototype teased "Generate criteria with AI" and "AI Comparative Review"; the brief
+lists AI as future scope, so they're excluded. Contract *automation* (generate-on-award,
+e-signature, counter-signature) is also out of v1 — but the `Award` carries a single
+`contract_file` field so a PM can **manually upload an out-of-band signed PDF** (the brief's
+v1 placeholder). Because it's just a file on the award, it also serves solicitations that
+were run off-platform. The fuller contract model (a dedicated `Contract` record +
+program-level `ContractTemplate`) is reserved for Phase 2.
+
+### 3.4 Data model
```
calellowitz, line 302: How much of this is copied from the "data model" tab in the design doc, vs your assessment of what data we need to implement the designed workflows?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    awarded_date = models.DateTimeField(auto_now_add=True)
+    award_amount = models.PositiveBigIntegerField()
+    currency = models.ForeignKey(Currency, on_delete=models.PROTECT, null=True)
+    notes = models.TextField(blank=True)
+    # v1 placeholder: PM manually uploads an out-of-band signed contract PDF.
+    # Also covers contracts for solicitations run off-platform. Phase 2 replaces this
+    # with generated documents + e-signature (a dedicated Contract model).
+    contract_file = models.FileField(upload_to="solicitations/contracts/", null=True, blank=True)
+    # Back-linked later once the existing flow spins up the opportunity (Part 2, Step 5).
+    linked_managed_opportunity = models.ForeignKey(
```
calellowitz: This implies linking to an opp, but other parts imply linking to a program? Do we need to mix both?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    application = models.ForeignKey(Application, on_delete=models.CASCADE)
+    awarded_by = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True)
+    awarded_date = models.DateTimeField(auto_now_add=True)
+    award_amount = models.PositiveBigIntegerField()
+    currency = models.ForeignKey(Currency, on_delete=models.PROTECT, null=True)
+    notes = models.TextField(blank=True)
+    # v1 placeholder: PM manually uploads an out-of-band signed contract PDF.
+    # Also covers contracts for solicitations run off-platform. Phase 2 replaces this
+    # with generated documents + e-signature (a dedicated Contract model).
+    contract_file = models.FileField(upload_to="solicitations/contracts/", null=True, blank=True)
```
calellowitz, line 543: Do we use this instead of S3 elsehwere? I would have guessed our default file storage is s3

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+- **Contract:** no contract *lifecycle* in v1 — the `Award` just holds a manually-uploaded `contract_file`. Generation, status tracking, and e-signature are Phase 2.
+
+### 3.6 Permissions & access (summary)
+
+| Surface | Gate |
+|---|---|
+| Public marketplace | none — unauthenticated; only `public` + `active` solicitations |
+| Apply flow | authenticated; applicant picks an affiliated `LLOEntity` or creates one inline (which also creates a backing org) |
+| PM workspace | `@org_program_manager_required` |
+| Review screens | non-org-scoped; a `ReviewerAssignment` for that solicitation (reviewer must also be a member of the posting org) |
```
calellowitz: Posted similarly above, but given that this already dependent on the org as described, I think treating this as non org scoped is confusing. These sentences read as contradictory.

If "reviewer must also be a member of the posting org," then isn't it "org scoped?"

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+
+### A. Assumptions made (brief silent or ambiguous → this doc decided)
+
+- **Who may post = `program_manager` org admins.** The brief says "any permissioned user"
+  with a generic "Funder" role "assigned to any Connect user (Dimagi or external) in a
+  funding organization." This doc (Decision 3) assumes that means **admins of an org flagged
+  `program_manager=True`** via the existing `@org_program_manager_required` decorator — no new
+  permission concept, and the "Funder" persona is collapsed into "PM."
+
+- **Inline LLO creation also creates a backing `Organization`.** The brief lists the applicant
```
calellowitz: I think we want to avoid the whole llo creation piece in this work. We are rebuilding all the signup flows in a different design doc (the permissions one), and I think that these two works will conflict if we implement the signup flow described here

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+  identity as "LLO entity, Organization, User" but never says where the `Organization` comes
+  from. This doc (Decision 2) assumes creating an `LLOEntity` inline **also provisions a
+  Connect `Organization`** (with the user as member), so a real org always exists by submit
+  time and the award handoff needs nothing created lazily. *Affiliation* is assumed to mean org
+  membership (the LLOs a user can pick = `llo_entity` of orgs they belong to).
+
+- **`Application` is keyed on `LLOEntity`, one per LLO per solicitation.** The brief implies
```
calellowitz: Mentioned above but I think we want to use org here

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+  `expected_start_date` / `expected_end_date`, and `hide_scores_until_submit` as a stored
+  toggle. The brief implies score-visibility is "configurable" but doesn't name the rest.
+
+- **Naming/placement specifics:** the Django app is `solicitation`, the feature switch is
+  `solicitations`, and the public nav label is provisionally **"Explore opportunities"**
+  (flagged TBC with Product in §3.2). The brief specifies none of these.
+
+### B. Where the two documents disagree
+
+- **Evaluation-criteria visibility.** The brief's PM story says criteria "are available to be
```
calellowitz: Can you explain more about this choice? Im not sure I understand what it is saying

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+  "EOI/RFP can be created with no Program affiliation," but the *Reuse vs. new entities* table
+  says "Each solicitation **is** scoped to a Program." **This doc makes `program` optional**
+  (`null=True`), siding with the Goals.
+
+- **Private-solicitation access model.** The brief defines private as "only visible to invited
+  orgs" — an access-control mechanism. The data model here has only a `public` boolean and **no
+  invited-org gating**; this doc leans toward private = *unlisted* (reachable by direct link).
+  *Unresolved — already flagged in §Part 2 Step 1 and noted as needing its own spec.*
+
+- **"Invite users/orgs to apply" as a formal mechanism.** The brief lists PM stories to "invite
```
calellowitz: Is this confirmed with product? That felt like a key workflow

calellowitz (commented):
I left a lot of questions on this, but also as a high level comment, this spec is very long and verbose and includes what appears to be a long summary of the whole design doc. That makes it quite hard to review. In the future it would be great to focus on keeping these concise and narrow to the technical implementation. We may need to work on how to keep claude more under control (even if that just means asking it to be less verbose), and I would be curious the process used to create this, as it seems much larger than other recent specs (which might be a a result of the prompt, a different claude model, or the skill used).

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+
+A new Django app, `commcare_connect/solicitation/`, modelled on the existing `program`
+app. The feature has four distinct audiences, so it has four URL surfaces:
+
+| Surface | URL | Why it's separate |
+|---|---|---|
+| **Public marketplace** | top-level `/solicitations/`, `/solicitations/<id>/` | Must be reachable with no login, like the existing `prelogin` marketing pages. Shows only public, active solicitations. |
+| **Apply flow** | `/solicitations/<id>/apply/` (authenticated) | Standard login; the applicant then picks an `LLOEntity` they're affiliated with or creates one inline (which also creates a backing org). The application is keyed on that LLO. |
+| **PM workspace** | `/a/<org_slug>/solicitations/…` | Org-scoped management screens; reuses Connect's standard per-org URL pattern and permissions. |
+| **Review screens** | top-level `/solicitations/reviews/` (list), `/solicitations/<id>/review/` (scoring) | Authorized by `ReviewerAssignment`, **not** org membership — so a reviewer assigned across several orgs sees *all* their assigned solicitations in one non-org-scoped list. |
```
mkangia: I wonder if this could get tricky and we should nest this url also under the org_slug and show solicitations of only the current organization.
I believe almost all pages on Connect expect a current organization. I don't see why this should be any different.
This is a minor concern.

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+the only non-org-scoped UI element.
+
+**Public marketplace** (unauthenticated; shows only `public` + `active` solicitations)
+
+| Page | Audience | Reached via | Purpose |
+|---|---|---|---|
+| Marketplace list | Public visitor / prospective applicant | New **"Explore opportunities"** item in the public site nav + home CTA | Browse + filter (type, country, delivery type, deadline) published solicitations as scannable cards. *(Part 2, Step 2)* |
+| Solicitation detail (public) | Public visitor / prospective applicant | Marketplace cards → detail | Read scope, budget range, deadline. **No questions, no criteria.** "Apply" CTA → sign-up/login. |
+
+**Apply flow** (authenticated applicant; standard sign-up/login, applies as an `LLOEntity` — picked or created inline)
```
mkangia: do we currently expose our list of LLOEntities to all users?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+org they already belong to) or **creates one inline** (`name`, `short_name`). Picking an
+existing LLO reuses the org they reached it through; creating a new LLO inline also creates a
+Connect `Organization` linked to it (`org.llo_entity = the new LLO`), with the user as a
+member. The **`Application` is keyed on the `LLOEntity`** — one application per LLO per
+solicitation — and also stores the resolved `Organization` so the award has a definite org.
+*Benefit:* the applicant identity matches how LLOs think of themselves, and a real
+`Organization` always exists by submit time, so there's nothing to create lazily at award —
+the award step simply creates/updates the `ProgramApplication` for that org.
+*Affiliation* is defined through org membership: the LLOs a user can pick are the
+`llo_entity` values of the orgs they belong to (a user with no orgs simply creates one
```
mkangia: so if its a new user, how would they do the following that was said above

> Picking an existing LLO reuses the org they reached it through

i.e selecting an existing LLO.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: > in the medium term we are doing away with the LLOEntity 

good info.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: I think so too.

I raised the same concern when the URLs were shared above. Showing solicitations of the current organization would match the behavior that exists on the UI for almost all pages that exist today.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: https://github.com/dimagi/commcare-connect/pull/1263/changes#diff-966a25cc7f8157c09e26ed64ca2a1d32147f6567c229eb3122705bdaea42072dR220

It is for the budget range to be mentioned on the solicitation.
Is there we have a range and not just the budget directly like we have for opportunity?

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: my guess is that we want to plugin into the existing workflow which uses ProgramApplication so we are keeping that model in the loop.

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+class Award(BaseModel):
+    """A decision to award an applicant. A solicitation may have multiple awards."""
+
+    solicitation = models.ForeignKey(Solicitation, on_delete=models.CASCADE, related_name="awards")
+    application = models.ForeignKey(Application, on_delete=models.CASCADE)
+    awarded_by = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.SET_NULL, null=True)
+    awarded_date = models.DateTimeField(auto_now_add=True)
+    award_amount = models.PositiveBigIntegerField()
+    currency = models.ForeignKey(Currency, on_delete=models.PROTECT, null=True)
+    notes = models.TextField(blank=True)
```
mkangia: nit: we tend to put "notes" on a lot of models. And it's usually a text field so its actually a "note" since you can't put more than one.
I like the "comment" in the model above, so I suggest we use "comment" here as well.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: my guess is that the linking is done based on what is available at that point in the workflow?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    # Phase 2 reserves: contract = models.OneToOneField("Contract", ...)
+```
+
+**Phase 2 only (not built in v1):** a `Contract` (per award; status, signed copy,
+e-signature provider fields) and a program-level versioned `ContractTemplate`.
+
+### 3.5 Status lifecycles
+
+- **Solicitation:** `draft` → `active` (on publish) → `closed` (deadline passed, or PM
+  closes early with a reason). `awarded` is set once ≥1 award is made, and can be reached from
```
mkangia: isn't award limited to only one application?

> once ≥1 award

so, this should be `== 1`?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+  either `active` (a PM may award before the deadline — Part 2, Step 4) or `closed`.
+  `cancelled` is a reason-tagged terminal action from `draft`/`active`. Drafts are freely
+  editable; once `active`, structural fields (questions/criteria) lock while descriptive copy
+  and deadline extensions stay editable.
+  - **Deadline close is automatic.** A daily Celery-beat task flips `active → closed` for any
+    solicitation whose `application_deadline` has passed (PMs can also close early by hand).
+- **Application:** `draft` → `submitted` → `under_review` → `shortlisted` →
+  `awarded` | `rejected`. The PM may also award or reject directly from `under_review`
+  (shortlisting is the normal path, not a gate). `withdrawn` is an applicant action allowed
```
mkangia: what is the expectation if someone withdraws after awarded?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+    application into `under_review` when they pick it up to score (this triggers the
+    applicant's "under review" email, §3.7); it is not flipped automatically.
+  - **Shortlisting** is a PM-only curation step, done **in bulk** from the per-solicitation
+    Applications dashboard (reviewers cannot shortlist). It moves the selected applications
+    `under_review → shortlisted` and **emails each affected applicant** (§3.7).
+  - **Not mandatory before award.** `under_review → awarded` is allowed, so a PM can award a
+    clear-cut RFP without shortlisting first; the dashboard still nudges toward the
+    shortlist-then-award path.
+  - **Reversible.** `shortlisted → under_review` (un-shortlist) is allowed any time before the
+    application is awarded or rejected. Un-shortlisting sends **no** applicant email, to avoid
```
mkangia: strange, I would expect an email in case of un-shortlisting too.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: this would be nice? so we limit this to just existing LLOs?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -0,0 +1,697 @@
...
+
+- **Added Solicitation fields not named in the brief:** `contact_email`, `estimated_scale`,
+  `expected_start_date` / `expected_end_date`, and `hide_scores_until_submit` as a stored
+  toggle. The brief implies score-visibility is "configurable" but doesn't name the rest.
+
+- **Naming/placement specifics:** the Django app is `solicitation`, the feature switch is
+  `solicitations`, and the public nav label is provisionally **"Explore opportunities"**
+  (flagged TBC with Product in §3.2). The brief specifies none of these.
+
+### B. Where the two documents disagree
```
mkangia: maybe @mmmary314 can help with some of these?

mkangia (commented):
Overall this looks a good first step. Understanding all of it in detail at this point seems unreasonable.

The LLO piece could be cleared out, it seems that is going to run into conflicts.

Even after we have tickets for these, we should continue to think through each implementation piece and ensure its the best way forward.

mkangia:
Also, Part 1 and 2 was really helpful. Thanks for adding that.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: I particularly [asked about this](https://docs.google.com/document/d/1QkbWoT0Ebqy4zZjlCfgvJra2mfR9SJo9RysRKapD_4U/edit?disco=AAAB89Mf3dg) in the spec and seems like it should ignore org boundaries.

Charl1996: As for implementation - I think we can just check whether the request user is authenticated before rendering the page.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: That's how I understand it, but just double-checking on the requirements doc if this is the intent.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: I'm under the impression "not necessarily true", meaning if there's a program linked we can still slot into that flow, otherwise not.

My understanding is that a central goal of the solicitation feature is to get more participation in programs in an organic manner, meaning the end goal is "this org is linked to this program" (otherwise why link the program to a solicitation at all?). Of course, there's cases where no program is linked, presumably because no program exists yet (I assume this is for EOI types).

Given the above, it makes sense to me to still reuse the ProgramApplication table if there's a program backing the solicitation, since the ProgramApplication is what links the org to the program. If an org applied and is accepted, that's just another path into the program than being invited to the program (which is currently the only path).

That said, did you have something else in mind here?

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: This would then not apply to the new user (since they are not a member of any org yet) and they would have to create one inline.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: We recently added it if I'm not mistaken, so curious what the future plans are?

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: > This is different than every other UI and solicitations are otherwise within an org context

Agreed, but I asked about this in the spec and it seems to be a requirement that we ignore the org boundaries. Given the responses to this I'm confirming in the doc whether ignoring org boundaries is what we want as opposed to considering only the current org context.

> Can you explain more about this?

Since a user can belong to multiple orgs the scenario exists where a user is a reviewer on multiple of these orgs, so what I'm suggesting (assuming we want to ignore org boundaries) is that the "My assigned solicitations" page lists all the solicitations the logged-in user is assigned to as a reviewer across all orgs.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: It's mostly similar, but there are slight differences (field additions and omissions) pertaining to certain decisions that was made as part of this doc. 

I do think the models setup in this doc is a fairly good representation of the solution and is a good starting point, but there are small things on some models that this doc adds which comes from the google doc but that I wonder about (like the `awarded_date` on `Solicitation` - we can just derive that on the fly, but I can also make a case that we want that on Solicitation). However, those are minor and I left it because it feels like we can iron that out during implementation.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: My take is that a solicitation does not award just one amount, hence the range.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: > plugin into the existing workflow

Exactly. See also [this comment](https://github.com/dimagi/commcare-connect/pull/1263/changes#r3394808622).

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: > I would have guessed our default file storage is s3

It is - `upload_to` is just the file path we upload it to **on S3**. S3 is defined as the default storage backend and `models.FileField` use that if no storage argument is given.
 
I don't think we specify FileFields anywhere else in the codebase, but it's a nice and clean way of linking a file to a model.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Good to challenge that. Confirming on the doc whether a solicitation (and by extension an award) only applies to program level or also opportunity level. I assumed it can be both, but this might not be true.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Award is 1:1 with application, but multiple awards can be given per solicitation.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: At the moment nothing happens except the application is marked as "withdrawn".

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: That was my own assumption, but happy to update if you think it should be. I'll just also confirm on the requirements doc quickly as well.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Also curious if this means we limit to existing LLOs for now.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Basically just that we don't show the evaluation criteria defined by the PM (which reviewers use to score the application).

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Confirmed that this will work differently than this design doc says. Will update.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: okay, does the solicitation become re-open or something?

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: https://github.com/dimagi/commcare-connect/pull/1263#discussion_r3394704841

So we show LLO entities to all users but its not really public. So they just need to sign up and not really need to be part of any organization to access LLOs list. Not sure I fully understand the concept of LLOs.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: or may be we need to revoke the access as a reviewer if their membership is removed as well?

Charl1996:
@calellowitz 

>  this spec is very long and verbose and includes what appears to be a long summary of the whole design doc. That makes it quite hard to review. 

I was afraid that would be the case, though I intentionally added Parts 1 - 3 for reviewers to not have to go back to the [requirements doc](https://docs.google.com/document/d/1QkbWoT0Ebqy4zZjlCfgvJra2mfR9SJo9RysRKapD_4U/edit?tab=t.0) and would be able to read top to bottom to gain context from a high level down to technical pieces.

But good to know you found it too long. 

> ... keeping these concise and narrow to the technical implementation.

I was hoping Part 3 would do that.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Updated the spec to move this to org-scoped page.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Updated the spec

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Updated the spec to scope this to org

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Updated the spec to align more closely with the new org permissions planned.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: any follow up on this @Charl1996 ?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -54,6 +54,34 @@ parallel system.
 - An award cleanly creates/updates the downstream onboarding record so the LLO can be
   brought into the program through the flow that already exists.
 
+### Definitions
+
+A quick glossary of the terms used throughout this doc.
+
+- **Solicitation** — the posting itself: an RFP or EOI shown on the marketplace. The core new
+  object in this module. Hey folks, just a reminder that I'm offline for the next two weeks.
```
mkangia: @Charl1996 where did that offline note come from? 😆

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -62,54 +90,51 @@ This section walks the whole lifecycle once, in order, before meeting the data m
...
-awarded/rejected*) and can withdraw before the deadline. They're emailed on every status
-change.
+deadline, and opens a solicitation to read its scope, budget range, and deadline. A **logged-in**
+user additionally sees an **"Invited to you"** section listing the private solicitations their
+orgs have been invited to. The **questions and the internal criteria are not shown** at this
+stage — questions appear only after they sign in to apply, and criteria are never shown to
+applicants at all (they're an internal scoring tool).
+
+When they click **Apply**, they go through the **standard Connect sign-up/login flow** (the
+shared signup owned by CCCT-2494) and then choose which **organization** they're applying *as*:
```
mkangia, line 124: Can user apply for the same solicitation from a different organization? 
or can they change the organization once application initiated?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -133,7 +158,7 @@ This is the hand-off to the existing system. When a winner is awarded:
   **ProgramApplication** record to *accepted* — which is exactly what *unlocks* the normal
   opportunity-setup flow the platform already has. The marketplace deliberately stops there;
   it does not try to auto-build the opportunity itself.
-- If the solicitation has **no Program** (a standalone EOI), the award is simply recorded.
+- If the solicitation has **no Program**, the award is simply recorded.
```
mkangia, line 161: what is the reason to have a `ProgramApplication` in case of a program linked and not otherwise?

Or in other words, what is the use of having a `ProgramApplication`? and if there is, then how does that use get cater to in case of no program application.

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -153,9 +178,9 @@ app. The feature has four distinct audiences, so it has four URL surfaces:
...
+| **Apply flow** | `/solicitations/<id>/apply/` (authenticated) | Standard login/signup (owned by CCCT-2494); the applicant picks one of their organizations or creates a new (probationary) org during signup. The application is keyed on that organization. |
 | **PM workspace** | `/a/<org_slug>/solicitations/…` | Org-scoped management screens; reuses Connect's standard per-org URL pattern and permissions. |
-| **Review screens** | top-level `/solicitations/reviews/` (list), `/solicitations/<id>/review/` (scoring) | Authorized by `ReviewerAssignment`, **not** org membership — so a reviewer assigned across several orgs sees *all* their assigned solicitations in one non-org-scoped list. |
+| **Review screens** | org-scoped `/a/<org_slug>/solicitations/reviews/` (list), `/a/<org_slug>/solicitations/<id>/review/` (scoring) | Lists only the current org's solicitations the user is assigned to. A `ReviewerAssignment` record determines which solicitations the reviewer sees. A reviewer assigned in several orgs switches org context to see each. |
```
mkangia: how does ReviewerAssignment work?

Is it created when a review reviews an assignment right?
There access to assignments is dependent if they are added a reviewer for the organization that owns the solicitation, right?

Then this does not seem exactly right
> A `ReviewerAssignment` record determines which solicitations the reviewer sees

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -166,26 +191,21 @@ The screens below, grouped by the four URL surfaces above. "Audience" is who the
...
+**Public marketplace** (`public` + `active` shown to everyone; logged-in users also see private solicitations their orgs are invited to)
 
 | Page | Audience | Reached via | Purpose |
 |---|---|---|---|
-| Marketplace list | Public visitor / prospective applicant | New **"Explore opportunities"** item in the public site nav + home CTA | Browse + filter (type, country, delivery type, deadline) published solicitations as scannable cards. *(Part 2, Step 2)* |
-| Solicitation detail (public) | Public visitor / prospective applicant | Marketplace cards → detail | Read scope, budget range, deadline. **No questions, no criteria.** "Apply" CTA → sign-up/login. |
+| Marketplace list | Public visitor / prospective applicant | New **"Explore opportunities"** item in the public site nav + home CTA | Browse + filter (type, country, delivery type, deadline) published `public` solicitations as scannable cards. *(Part 2, Step 2)* |
+| "Invited to you" section | Logged-in member of an invited org | Marketplace (logged-in only) | Lists `private` solicitations the user's orgs hold a `SolicitationInvitation` for. *(Part 2, Step 2)* |
```
mkangia: nit: not sure I agree with that phrasing "Invited to you". May be just "Invited" would suffice?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -166,26 +191,21 @@ The screens below, grouped by the four URL surfaces above. "Audience" is who the
...
 
-**Apply flow** (authenticated applicant; standard sign-up/login, applies as an `LLOEntity` — picked or created inline)
+**Apply flow** (authenticated applicant; standard sign-up/login via CCCT-2494, applies as an `Organization` — an existing membership or a new probationary org)
 
 | Page | Audience | Reached via | Purpose |
 |---|---|---|---|
-| Application form | Applicant (LLO) | **"Apply" CTA** on the public detail page (routes through login) | Pick an affiliated `LLOEntity` or create one inline (name, short name), then answer the question template; save draft or submit. Questions visible here for the first time. *(Part 2, Step 2;)* |
-| "My applications" list | Applicant (LLO) | **Header user-profile dropdown** | All applications the user submitted, with status; filter by type. |
+| Application form | Applicant (LLO) | **"Apply" CTA** on the public detail page (routes through login/signup) | Pick which organization to apply as (or create one during signup), then answer the question template; save draft or submit. Questions visible here for the first time. *(Part 2, Step 2)* |
```
mkangia: are we still referring to applications as LLOs and not just orgs?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -214,74 +234,54 @@ the only non-org-scoped UI element.
...
-> constraint would block a second, genuinely distinct org that shares the LLO from applying.
-> We need to decide whether to assume/enforce 1:1 org↔LLO for the apply flow or key the
-> `Application` on `Organization` instead. **Deferred.**
+Solicitation. A solicitation can produce several awards (especially EOIs), so the **sum of all
+**active** award amounts on a solicitation must fit within its `budget_max`** — an award that
+would push the cumulative awarded total over `budget_max` is **rejected (hard block)**. (Awards
+released by a post-award withdrawal don't count toward the total — see §3.5.) The cap is enforced
+on the actual `Award.award_amount` totals at award time; the award amount is set by the PM
+(reviewers do not suggest amounts). When a Program *is*
+linked, the award must **also** fit within the Program's remaining budget — the Program remains the
```
mkangia, line 247: how does this fit in when there is no program?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -214,74 +234,54 @@ the only non-org-scoped UI element.
...
+financial source of truth. Both caps apply where relevant; an award must satisfy whichever is
+binding.
+
+**Decision 2 — Applicants apply as an `Organization`; signup and org creation are delegated to CCCT-2494.**
+*Problem it solves:* a marketplace applicant must resolve to a Connect `Organization` (the
+downstream `ProgramApplication` is `Organization`-keyed), but a brand-new external applicant may
+have no account and no org yet — and historically such signups were **blocked pending manual
```
mkangia, line 254: is there any email confirmation? just checking if we would get spammed

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -345,9 +337,9 @@ class Solicitation(BaseModel):
 
     description = models.TextField()  # the scope of work; UI may label this "Scope of work" or "Description"
 
-    budget_min = models.PositiveBigIntegerField(null=True, blank=True)
-    budget_max = models.PositiveBigIntegerField(null=True, blank=True)
-    currency = models.ForeignKey(Currency, on_delete=models.PROTECT, null=True)
+    budget_min = models.PositiveBigIntegerField(null=True, blank=True)  # informational lower bound shown to applicants
```
mkangia, line 340: that is strange, the applicants can see the budget the solicitation has?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -409,30 +400,51 @@ class SolicitationAttachment(BaseModel):
...
         WITHDRAWN = "withdrawn", _("Withdrawn")
 
     application_id = models.UUIDField(editable=False, default=uuid4, unique=True)
     solicitation = models.ForeignKey(Solicitation, on_delete=models.CASCADE, related_name="applications")
-    # The applicant identity: an LLOEntity, picked or created inline at apply.
-    llo_entity = models.ForeignKey(LLOEntity, on_delete=models.PROTECT, related_name="solicitation_applications")
-    # The backing Connect org (created or linked inline at apply); carried for the award handoff.
+    # The applying Connect Organization (an existing membership, or a new probationary org
+    # created during CCCT-2494 signup). This is the application's KEY — it matches the
+    # Organization-keyed ProgramApplication downstream. No LLOEntity is involved (Decision 2).
```
mkangia: it seems this spec is also mentioning that it changed its approach from LLO to Organization.
Is that really useful information to retain?

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -481,17 +498,13 @@ class Review(BaseModel):
         APPROVE = "approve", _("Approve")
         REJECT = "reject", _("Reject")
         NEEDS_REVISION = "needs_revision", _("Needs Revision")
-        UNDER_REVIEW = "under_review", _("Under Review")
```
mkangia: why was this removed?
It seems the spec still refers to this being [used](https://github.com/dimagi/commcare-connect/pull/1263/changes/4fe016c868c5793472e67ada32d12f6d50a7c0da..53de691534c54ac1c0bc20692f37d36134a1d95d#diff-966a25cc7f8157c09e26ed64ca2a1d32147f6567c229eb3122705bdaea42072dR445)

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -536,162 +562,63 @@ e-signature provider fields) and a program-level versioned `ContractTemplate`.
...
-  - **Does not freeze review.** Reviewers can keep scoring a shortlisted application —
-    shortlisting is PM curation, not a review lock.
+  - **Award requires a verified org.** A probationary (unverified) applicant org can apply and be
+    reviewed, but `→ awarded` is **blocked until a System Admin verifies the org** (CCCT-2494) —
+    the award commits budget and flips `ProgramApplication → accepted` (Decision 2).
+  - **Post-award withdrawal releases the award.** An awarded applicant may withdraw; doing so
+    **releases the `Award`** (sets `released_date`, frees its budget) and moves the application to
+    `withdrawn`. Any `ProgramApplication` created by that award is reverted out of `accepted`. The
+    PM may then **re-award** another applicant. The solicitation drops from `awarded` back to its
+    prior state (`active`/`closed`) if no active awards remain, and stays `awarded` if other awards
```
mkangia, line 599: seeing a solicitation as "awarded" because it has at least one reward but can then still be awarded if the budget allows seems confusing to me.

"partially awarded" , "awarding in progress" or something better would make sense.

docs/superpowers/specs/2026-06-09-solicitations-design.md
```diff
@@ -536,162 +562,63 @@ e-signature provider fields) and a program-level versioned `ContractTemplate`.
...
-integration with status-callback webhooks, and two-party counter-signature. Designed to also
-support contracts for solicitations that happened off-platform. (Google Docs signing is one
-option to explore.) Other future scope from the brief: AI criteria generation and AI
-application/review helpers, reviewer blinding, opt-in notifications for new solicitations,
-external-funder posting, per-RFP FAQs, and a funder analytics dashboard.
-
----
+- **To invited orgs** — when invited to a **private** solicitation (a `SolicitationInvitation`
+  is created), an email with a link to it; for **public** solicitations the PM can still broadcast
+  a link to selected orgs.
```
mkangia, line 623: I assume the link is same as the public link or would it need to be org specific?

mkangia (commented):
Thanks for the revision @Charl1996 

I reviewed just the changes since last time since reviewing it all would be a challenge.

I see the spec is keeping some redundant information. Not sure why that context is useful

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Huh...not sure. Good catch though! I'll see if I can find out when that was added. Maybe I accidentally copied the text from somewhere.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: > Can user apply for the same solicitation from a different organization?

Yes, I do think so, since an application is linked to an organization.

> can they change the organization once application initiated

I wouldn't think so. The user can then withdraw and apply as the other organization I suppose.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: As far as I understand the program application is in case there is already a program, then the successful application automatically gets linked to the program and don't have to go through the "invite to program" flow.

For an Expression of Interest (EOI) I think it might be the case that no program is ever created.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: The `ReviewerAssignment` model links a particular user to an application as a reviewer. By default not all users see solicitation applications, unless they are selected to be a reviewer, hence why the `ReviewerAssignment` gates access to solicitations for a particular user.

I've updated the model name which I think makes the purpose more clearer (https://github.com/dimagi/commcare-connect/pull/1263/commits/fc0969f9dc5f692424a476bf33ca5238a3be58c6).

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: "LLO" here simply refers to the fact that the applicant is a "locally led organization".

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: There's already an min/max budget on the solicitation model, so there's always something to compare against. 

I wonder now if we should not rather at solicitation create time compare the solicitation against the program's remaining budget (assuming there's program linked). If we do that we can always just compare against the solicitation budget at award time.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Applicants (new included) will always have to apply with an org, so they will have to have gone through the sign up process, which includes email verification.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Yes, as per the requirements doc prospects should see this information on the public page.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Maybe not _as_ useful, but I decided to keep it since the requirements doc mentions LLOEntities but we're changing that in light of new organization structure changes in work that's coming up (where we remove LLOEntities it seems).

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Applications can be "under review", but it's not a sensible "recommendation" which is where I removed it from.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: I think this is a fair point. We will probably decide the exact wording at implementation time.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Since we show solicitations an org is invited to on the marketplace page when the user is logged in to an org I think it's probably fine if this is the same link as the public link? Unless we want to show the invited solicitations for an org on a different (dedicated) page.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: not sure I understand

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: I would think a spec would keep information relevant for the implementation.
During design phase we do go through iterations, which is expected. Though, if, the final state of the spec would keep track of all possible ways things could have been done, it would be a lot of information.
At times it could be useful to list down some specific paths that were intentionally avoided but keeping such alternatives spread through the spec would cause information overload.

I am thinking there possibly could not be a right or wrong here but just doing things that keeps the process simple and effective for all would be good for us.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: Cool, thanks

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: Not sure I understand what you meant but it seems like you have something relevant going on there.

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: I thought application was an org?

multiple orgs can have the same LLO right?

docs/superpowers/specs/2026-06-09-solicitations-design.md
mkangia: `ApplicationReviewer` sounds good to me.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: This is an out-of-date flow, so not relevant anymore. The new apply flow only shows the orgs the logged-in user has access to.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Also noting this is an out of date question .

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: I think both of your statements is true, but for the technical implementation we only ever care about org. I've removed the "LLO" part [here](https://github.com/dimagi/commcare-connect/pull/1263/commits/6ac09f485f1b715629b9d0916747e9441f017461).

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: > I wonder now if we should not rather at solicitation create time compare the solicitation against the program's remaining budget (assuming there's program linked). If we do that we can always just compare against the solicitation budget at award time.

Confirmed we want to compare with program budget at award time since a program's budget can be updated on the fly.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Good point. I'll update.

docs/superpowers/specs/2026-06-09-solicitations-design.md
Charl1996: Your initial question of why the `under_review` was removed was pointing to the reviewer's recommendation options - the "under_review" status does not belong there. However, the application status options still has "under_review" as a status.

mkangia (approved):
I believe this is good to get started with and should be open to modifications as we start to implement this, specially when we get into details.

Can you share why the ticket breakdown was removed?

Charl1996:
> Can you share why the ticket breakdown was removed?

I just thought it's probably not needed to include here (I don't see other specs include it). It also keeps this already-massive PR smaller.

calellowitz (commented):
Generally looks good. Two questions

1) I know we went back and forth on the design doc a bit about the tie to the `ProgramApplication` model, but I'm still pretty confused how that fits in here. I thought we decided that was a claude error in the design doc and we didn't want to tie the solicitations models to explicit program applications, but it still appears throughout this doc. Is that no longer true?

2) At various times the doc says that only invited organizations may view or apply to solicitations, but also that anyone can create a new org and apply via that org. These seem to contradict. Which of those is true?

Charl1996:
> I know we went back and forth on the design doc a bit about the tie to the ProgramApplication model, but I'm still pretty confused how that fits in here. I thought we decided that was a claude error in the design doc and we didn't want to tie the solicitations models to explicit program applications, but it still appears throughout this doc. Is that no longer true?

I'm not sure why you think we shouldn't optionally link programs?

> At various times the doc says that only invited organizations may view or apply to solicitations, but also that anyone can create a new org and apply via that org. These seem to contradict. Which of those is true?

I'm curious which instances you're referring to? The only time "invited only" orgs can apply is for private solicitations - otherwise it's open to everyone, provided you belong to an org.

Charl1996:
Closing as the tickets have long been created.

PR #1265, opened 2026-06-10
update defaults for deprecated verfication checks

calellowitz:
## Technical Summary

Since we no longer want to allow the duplicate and gps flags, this sets their default value to `False` on the model level.

### Safety story

These aren't used anymore and only affect new opportunities since it doesn't change existing data, just default values.

### Automated test coverage

Tests updated to reflect new behavior.

### Labels & Review


commcare_connect/opportunity/models.py
```diff
@@ -239,8 +239,8 @@ def has_ended(self):
 class OpportunityVerificationFlags(models.Model):
     opportunity = models.OneToOneField(Opportunity, on_delete=models.CASCADE)
     duration = models.PositiveIntegerField(default=1)
-    gps = models.BooleanField(default=True)
-    duplicate = models.BooleanField(default=True)
+    gps = models.BooleanField(default=False)
+    duplicate = models.BooleanField(default=False)
     location = models.PositiveIntegerField(default=10)
```
hemant10yadav: The location value also needs to be 0.
https://github.com/dimagi/commcare-connect/blob/main/commcare_connect/form_receiver/processor.py#L365

hemant10yadav (approved):

PR #1269, opened 2026-06-12
Fixed two broken actions in the audit review modal

hemant10yadav:
## Product Description
<video controls class="w-full" src="https://github.com/user-attachments/assets/cb3e759f-d0fb-4f9b-904c-f58287e64530"></video>

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-2478
https://dimagi.atlassian.net/browse/CCCT-2477

Fixed the "No Action Needed" flow to submit directly with action=none instead of incorrectly showing the confirmation modal.
Replaced the broken hx-on:htmx:after-request handler with an HX-Trigger response header and native HTMX modal close/page refresh behaviour.

### Safety story

UI-layer only. No data logic changed. The one server-side change wraps the existing assign call in a try/except for the already-assigned case, returning a 400 with a message rather than failing silently.

### Automated test coverage
Test added for the duplicate assignment 400 response; all existing audit view tests pass.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/audit/views.py
+++ b/commcare_connect/audit/views.py
@@ -15,6 +15,7 @@ from commcare_connect.audit.models import AuditReport, AuditReportEntry
 from commcare_connect.audit.tables import AuditReportEntryTable, AuditReportTable
 from commcare_connect.flags.flag_names import WEEKLY_PERFORMANCE_REPORT
 from commcare_connect.flags.models import Flag
+from commcare_connect.opportunity.exceptions import TaskAlreadyAssignedError
 from commcare_connect.opportunity.models import AssignedTask, TaskType
 from commcare_connect.organization.decorators import opportunity_required, org_program_manager_required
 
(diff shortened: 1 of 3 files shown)


commcare_connect/templates/audit/audit_report_task_modal.html
```diff
@@ -1,22 +1,26 @@
...
-<div x-data="{ open: true, confirm: false, action: null, error: '' }"
+<div x-data="{ open: true, confirm: false, error: '' }"
+     x-ref="modal"
      x-show="open"
      x-cloak
+     @click.self="open = false"
      @keydown.escape.window="open = false"
+     @htmx:after-request="if (!event.detail.successful) { error = event.detail.xhr.responseText || '{% trans "Something went wrong. Please try again." %}'; confirm = false; }"
      class="fixed inset-0 z-50 flex items-center justify-center bg-black/50">
-  <div @click.outside="open = false"
```
zandre-eng, line 7: Is there a specific reason this was removed? This seems to be a pattern we do on a lot of other modals.

zandre-eng (approved):
Looks good, just a non-blocking comment

commcare_connect/templates/audit/audit_report_task_modal.html
hemant10yadav: We switched from @click.outside="open = false" to @click.self="open = false" on the outer backdrop div.

Before: @click.outside — closes when clicking outside the element (document-level listener, caused ghost-click re-open)

commcare_connect/audit/views.py
```diff
@@ -207,25 +208,30 @@ def audit_report_task_action(request, org_slug, opp_id, audit_report_id, entry_i
...
+        return HttpResponseBadRequest(_("Unknown action."))
 
     entry.reviewed = True
     entry.save(update_fields=["reviewed", "review_action", "date_modified"])
 
-    # Empty 200. The modal's client-side handler closes the modal and triggers
-    # `refreshDetail` on document.body, which causes `#detail-body` to re-fetch
-    # itself and render the updated tables.
-    return HttpResponse(status=200)
+    response = HttpResponse(status=200)
```
ajeety4, line 234: :+1:

ajeety4 (approved):

PR #1274, opened 2026-06-12
Surface org invite form errors instead of silently redirecting

Charl1996:
(The PR is an ad-hoc improvement, so no ticket)

Org member invites previously redirected with no feedback on an invalid form, so inviting an unregistered/already-member email just refreshed the page. `add_members_form` now shows a success or error message after submission.

Example error when no user is found on the system:
image

### DIFF
--- a/commcare_connect/organization/views.py
+++ b/commcare_connect/organization/views.py
@@ -69,6 +69,10 @@ def add_members_form(request, org_slug):
         form.instance.organization = org
         form.save()
         send_org_invite(membership_id=form.instance.pk, host_user_id=request.user.pk)
+        messages.success(request, gettext("Invite sent to {email}.").format(email=form.cleaned_data["email"]))
+    else:
+        error = next(iter(form.errors.values()))[0] if form.errors else gettext("Unable to send invite.")
+        messages.error(request, error)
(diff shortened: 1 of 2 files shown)


hemant10yadav (approved):

commcare_connect/organization/views.py
```diff
@@ -69,6 +69,10 @@ def add_members_form(request, org_slug):
         form.instance.organization = org
         form.save()
         send_org_invite(membership_id=form.instance.pk, host_user_id=request.user.pk)
+        messages.success(request, gettext("Invite sent to {email}.").format(email=form.cleaned_data["email"]))
+    else:
+        error = next(iter(form.errors.values()))[0] if form.errors else gettext("Unable to send invite.")
```
mkangia, line 74: it's unusual to show only the first error message. Is it because you are expecting/assuming only one error to be present?
also, is there a better way to find it instead of using the (next + iter)?

and is it that the form isn't visible on redirect so the error messages are not showed on the form?

mkangia:
Thanks for the effort here @Charl1996 

Do you think we can add an interrupt ticket so this gets tracked?

zandre-eng (approved):

PR #1276, opened 2026-06-15
fix the wrong org_slug for pm action

hemant10yadav:
## Issue

PM users received a 404 error when trying to assign tasks from the Audit Report Review modal. The page loaded correctly, but submitting the action failed.

## Technical Summary
https://dimagi.atlassian.net/browse/CI-721
The audit report templates were hardcoding `opportunity.organization.slug` in the HTMX action URLs, which always resolves to the NM organization slug. As a result, when a PM user submitted the action, the `org_slug` parameter was set to the NM organization. Our decorator then rejected the request and returned a 404 because the user did not belong to that organization.

### Safety story

The @org_program_manager_required decorator already validates the slug against the user's membership, so using the request's org_slug can't grant access beyond what was already permitted to load the page.

### Automated test coverage
Existing audit view tests pass with no changes needed.

### QA Plan

Verified locally

### Labels & Review

### DIFF
--- a/commcare_connect/audit/views.py
+++ b/commcare_connect/audit/views.py
@@ -155,6 +155,7 @@ def audit_report_detail(request, org_slug, opp_id, audit_report_id):
         "can_complete": total_flagged == reviewed_count and report.status == AuditReport.Status.PENDING,
         "name_filter": name_filter,
         "path": path,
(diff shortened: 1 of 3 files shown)


mkangia (approved):
oh lord!

PR #1279, opened 2026-06-15
Move program FK from ManagedOpportunity to Opportunity

hemant10yadav:
## Product Description

No user-facing change.

## Technical Summary

[CCCT-2500](https://dimagi.atlassian.net/browse/CCCT-2500)

- Moves `program` from `ManagedOpportunity` to the base `Opportunity` model, so all opportunities have a direct program reference rather than going through the proxy table.
- The migration backfills managed opportunities from `ManagedOpportunity.program_old` and creates two legacy programs for non-managed ones (test and real).
- `ManagedOpportunity.program` is renamed to `program_old` and kept in sync on every write, so a rollback won't leave the old column stale. The field will be removed in CCCT-2501.

### Safety story

`program_old` is written alongside program on every create and save path, so a revert will not leave any stale data.

The migration has been tested locally and deployed to staging. So far, no 500 errors have surfaced during QA, and everything appears to be working as expected.

### Automated test coverage

Tests updated and passing.

### QA Plan

https://dimagi.atlassian.net/browse/QA-8559

### Labels & Review

[CCCT-2500]: https://dimagi.atlassian.net/browse/CCCT-2500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/models.py
```diff
@@ -115,6 +115,7 @@ class Opportunity(BaseModel):
     is_test = models.BooleanField(default=True)
     delivery_type = models.ForeignKey(DeliveryType, null=True, blank=True, on_delete=models.DO_NOTHING)
     managed = models.BooleanField(default=False)
+    program = models.ForeignKey("program.Program", on_delete=models.DO_NOTHING, null=True)
```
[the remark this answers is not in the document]
hemant10yadav: The backfill will update the programs for all the opportunities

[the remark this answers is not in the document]
hemant10yadav: I think it is not needed

commcare_connect/opportunity/helpers.py
```diff
@@ -454,7 +454,7 @@ def get_base_qs(organization, filters, program_manager=False):
         if programs:
             base_filter &= Q(managedopportunity__program__slug__in=programs)
         queryset = Opportunity.objects.filter(base_filter).annotate(
-            program=F("managedopportunity__program__name"),
+            program_name=F("managedopportunity__program__name"),
```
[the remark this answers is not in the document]
hemant10yadav: No, we want to keep it this way for now. It will be updated in an upcoming PR. If we use `Opportunity.program`, it will show programs for non-managed opportunities as well, which we do not want until the next PR is merged. That PR will address this change.

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def backfill_program_from_managed(apps, schema_editor):
+    ManagedOpportunity = apps.get_model("program", "ManagedOpportunity")
+    Opportunity = apps.get_model("opportunity", "Opportunity")
+    Opportunity.objects.filter(managed=True).update(
+        program_id=Subquery(
+            ManagedOpportunity.objects.filter(pk=OuterRef("pk")).values("program_old_id")[:1]
+        )
+    )
```
zandre-eng, line 16: I think this can be replaced with:
```
Opportunity.objects.filter(managed=True).update(
    program_id=F('managedopportunity__program_old_id')
)
```
which should generate a single `UPDATE ... FROM` rather than a correlated subquery per row.

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
```diff
@@ -0,0 +1,113 @@
...
+            ),
+        ),
+        migrations.RunPython(
+            backfill_program_from_managed,
+            migrations.RunPython.noop,
+            hints={"run_on_secondary": False},
+        ),
+        migrations.RunPython(
+            create_legacy_programs,
+            migrations.RunPython.noop,
```
zandre-eng, line 110: Minor nit: Since this is a noop reversing and running this migration again will cause duplicate legacy programs to be created. Just checking whether we are fine with that?

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
```diff
@@ -0,0 +1,113 @@
...
+    today = datetime.date.today()
+    start_date = datetime.date(2021, 1, 1)
+
+    delivery_type = DeliveryType.objects.create(
+        name="Non Managed Opportunity",
+        slug=slugify("Non Managed Opportunity"),
+        description="Auto-created delivery type for legacy non-managed opportunities by system",
+    )
+
+    llo_entity = LLOEntity.objects.create(name="Legacy Connect llo")
```
zandre-eng: Minor nit: Capitalising the acronym
```suggestion
    llo_entity = LLOEntity.objects.create(name="Legacy Connect LLO")
```

commcare_connect/templates/program/email/opportunity_expiry_reminder.html
```diff
@@ -9,7 +9,7 @@
       <li>
         <a href="{{ opp.opportunity_url }}">{{ opp.name }}</a>
         <br>
-        Program: {{ opp.program_name }}
+        Program: {{ opp.program.name }}
```
zandre-eng: Nit: Do we still need to pass `program_name` from the task that uses this template? ([code ref](https://github.com/dimagi/commcare-connect/blob/473d445cd758841327f51fadafcfec2027a99ab0/commcare_connect/program/tasks.py#L215))

zandre-eng (approved):
Overall looks good, considering QA is being done here. Just one question and nits from my side.

mkangia:
Do we need to worry about any managed opportunities created during the deploy after the migration but before the code switch? I understand its an edge case but can we run a sanity query after deploy to be sure? may be query for any opportunities with missing program id and ensuring there is none?

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
hemant10yadav: `F()` expressions can traverse joins in `filter()`, but they cannot traverse joins in `update()`

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
hemant10yadav: I think it's better to avoid a reverse migration here. By the time a rollback is considered, those programs/orgs already have live data associated with them, making automatic deletion risky. Once the new program column is dropped, they will no longer be referenced and can be cleaned up manually if needed. That's why I chose to make the reverse migration a no-op.

commcare_connect/templates/program/email/opportunity_expiry_reminder.html
hemant10yadav: It wasn't a model object, so I've changed it back to program_name. [7e1b519](https://github.com/dimagi/commcare-connect/pull/1279/commits/7e1b519ccec5596c5b2253dd92b8c1426b0c979a)

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
hemant10yadav: Thanks! Added-[7e1b519](https://github.com/dimagi/commcare-connect/pull/1279/commits/7e1b519ccec5596c5b2253dd92b8c1426b0c979a)

hemant10yadav:
> Do we need to worry about any managed opportunities created during the deploy after the migration but before the code switch? I understand its an edge case but can we run a sanity query after deploy to be sure? may be query for any opportunities with missing program id and ensuring there is none?

@mkangia Thanks for bringing that up. It is a rare edge case, but I did consider it while writing the tech spec.

My understanding is that it should not result in inconsistent data. If the migration has been applied but the old code is still running on an app server, any request that attempts to create a managed opportunity using the old code path would fail because the expected program relationship no longer exists in the schema. Since we have ATOMIC_REQUESTS enabled, the entire request would be rolled back, preventing partial or invalid data from being persisted.

I agree that a sanity check after deployment is worthwhile, and I'll run one once the deployment is complete.

commcare_connect/opportunity/migrations/0134_add_program_fk_to_opportunity.py
zandre-eng: That sounds fair, thanks for sharing your reasoning!

zandre-eng (approved):

pxwxnvermx (approved):

hemant10yadav:
> I agree that a sanity check after deployment is worthwhile, and I'll run one once the deployment is complete.

@mkangia I verified the production data in Superset, and no opportunities currently have a NULL value for program_id.

image

mkangia:
thanks @hemant10yadav 
I trust your words but that image is helpful too :+1:

PR #1283, opened 2026-06-16
Backfill hq_user_uuid on ConnectIDUserLink

mkangia:
## Product Description

  Internal/operational change only — no user-facing effect.

  ## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2429

  Adds the `backfill_hq_user_uuid` management command to populate the recently added `ConnectIDUserLink.hq_user_uuid` for existing records where it is missing, looked up from the CommCare HQ
  user API. See the command's module docstring for the approach and flow.

  ## Safety Assurance

  ### Safety story

  - Gated behind two interactive confirmations; `--dry-run` reports and writes the CSV without any DB writes.
  - A timestamped CSV of every record to be updated is written before saving.
  - Only the empty `hq_user_uuid` field is written; rows that already have a value are never re-queried or modified.
  - Saves are batched and resumable — already-saved rows drop out of the "missing" filter on a rerun — with a small retry.

Tested on staging: Log [here](https://dimagi.atlassian.net/browse/CCCT-2429?focusedCommentId=473963)
For production: I would run this outside the deploy to avoid blocking deploy with this and see the dry run before running it.

  ### Automated test coverage
Tests added for changes

  ### QA Plan

  - Run with `--dry-run` on staging; confirm the reported counts and generated CSV look correct.
  - Run for real; verify affected rows have `hq_user_uuid` matching HQ

  ### Labels & Review


commcare_connect/users/helpers.py
```diff
@@ -41,8 +41,22 @@ def create_hq_user_and_link(user, domain, opportunity):
...
+    return None
+
+
+def fetch_hq_user_uuids(api_key: HQApiKey, domain: str) -> dict[str, str]:
+    """Return a {commcare_username: hq_user_uuid} map for every user in the domain.
+
+    The API key is an opportunity-level admin credential, so a single sweep lists all
+    mobile workers in the domain and backfills them in one call.
+    """
+    return {obj["username"]: obj["id"] for obj in _iter_hq_users(api_key, domain) if obj.get("username")}
```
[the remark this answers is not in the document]
mkangia: it is safe to assume here that id would always be present.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,205 @@
...
+            for link in to_be_updated:
+                writer.writerow([link.user_id, link.id, link.commcare_username, link.domain, link.hq_user_uuid])
+        return path
+
+    def _bulk_update_with_retry(self, batch):
+        for attempt in range(1, SAVE_ATTEMPTS + 1):
+            try:
+                ConnectIDUserLink.objects.bulk_update(batch, ["hq_user_uuid"])
+                return
+            except Exception as e:
```
zandre-eng, line 187: Is there a specific exception you suspect could happen here? Since this is for a DB update, if something went wrong I'm not sure if a retry would help (unless the DB was temporarily unreachable).

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,209 @@
...
+            len(links) for links in resolvable.values()
+        )
+        reference_path = self._save(to_be_updated, options["batch_size"], dry_run)
+        self._report(len(to_be_updated), not_found, no_key, errors, dry_run)
+        if reference_path:
+            self.stdout.write(f"Reference of updated users: {reference_path}")
+
+    def _links_missing_uuid_by_server_and_domain(self):
+        links = (
+            ConnectIDUserLink.objects.filter(
```
zandre-eng, line 79: Do we want to flag/output any links that have null UUIDs and have no HQ server link? These sound like they would never get linked and should maybe be removed instead.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,209 @@
...
+        return None
+
+    def _confirm_lookups(self, links_by_server_and_domain, resolvable):
+        total = sum(len(links) for links in links_by_server_and_domain.values())
+        unique_users = len({link.user_id for links in links_by_server_and_domain.values() for link in links})
+        covered = sum(len(links) for links in resolvable.values())
+        self.stdout.write(f"Found {total} ConnectIDUserLink records missing hq_user_uuid across {unique_users} users.")
+        self.stdout.write(
+            f"{covered} are covered by {len(resolvable)} domain(s) (one API call each); "
+            f"{total - covered} will be skipped (no API key)."
```
zandre-eng, line 122: Do want to do any further investigation into these skipped ones? Since there is no associated API keys these links will always remain unlinked in the DB table (and possibly hints to a broken configuration for an opp).

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,209 @@
...
+        parser.add_argument(
+            "--batch-size", type=int, default=100, help="Save resolved UUIDs in batches of this size (default 100)."
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Look up UUIDs and report what would change, without saving.",
+        )
+
+    def handle(self, *args, **options):
```
zandre-eng, line 50: Nit: Do we want to validate through doing a filter query that a valid opp and/or org UUID was passed in the args? This could save unnecessary processing and potential confusion by raising an error early if these don't match any opp/org in the system.

zandre-eng (approved):
Implementation looks good, just a few non-blocking queries from my side.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,205 @@
...
+                ConnectIDUserLink.objects.bulk_update(batch, ["hq_user_uuid"])
+                return
+            except Exception as e:
+                if attempt == SAVE_ATTEMPTS:
+                    raise
+                self.stderr.write(
+                    self.style.WARNING(f"Save failed (attempt {attempt}/{SAVE_ATTEMPTS}), retrying: {e}")
+                )
+
+    def _report(self, updated, not_found, no_key, errors, dry_run):
```
ajeety4: super nit: this will be reported even when the user declines the confirmation to update

ajeety4 (commented):
Overall looks good.

Just below query on scoping.

**The --opp/--org filter**:  I don't think it scopes the blast radius the way the PR description suggests. `_links_missing_uuid_by_server_and_domain()` pulls in all missing links with no filter, so when you intersect with the API keys (which are filtered by --opp/--org), you end up updating all users in the matched domain,  not just users belonging to the specified opportunity.

Concretely: if Opp X and Opp Y both live on domain-alpha, running --opp X will also update Opp Y's users. The flag is really just picking which API key to use (and therefore which domain to resolve), rather than limiting which users actually get touched.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
mkangia: I believe its only timeout if there are a lot of updates to be done or a lock on records being updated by another update happening on the environment.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
mkangia: Good point.

I agree we might end up with some links not having a UUID though it's not obvious how those should be fixed. These could be users/records that are now not active, however, removing data does not seem the right way forward. The decision about what to do with them stays with the organization they are a part of.

So, not addressing such records for now. 
Though I will consider running a query to check this count after the backfil.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
mkangia: Similar to https://github.com/dimagi/commcare-connect/pull/1283#discussion_r3467460984

I expect that even for some there is an API key, the key might have expired so fetching the UUID might fail.
The scope of this effort to backfil data but not address any gaps in the data for now.

mkangia:
Thanks @ajeety4 

> if Opp X and Opp Y both live on domain-alpha, running --opp X will also update Opp Y's users

The user should get updated only once. I intended for blast radius to refer to number of users getting updated and using opportunity and org was just a way to filter for less users but it seems opportunity was not a good choice.
Though the same overlap can happen for organizations too.

I can drop these filters, I don't intend to use them for running this command and were added a good to have. If needed they can be added in future.

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
mkangia: aha! 
The report should happen in either case but the wording was definitely off.

Fixed it here, https://github.com/dimagi/commcare-connect/pull/1283/commits/18aef88d503bf618d49ab05c3797e1e867e84679

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
mkangia: Actually went ahead and removed the filters https://github.com/dimagi/commcare-connect/pull/1283/commits/ae2040c19daccb843a24068a3c217ddd39a8ae4d

I don't intend to use them for the run that I need to do and it seems (as raised by Ajeet) that they were causing confusion.

mkangia:
Thanks for the review @ajeety4 @zandre-eng 

Addressed feedback, requesting for another pass

ajeety4 (commented):
Thanks for the changes :+1:

ajeety4 (approved):

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
zandre-eng: Thanks for sharing the context. It would be nice to scope this exception with specific classes then so that it's clear what we expect could fail here (or even just a comment to make it clear for another dev reading this). 

Also fine to leave this as is if you would prefer to throw a broad net over the exception catching here.

zandre-eng (approved):

commcare_connect/users/management/commands/backfill_hq_user_uuid.py
```diff
@@ -0,0 +1,201 @@
...
+        total = len(to_be_updated)
+        for start in range(0, total, batch_size):
+            batch = to_be_updated[start : start + batch_size]  # noqa: E203
+            self._bulk_update_with_retry(batch)
+            self.stdout.write(self.style.SUCCESS(f"Updated {min(start + batch_size, total)}/{total} records."))
+        return path
+
+    def _write_reference_file(self, to_be_updated):
+        timestamp = timezone.now().strftime("%Y%m%d_%H%M%S")
+        path = Path(tempfile.gettempdir()) / f"hq_user_uuid_backfill_{timestamp}.csv"
```
Charl1996, line 174: @mkangia  Will `tempfile` clean up automatically?

mkangia: Good point.

I deleted it manually on staging and I will do so on production as well once I have used it.
But also, I believe, temp files are deleted periodically by OS. Additionally, there is no sensitive info on this file.

PR #1286, opened 2026-06-17
Fix a time-related bug in audit report

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CI-728

This is not directly a fix to the above ticket, but as I debugged it, I found that the weekly performance report is not idempotent i.e. if it’s re-generated with same date-filters today, it should generate the same data, which it doesn’t. This is because the report doesn’t apply date-filters on all components of the calculation.

This fix calculates visit records filtered to the report's end date, so neither number can change when report is re-generated.

### Safety story

All tests pass.

### Automated test coverage

New test added.

### Labels & Review

### DIFF
--- a/commcare_connect/audit/chc_indicators.py
+++ b/commcare_connect/audit/chc_indicators.py
@@ -265,25 +265,25 @@ class WACoverageToVisitRatio(AuditCalculation):
         _eligible = ~Q(status__in=[WorkAreaStatus.EXCLUDED])
         wa_stats = WorkArea.objects.filter(opportunity_access=opportunity_access).aggregate(
             total_eligible=Count("id", filter=_eligible),
-            visited_count=Count(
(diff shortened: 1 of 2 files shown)


sravfeyn:
Open question: Should we regenerate the already generated reports?

sravfeyn:
Also another question: given this is a sensitive report, is there any QA plan for this feature?

hemant10yadav (approved):

hemant10yadav:
> Open question: Should we regenerate the already generated reports?

I don't think so. They might have already done some reviews and assigned tasks to the users.

> Also another question: given this is a sensitive report, is there any QA plan for this feature?

The report is hard to test manually because the minimum sample size thresholds are high.

Charl1996 (approved):

Charl1996:
> Open question: Should we regenerate the already generated reports?

Agreed with Hemant, but maybe good to check with Mary or so as well.

sravfeyn:
Closing this since, the visits could always come later to when the report gets generated.

PR #1288, opened 2026-06-18
add program redesign spec

calellowitz:
## Technical Summary

Tech Spec 1 of 3 for the permission and org structure updates [design doc](https://docs.google.com/document/d/1oORKTbPcjOUn2yr7jK5dbpvKEU8qNedBSdGpCl4QZh0/edit?tab=t.5bl3pzocqv7l)

This covers changes to the program structure. We will be adding a few new levels to the current PM/NM split, allowing the assignment of PM privileges per opportunity rather than a whole program (what we will call the "supervising org"), as well as sharing readonly views to as many orgs as desired. This spec should be self explanatory but requirements are in the linked doc, if more context is required

### DIFF
--- a/docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
+++ b/docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
@@ -1,0 +1,173 @@
+# Program / Opportunity Organization Role Restructuring
+
+**Date:** 2026-06-17
+**Status:** Design — pending review
+**Scope:** Subsystem #3 of 3 in the larger organization/permission rework. This spec covers only the program/opportunity role restructuring. The LLOEntity merge + org-merging (#1) and the signup/invite overhaul (#2) are separate specs.
+
+## Problem
+
+Today the organization that owns a `Program` (`Program.organization`, an org with `program_manager=True`) holds blanket "program manager" (PM) oversight over every `ManagedOpportunity` in that program. PM-ness is derived from `request.org.program_manager` + an admin membership and is checked in `program/utils.py`, `organization/decorators.py`, the `OrganizationMiddleware`, and several views/mixins.
+
+We need a richer set of org-to-program and org-to-opportunity relationships:
+
+- A program may have an optional **funder** organization with the same access as the program org.
+- A program may have any number of **watcher** organizations with read-only access to the program and its opportunities.
+- PM-level oversight is **pushed down** to a per-opportunity **supervising organization** (an additional delegate that gets PM-level access to that one opportunity).
+
+## Decisions (from brainstorming)
+
+1. **Supervising org is a separate new role**, distinct from `opportunity.organization` (the executing Network Manager). An opportunity therefore has both an executor and a supervisor, often different orgs.
+2. **Program org keeps blanket access** to all its opportunities; the supervising org is an *additional* delegate that also gets PM-level access to its opportunity.
+3. **Data model: dedicated relations (Approach A)** — direct FKs / M2M, matching the existing FK-based design, rather than generalized role tables.
+4. **Access = min(relationship ceiling, internal org role).** Manager/funder/supervisor mirror today's program-org behavior (admin→manage, member→standard, viewer→read). Watcher caps *every* member at read-only regardless of internal role.
+5. **Supervisor choices:** the program org, its funder, or orgs with an accepted `ProgramApplication` (participating Network Managers).
+6. **funder is creation-only** (immutable after the program is created). **watchers are editable** post-creation by users with manage access.
+
+## Out of scope
+
+- The `funder` boolean is added to `Organization` here because this spec depends on it; the rest of the `Organization` structural changes (LLOEntity merge, org merging, `verified`) live in spec #1.
+- No changes to opportunity execution, payments, or the Network Manager application flow beyond what permission rewiring requires.
+
+---
+
+## Data model changes
+
+### `Organization` (`commcare_connect/organization/models.py`)
+
+Add a boolean alongside the existing `program_manager`:
+
+```python
+funder = models.BooleanField(default=False)
+```
+
+Editable in Django admin (add to `OrganizationAdmin` list/fields). Default `False`.
+
+### `Program` (`commcare_connect/program/models.py`)
+
+```python
+funder = models.ForeignKey(
+    Organization,
+    null=True,
+    blank=True,
+    on_delete=models.SET_NULL,
+    related_name="funded_programs",
+)
+watchers = models.ManyToManyField(
+    Organization,
+    blank=True,
+    related_name="watched_programs",
+)
+```
+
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+)
+```
+
+- `funder` must reference an org with `funder=True` — enforced in `ProgramForm.clean()` and by limiting the form field queryset. Set only at creation (see UI section).
+- An org may not be both `funder` and a `watcher` of the same program — enforced in `ProgramForm.clean()`.
+
+### `ManagedOpportunity` (`commcare_connect/program/models.py`)
+
+```python
+supervising_organization = models.ForeignKey(
```
Charl1996, line 68: Only one supervisor org per opportunity?

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+    Organization,
+    blank=True,
+    related_name="watched_programs",
+)
+```
+
+- `funder` must reference an org with `funder=True` — enforced in `ProgramForm.clean()` and by limiting the form field queryset. Set only at creation (see UI section).
+- An org may not be both `funder` and a `watcher` of the same program — enforced in `ProgramForm.clean()`.
+
+### `ManagedOpportunity` (`commcare_connect/program/models.py`)
```
Charl1996, line 65: What do you think about renaming the existing `organization` field on the model to `executor_organization` so the dichotomy is clearer between these two orgs?

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+- PM-level oversight is **pushed down** to a per-opportunity **supervising organization** (an additional delegate that gets PM-level access to that one opportunity).
+
+## Decisions (from brainstorming)
+
+1. **Supervising org is a separate new role**, distinct from `opportunity.organization` (the executing Network Manager). An opportunity therefore has both an executor and a supervisor, often different orgs.
+2. **Program org keeps blanket access** to all its opportunities; the supervising org is an *additional* delegate that also gets PM-level access to its opportunity.
+3. **Data model: dedicated relations (Approach A)** — direct FKs / M2M, matching the existing FK-based design, rather than generalized role tables.
+4. **Access = min(relationship ceiling, internal org role).** Manager/funder/supervisor mirror today's program-org behavior (admin→manage, member→standard, viewer→read). Watcher caps *every* member at read-only regardless of internal role.
+5. **Supervisor choices:** the program org, its funder, or orgs with an accepted `ProgramApplication` (participating Network Managers).
```
Charl1996, line 23: > or orgs with an accepted `ProgramApplication` (participating Network Managers)

This is an interesting one for me. What does "participating" mean in this context? It seems like supervisor orgs have edit access on opportunities, so does that mean they might e.g. also participate in reviewing visits and so?

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
calellowitz: yep!

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
calellowitz: do you mean on the `Opportunity` model? I would probably go with `implementing_organization` to match the language in the design doc, but would be open to it. I am a bit concerned about the scope of that change though. I think we use the `organization` field for permission checks in a lot of places (across models), so it might be more complex than expected.

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
calellowitz: This is just the dropdown for the supervising org choices in the opportunity creation flow, so here it just means "are involved in the program in some way," but yes supervising orgs definitely participate. It's roughly equivalent to the PM now, but scoped only to a single opportunity (none of the program admin, just the admin level oversight of the single opportunity)

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
Charl1996: Yeah, that would increase scope a bit. We can probably defer that to some other time.

Charl1996 (approved):

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+
+### `Program` (`commcare_connect/program/models.py`)
+
+```python
+funder = models.ForeignKey(
+    Organization,
+    null=True,
+    blank=True,
+    on_delete=models.SET_NULL,
+    related_name="funded_programs",
```
mkangia, line 53: self note: a program can have only one funder but a funder can fund multiple programs, so this is funded_programs, makes sense.

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+
+```python
+supervising_organization = models.ForeignKey(
+    Organization,
+    on_delete=models.PROTECT,
+    related_name="supervised_opportunities",
+)
+```
+
+- **Non-nullable.** `ManagedOpportunity` uses multi-table inheritance, so this column lives on the child table (`program_managedopportunity`) and does **not** exist on the base `opportunity` table — making it non-null has no effect on unmanaged opportunities. Every managed opportunity must have a supervisor.
```
mkangia, line 75: @calellowitz 

does this date back to when we had "ManagedOpportunity" ?
how the approach/spec get revised? or the change is deferred to implementation?

docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Program create/edit (`ProgramForm`, `program/views.py: ProgramCreateOrUpdate`)
+
+- Add `funder` select — queryset filtered to `Organization.objects.filter(funder=True)`, excluding the program org. **Shown only on create**; on edit the field is rendered read-only / omitted (funder is immutable).
+- Add `watchers` multi-select — editable on both create and edit. Excludes the program org and the funder.
+- `clean()` validates: funder has `funder=True`; funder ∉ watchers.
+
+### Managed opportunity create/edit (`ManagedOpportunityInitForm`, `program/views.py`)
+
+- Add `supervising_organization` select — queryset = {program org} ∪ {funder if set} ∪ {orgs with an accepted `ProgramApplication` for this program}. Default = program org.
```
mkangia, line 153: > {program org}

Is it really needed for an opportunity to add it's program org as the supervising org? isn't that implicit already.

PR #1290, opened 2026-06-18
add signup update spec

calellowitz:
## Technical Summary

Tech Spec 3 of 3 for the permission and org structure updates [design doc](https://docs.google.com/document/d/1oORKTbPcjOUn2yr7jK5dbpvKEU8qNedBSdGpCl4QZh0/edit?tab=t.5bl3pzocqv7l)

This spec includes the signup workflow updates from the linked design doc. This corresponds to the flow charts at the bottom of the new signup flow page. Notably users are now able to create their own organization again, and users who signup without orgs are prompted to create one or seek an invite. Invites to an org are also updated to be more secure, to expire, and to be able to be sent to emails that have not yet created an account.

### DIFF
--- a/docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
+++ b/docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
@@ -1,0 +1,125 @@
+# Signup & Org-Invite Overhaul
+
+**Date:** 2026-06-17
+**Status:** Design — pending review
+**Scope:** Subsystem #2 of 3 in the organization/permission rework. Covers open org creation on signup, the no-org landing page, and a rebuilt org-invitation flow (new + existing users, 7-day expiry, accept-only-by-invited-email, admin notification on accept). The program/opportunity roles (#3) and the LLOEntity merge + org-merging (#1) are separate specs.
+
+## Dependencies & sequencing
+
+- **Depends on spec #1** for `Organization.verified` (consumed here) and for the simplified `OrganizationSelectOrCreateForm` (spec #1 removes the LLO-entity select from org creation). This spec only changes the *gate* and *defaults* of org creation, not the form's field structure.
+- Independent of spec #3.
+
+## Decisions (from brainstorming)
+
+- **Dedicated `OrganizationInvitation` model.** No `UserOrganizationMembership` is created until acceptance.
+- **`verified` is purely informational** — tracked/editable in Django admin only. No capability gating, **no UI banner**.
+- Any authenticated user may create an organization; new orgs default `verified=False` and the creator becomes ADMIN.
+
+---
+
+## Part A — Open org creation
+
+`organization/views.py: organization_create` currently requires `WORKSPACE_ENTITY_MANAGEMENT_ACCESS`.
+
+- **Remove** the `@permission_required(WORKSPACE_ENTITY_MANAGEMENT_ACCESS, ...)` decorator; keep `@login_required`. Any authenticated user may create an org.
+- New org: `verified=False` (default), creator added via `UserOrganizationMembership` as `ADMIN` (unchanged behavior).
+- No limit on orgs-per-user (creating multiple is allowed).
+- "Create on signup" is **offered, not forced**: a freshly-signed-up user with no membership lands on the no-org page (Part B), which links to org creation. Org creation is not a mandatory signup step.
+
+---
+
+## Part B — No-org landing page
+
+Today membership-less users are sent to `home` (`pages/home.html`); the `create_org_for_user` signal that tries to redirect them is a **no-op** (returning a redirect from a signal receiver does nothing) and is removed.
+
+- New view + template `organization/no_organization.html`: explains the user isn't a member of any organization and offers two paths — **(1)** reach out to your organization's administrator to be invited, and **(2)** create a new organization (links to `organization_create`).
+- Wire-up: `users/views.py: UserRedirectView.get_redirect_url()` — when `request.user.memberships` is empty, redirect to this page (replacing the current `reverse("home")`).
+- Remove the dead `create_org_for_user` signal (or repurpose it into the redirect logic; the redirect already lives in `UserRedirectView`, so deletion is cleanest).
+- The page is reachable by any authenticated user with no memberships; a user who later gains a membership is routed normally.
+
+---
+
+## Part C — Rebuilt invitation flow
+
+### `OrganizationInvitation` model (`organization/models.py`)
+
+```python
+class OrganizationInvitation(BaseModel):
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="invitations")
+    email = models.EmailField()
+    role = models.CharField(max_length=20, choices=UserOrganizationMembership.Role.choices,
+                            default=UserOrganizationMembership.Role.MEMBER)
+    token = models.CharField(max_length=64, unique=True, default=<secrets.token_urlsafe>)
+    invited_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name="sent_invitations")
+    expires_at = models.DateTimeField()        # set to created + 7 days on creation
+    accepted_at = models.DateTimeField(null=True, blank=True)
+```
+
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+- `expires_at` defaults to creation time + 7 days (set in `save()`/at creation).
+- Helper props: `is_expired` (`now > expires_at`), `is_accepted` (`accepted_at is not None`), `is_pending` (not expired, not accepted).
+- Index on `token` (unique already) and on `(organization, email)`.
+- The old `UserOrganizationMembership.invite_id` field becomes vestigial — drop it (and its index) in a migration; nothing in the new flow reads it.
+
+### Sending an invite (rewrite `MembershipForm` + `add_members_form`)
+
+- `MembershipForm` (rename to `OrganizationInvitationForm`): fields `email`, `role`. **Drop** the "user must already exist" validation — invites may target any email, new or existing.
```
zandre-eng, line 65: I wonder if a cap or rate limiting here per user would make sense. With open org creation (Part A) plus allowing invites to any arbitrary email, a malicious user could sign up -> create an org -> invite 10,000 arbitrary addresses, adding strain to our email service.

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
```diff
@@ -0,0 +1,125 @@
...
+- "Create on signup" is **offered, not forced**: a freshly-signed-up user with no membership lands on the no-org page (Part B), which links to org creation. Org creation is not a mandatory signup step.
+
+---
+
+## Part B — No-org landing page
+
+Today membership-less users are sent to `home` (`pages/home.html`); the `create_org_for_user` signal that tries to redirect them is a **no-op** (returning a redirect from a signal receiver does nothing) and is removed.
+
+- New view + template `organization/no_organization.html`: explains the user isn't a member of any organization and offers two paths — **(1)** reach out to your organization's administrator to be invited, and **(2)** create a new organization (links to `organization_create`).
+- Wire-up: `users/views.py: UserRedirectView.get_redirect_url()` — when `request.user.memberships` is empty, redirect to this page (replacing the current `reverse("home")`).
```
zandre-eng, line 36: Just confirming that this also means `home.html` should be deleted as part of this effort?

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
```diff
@@ -0,0 +1,125 @@
...
+
+- New token-based route **outside** `/a/<org_slug>/`: e.g. `path("invitations/<str:token>/accept/", ...)` (invitee may be unauthenticated and is not yet an org member).
+- `accept_invitation(request, token)` logic:
+  1. Look up invitation by `token` (404 if unknown). If `is_accepted` → friendly "already accepted" message. If `is_expired` → "this invite has expired, ask the admin to resend" message.
+  2. **Unauthenticated** + email has **no account** → send to allauth **signup** with the email pre-filled and locked to `invitation.email`; carry the token through so acceptance completes after signup.
+  3. **Unauthenticated** + email **has an account** → send to login (as that email), then return to acceptance.
+  4. **Authenticated**: enforce **accept-only-by-invited-email** — `request.user.email` must equal `invitation.email` (case-insensitive); otherwise reject with a clear message (e.g. "this invite was sent to a different address").
+  5. On success: create `UserOrganizationMembership(user, organization, role=invitation.role)` (idempotent if somehow already a member), set `accepted_at=now`, and **notify admins** (below). Redirect to `opportunity:list` for the org.
```
zandre-eng, line 83: Once a user has successfully accepted an invite and created a membership, do we still want to keep the invitation record around? I imagine yes as this could be useful for debugging, but just confirming here.

zandre-eng (approved):

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
calellowitz: yeah, as usual, I think we should keep them around until they start to cause problems. If disk space becomes a concern, we can add regular cleanup, but I think we are pretty far from that at this point.

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
calellowitz: Oh good point. Since that page is basically the same thing already, I think we can just reuse it. Maybe we can leave that to the implementer, but either a new page or the existing `home.html` seem reasonable to me.

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
calellowitz: That's a great question. I poked around HQ a bit and they don't seem to do that despite having a similar system in place, so I'm tempted to say that is a problem we can hold off on for now, but happy to also leave that up to the implementer

zandre-eng: It could be worth having the implementer at least see how complex it would be to implement this, as I think it would be a good defensive addition. As they say "prevention is the best cure". With that said, I agree and am happy to defer this to the implementer.

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
```diff
@@ -0,0 +1,125 @@
...
+    email = models.EmailField()
+    role = models.CharField(max_length=20, choices=UserOrganizationMembership.Role.choices,
+                            default=UserOrganizationMembership.Role.MEMBER)
+    token = models.CharField(max_length=64, unique=True, default=<secrets.token_urlsafe>)
+    invited_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name="sent_invitations")
+    expires_at = models.DateTimeField()        # set to created + 7 days on creation
+    accepted_at = models.DateTimeField(null=True, blank=True)
+```
+
+- `expires_at` defaults to creation time + 7 days (set in `save()`/at creation).
```
Charl1996, line 58: Any particular reason we don't calculate this on the fly (we persist `date_created` so calculating it is trivial)?

docs/superpowers/specs/2026-06-17-signup-and-invite-overhaul-design.md
```diff
@@ -0,0 +1,125 @@
...
+  2. **Unauthenticated** + email has **no account** → send to allauth **signup** with the email pre-filled and locked to `invitation.email`; carry the token through so acceptance completes after signup.
+  3. **Unauthenticated** + email **has an account** → send to login (as that email), then return to acceptance.
+  4. **Authenticated**: enforce **accept-only-by-invited-email** — `request.user.email` must equal `invitation.email` (case-insensitive); otherwise reject with a clear message (e.g. "this invite was sent to a different address").
+  5. On success: create `UserOrganizationMembership(user, organization, role=invitation.role)` (idempotent if somehow already a member), set `accepted_at=now`, and **notify admins** (below). Redirect to `opportunity:list` for the org.
+- **Email verification:** because the token was delivered to `invitation.email`, acceptance proves control of that address — mark the allauth `EmailAddress` verified on acceptance to avoid a redundant verification step. *(Flag for review — alternative is to keep standard mandatory verification.)*
```
Charl1996, line 84: > acceptance proves control of that address

Agreed here, but how would that work in code? Do we populate the `next` query param so in the signup view we mark the user's email as verified (if `next` query param exists and token is valid) and redirect to the "accept your invite" page? Or something else?

The standard flow:
(unauthenticated) user clicks link -> redirect to sign up -> confirm email -> user must click accept link again (now authenticated).

The above flow is simpler in code, but might be more confusing for the user - I'm not sure.

Charl1996 (approved):
LGTM. Just one [query](https://github.com/dimagi/commcare-connect/pull/1290/changes#r3473570273) I want to flag.

PR #1293, opened 2026-06-18
handle bom char when importing csv generated on windows machines

sravfeyn:
## Product Description

https://dimagi.atlassian.net/browse/CI-711

Fixes a CSV import bug

## Technical Summary

This fixes a bug where first column of CSV (exported as CSV from Windows Excel) is ignored due additional Bom character inserted by windows excel.

### Safety story

Tested locally

### Automated test coverage

Added
### QA Plan

NA

### Labels & Review

### DIFF
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -43,7 +43,10 @@ class WorkAreaCSVImporter:
 
     def __init__(self, opp_id, csv_source):
         self.opp_id = opp_id
-        self.csv_source = csv_source
+        content = csv_source.read()
+        if content.startswith("﻿"):  # strip UTF-8 BOM added by Windows Excel CSV export
+            content = content[1:]
+        self.csv_source = io.StringIO(content)
         self.errors = defaultdict(list)
         self.seen_slugs = set()
         self.created_count = 0
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -43,10 +43,7 @@ class WorkAreaCSVImporter:
 
     def __init__(self, opp_id, csv_source):
         self.opp_id = opp_id
-        content = csv_source.read()
-        if content.startswith("﻿"):  # strip UTF-8 BOM added by Windows Excel CSV export
-            content = content[1:]
-        self.csv_source = io.StringIO(content)
+        self.csv_source = csv_source
         self.errors = defaultdict(list)
         self.seen_slugs = set()
         self.created_count = 0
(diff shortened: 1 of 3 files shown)


Charl1996 (approved):
> Tested locally

Did you test this on a windows machine?

commcare_connect/microplanning/tests/test_tasks.py
```diff
@@ -31,9 +31,11 @@ class TestWorkAreaCSVImporter:
         "State",
     ]
 
-    def build_csv(self, rows, headers=None):
+    def build_csv(self, rows, headers=None, bom=False):
         headers = headers or self.HEADERS
         output = io.StringIO()
+        if bom:
+            output.write("﻿")
```
snopoke: You should use a constant and the escaped unicode char to make this clearer: `UTF8_BOM = "\ufeff"`

commcare_connect/microplanning/tasks.py
```diff
@@ -43,7 +43,10 @@ class WorkAreaCSVImporter:
 
     def __init__(self, opp_id, csv_source):
         self.opp_id = opp_id
-        self.csv_source = csv_source
+        content = csv_source.read()
+        if content.startswith("﻿"):  # strip UTF-8 BOM added by Windows Excel CSV export
+            content = content[1:]
+        self.csv_source = io.StringIO(content)
```
snopoke: instead of handling this here you should read the file / stream using the `utf-8-sig` encoding which will consume the BOM automatically.

See https://docs.python.org/3/library/codecs.html#encodings-and-unicode and https://theproductguy.in/blogs/csv-encoding-issues/

snopoke (commented):
To avoid having to fix this again, you should use the same fix on any other places that read CSV files.

sravfeyn:
Thanks Simon, I have updated code per your comments.

commcare_connect/microplanning/tests/test_tasks.py
```diff
@@ -162,6 +162,20 @@ def test_random_column_order(self, opportunity):
 
         assert result["created"] == 1
 
+    def test_utf8_bom_csv(self, opportunity):
+        UTF8_BOM = "﻿"
```
snopoke: Use the escaped unicode representation for clarity
```suggestion
        UTF8_BOM = "\ufeff"
```

commcare_connect/microplanning/tests/test_tasks.py
```diff
@@ -162,6 +162,20 @@ def test_random_column_order(self, opportunity):
 
         assert result["created"] == 1
 
+    def test_utf8_bom_csv(self, opportunity):
```
snopoke: This test isn't actually testing the code you changed. It's just duplicating the code.

commcare_connect/microplanning/tests/test_tasks.py
sravfeyn: Updated

commcare_connect/microplanning/tests/test_tasks.py
sravfeyn: Yeah, I was testing the common part `.decode("utf-8-sig")` that I have added in multiple imports. But that's standard library thing that doesn't need to be tested. I have replaced this now with a test that tests the task itself.

sravfeyn:
@snopoke Thanks for the review, I have updated the code

snopoke (approved):

PR #1296, opened 2026-06-18
feat(audit): download weekly performance report as CSV

Charl1996:
## Product Description

Adds a CSV download for the Weekly Performance Report. A download button sits inline with the worker-name filter; clicking it downloads the report as a CSV. The export includes all data elements, respects the active worker-name filter, and appends each calculation's acceptable range to its column header (e.g. `Visit Accuracy (0.5 - 1.0)`).

## Technical Summary

Ticket: https://dimagi.atlassian.net/browse/CCCT-2483

The report is already computed, so the export is served synchronously via a single `StreamingHttpResponse` GET view . Entries are read with `.iterator()` and written one CSV row at a time.

### Safety story

Read-only export, gated behind the `WEEKLY_PERFORMANCE_REPORT` flag and program-manager access; no impact on existing data.

### Automated test coverage

Unit tests cover the streamed CSV (filtered output, `N/A` handling, reference-range header) and the export view (streaming response + filter, 404 when flag disabled).

### QA Plan

No formal QA.


commcare_connect/audit/tests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+def make_audit_entry():
+    """Factory fixture: build an AuditReportEntry for ``report`` with a single
+    ``calc_a`` result, for a worker named ``name``.
+    """
+
+    def _make(report, name, value, has_data=True, in_range=True):
+        access = OpportunityAccessFactory(user=UserFactory(name=name))
+        return AuditReportEntryFactory(
+            audit_report=report,
+            opportunity_access=access,
```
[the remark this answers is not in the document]
Charl1996: Not an issue. The report get its opportunity from the access record.

zandre-eng (approved):

commcare_connect/audit/views.py
```diff
@@ -250,3 +240,63 @@ def audit_report_complete(request, org_slug, opp_id, audit_report_id):
...
+@require_POST
+def export_audit_report(request, org_slug, opp_id, audit_report_id):
+    opportunity = request.opportunity
+    _require_feature_flag(opportunity)
+    report = get_object_or_404(AuditReport, audit_report_id=audit_report_id, opportunity=opportunity)
+
+    name_filter = request.POST.get("filter", "").strip()
+    task = export_audit_report_task.delay(str(report.audit_report_id), name_filter)
+
+    return _render_audit_export_status(request, org_slug, opp_id, audit_report_id, task.id)
```
hemant10yadav: How do you feel about streaming the CSV instead, given that this report is already calculated? Using a `StreamingHttpResponse` would allow us to replace the Celery task, S3 storage, status-polling endpoint, and task-ownership checks with a single view.

commcare_connect/opportunity/views.py
```diff
@@ -604,8 +606,10 @@ def ownership_check(request, task_meta):
 @require_GET
 def download_export(request, org_slug, task_id):
     task_meta = AsyncResult(task_id)._get_task_meta()
-    opportunity_id = task_meta.get("args")[0]
-    opportunity = get_opportunity_or_404(org_slug=org_slug, pk=opportunity_id)
+    args = task_meta.get("args") or []
```
hemant10yadav: Can we use the public API here instead of `_get_task_meta()`?

```python
task = AsyncResult(task_id)
args = task.args or []
```

`task.args` should give us the same information without relying on a private method.

hemant10yadav (commented):
Looks good overall. I have a few questions:

commcare_connect/audit/views.py
Charl1996: I was leaning towards that also initially but decided against it because the ticket particularly asks for a download experience similar to invoices download (which has a progress bar). I didn't feel strongly enough about the streaming response to challenge the progress bar approach.

Charl1996: Although now that you mention it I do think it is a simpler approach.

Charl1996: See https://github.com/dimagi/commcare-connect/pull/1296/changes/a06788e6000ff3d94273bd693b88090c8264eb36. It's a large refactor, so probably better to view the PR [from here](https://github.com/dimagi/commcare-connect/pull/1296/changes).

commcare_connect/audit/services.py
```diff
@@ -1,12 +1,80 @@
...
+def stream_audit_report_csv(report, name_filter=""):
+    columns = column_specs(list(report.entries.all()))
+    writer = csv.writer(EchoWriter())
+
+    yield writer.writerow([gettext("Connect Worker"), *_column_headers(columns)])
+
+    entries = entries_for_export(report, name_filter)
+    for entry in entries.iterator(chunk_size=STREAM_CHUNK_SIZE):
+        cells = [_export_cell_value(entry.results, name) for name, _label in columns]
+        yield writer.writerow([entry.opportunity_access.user.name, *cells])
```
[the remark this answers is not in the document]
Charl1996: I think this is not a problem as this type of text will always only be injected after a text title

commcare_connect/audit/services.py
```diff
@@ -1,12 +1,80 @@
...
+from commcare_connect.audit.calculations import get_registered_calculations
 from commcare_connect.audit.models import AuditReport, AuditReportEntry
 from commcare_connect.opportunity.models import Opportunity, OpportunityAccess
+from commcare_connect.utils.file import EchoWriter
+
+STREAM_CHUNK_SIZE = 2000
+
+
+def stream_audit_report_csv(report, name_filter=""):
+    columns = column_specs(list(report.entries.all()))
```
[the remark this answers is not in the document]
Charl1996: Ah, good catch.

hemant10yadav (approved):
Thanks!

hemant10yadav (approved):

commcare_connect/audit/views.py
```diff
@@ -249,3 +237,23 @@ def audit_report_complete(request, org_slug, opp_id, audit_report_id):
...
+
+@opportunity_required
+@org_program_manager_required
+@require_GET
+def export_audit_report(request, org_slug, opp_id, audit_report_id):
+    opportunity = request.opportunity
+    _require_feature_flag(opportunity)
+    report = get_object_or_404(AuditReport, audit_report_id=audit_report_id, opportunity=opportunity)
+
+    name_filter = request.GET.get("filter", "").strip()
```
zandre-eng: Should this `name_filter` be removed? I'm assuming this is related to the worker search bar, however that has been switched to a multi-select dropdown which is being fetched below with `request.GET.getlist("worker")`.

zandre-eng (approved):
Just one query, but otherwise looks good.

commcare_connect/audit/views.py
Charl1996: Yes, you're right. Thanks for catching that. That was the previous implementation.

zandre-eng (approved):

PR #1294, opened 2026-06-18
CCCT-2505: Count only PM-agreed visits for managed opp billable count

ajeety4:
## Product Description

No user-facing change. Fixes how billable visit counts are computed for **managed opportunities** so only PM-agreed visits count, plus regression tests for the review-triggered payment accrual flow.

## Technical Summary

Ticket: [CCCT-2505](https://dimagi.atlassian.net/browse/CCCT-2505)

Managed-opp billable count now uses the PM `agree` count instead of the NM `approved` count, so a visit that is NM-approved but not yet PM-agreed no longer inflates `saved_approved_count` / payment accrual. Non-managed opps are unchanged. Includes characterization tests for the bug and regression tests asserting that agreeing/disagreeing a visit always triggers the CompletedWork status + payment accrual recompute, on both the UI and bulk-CSV review paths.

> Note: the one-time backfill command (`fix_managed_opp_approved_counts`) for already-stale rows is intentionally **not** in this PR — it ships separately so it can be reverted after running in prod.

### Safety story

Logic change is scoped to managed opportunities only (`opportunity.managed` branch in `CompletedWorkUpdater`); non-managed flows take the identical pre-existing path. The accrual recompute reads from current visit/review state (source of truth), so it is idempotent. Existing managed-opp rows with stale counts are addressed separately by the backfill command (not in this PR).

### Automated test coverage

- `test_cw_status_updates.py` — managed-opp status/payment computation (agree-gated billable count).
- `test_views.py::test_user_visit_review` — UI agree/disagree triggers accrual with correct users.
- `test_visit_import.py::TestBulkReviewVisitImport` — bulk review triggers (and suppresses) `bulk_update_payment_accrued.delay`.

### QA Plan
https://dimagi.atlassian.net/browse/QA-8560

### Labels & Review

- [x] The set of people pinged as reviewers is appropriate for the level of risk of the change

[CCCT-2505]: https://dimagi.atlassian.net/browse/CCCT-2505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1297, opened 2026-06-18
CCCT-2505: One-time backfill for managed opp unbilled completed works

ajeety4:
## Product Description

No user-facing change. One-time data-correction command to backfill `saved_approved_count` / payment accrual for managed-opportunity completed works that were over-counted by the pre-fix logic for **duplicate works**.

> **Stacked on #1294.** Review/merge that fix first — this command depends on the PM-agreed billable logic being live.

## Technical Summary

Ticket: CCCT-2505

- `fix_managed_opp_approved_counts` re-runs the live `update_status` over unbilled, duplicated (`saved_approved_count > 1`) completed works on active managed opps, so the correction inherits the PM-agreed billable logic from #1294. 
- Idempotent. 
- Required `--dry-run` (read-only preview) / `--no-dry-run` (apply) flag. One-time command — ship separately, run once after #1294 deploys.

### Safety story

Correction logic is not reimplemented — it delegates to the same `update_status` used in production, scoped to managed + active + unbilled + `count > 1` rows. Idempotent, so a partial/interrupted run is safe to re-run. Each access is wrapped in its own transaction. `--dry-run` previews the candidate set without any writes.

This was tested on staging , and it succesfully corrected an unbilled completed work.

### Automated test coverage

`test_management_commands.py::TestFixManagedOppApprovedCounts` covers: count corrected (incl. access-level `payment_accrued` total), dry-run makes no changes, re-run idempotency, and negative cases — `count == 1`, billed, inactive, and non-managed works are all left untouched.

### QA Plan
None.

- On staging, seed a managed opp with a duplicated unbilled completed work (`saved_approved_count = 2`, one PM-agreed visit), run `--dry-run` (confirm it's listed, no change), then `--no-dry-run` (confirm count → 1 and access `payment_accrued` recomputed).

### Labels & Review


commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
```diff
@@ -0,0 +1,84 @@
...
+        ).select_related("opportunity")
+
+        affected = 0
+        for access in accesses:
+            works = list(
+                CompletedWork.objects.filter(
+                    opportunity_access=access,
+                    invoice__isnull=True,
+                    saved_approved_count__gt=1,  # the correction is needed only for duplicated completed works
+                ).select_related("payment_unit")
```
zandre-eng: This will currently do a DB query for each access in `accesses`. You could do a prefetch on the `OpportunityAccess` query:
```
from django.db.models import Prefetch
accesses = OpportunityAccess.objects.filter(
    opportunity__managed=True,
    opportunity__active=True,
).select_related("opportunity").prefetch_related(
    Prefetch(
        "completedwork_set",
        queryset=CompletedWork.objects.filter(invoice__isnull=True, saved_approved_count__gt=1),
    )
)
```
And then here in the loop you can access by:
```
for access in accesses:
    completed_works = access.completedwork_set.all()
```

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
```diff
@@ -0,0 +1,84 @@
...
+            )
+        return len(works)
+
+    def _recalculate(self, access, works):
+        """Recompute saved counts, logging each before/after delta.
+
+        Wrapped in a transaction so a crash can't leave a single access half-corrected.
+        """
+        work_ids = [work.id for work in works]
+        before = {work.id: (work.saved_approved_count, work.saved_payment_accrued) for work in works}
```
zandre-eng: Nit: These two could be collapsed into a single loop:
```
before = {work.id: (work.saved_approved_count, work.saved_payment_accrued) for work in works}
work_ids = before.keys()
```

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
```diff
@@ -0,0 +1,84 @@
...
+
+        with transaction.atomic():
+            update_status(
```
zandre-eng: I'm wondering if there is a risk of a race condition between querying the completed works and updating their status here. If a user adds a completed work to an invoice in-between the two then the status will get incorrectly updated here.


This feels like a low probability edge case, but curious to hear if you think this should be accounted for?

zandre-eng (commented):
Looks good. Just one question about a race condition, but otherwise non-blocking feedback.

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
ajeety4: Addressed in https://github.com/dimagi/commcare-connect/pull/1297/commits/c0b6fbd9244aa2364a1aac13894ced9d719d204a

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
ajeety4: Good point.
As the completed work set count will be low in practice, we can add prefetch and not worry about memory. Also added chunk as a safety net.

Added in https://github.com/dimagi/commcare-connect/pull/1297/commits/c0b6fbd9244aa2364a1aac13894ced9d719d204a

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
ajeety4: Good catch. Agreed it's extremely low-probability but since it touches billing data I went ahead and closed the window rather than leave it to chance.
`_recalculate` now re-fetches the works inside the transaction with select_for_update and re-applies the `invoice__isnull=True` filter before recalculating.
https://github.com/dimagi/commcare-connect/pull/1297/commits/f8658603fae52045d2b92be7111b5dfc0f60bbba

zandre-eng (dismissed):
Thanks for following up on my feedback, looks good!

commcare_connect/opportunity/management/commands/fix_managed_opp_approved_counts.py
```diff
@@ -0,0 +1,98 @@
...
+        (update_payment_accrued_for_user) so the access's payment_accrued isn't clobbered by a
+        concurrent recompute reading a stale sum.
+        """
+        before = {work.id: (work.saved_approved_count, work.saved_payment_accrued) for work in works}
+        work_ids = before.keys()
+
+        with cache.lock(f"update_payment_accrued_lock_{access.id}", timeout=900), transaction.atomic():
+            still_uninvoiced = (
+                CompletedWork.objects.filter(id__in=work_ids, invoice__isnull=True)
+                .select_for_update(of=("self",))
```
Charl1996, line 80: First time I've seen this - nice!

zandre-eng (approved):

ajeety4:
Did an empty force push to trigger the stuck CI job

zandre-eng (approved):

PR #1312, opened 2026-06-19
Web Claude Spec: Trigger an OCS conversation from a task in Connect

pxwxnvermx:
## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2467)

### Labels & Review

### DIFF
--- a/docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
+++ b/docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
@@ -66,8 +66,8 @@ Relevant existing code (all in `commcare_connect/opportunity/` unless noted):
 |---|---|
 | **Outbound auth (Connect→OCS)** | **OAuth2 via django-allauth**, using the logged-in user's connected OCS account. No static `X-api-key`, no service credential. |
 | Why not `X-api-key` | Reversed in favor of per-user scoped, revocable OAuth tokens; avoids a pasted shared secret. |
-| Why not a background service flow | OCS offers **no `client_credentials` grant** (only Authorization Code + PKCE and refresh). A non-interactive flow is not available, which is what pushed the trigger to be user-facing (below). |
-| Trigger flow | **Synchronous and user-facing** within the assignment request, using the acting user's OCS token. No Celery task, no retry, no background reconciliation. |
+| Why user-facing (not background) | A background trigger **is** technically possible by storing and refreshing the user's token (OCS issues refresh tokens — confirmed by the OCS team in PR review). We still chose a **synchronous, user-facing** trigger: it avoids persisting/refreshing long-lived tokens and the refresh-token rotation races that come with background workers, and it gives the assigning user immediate success/failure feedback. |
+| Trigger flow | **Synchronous and user-facing** within the assignment request, using the acting user's OCS token. No Celery task, no retry. |
 | On trigger failure | The per-assign transaction **rolls back** — no `AssignedTask` persisted; the user sees the error. A failed trigger creates nothing on OCS, so nothing dangles. |
 | Bulk assignment transactions | **Refactor the audit bulk path to one transaction per `assign()`** so a failure (or failed trigger) rolls back only that one task, not the batch. |
 | Chatbot access mismatch | If the acting user's OCS token can't access the configured chatbot/team, **let the OCS call fail and surface the error** to the user. No pre-validation of team membership. |
@@ -104,8 +104,8 @@ endpoint.
    the OCS error. We do not attempt to guarantee team consistency.
 3. **Rare ambiguous-failure dangling.** If OCS creates a conversation but the HTTP call errors/
    times out before responding, we roll back the `AssignedTask` while the conversation exists on
-   OCS. With a synchronous, no-background-reconciliation design this is unrecoverable but rare;
-   accepted.
+   OCS. The completion-reconciliation job can't recover this case (no task and no `ocs_session_id`
+   were persisted to reconcile against), so it remains unrecoverable but rare; accepted.
 
 ## Data Model Changes
 
@@ -134,11 +134,14 @@ ocs_chatbot_id = models.CharField(max_length=255, null=True, blank=True)  # OCS 
 ### `AssignedTask` additions
 
 ```python
-connect_channel_id = models.CharField(max_length=255, null=True, blank=True)
+connect_channel_id = models.CharField(max_length=255, null=True, blank=True)  # OCS `channel`
+ocs_session_id = models.CharField(max_length=255, null=True, blank=True)       # OCS `session_id`
 ```
 
-- Stores the channel/session reference returned by the OCS `trigger_bot` response, written
-  synchronously at assignment time.
+- Both are taken from the OCS `trigger_bot` response and written synchronously at assignment
+  time. We store **both** (per PR review): `connect_channel_id` is the messaging channel, and
+  `ocs_session_id` lets us cross-correlate an assigned task to its OCS session — needed for the
+  completion-reconciliation fallback (below) and any future session lookups.
 - `completed_at` and `status` are reused unchanged — for OCS tasks they are set by the
   callback rather than by xform processing.
 
@@ -182,8 +185,8 @@ Notes from the OCS schema:
 - List endpoint is `/api/v2/chatbots/` (NOT `/api/chatbots/` as the requirements draft said) and
   is **paginated** — `list_chatbots` must page through results.
 - `experiment` is a UUID (the chatbot id).
-- `trigger_bot` response fields: `session_id`, `url`, `team`, `channel`. Our `connect_channel_id`
-  will store **`channel`** (confirm vs `session_id` during implementation).
+- `trigger_bot` response fields: `session_id`, `url`, `team`, `channel`. We store **both**
+  `channel` (→ `connect_channel_id`) and `session_id` (→ `ocs_session_id`).
 - `trigger_bot` also accepts `start_new_session` (bool) and `prompt_text`/`message_text`. The
   bare payload may not actually *start* a conversation; confirm whether we need
   `start_new_session=true` and/or an initial message.
@@ -208,7 +211,8 @@ Synchronous and user-facing, inside `AssignedTask.assign(...)` (which already op
 2. **If `task_type.type == ocs`:**
    - Call `ocs.trigger_bot(acting_user, ...)` **synchronously** with the payload above, using
      the acting user's OCS token.
-   - On success: set `assigned_task.connect_channel_id` from the response and save.
+   - On success: set `assigned_task.connect_channel_id` and `assigned_task.ocs_session_id`
+     from the response and save.
    - On failure (`OCSAPIException` / httpx errors / token-not-connected / chatbot-not-accessible):
      let it propagate so the per-assign atomic block rolls back — **no row persisted**, and the
      user sees the error. A failed trigger created nothing on OCS.
@@ -266,10 +270,39 @@ Behavior:
 - Validation via a DRF serializer; completion logic in a standalone function (testable without
   HTTP, per repo convention).
 
-**Authentication — OPEN DECISION.** Connect must issue its own credential for OCS to present
-(the OAuth token above is *OCS's* — it authenticates Connect→OCS, not the reverse). Leaning
-toward a Connect-issued key/secret in a request header, validated by a small DRF auth/permission
-class. See Open Decisions.
+**Authentication — RESOLVED (OAuth2, Client Credentials grant).** OCS authenticates to Connect
+the same way **CommCare HQ authenticates to the Form Processor** today: OAuth2 with the
+**Client Credentials** grant. OCS is registered as an OAuth2 client of Connect (Connect is
+already an OAuth2 provider via `oauth2_provider`), and the callback view reuses the
+`FormReceiver` pattern — `OAuth2Authentication` + a read/write scope check. This keeps the
+inbound auth consistent with the existing HQ→Connect integration and requires no bespoke
+key-management. (Confirmed with the OCS team in PR review; OCS will add the matching OAuth
+support — tracked in [open-chat-studio#2179](https://github.com/dimagi/open-chat-studio/issues/2179).)
+
+**How OCS decides a conversation is "done" (OCS-side, informational).** The completion callback
+is fired from OCS via one of their pipeline mechanisms — custom actions (a bot "tool" the LLM
+invokes), Python nodes (external API call), a router node evaluating conversation state, or a
+timeout trigger. Choosing/among these is the OCS team's configuration concern, outside Connect's
+scope; Connect only needs to receive the resulting callback.
+
+**Completed-task behavior (confirmed in review).** Connect does not clean up completed assigned
+tasks today, and the API continues to return them to mobile. OCS channels also remain alive
+after completion (no archival mechanism currently). No change to this behavior here.
+
+### Completion reconciliation (fallback for missed callbacks)
+
+The callback is the primary completion signal, but it can be lost (OCS fails to send it, it
+doesn't reach us, or Connect errors while consuming it). To get definitive closure on in-flight
+OCS tasks, add a **periodic reconciliation** (Celery beat, e.g. daily):
+
+- Find OCS `AssignedTask`s still `assigned` that have an `ocs_session_id`.
+- For each, fetch the OCS session status via the session-retrieve endpoint
+  (`GET /api/v1/sessions/{id}/` — see OCS docs) using the assigning user's OAuth token.
+- If OCS reports the session complete, call `mark_completed(...)` (idempotent with the callback).
+
+This is the one place a background OCS call is acceptable — it's read-only status polling, not
+the trigger, so it doesn't reintroduce the dangling-conversation risk. (Suggested by the OCS
+team in PR review.)
 
 ## Task Completion Notification
 
@@ -342,8 +375,12 @@ responses; prefer fixtures.
   `AssignedTask`, a fixture for a user with a connected OCS `SocialToken`.
 - **OCS client:** `list_chatbots` (pagination) and `trigger_bot` — success and failure
   (timeout / non-2xx → `OCSAPIException`); Bearer auth header; token refresh on expiry.
-- **`AssignedTask.assign` (OCS):** success stores `connect_channel_id`; trigger failure rolls
-  back so no `AssignedTask` exists; not-connected user blocked; Re-Learn path unchanged.
+- **`AssignedTask.assign` (OCS):** success stores `connect_channel_id` **and `ocs_session_id`**;
+  trigger failure rolls back so no `AssignedTask` exists; not-connected user blocked; Re-Learn
+  path unchanged.
+- **Completion reconciliation:** the periodic job finds `assigned` OCS tasks with an
+  `ocs_session_id`, polls OCS session status, and `mark_completed`s those OCS reports done;
+  idempotent with the callback; ignores tasks without a session id.
 - **Bulk audit path:** per-assign isolation — one failing task doesn't roll back the others;
   partial-success messaging.
 - **Callback logic:** marks completed; username mismatch rejected; optional `completed_at`;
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,298 @@
...
+  conversation; confirm whether we need `start_new_session=true` and/or an initial message.
+
+`trigger_bot` request body:
+
+```json
+{
+  "identifier": "<flw username>",
+  "experiment": "<task_type.ocs_chatbot_id>",
+  "platform": "commcare_connect",
+  "participant_data": { "taskId": "<assigned_task.assigned_task_id (UUID)>" }
```
snopoke: nit: namespace taskId

```suggestion
  "participant_data": { "connectTaskId": "<assigned_task.assigned_task_id (UUID)>" }
```

snopoke (commented):
I didn't review all of it closely. I'd like to understand more about the decision to use an API key instead of oauth for the authentication.

From my read of the doc, a task is always manually triggered by a user which means that the user is always available to authenticate the request if necessary. Using oauth the user would never have to create an API key. CCC could authenticate the first time and store the oauth token and refresh token. If the token is expired and can't be refreshed the user could re-authenticate.

pxwxnvermx:
> I'd like to understand more about the decision to use an API key instead of oauth for the authentication.
 From my read of the doc, a task is always manually triggered by a user which means that the user is always available to authenticate the request if necessary. Using oauth the user would never have to create an API key. CCC could authenticate the first time and store the oauth token and refresh token. If the token is expired and can't be refreshed the user could re-authenticate.

I was exploring options to do async background trigger_bot API call in case of failures. That did not seem possible with OAuth.

snopoke:
> > I'd like to understand more about the decision to use an API key instead of oauth for the authentication.
> > From my read of the doc, a task is always manually triggered by a user which means that the user is always available to authenticate the request if necessary. Using oauth the user would never have to create an API key. CCC could authenticate the first time and store the oauth token and refresh token. If the token is expired and can't be refreshed the user could re-authenticate.
> 
> I was exploring options to do async background trigger_bot API call in case of failures. That did not seem possible with OAuth.

It is definitely possible if you have a [refresh token](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5) (which OCS does provide).

pxwxnvermx:
> It is definitely possible if you have a [refresh token](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5) (which OCS does provide).

Thanks. I will update the spec to use Oauth2 for OCS API requests.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,322 @@
...
+- Created via a `HQApiKeyCreateForm`-style form; the view sets `api_key.user = request.user`
+  (mirrors `add_api_key`, `views.py:3280`).
+- Plaintext storage, matching the `HQApiKey` precedent (not changing the security posture).
+
+### `TaskType` additions
+
+```python
+class TaskTypeChoices(models.TextChoices):
+    RELEARN = "relearn", gettext("Re-Learn")
+    OCS = "ocs", gettext("OCS")
```
shubham1g5, line 120: nit: thoughts on generalising this more to call it `messaging` ? I am not sure if we want to be platform agnostic here.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,322 @@
...
+
+- Default `relearn` so existing rows are unaffected (data migration not required; default
+  covers backfill).
+- For OCS task types: `app` stays `opportunity.deliver_app` (keeps them visible in the
+  existing `TaskTable` filter), `slug` is generated from `ocs_chatbot_id`, `case_property`
+  stays null. The `(app_id, slug)` unique constraint then naturally prevents the same chatbot
+  being added twice to one app.
+- `ocs_chatbot_id` is required when `type == ocs` (enforced in the form's `clean()`).
+
+### `AssignedTask` additions
```
shubham1g5, line 134: Once a task is completed, does web keep on sending the tasks in the API response to mobile ? Relatedly, do we do any cleanup for completed assigned tasks today ?

pxwxnvermx: Currently all tasks are sent through the API. We don't do any cleanup for completed tasks.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: I think the Tasks are heavily tied to OCS on Connect. If we refactor them to trigger messaging directly only then it makes sense to call it `messaging`.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
shubham1g5: Got it, another related aquestion -> Once the OCS task is completed, does the channel stays alive for user to interact with or it gets archived or something similar ?

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,388 @@
...
+These require input from outside this design — primarily the OCS team — and must be resolved
+before the dependent work can be finalized.
+
+1. **Callback authentication (OCS→Connect).** How does OCS authenticate to Connect's
+   completion callback? Connect must issue its own credential for OCS to present (the OAuth
+   token in this design is *OCS's* — it authenticates Connect→OCS, not the reverse). This is a
+   cross-team decision: the mechanism has to be something the OCS team can implement and
+   configure on their side. Current lean is a Connect-issued key/secret in a request header,
+   validated by a small DRF auth/permission class — but the choice depends on OCS-team input.
+   Blocks implementation of the callback endpoint.
```
pxwxnvermx: @snopoke Looking for your inputs on this.
I was trying to figure out how OCS can auth to Connect for marking task completions.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: The channels remain alive for the user. There is no way for us currently to mark channels as archived after task completion.

shubham1g5: Great, thanks for confirming.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,388 @@
...
+**Ticket:** CCCT-2467
+**Date:** 2026-06-19
+**Status:** Draft (open decisions — see Open Decisions)
+
+## Summary
+
+Connect lets Network Managers (NMs) assign tasks to frontline workers (FLWs). Today a
+"task" is effectively a CommCare re-learn activity tracked via xform submissions. This
+feature adds a second kind of task: an **OCS (Open Chat Studio) conversation**. When a user
+assigns an OCS task type to an FLW, Connect triggers a chatbot conversation in OCS for that
```
Charl1996, line 12: Does the bot initiate the conversation at the FLW's side based on the trigger?

pxwxnvermx: Yep, when the API request to OCS is made it triggers the bot to send the user a message using Connect messaging.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,388 @@
...
+  exception *outside* it rolls back only that task and keeps the request transaction valid, so
+  successful assigns still commit. The failed task's `on_commit` (registered in the rolled-back
+  savepoint) is discarded — no push, no trigger side effects.
+- **Partial-success behavior (open nit):** decide whether to mark the entry
+  `reviewed`/`TASKS_ASSIGNED` when ≥1 task assigned, and show an "assigned N, skipped M" message.
+  Recommended: yes.
+
+## Callback API (Completion)
+
+New DRF endpoint (e.g. `POST /api/ocs/task_completed/`) that OCS calls when an FLW finishes a
```
Charl1996: Should we think about what happens if this callback fails for some reason, either OCS failing to send it, it fails to reach us or an error occurs during Connect consuming the payload? 

I'm thinking it would be nice if we're able to poll OCS periodically (say, once a day) to ask about in-flight conversations in case we miss the webhook - there would at least be a definitive way for us to get closure on in-flight tasks.

Not sure if this is the right endpoint: https://chatbots.dimagi.com/api/v1/docs/#tag/Chat/operation/chat_poll_task_response

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,388 @@
...
+## Trigger Flow (Assignment)
+
+Synchronous and user-facing, inside `AssignedTask.assign(...)` (which already opens its own
+`transaction.atomic()`).
+
+1. Create the `AssignedTask` row.
+2. **If `task_type.type == ocs`:**
+   - Call `ocs.trigger_bot(acting_user, ...)` **synchronously** with the payload above, using
+     the acting user's OCS token.
+   - On success: set `assigned_task.connect_channel_id` from the response and save.
```
Charl1996: I wonder if we shouldn't store the session_id here as well? This way we'll be better positioned to cross-correlate assigned tasks to sessions on OCS.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
snopoke: You'd want to use this endpoint to fetch the session and check it's status: https://www.openchatstudio.com/api/v1/docs/#tag/Experiment-Sessions/operation/session_retrieve

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
snopoke: You should definitely be storing the OCS session ID for future use

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,388 @@
...
+
+- `assign()`'s internal `atomic()` is a savepoint under `ATOMIC_REQUESTS`; catching the
+  exception *outside* it rolls back only that task and keeps the request transaction valid, so
+  successful assigns still commit. The failed task's `on_commit` (registered in the rolled-back
+  savepoint) is discarded — no push, no trigger side effects.
+- **Partial-success behavior (open nit):** decide whether to mark the entry
+  `reviewed`/`TASKS_ASSIGNED` when ≥1 task assigned, and show an "assigned N, skipped M" message.
+  Recommended: yes.
+
+## Callback API (Completion)
```
Charl1996, line 244: I think the OCS team can weigh in a here on what mechanisms are possible today. I know they have the concept of a pipeline where you can trigger actions based on events, e.g. "post to this webhook when conversation is done" (whatever "done" means).

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
snopoke: We can add oauth support to OCS - it's been on the list for a while: https://github.com/dimagi/open-chat-studio/issues/2179

pxwxnvermx: Got it. Thanks. Would this work in the same way form processor works on Connect?

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: I think the custom actions on OCS can be used here. 
https://docs.openchatstudio.com/concepts/team/custom_actions/

snopoke: You can also do integrations using Python nodes which I think is the current method being used by the Connect team: https://docs.openchatstudio.com/tech-hub/external-api-calls/

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
snopoke: I don't know how those work - are there docs I can look at?

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
SmittieC: Custom actions is one way to do it, but you can also consider timeout triggers. In the custom actions approach, you give the bot a "tool" that makes an API call. The bot should be instructed in when to use this tool.

In the timeout trigger approach, the conversation times out after a certain period after the last user message. It then executes a pipeline which then makes an API call. This does not guarentee that the conversation is really "done" (the user could have stopped replying).

A third way is if you use a Router Node that evaluates the conversation's "state" at each turn. When the router - which is just an LLM call really - determines that the conversation is at its end, you can call a python node before replying with the final thank you message.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: The docs don't contain the auth. The Form Processor also uses OAuth with the Client Credentials grant for authentication.

snopoke: Ok, yes, we will support the same flow. I've added more details to the [ticket](https://github.com/dimagi/open-chat-studio/issues/2179)

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: Added.

Charl1996 (approved):
LGTM.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+Connect lets Network Managers (NMs) assign tasks to frontline workers (FLWs). Today a
+"task" is effectively a CommCare re-learn activity tracked via xform submissions. This
+feature adds a second kind of task: an **OCS (Open Chat Studio) conversation**. When a user
+assigns an OCS task type to an FLW, Connect triggers a chatbot conversation in OCS for that
+FLW. When the FLW finishes the conversation, OCS calls back to Connect to mark the task
+complete.
+
+Concretely this requires:
+
+- **OAuth2-based** authentication to OCS, via django-allauth, using the **logged-in user's**
```
snopoke, line 18: Just confirming that this will require any user who wants to trigger a task to have an OCS account too.

An alternative is to use a client credentials application which can be set up by an 'admin' user and used by anyone with access to the opportunity.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+  **`SOCIALACCOUNT_STORE_TOKENS = True`** (so `SocialToken` access/refresh tokens are persisted).
+- `ATOMIC_REQUESTS = True` — every request is already a DB transaction.
+
+## Design Decisions (resolved during brainstorming / grilling)
+
+| Decision | Choice |
+|---|---|
+| **Outbound auth (Connect→OCS)** | **OAuth2 via django-allauth**, using the logged-in user's connected OCS account. No static `X-api-key`, no service credential. |
+| Why not `X-api-key` | Reversed in favor of per-user scoped, revocable OAuth tokens; avoids a pasted shared secret. |
+| Why user-facing (not background) | A background trigger **is** technically possible by storing and refreshing the user's token (OCS issues refresh tokens — confirmed by the OCS team in PR review). We still chose a **synchronous, user-facing** trigger: it avoids persisting/refreshing long-lived tokens and the refresh-token rotation races that come with background workers, and it gives the assigning user immediate success/failure feedback. |
```
snopoke, line 69: This implies that the user will have to auth with OCS every time they assign a task. Is that correct?

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+| Task identifier sent to OCS / accepted in callback | **UUID** (`assigned_task_id`), not the integer PK. |
+| OCS client placement | Single small **util module** `commcare_connect/opportunity/ocs.py`. |
+| Completion notification | Fire an FCM push (`"key": "task_completion"`) on completion of **any** task type, via a shared `AssignedTask.mark_completed(...)`. No backfill. |
+
+## OAuth2 Integration (django-allauth)
+
+OCS supports OAuth2 **Authorization Code + PKCE** and **Refresh Token** grants (confirmed from
+OCS docs). Endpoints: authorize `https://www.openchatstudio.com/o/authorize/`, token
+`/o/token/`, userinfo `/o/userinfo/`. Access tokens last ~1 hour; refresh tokens rotate.
+Scopes needed: `chatbots:read` (list) and `chatbots:interact` (trigger). No OIDC discovery
```
snopoke: you'll also need `sessions:read` to check session status

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+
+Notes from the OCS schema:
+- List endpoint is `/api/v2/chatbots/` (NOT `/api/chatbots/` as the requirements draft said) and
+  is **paginated** — `list_chatbots` must page through results.
+- `experiment` is a UUID (the chatbot id).
+- `trigger_bot` response fields: `session_id`, `url`, `team`, `channel`. We store **both**
+  `channel` (→ `connect_channel_id`) and `session_id` (→ `ocs_session_id`).
+- `trigger_bot` also accepts `start_new_session` (bool) and `prompt_text`/`message_text`. The
+  bare payload may not actually *start* a conversation; confirm whether we need
+  `start_new_session=true` and/or an initial message.
```
snopoke: you should always use `start_new_session` for tasks to ensure that each task get's a distinct session

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
snopoke: One challenge with using 'session status' as an indicator of task success is that sessions can be ended in other ways (manually by an admin etc). We don't have a state that maps cleanly to 'task completed successfully'.

We'll just need to be aware of this.

Update:

One option we do have is to use 'session state' for this rather than the session status. Session state is not user editable and it's fields and meaning are fully determined by the bot developer. The downside is that the field name would need to be consistent across all chatbots used in this way.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+  exception *outside* it rolls back only that task and keeps the request transaction valid, so
+  successful assigns still commit. The failed task's `on_commit` (registered in the rolled-back
+  savepoint) is discarded — no push, no trigger side effects.
+- **Partial-success behavior (open nit):** decide whether to mark the entry
+  `reviewed`/`TASKS_ASSIGNED` when ≥1 task assigned, and show an "assigned N, skipped M" message.
+  Recommended: yes.
+
+## Callback API (Completion)
+
+New DRF endpoint (e.g. `POST /api/ocs/task_completed/`) that OCS calls when an FLW finishes a
```
snopoke: Is there a reason this is `ocs` scoped? Seems like a generic API that could be used by other systems in future.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+is fired from OCS via one of their pipeline mechanisms — custom actions (a bot "tool" the LLM
+invokes), Python nodes (external API call), a router node evaluating conversation state, or a
+timeout trigger. Choosing/among these is the OCS team's configuration concern, outside Connect's
+scope; Connect only needs to receive the resulting callback.
+
+**Completed-task behavior (confirmed in review).** Connect does not clean up completed assigned
+tasks today, and the API continues to return them to mobile. OCS channels also remain alive
+after completion (no archival mechanism currently). No change to this behavior here.
+
+### Completion reconciliation (fallback for missed callbacks)
```
snopoke, line 289: This relies on there being a valid auth token available (especially since you've decided not to store refresh tokens)

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
```diff
@@ -0,0 +1,419 @@
...
+  authenticate to Connect's callback via OAuth2 (Client Credentials grant), the same mechanism HQ
+  uses for the Form Processor (see Callback API). This depends on OCS adding OAuth support,
+  tracked in [open-chat-studio#2179](https://github.com/dimagi/open-chat-studio/issues/2179).
+  The callback endpoint can be built against the existing `FormReceiver` OAuth2 pattern once that
+  lands.
+
+## Open Decisions
+
+1. **`start_new_session` / initial message** — confirm whether the bare `trigger_bot` payload
+   starts a conversation or needs `start_new_session=true` / `prompt_text`.
```
snopoke: Confirm - always pass 'start_new_session'

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: We are going to use Django Allauth for the OCS auth management. It will save the refresh token, so the user would need to auth whenever the refresh token expires.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: Since we are using django all auth for the OCS Authentication, we are also persisting the refresh token in the database.

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1312/commits/5e0a032b4037f0f0a88fd4dba66ee4c03c57b378

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: No reason for that. I will remove the ocs scope from this.
https://github.com/dimagi/commcare-connect/pull/1312/commits/5e0a032b4037f0f0a88fd4dba66ee4c03c57b378

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: Done. https://github.com/dimagi/commcare-connect/pull/1312/commits/5e0a032b4037f0f0a88fd4dba66ee4c03c57b378

docs/superpowers/specs/2026-06-19-ocs-conversations-design.md
pxwxnvermx: > Just confirming that this will require any user who wants to trigger a task to have an OCS account too.

Yes, that is correct. I will add this in the documentation for this feature.

snopoke (approved):

PR #1314, opened 2026-06-19
CCCT-2109: Fix daily limit race condition in deliver form processing

ajeety4:
## Product Description

Fixes a bug where visits could slip past the daily limit. For example, with a daily limit of 5, the 6th visit was sometimes **not** flagged as `over_limit` while later visits (7th, 8th) were.

This happened when two forms were submitted at nearly the same time: both checked the daily count before either was saved, so both saw a count below the limit and were accepted. Under production concurrency (gevent workers) this let an extra visit through.

## Technical Summary

Ticket: [CCCT-2109](https://dimagi.atlassian.net/browse/CCCT-2109)

The limit check counted existing visits and then saved the new one in a transaction. Two concurrent submissions for **different entities** read the same (stale) count before either committed, both passed the check, and both were saved as `pending`.

The previous guard was a Redis lock keyed on `entity_id`, so submissions for *different* entities took *different* locks and ran concurrently — never serialized.

**Fix:** take a `select_for_update` lock on the per-`(user, payment_unit)` `OpportunityClaimLimit` row inside the transaction, so all submissions sharing a payment unit are serialized and the limit check always sees committed counts.

**Why `select_for_update` instead of changing the Redis key:**
- **Right granularity** — the limit is per `(user, payment_unit)`, which is exactly the `OpportunityClaimLimit` row's key.
- **Correctness lives with the data** — the invariant is a DB count, serialized in the same transaction that reads it.
- **No mid-transaction expiry** — the Redis lock's 30s timeout can release while the transaction is still running (the critical section includes an O(visits) location loop), reopening the race. A row lock lasts exactly as long as the transaction.
- **Consistency** — `select_for_update` is already the pattern in this module (task and work-area processing), including in this same function.

The entity-scoped Redis lock is removed; the payment-unit row lock is broader and covers everything it did, including the duplicate-entity check. Errors propagate (500) like the other processing paths.

> Note: I am open to consider using the redis approach if there are concerns with the existing one.

### Safety story

The change is small and localized to `process_deliver_unit`, and reuses a locking pattern already established in this module. Contention is scoped to the same user + payment unit (different users/payment units never block each other). The lock is held only for the transaction, so there is no risk of a stale/expired lock. No data migration; existing visits are unaffected.

### Automated test coverage

Full `form_receiver` suite passes.

### QA Plan
https://dimagi.atlassian.net/browse/QA-8562

- With a daily limit of N, submit several forms in quick succession and confirm the (N+1)th visit is flagged `over_limit`.
- Confirm normal single submissions and duplicate-entity handling still behave as before.

### Labels & Review

[CCCT-2109]: https://dimagi.atlassian.net/browse/CCCT-2109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -45,7 +45,6 @@ from commcare_connect.opportunity.tasks import (
 )
 from commcare_connect.opportunity.visit_import import update_payment_accrued_for_user
 from commcare_connect.users.models import User
-from commcare_connect.utils.lock import try_redis_lock
 
 logger = logging.getLogger(__name__)
 
@@ -434,7 +433,8 @@ def clean_form_submission(access: OpportunityAccess, user_visit: UserVisit, xfor
 def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Opportunity, deliver_unit_block: dict):
     """Process a delivery form submission into a UserVisit and update CompletedWork.
 
-    Acquires a Redis lock per entity to handle concurrent submissions, then:
+    Locks the claim limit row (select_for_update) to serialize concurrent
+    submissions for the same user + payment unit, then:
     1. Creates a UserVisit with initial status based on daily/total/claim limits
     2. Runs verification flag checks via clean_form_submission()
     3. Auto-rejects flagged visits if automatic_visit_verification is enabled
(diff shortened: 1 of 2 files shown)


hemant10yadav (approved):
LGTM! One question do we already have tests for this? If not, it might be worth adding a test case for this condition so that we don't introduce a regression in the future.

zandre-eng (approved):

ajeety4:
@hemant10yadav Added a regression test in e6b9e88de: `test_deliver_form_locks_claim_limit_row`. It processes a deliver form and asserts a `SELECT ... FOR UPDATE` is issued against the `OpportunityClaimLimit` row — verified it fails if the `select_for_update` is removed, so it guards against the lock being silently dropped in a future refactor.

I intentionally avoided a full threaded race reproduction. Deterministically triggering the race needs two threads plus a mid-transaction synchronization hook (monkeypatching the count query with events), which is brittle and slow and pollutes the test. This query-level assertion catches the realistic regression (the lock being removed) while staying clean and fast.

hemant10yadav (approved):
Thanks!

PR #1315, opened 2026-06-22
CCCT-2470: Split Org Pay vs FLW Pay on invoices

Charl1996:
## Product Description

Service-delivery invoices now break payment into **FLW Pay**, **Org Pay**, and **Total Pay**, and the invoice's payable amount is the total (FLW + Org) rather than FLW pay alone.

On the invoice **Review** screen and the **create-invoice** line-items table, managed opportunities (those with a non-zero org amount) show three columns — FLW Pay, Org Pay, Total Pay — each localized to the opportunity's currency. Non-managed opportunities (no org pay) show a single column, now labelled **Total Pay**. The **line-items CSV download** carries the same breakdown.

### Line-items table on a managed opportunity showing FLW Pay / Org Pay / Total Pay
image

### Line-items of the same table if no org pay is specified for opportunity
image

## Technical Summary

[Link to ticket](https://dimagi.atlassian.net/browse/CCCT-2470)

Org pay was already cached per `CompletedWork` (`saved_org_payment_accrued` / `_usd`) but dropped at invoicing. This change surfaces it through the line-item pipeline and folds it into the invoice amount.

- **Line-item data:** `get_invoice_items` now returns separate `flw_/org_/total_` components for both local currency and USD, with `total = flw + org`. The obsolete `amount_per_unit` key was removed.
- **Table:** `InvoiceLineItemsTable(currency, data, show_org=False)` conditionally renders the FLW/Org columns (Total Pay always shown) with currency-localized headers. The Review view and the create-page `invoice_items` endpoint set `show_org` when any line item has a non-zero org amount.
- **CSV export:** `InvoiceDeliveriesTable` (the line-items download) now mirrors the on-screen table — FLW Pay / Org Pay / Total Pay in local currency plus Total Pay (USD), replacing the single Payment Accrued columns. As this table is export-only, the FLW/Org columns are dropped via `exclude_from_export` (not `visible`) when the export contains no org pay; the download view sets `show_org` from the deliveries being exported.

### Note
This change of including org payment in the invoice amount applies to **newly created** service-delivery invoices. This was not done previously (although it should have been). I've raised this with the necessary people already and identified the invoices on production that are incorrect. We will not attempt to rectify older invoices at this point.

One other thing to note - for these incorrectly calculated invoices, reviewing the invoice will show the org payment in the table but it won't add up to the invoice amount since the invoice amount is fixed at invoice creation while the line items table is calculated on the fly. All affected invoices have been identified and the necessary communication will be made to the stakeholders involved.

### Safety story

The change is scoped to the service-delivery invoice line-items table, its CSV export, and the service-delivery invoice amount.

Existing invoices are not modified, so already-created invoices (pending, frozen, or paid) keep their original amounts as a historical record.

### Automated test coverage

- `test_completed_work.py::test_get_invoice_items_total_includes_org_pay` — total folds in org pay for both local and USD; raw FLW/Org components are surfaced separately.
- `test_tables.py` — `InvoiceLineItemsTable` hides FLW/Org columns when `show_org=False` and shows them with currency-localized headers when `show_org=True`.
- `test_tables.py` — `InvoiceDeliveriesTable` (CSV export) hides FLW/Org columns when `show_org=False`, shows them when `show_org=True`, and its Total columns fold in org pay through the export path.

### QA Plan

No QA planned, as this change is mostly rendering.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -1729,31 +1729,33 @@ class PaymentUnitTable(OrgContextTable):
 
 
 class InvoiceLineItemsTable(tables.Table):
-    month = tables.Column()
-    payment_unit_name = tables.Column(verbose_name="Payment Unit")
-    number_approved = tables.Column()
-    flw_amount_local = tables.Column(verbose_name="FLW Pay (local)")
-    org_amount_local = tables.Column(verbose_name="Org Pay (local)")
-    total_amount_local = tables.Column(verbose_name="Total Pay (local)")
-    exchange_rate = tables.Column()
-    total_amount_usd = tables.Column(verbose_name="Total Pay (USD)")
+    month = tables.Column(verbose_name=gettext_lazy("Month"))
+    payment_unit_name = tables.Column(verbose_name=gettext_lazy("Payment Unit"))
+    number_approved = tables.Column(verbose_name=gettext_lazy("Number Approved"))
+    flw_amount_local = tables.Column(verbose_name=gettext_lazy("FLW Pay (local)"))
+    org_amount_local = tables.Column(verbose_name=gettext_lazy("Org Pay (local)"))
+    total_amount_local = tables.Column(verbose_name=gettext_lazy("Total Pay (local)"))
+    exchange_rate = tables.Column(verbose_name=gettext_lazy("Exchange Rate"))
+    total_amount_usd = tables.Column(verbose_name=gettext_lazy("Total Pay (USD)"))
 
     def __init__(self, currency, *args, show_org=False, **kwargs):
         super().__init__(*args, **kwargs)
         if currency:
-            self.columns["flw_amount_local"].column.verbose_name = f"FLW Pay ({currency})"
-            self.columns["org_amount_local"].column.verbose_name = f"Org Pay ({currency})"
-            self.columns["total_amount_local"].column.verbose_name = f"Total Pay ({currency})"
+            self.columns["flw_amount_local"].column.verbose_name = _("FLW Pay (%(currency)s)") % {"currency": currency}
+            self.columns["org_amount_local"].column.verbose_name = _("Org Pay (%(currency)s)") % {"currency": currency}
+            self.columns["total_amount_local"].column.verbose_name = _("Total Pay (%(currency)s)") % {
+                "currency": currency
+            }
         if show_org:
-            usd_tooltip = (
+            usd_tooltip = _(
                 "Sum of FLW pay and org pay (USD), each converted at the exchange rate "
                 "at the delivery's approval time."
             )
         else:
             self.columns["flw_amount_local"].column.visible = False
             self.columns["org_amount_local"].column.visible = False
-            usd_tooltip = "FLW pay (USD), converted at the exchange rate at the delivery's approval time."
-        self.columns["total_amount_usd"].column.verbose_name = header_with_tooltip("Total Pay (USD)", usd_tooltip)
+            usd_tooltip = _("FLW pay (USD), converted at the exchange rate at the delivery's approval time.")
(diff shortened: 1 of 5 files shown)


zandre-eng:
> I'm not sure what the best course of action is:

This could be a good question for product, but I would advocate for (2) with showing a warning to the user that there is a discrepancy. This can at least help make it obvious to the user and avoid unnecessary support queries where users think there is a bug due to the discrepancy.

commcare_connect/opportunity/tables.py
```diff
@@ -1732,26 +1732,28 @@ class InvoiceLineItemsTable(tables.Table):
...
     number_approved = tables.Column()
-    amount_per_unit = tables.Column(
-        verbose_name="Payment Unit Amount (local)",
-    )
-    total_amount_local = tables.Column(
-        verbose_name="Total Amount (local)",
-    )
+    flw_amount_local = tables.Column(verbose_name="FLW Pay (local)")
+    org_amount_local = tables.Column(verbose_name="Org Pay (local)")
+    total_amount_local = tables.Column(verbose_name="Total Pay (local)")
```
zandre-eng: Nit: Translations for the headers.

Charl1996:
### Code review

Found 1 issue:

1. The four new `show_org` column-visibility tests are near-identical and should be collapsed with `pytest.mark.parametrize` (CLAUDE.md says "Use `pytest.mark.parametrize` instead of writing multiple near-identical tests"). The two `InvoiceLineItemsTable` tests differ only by the `show_org` boolean and mirrored assertions, and the two `InvoiceDeliveriesTable` tests do the same — each pair is a textbook parametrize candidate.

https://github.com/dimagi/commcare-connect/blob/71444218ccd3aac485e7c6f480a722dd2eaf01d5/commcare_connect/opportunity/tests/test_tables.py#L21-L55

zandre-eng (approved):
Just a minor nit.

Charl1996:
> > I'm not sure what the best course of action is:
> 
> This could be a good question for product, but I would advocate for (2) with showing a warning to the user that there is a discrepancy. This can at least help make it obvious to the user and avoid unnecessary support queries where users think there is a bug due to the discrepancy.

Since Mary is reaching out individually to the affected programs I think we can probably leave it. I'll just double check with her if we want to additionally show a message.

zandre-eng (approved):

commcare_connect/opportunity/tables.py
```diff
@@ -1732,26 +1732,28 @@ class InvoiceLineItemsTable(tables.Table):
...
+    def __init__(self, currency, *args, show_org=False, **kwargs):
         super().__init__(*args, **kwargs)
         if currency:
-            self.columns["amount_per_unit"].column.verbose_name = f"Payment Unit Amount ({currency})"
-            self.columns["total_amount_local"].column.verbose_name = f"Total Amount ({currency})"
+            self.columns["flw_amount_local"].column.verbose_name = f"FLW Pay ({currency})"
+            self.columns["org_amount_local"].column.verbose_name = f"Org Pay ({currency})"
+            self.columns["total_amount_local"].column.verbose_name = f"Total Pay ({currency})"
+        if show_org:
+            usd_tooltip = (
```
ajeety4: I would reccomend to also indicate the rounding off part in the tooltip message if possible.  The intention of the original tooltip was to clearly state the formula used especially for conversion.

ajeety4 (commented):
LGTM
Just a single query on the tooltip

commcare_connect/opportunity/tables.py
Charl1996: Good point. I'll update.

Charl1996: https://github.com/dimagi/commcare-connect/pull/1315/commits/2282a5d975755596ce1eb514bc3561675cc100fc

ajeety4 (approved):
Thanks for the updates

PR #1317, opened 2026-06-22
Audit Report Table Improvements

zandre-eng:
## Product Description

The weekly performance audit report's detail view has been reworked so reviewers can scan and triage all FLWs from a single, sortable table.

The previous two separate tables ("Workers to Review" and "No Action") are now one merged table listing every FLW, with workers still needing review floated to the top:
Screenshot_20260622_121356

The worker search box in the upper-left has been replaced with a multi-select dropdown, so reviewers can filter the table to one or several specific FLWs at once:
Screenshot_20260622_121412

Every column is now sortable (previously only the worker-name column was), and the metric column headers wrap onto multiple lines to keep the columns narrow instead of stretching wide.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2482)

The detail view assembles all entries into a single Python list (it must materialize the full set anyway to build the filter dropdown options and the reviewed/flagged counts), so filtering and the default "review-needed first" ordering are done in memory rather than via additional queries. The worker filter is a TomSelect multi-select that live-updates only the results table via htmx, leaving the dropdown intact while selecting.

### Safety story

This change is low-risk: it is read-only (presentation and filtering of existing audit data only — no writes or data migrations), scoped entirely to the weekly performance report detail view, and the report itself remains gated behind the existing `WEEKLY_PERFORMANCE_REPORT` feature flag. It is fully reversible by revert.

### Automated test coverage

Tests cover the merged single table (all FLWs present, review-needed ordered first), server-side filtering by selected worker with the full worker list still offered as filter options, and ascending/descending sorting by a calculation column. Existing detail, task-modal, and complete-audit tests remain green.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/audit/views.py
+++ b/commcare_connect/audit/views.py
@@ -88,14 +88,18 @@ def audit_report_detail(request, org_slug, opp_id, audit_report_id):
     _require_feature_flag(opportunity)
     report = get_object_or_404(AuditReport, audit_report_id=audit_report_id, opportunity=opportunity)
 
-    name_filter = request.GET.get("filter", "").strip()
-    qs = report.entries.select_related("opportunity_access__user")
-    if name_filter:
-        qs = qs.filter(opportunity_access__user__name__icontains=name_filter)
-    entries = list(qs.order_by("opportunity_access__user__name"))
-
-    all_entries = list(report.entries.all())
+    all_entries = list(
+        report.entries.select_related("opportunity_access__user").order_by("opportunity_access__user__name")
+    )
     columns_spec = _column_specs(all_entries)
+
+    worker_filter_choices = [(str(e.opportunity_access_id), e.opportunity_access.user.name) for e in all_entries]
+
+    selected_workers = request.GET.getlist("worker")
+    if selected_workers:
+        entries = [e for e in all_entries if str(e.opportunity_access_id) in selected_workers]
+    else:
+        entries = list(all_entries)
 
     # If no sorting given, apply default to float workers that need review to the top of the table
     if not request.GET.get("sort"):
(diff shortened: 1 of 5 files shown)


commcare_connect/audit/views.py
```diff
@@ -88,39 +88,34 @@ def audit_report_detail(request, org_slug, opp_id, audit_report_id):
...
+    )
     columns_spec = _column_specs(all_entries)
 
-    to_review = [e for e in entries if e.flagged and not e.reviewed]
-    no_action = [e for e in entries if not e.flagged or e.reviewed]
+    worker_filter_choices = [(str(e.opportunity_access_id), e.opportunity_access.user.name) for e in all_entries]
+
+    selected_workers = request.GET.getlist("worker")
+    if selected_workers:
+        entries = [e for e in all_entries if str(e.opportunity_access_id) in selected_workers]
```
zandre-eng, line 100: Filtering in memory here as `all_entries` is already loaded for the dropdown options, column spec, and the all-worker flagged/reviewed counts, so doing a query level filter instead would have just added a redundant roundtrip for data that is already in memory.

commcare_connect/templates/audit/audit_report_body.html
```diff
@@ -15,28 +16,38 @@ <h1 class="text-2xl font-medium text-white">{% trans "Weekly Performance Report"
...
-      {% trans "Complete Audit" %}
-    </button>
+  <div id="audit-table-region" class="flex flex-col gap-4">
+    {% render_table table %}
+    <div class="flex justify-end mt-6">
+      <button type="button"
+              class="button button-md primary-dark"
+              {% if not can_complete %}disabled{% endif %}
+              hx-post="{% url 'opportunity:audit:audit_report_complete' org_slug=org_slug opp_id=opportunity.opportunity_id audit_report_id=report.audit_report_id %}"
+              hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'
```
ajeety4: nit: this can be skipped as csrf token is automatially included for htmx requests

ajeety4 (approved):

commcare_connect/templates/audit/audit_report_body.html
zandre-eng: Good catch, removed in [d8ea854](https://github.com/dimagi/commcare-connect/pull/1317/commits/d8ea85499b7b2caed909f27156441638332fa8a5).

hemant10yadav (approved):

PR #1319, opened 2026-06-23
CCCT-2411: Keep worker/audit table nav arrow visible regardless of zoom or width

ajeety4:
## Product Description

Issue: The row navigation arrow (chevron) in the worker **Deliver**, worker **Learn**, and **audit report list** tables lives in the rightmost column, which sits inside a horizontally-scrolling area. At higher zoom levels or narrow widths the column scrolled off the right edge, so the arrow was only reachable by scrolling sideways.

Fix: This pins that action column to the right edge so the arrow stays visible at any zoom level or table width — no horizontal scrolling needed to discover it. The hover-reveal behavior is preserved; the arrow now simply appears in a spot that's always on-screen.

See the first row in the below screenshots
**Before**

image

**After**
image

## Technical Summary

Jira: [CCCT-2411](https://dimagi.atlassian.net/browse/CCCT-2411)

- **`tailwind/tailwind.css`** — new scoped `col-action` modifier on `.base-table`:
  - `position: sticky; right: 0` pins the column to the right edge of the scroll container.
  - The body cell gets an opaque background (`bg-white`, `bg-gray-100` on hover/active) so columns scrolling underneath don't show through. The header cell is already opaque via the existing `bg-gray-200`.
- **`commcare_connect/opportunity/tables.py`** — apply `col-action` to the `action` column of `WorkerDeliveryTable` and `WorkerLearnTable`.
- **`commcare_connect/audit/tables.py`** — apply `col-action` to the `view` column of `AuditReportTable`.

### Safety story

CSS/markup-only change, scoped to three specific table columns via a dedicated `col-action` class. 

### Automated test coverage

None added — this is a presentational CSS/markup change with no testable business logic. The existing table tests still cover rendering of these tables.

### QA Plan
None

- Open the worker Deliver, worker Learn, and audit report list tables.
- Zoom in / narrow the window so the table overflows horizontally.
- Confirm the navigation arrow stays pinned to the right edge and is reachable on row hover without scrolling sideways.

### Labels & Review

[CCCT-2411]: https://dimagi.atlassian.net/browse/CCCT-2411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/audit/tables.py
+++ b/commcare_connect/audit/tables.py
@@ -28,6 +28,7 @@ class AuditReportTable(OrgContextTable):
         accessor="pk",
         verbose_name="",
         orderable=False,
+        attrs={"th": {"class": "col-action"}, "td": {"class": "col-action"}},
     )
 
     class Meta:
--- a/tailwind/tailwind.css
+++ b/tailwind/tailwind.css
@@ -613,22 +613,16 @@
 
   .base-table tbody tr::after {
     content: '';
-    @apply -translate-1/2 absolute left-1/2 -z-10 w-[99%] -translate-y-1/2 rounded-lg bg-gray-100 opacity-0;
+    @apply -translate-1/2 absolute left-1/2 -z-10 w-[99%] rounded-lg bg-gray-100 opacity-0;
     top: var(--row-hl-top);
     height: var(--row-hl-height);
   }
 
-  .base-table tbody tr:hover::after {
-    @apply opacity-100;
-  }
-
+  .base-table tbody tr:hover::after,
   .base-table tbody tr.active::after {
     @apply opacity-100;
   }
 
-  /* .base-table tbody tr:hover{
-      @apply bg-gray-100 rounded-lg;
-  }  */
   .base-table tbody td {
     @apply whitespace-nowrap border-b-[1px] border-gray-100 p-4 text-left;
   }
--- a/tailwind/tailwind.css
+++ b/tailwind/tailwind.css
@@ -602,16 +602,20 @@
 
   .base-table tbody tr {
     @apply relative;
+    --row-hl-top: 50%;
+    --row-hl-height: 90%;
+  }
+
+  .base-table tbody tr:first-child {
+    --row-hl-top: 53.8%;
+    --row-hl-height: 80%;
   }
 
   .base-table tbody tr::after {
     content: '';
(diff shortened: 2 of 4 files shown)


zandre-eng (approved):

tailwind/tailwind.css
```diff
@@ -602,32 +602,55 @@
...
     content: '';
-    @apply -translate-1/2 absolute left-1/2 top-[53.8%] -z-10 h-[80%] w-[99%] -translate-y-1/2 rounded-lg bg-gray-100 opacity-0;
-  }
-
-  .base-table tbody tr:hover::after {
-    @apply opacity-100;
+    @apply -translate-1/2 absolute left-1/2 -z-10 w-[99%] rounded-lg bg-gray-100 opacity-0;
+    top: var(--row-hl-top);
+    height: var(--row-hl-height);
   }
```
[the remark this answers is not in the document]
ajeety4: resolving as the linter tool ci run has passed

Charl1996:
I'm going to be pedantic here, but I see the chevron colours the whole background where the column left of it has a slight space from the top and bottom of the cell. 

image

I'm not sure how hard/ugly it would be to implement, but maybe worth having a look now rather than later?

ajeety4:
> I'm going to be pedantic here, but I see the chevron colours the whole background where the column left of it has a slight space from the top and bottom of the cell.
> 
> image
> I'm not sure how hard/ugly it would be to implement, but maybe worth having a look now rather than later?

This was fixed, however forgot to update the screenshot. Apologies.
Just updated it now.

commcare_connect/audit/tables.py
```diff
@@ -28,6 +28,7 @@ class AuditReportTable(OrgContextTable):
         accessor="pk",
         verbose_name="",
         orderable=False,
+        attrs={"th": {"class": "col-action"}, "td": {"class": "col-action"}},
```
sravfeyn, line 31: Is there no-way to target these with other css selectors like `first-child`, `nth-child` etc without having to add `col-action` class?

ajeety4: We could, since the action column is always last (td:last-child), but these aren't generic last-column styles. They pin the column and add a highlight overlay, which only makes sense for action columns. Not every `.base-table` has one, so `:last-child` would wrongly pin/highlight the last data column in those tables. `col-action` keeps the styling tied to what the column is rather than its position.

sravfeyn (approved):

Charl1996 (approved):
Thanks!

PR #1323, opened 2026-06-26
Apply Correct Auth for Adding Credentials

zandre-eng:
## Product Description

No user-facing changes. This is a backend-only change to how CommCare Connect authenticates with the PersonalID service when adding user credentials.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-1942)

The `add_credential` endpoint on ConnectID requires a dedicated set of client credentials rather than the default ConnectID client used by all other requests. To support this, `_make_request` now accepts an optional `auth` argument, falling back to the default `CONNECTID_CLIENT_ID`/`CONNECTID_CLIENT_SECRET` when none is supplied, while `add_credentials` passes a `BasicAuth` built from the new `CONNECTID_CREDENTIALS_CLIENT_ID`/`CONNECTID_CREDENTIALS_CLIENT_SECRET` settings.

Without applying this specific client credentials, the user credentials on PersonalID default to having the issuing auth be staging, which is incorrect for user creds added from Connect prod.

### Safety story

The change is low-risk and narrowly scoped: it only alters which credentials are sent for the single `add_credential` call, and all other PersonalID requests continue to use the default auth via the unchanged fallback path. The new settings default to empty strings, so deployment requires the corresponding environment variables to be configured.

### Automated test coverage

No new automated tests were added; the change relies on the existing `_make_request` behaviour, which is unchanged for all callers that do not pass `auth`.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -44,7 +44,8 @@ def send_message_bulk(messages: list[Message]) -> MessagingBulkResponse:
 
 def add_credentials(credentials_items: list[dict]):
     json = {"credentials": credentials_items}
-    response = _make_request(POST, "/users/add_credential", json=json, timeout=30)
+    auth = BasicAuth(settings.CONNECTID_CREDENTIALS_CLIENT_ID, settings.CONNECTID_CREDENTIALS_CLIENT_SECRET)
+    response = _make_request(POST, "/users/add_credential", json=json, timeout=30, auth=auth)
     return response.json()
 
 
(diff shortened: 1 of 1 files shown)


commcare_connect/connect_id_client/main.py
```diff
@@ -44,7 +44,8 @@ def send_message_bulk(messages: list[Message]) -> MessagingBulkResponse:
 
 def add_credentials(credentials_items: list[dict]):
     json = {"credentials": credentials_items}
-    response = _make_request(POST, "/users/add_credential", json=json, timeout=30)
+    auth = BasicAuth(settings.CONNECTID_CREDENTIALS_CLIENT_ID, settings.CONNECTID_CREDENTIALS_CLIENT_SECRET)
```
sravfeyn, line 47: Why do we need separate auth only for this endpoint? Is it a matter of scope? If so, aren't there any more URLs related to credentials?

zandre-eng: Yeah I believe the original intent behind having a separate auth for credentials was to keep the scope constrained, and was a pattern we wanted to implement more as additional integrations were developed. There is another endpoint `filter_users` but this is a user-filtering endpoint that takes a credential param, not a credential CRUD endpoint.

Do you think we should remove this and rather use the normal basic auth for this endpoint?

sravfeyn (approved):
LGTM

Charl1996 (approved):

PR #1325, opened 2026-06-26
CCCT-2594: Email PMs when new audit reports are generated

Charl1996:
**RP: 1** · Ticket: [CCCT-2594](https://dimagi.atlassian.net/browse/CCCT-2594)

## Product Description

When the weekly job generates audit reports, PM-org admins are emailed a notification listing the new report(s) — opportunity name, period, and a link to each. Batched one email per PM org (multiple reports → a single email, not one each).

Example email:
image

## Technical Summary

`generate_audit_reports` now calls a new `send_new_audit_report_notifications(reports)` helper after its loop (`commcare_connect/audit/tasks.py`). It groups reports by PM org (`opportunity.program.organization`), emails the org's **admins** via the existing `send_mail_async` task, and skips opportunities with no program or orgs with no admins. Notification and per-org failures are isolated in `try/except` so report generation is never affected. Adds two `email/base.*` templates.

### Safety story

Additive only — no schema/migration changes and no change to report generation. The email is gated behind the existing `WEEKLY_PERFORMANCE_REPORT` flag and fully wrapped in exception handling, so it cannot interrupt generation.

### Automated test coverage

11 tests in `audit/tests/test_tasks.py`: batching per org, admins-only recipients, no-program/no-admin skips, and both failure-isolation paths.

### QA Plan

No QA required.

### Labels & Review

[CCCT-2594]: https://dimagi.atlassian.net/browse/CCCT-2594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/audit/tasks.py
```diff
@@ -24,17 +30,101 @@ def generate_audit_reports() -> None:
...
+            ),
+        )
+        reports_by_org[pm_org].append(
+            {
+                "opportunity_name": opportunity.name,
+                "period_start": report.period_start.isoformat(),
+                "period_end": report.period_end.isoformat(),
+                "url": report_url,
+            }
+        )
```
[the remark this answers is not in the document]
Charl1996: Good catch.

commcare_connect/audit/tasks.py
```diff
@@ -24,17 +31,108 @@ def generate_audit_reports() -> None:
...
+        .values_list("organization_id", "emails")
+    )
+
+    for pm_org, items in reports_by_org.items():
+        recipient_emails = org_emails.get(pm_org.id)
+        if not recipient_emails:
+            continue
+        try:
+            _send_org_audit_report_email(pm_org, items, recipient_emails)
+        except Exception:
```
zandre-eng, line 119: Is there a specific set of exception classes we expect could be raised here? If so, it could be nicer to have those explicitly listed here instead of the generic exception class to make it easier to understand why this try block is here.

zandre-eng (approved):
Nitpick about exception class, but otherwise looks good.

commcare_connect/audit/tasks.py
Charl1996: Not really any that I expect, so I added the generic one so any one failure doesn't stop the whole load. I've added a logger.exception so we can see in Sentry if any pops up.

hemant10yadav (approved):

zandre-eng (approved):

PR #1326, opened 2026-06-26
Show payment import result as an on-page banner

mkangia:
## Product Description

This changes payment status updates from popups to banners.

-------------------------------------------
**Before**

Screenshot from 2026-07-08 00-29-11

-------------------------------------------
Screenshot from 2026-07-08 00-29-33

-------------------------------------------
Screenshot from 2026-07-08 00-29-45

-------------------------------------------
Screenshot from 2026-07-08 00-31-15

-------------------------------------------
Screenshot from 2026-07-08 00-32-04

------------------------------------------
**After**

Screenshot from 2026-07-08 00-33-28

-------------------------------------------
Screenshot from 2026-07-08 00-34-49

-------------------------------------------
image

-------------------------------------------
image

## Technical Summary

Used the message banners already available by reloading the page and adding the message on `request` for render.
In order to improve on the polling for the result, had to create a fresh template. It did seem we were using "export" vocabulary even for import.

### Safety story

Tested locally.

### Labels & Review


commcare_connect/templates/opportunity/payment_import_progress.html
```diff
@@ -0,0 +1,35 @@
+{% load i18n %}
```
sravfeyn, line 2: I think we should use already existing `upload_progress_bar.html` which this seems to be mostly similar with.

commcare_connect/opportunity/views.py
```diff
@@ -799,7 +806,31 @@ def payment_import(request, org_slug=None, opp_id=None):
     saved_path = default_storage.save(file_path, file)
 
     result = bulk_update_payments_task.delay(request.opportunity.pk, saved_path, file_format)
-    return redirect(f"{redirect_url}?export_task_id={result.id}")
+    return redirect(f"{redirect_url}?payment_import_task_id={result.id}")
+
+
+@org_member_required
+@require_GET
+def payment_import_status(request, org_slug, task_id):
```
sravfeyn: This is almost exactly same as `render_export_status`. The name might be misleading, it's actually a `render_task_status`

commcare_connect/opportunity/tests/test_visit_import.py
```diff
@@ -1011,3 +1012,15 @@ def test_locked_visits(self):
...
         )
+
+
+@pytest.mark.parametrize("missing_users", [{"ghost"}, {"ghost1", "ghost2"}])
+def test_payment_import_missing_message_has_no_leading_break(missing_users):
+    status = PaymentImportStatus(seen_users={"alice"}, missing_users=missing_users)
+
+    message = status.get_missing_message()
+
+    assert not message.startswith("<br>")
```
sravfeyn: nit: I would remove this line

commcare_connect/opportunity/tests/test_views.py
```diff
@@ -2971,3 +2973,68 @@ def test_tasks_tile_shown_only_with_switch_and_configured_task_type(
         with override_switch(WORKER_VISITS_TASKS, active=True):
             response = client.get(self._url(organization, opportunity))
         assert b"Tasks Assigned to Connect Workers" in response.content
+
+
+@mock.patch("commcare_connect.opportunity.views.bulk_update_payments_task.delay")
+def test_payment_import_redirects_with_payment_task_id(mock_delay, client, organization, opportunity, org_user_member):
```
sravfeyn, line 2979: nit: I think this test is not super necessary. 

If you would rather want to keep this and below. They could be written as a class and DRYed up significantly

commcare_connect/opportunity/views.py
mkangia: Agreed,

refactored these here: https://github.com/dimagi/commcare-connect/pull/1326/commits/1c0801b1e921b64d7b25f582243a0607078fdf97

commcare_connect/opportunity/tests/test_visit_import.py
mkangia: this entire commit was reverted, it turned out that the space was needed.
https://github.com/dimagi/commcare-connect/pull/1326/commits/89f7ffb48a782495a02eb7a92632b39c527562e0

commcare_connect/templates/opportunity/payment_import_progress.html
mkangia: Using it the same template would introduce some more if-else to this template but I went with a refactor that makes sense

https://github.com/dimagi/commcare-connect/pull/1326/commits/c7c33f7177d14c8aba5bd34845d6f1e6d9f7fc16

commcare_connect/opportunity/views.py
mkangia: renamed here: https://github.com/dimagi/commcare-connect/pull/1326/commits/e998f23c5ebffa83aca5e772f58e6632497babcb

commcare_connect/opportunity/tasks.py
```diff
@@ -481,6 +481,22 @@ def get_payment_upload_key(opp_id):
     return f"bulk_update_payments_opportunity_{opp_id}"
 
 
+def get_payment_import_result_message(status):
+    """Build the (message, is_error) shown after a payment import completes.
+
+    No payments uploaded is treated as an error so it surfaces as a red banner.
+    """
+    if len(status):
+        lines = [f"Payment status uploaded successfully for {len(status)} users."]
```
sravfeyn: nit: Does this (and below) need translations?

commcare_connect/opportunity/tasks.py
```diff
@@ -494,18 +510,19 @@ def bulk_update_payments_task(self, opportunity_id: int, file_path: str, file_fo
...
+            is_error = True
         except Exception as e:
-            messages = [f"Unexpected error during payment import: {e}"]
+            message = f"Unexpected error during payment import: {e}"
+            is_error = True
         finally:
             default_storage.delete(file_path)
 
-        set_task_progress(self, "<br>".join(messages), is_complete=True)
+        self.update_state(state=CELERY_TASK_SUCCESS, meta={"message": message, "is_error": is_error})
```
sravfeyn: Wouldn't be better to add `is_error` support iself to `set_task_progress`?

commcare_connect/opportunity/tasks.py
```diff
@@ -481,6 +481,22 @@ def get_payment_upload_key(opp_id):
     return f"bulk_update_payments_opportunity_{opp_id}"
 
 
+def get_payment_import_result_message(status):
```
sravfeyn, line 484: `nit`: I believe nesting this into a separate method to keep the `bulk_update_payments_task` small is not super necessary here. The readability loss and context switching between two methods may not be worth in this case.

sravfeyn:
I believe we are adding a new type of task-status broadcast via messages instead of the progress-bar based one. We could do this in a modular way so that the celery and payment task specific code is not tied together (like in `_add_payment_import_message`) and more importantly when we want to do this for any other tasks, we could use the shared utils. Otherwise, its easy for someone latter to continue copy-pasting your pattern for their own tasks. I think this can be done by doing below.

1. Extend `set_task_progress` to accept `is_error`, storing it in meta alongside message. This should help in places where you needed to do `self.update_state`.
2. Extend `get_task_progress`'s returned dict to include is_error (read from the same meta), so it becomes the single place that knows how to interpret a task's result.  
3. Add new status broadcasting helper in utils/celery.py (e.g. add_task_result_message(request, task_id, ownership_check=None)) using messages.success/messages.error. This and above point should help in avoiding fetching task in `_add_payment_import_message`.

Though, if you think this is clearly one-off 1 and 2 might still be worth it.

commcare_connect/opportunity/tasks.py
mkangia: These are in celery tasks, I don't think translations work here?

mkangia:
Thanks for the thorough review @sravfeyn 

> Though, if you think this is clearly one-off 1 and 2 might still be worth it.

This is not a one-off as I think product wants to move to banners, so we need something reusable.

And for that, I did want to keep this separate from the current implementations so we can just switch to new code and scrap off the old one instead of branching/extending from the old code. The refactor/extraction helps use common code.
Though let me take another look.

commcare_connect/opportunity/tasks.py
mkangia: Makes sense, updated.

https://github.com/dimagi/commcare-connect/pull/1326/commits/de018ba20c2b164c56dc26442faabdf134f58055

commcare_connect/opportunity/tasks.py
sravfeyn: I think translations still need to be worked through celery tasks. Some of the task seem to have translations

mkangia: > Some of the task seem to have translations

I don't expect them to work.

https://stackoverflow.com/a/7208163

mkangia:
Hey @sravfeyn 

Can you let me know if there is any blocking feedback that you have on this PR? If possible I would like to get this in next deploy?

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,50 @@ class WorkerPaymentsView(BaseWorkerListView):
...
+        return task
+
+    def _payment_import_complete(self):
+        task = self._payment_import_task()
+        return bool(task) and task._get_task_meta().get("status") in (CELERY_TASK_SUCCESS, CELERY_TASK_FAILURE)
+
+    def _add_payment_import_message(self):
+        """Surface the finished import task's result as a standard banner."""
+        task = self._payment_import_task()
+        if task._get_task_meta().get("status") == CELERY_TASK_FAILURE:
```
pxwxnvermx: We should check here for a valid task before fetching the status. The code for `_payment_import_complete` checks before accessing any methods on the task.

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,50 @@ class WorkerPaymentsView(BaseWorkerListView):
...
+        if not request.htmx and self._payment_import_complete():
+            self._add_payment_import_message()
+        return super().get(request, org_slug, opp_id)
+
+    def _payment_import_task(self):
+        task_id = self.request.GET.get("payment_import_task_id")
+        if not task_id:
+            return None
+        task = AsyncResult(task_id)
+        args = task._get_task_meta().get("args") or []
```
pxwxnvermx: Almost all AsyncResult task meta properties have direct accessors defined on the class. These accessors use the same underlying data as above, so we can refactor to use the direct accessors instead of the "_get_task_meta()" internal method.

[Celery AsyncResult Docs](https://docs.celeryq.dev/en/main/reference/celery.result.html)

```suggestion
        args = task.args or []
```

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,50 @@ class WorkerPaymentsView(BaseWorkerListView):
...
+        if not args or args[0] != self.get_opportunity().pk:
+            return None
+        return task
+
+    def _payment_import_complete(self):
+        task = self._payment_import_task()
+        return bool(task) and task._get_task_meta().get("status") in (CELERY_TASK_SUCCESS, CELERY_TASK_FAILURE)
```
pxwxnvermx: Same comment as above about using the properties.

nit: The task can only be AsyncResult or None based on the `_payment_import_task` function, so the check can also be refactored to limit the number of permitted falsy values.

```suggestion
        return task is not None and task.status in (CELERY_TASK_SUCCESS, CELERY_TASK_FAILURE)
```

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,50 @@ class WorkerPaymentsView(BaseWorkerListView):
...
+    def _add_payment_import_message(self):
+        """Surface the finished import task's result as a standard banner."""
+        task = self._payment_import_task()
+        if task._get_task_meta().get("status") == CELERY_TASK_FAILURE:
+            messages.error(self.request, _("The payment import failed. Please try again."))
+            return
+        message = get_task_progress_message(task)
+        if not message:
+            return
+        is_error = (task._get_task_meta().get("result") or {}).get("is_error")
```
pxwxnvermx: Same as above about using the properties. 

nit: the or can be replaced with getattr, it will remove the need for wrapping in parentheses.

```suggestion
        is_error = getattr(task, "result", {}).get("is_error")
```

commcare_connect/templates/components/task_progress_base.html
```diff
@@ -0,0 +1,35 @@
...
+      </p>
+      <p class="text-sm mt-1">{{ progress.error }}</p>
+    </div>
+  {% elif not progress.complete %}
+    <div class="w-full">
+      <div class="w-full h-2 bg-gray-200 rounded-full overflow-hidden mb-4">
+        <div class="h-full bg-purple-600 rounded-full animate-[progress_2s_linear_infinite]"
+             style="width: 80%;
+                    background-image: linear-gradient(to right, #8b5cf6, #a78bfa, #8b5cf6);
+                    background-size: 200% 100%"></div>
```
pxwxnvermx: nit: we usually avoid inline styles. This can be replaced with tailwind classes or by defining custom classes.

commcare_connect/opportunity/views.py
sravfeyn: Can we not use `get_task_progress` (extend as needed) instead of duplicating some of its code in this method, in `_payment_import_task` and in `_add_payment_import_message`? This is what I suggested in point 2 in the [previous review](https://github.com/dimagi/commcare-connect/pull/1326#issuecomment-4923684385)

sravfeyn:
@mkangia 
Thanks for following up through the review. 

I have added [one comment](https://github.com/dimagi/commcare-connect/pull/1326/changes#r3568607895)

commcare_connect/opportunity/views.py
mkangia: Hey @sravfeyn 

https://github.com/dimagi/commcare-connect/pull/1326#issuecomment-4926982489

Just wanted to keep things separate. But if you feel strongly about extending the existing code, I will give it a try.

commcare_connect/opportunity/views.py
mkangia: https://github.com/dimagi/commcare-connect/pull/1326/commits/cfd5f552782c791a83015c42d200072671550485

commcare_connect/opportunity/views.py
mkangia: https://github.com/dimagi/commcare-connect/pull/1326/commits/cfd5f552782c791a83015c42d200072671550485

commcare_connect/opportunity/views.py
mkangia: Great, Updated this and rest

https://github.com/dimagi/commcare-connect/pull/1326/commits/cfd5f552782c791a83015c42d200072671550485

commcare_connect/opportunity/views.py
mkangia: oh yes absolutely,

https://github.com/dimagi/commcare-connect/pull/1326/commits/cfd5f552782c791a83015c42d200072671550485

commcare_connect/templates/components/task_progress_base.html
mkangia: https://github.com/dimagi/commcare-connect/pull/1326/commits/fecf2e97dd8c67ae4238ae4ebcbfe3593e6f68e4

looks like something we should add for claude in the repo

mkangia:
thanks for the review @pxwxnvermx , updated. Please re-review.

commcare_connect/opportunity/views.py
sravfeyn: I think its usually better to have DRY/unified methods to interact with celery, otherwise this could set a pattern for task/progress fetching code in more places in future. Feel free to do it in a followup PR if this is urgent.

mkangia: Not sure I understand the concern here.

> unified methods to interact with celery, otherwise this could set a pattern for task/progress fetching code in more places in future

After the refactor we now have `set_task_progress` and `get_task_progress` to interact with celery, under celery utils. And then those are used where needed accordingly.
Can you share what repetition you are referring to that is not resolved?

sravfeyn: Ha, I see your latest commits now, I thought you didn't do it based on above comment. Thanks.

I will comment on the latest code since this is outdated.

mkangia: Thanks, do you mind sharing an approval if this looks okay now? I believe there is no blocking review anymore.

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,52 @@ class WorkerPaymentsView(BaseWorkerListView):
...
+        # polling progress bar (added to context in get_extra_context).
+        if not request.htmx and self._payment_import_complete():
+            self._add_payment_import_message()
+        return super().get(request, org_slug, opp_id)
+
+    def _payment_import_task(self):
+        task_id = self.request.GET.get("payment_import_task_id")
+        if not task_id:
+            return None
+        task = AsyncResult(task_id)
```
sravfeyn, line 2835: My suggestion from [earlier comment](https://github.com/dimagi/commcare-connect/pull/1326#discussion_r3568607895) is to extend `get_task_progress` to return all the info you require in this view like below

```
{
    "status": status,
    "complete": status == CELERY_TASK_SUCCESS,
    "message": get_task_progress_message(task),
    "result": task_meta.get("result"),
}
```
That way the view here can just use it without querying for the task here.

At this point, I would leave this up to you.

commcare_connect/opportunity/views.py
```diff
@@ -2795,8 +2820,52 @@ class WorkerPaymentsView(BaseWorkerListView):
...
     active_tab = "payments"
 
+    def get(self, request, org_slug, opp_id):
+        # A finished import surfaces its result as a banner; a running one keeps the
+        # polling progress bar (added to context in get_extra_context).
+        if not request.htmx and self._payment_import_complete():
+            self._add_payment_import_message()
+        return super().get(request, org_slug, opp_id)
+
+    def _payment_import_task(self):
```
sravfeyn, line 2831: This is being called multiple times, is a cached_property decorator worth it?

pxwxnvermx (approved):

commcare_connect/opportunity/views.py
mkangia: Good point, 

https://github.com/dimagi/commcare-connect/pull/1326/commits/f8cd7b35c2a03b952c908b3bb924668cbdc7950d

mkangia:
Hey @pxwxnvermx @sravfeyn 

Made couple of improvements. Please help with hopefully one final review pass.

sravfeyn:
Are you planning to address [this comment](https://github.com/dimagi/commcare-connect/pull/1326/changes#r3614028136)? Otherwise I can approve.

commcare_connect/opportunity/views.py
mkangia: Thanks for the suggestion @sravfeyn 

I am going to leave it as it is.

mkangia:
Hey @sravfeyn 

I am going to leave it as it is now. I believe the feedback is non blocking. In your opinion if its a big improvement I can take as a follow up once I am back.

sravfeyn (approved):
If possible I would [followup](https://github.com/dimagi/commcare-connect/pull/1326/changes#r3614028136) in separate PR as that's high value.

mkangia:
Appreciate it, thank you @sravfeyn

PR #1328, opened 2026-06-30
Fix Visit Panel Height

hemant10yadav:
## Product Description

<video controls width="100%" src="https://github.com/user-attachments/assets/5217e480-328e-4585-b94c-0006a2beb424"></video>

## Technical Summary
https://dimagi.atlassian.net/browse/CI-726

While investigating the issue, I traced it back to PR #1184, where `min-height` was removed to keep the task actions visible. The real issue was the button placement, not the panel height.

This change:

* Restores `min-h-[600px]` on the visit and task detail panels.
* Moves "Create Task" and "Delete Tasks" into the tab bar as icon buttons with tooltips, matching the existing pattern.

### Safety story

Template-only change. No backend logic changed; the buttons still use the same Alpine.js handlers and HTMX endpoints.

### Automated test coverage

No tests added. UI-only change.

### QA Plan

Verified locally that the panel height is consistent, the actions remain visible, and both buttons work as expected.

### Labels & Review

* [x] The set of people pinged as reviewers is appropriate for the level of risk of the change.

### DIFF
--- a/commcare_connect/templates/opportunity/user_tasks.html
+++ b/commcare_connect/templates/opportunity/user_tasks.html
@@ -26,12 +26,14 @@
         {% if can_manage_tasks %}
           <button type="button"
                   class="button-icon shadow-sm"
+                  aria-label="{% trans 'Create Task' %}"
                   @click.stop="showCreateTaskModal = true"
                   x-tooltip.raw="{% trans 'Create Task' %}">
             <i class="fa-solid fa-plus text-brand-deep-purple"></i>
           </button>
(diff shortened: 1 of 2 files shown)


Charl1996 (approved):
Would you mind just linking the ticket to the description of the PR?

zandre-eng (approved):

hemant10yadav:
> Would you mind just linking the ticket to the description of the PR?

Sorry, I forgot to add the ticket link. I've updated it now.

PR #1333, opened 2026-06-30
CCCT-2589: Fix FLW summary double-counting already-assigned areas

ajeety4:
## Product Description

Fixes incorrect counts and clarifies wording in the work-area assignment map's **FLW Summary**, **Pending Assignments**, and confirmation modal.

Selecting an FLW auto-selects that FLW's already-assigned areas (for context / unassignment). Those areas were being double-counted, and the counts weren't clearly labelled. 

Changes:

- **FLW Summary** and **Pending Assignments** no longer double-count already-assigned or already-queued areas.
- **Pending Assignments** now shows only the *delta* (the new assignments being made), and now displays **Buildings** and **Expected Visits**.

image
- Delta shown in **FLW Summary**  as + delta (new)

image

- Message on the **confirmation modal** clarifying what its count represents.

image

## Technical Summary

Ticket:[ CCCT-2589](https://dimagi.atlassian.net/browse/CCCT-2589)

- `addToQueue()` queues only newly-selected areas (excluding `flwFilterWorkAreaIds`); FLW Summary = server base + queued + live delta.
- `getSelectedStatsForSummary()` also excludes areas already queued for the FLW (new `getQueuedWorkAreaIdsForAssignee()` helper), so re-selecting a queued area isn't counted twice.
- Selecting an FLW clears `flwFilterWorkAreaIds` before an async fetch repopulates it; the summary now shows its spinner (`flwSummaryLoading`) during the fetch so no total renders against a half-updated filter (fixes a transient inflated count).

### Safety story

Front-end-only change to count computation and labelling in the assignment map. The set of work areas persisted on save is unchanged; only displayed/queued counts and wording are corrected. Verified by tracing selection state through `selectByAssignee` → `addToQueue` → FLW Summary / Pending Assignments / confirm modal rendering.

### Automated test coverage

None — Alpine.js logic embedded in a Django template with no JS test harness in the repo. Manually verified.

### QA Plan

None.

### Labels & Review

### DIFF
--- a/commcare_connect/templates/microplanning/home.html
+++ b/commcare_connect/templates/microplanning/home.html
@@ -341,11 +341,7 @@
             </template>
             <!-- Panel 3: FLW Summary -->
             <div class="border-t border-gray-200 pt-3">
-              <h3 class="text-sm font-semibold mb-2">
-                {% translate "FLW Summary" %}
-                <i class="fa-solid fa-circle-info text-gray-400 ml-1 cursor-help"
-                   x-tooltip.raw="{% translate 'Totals reflect the counts currently assigned to this FLW, including any pending (unsaved) assignments.' %}"></i>
-              </h3>
+              <h3 class="text-sm font-semibold mb-2">{% translate "FLW Summary" %}</h3>
               <div class="mb-2">
                 <label class="block text-xs text-gray-500 mb-1">{% translate "Select FLW" %}</label>
                 <select x-ref="flwSummarySelect"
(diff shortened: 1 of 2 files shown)


sravfeyn (approved):
LGTM

commcare_connect/templates/microplanning/home.html
```diff
@@ -361,15 +365,15 @@ <h3 class="text-sm font-semibold mb-2">{% translate "FLW Summary" %}</h3>
                 <div>
                   <dl class="space-y-2">
                     <div class="grid grid-cols-2 items-center gap-2 text-sm">
-                      <dt class="text-gray-500">{% translate "Count of Assigned Buildings" %}</dt>
+                      <dt class="text-gray-500">{% translate "Total Buildings" %}</dt>
```
zandre-eng, line 368: Would users want to see the breakdown between currently assigned and pending here? We could display the value as something like `x + y` where x is currently assigned and y is pending.

With the above said, happy to leave this as is but just thought I'd bring it up for discussion.

commcare_connect/templates/microplanning/home.html
```diff
@@ -331,7 +331,7 @@ <h3 class="text-sm font-semibold mb-2">
                       </div>
                       <button type="button"
                               @click="removeFromQueue(index)"
-                              class="text-red-500 hover:text-red-700 ml-2 shrink-0">
+                              class="text-red-500 hover:text-red-700 ml-2 shrink-0 cursor-pointer">
```
zandre-eng, line 334: What's the reasoning behind this change? Do our other delete buttons in Connect display a cursor on hover?

zandre-eng (approved):
Looks good, just two non-blocking queries from my side.

hemant10yadav (approved):

commcare_connect/templates/microplanning/home.html
ajeety4: This is a good idea. We do show the changes in "Pending Assignments" section after saving. I think this will a good preview for the end user before clicking on Save. 
I will check once with product and get their inputs on it.

commcare_connect/templates/microplanning/home.html
ajeety4: Added as a good UX practice to signal that the icon is clickable. Connect currently follows a mixed approach here -  buttons built on our design-system classes (.button, .button-icon, etc.) get cursor-pointer for free, but several bare <button> elements don't have it. This keeps the remove icon consistent with the former; happy to standardize the rest separately if we want.

commcare_connect/templates/microplanning/home.html
ajeety4: Included in https://github.com/dimagi/commcare-connect/pull/1333/commits/78bcb9afe3b6a259e4dcc5c127556d747e754202 and updated the screenshot.

I removed the tooltip as now the numbers are self explanatory.

zandre-eng: Thanks for following up on this!

zandre-eng (approved):

PR #1339, opened 2026-07-01
Add audio attachment playback on visit verification page

hemant10yadav:
## Product Description
Screenshot 2026-07-02 at 12 30 06 AM

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-2607

Audio attachments in the download task are now routed to a new `AudioAttachment` model instead of the generic `BlobMeta` table, keyed by `content_type` starting with `audio/`. Download stays idempotent since it's a `get_or_create` on `(user_visit, name)`, so re-processing a visit's attachments won't create duplicates. Transcript and translation fields live on the model now so LLOs can attach that data later, but nothing populates them yet in this PR.

If you'd like to better understand the requirements, you can review the spec linked in this [branch](https://github.com/dimagi/commcare-connect/pull/new/ce/audio-attachments.).

Additional information can be found in this [document](https://docs.google.com/document/d/1dsozUHyEIMFrfg3zXFM6MIAkl-1bp9Y16KkNFSgOATs/edit?tab=t.g3ve57aqbc5o.)

### Safety story

Tested locally against sample visits with audio attachments. The new fetch view checks that the attachment's visit belongs to the requesting opportunity before serving the file, so it can't leak audio across orgs. The migration only adds a new table, so existing data is untouched.

### Automated test coverage

Added tests for the model, the download routing logic, and the fetch view (including the opportunity-scoping check); all passing.

### QA Plan

### Labels & Review


commcare_connect/opportunity/models.py
```diff
@@ -993,6 +997,26 @@ class Meta:
...
+    content_type = models.CharField(max_length=255, null=True)
+    content_length = models.IntegerField()
+    transcript = models.TextField(blank=True, default="")
+    translation = models.TextField(blank=True, default="")
+    date_created = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        unique_together = [
+            ("user_visit", "name"),
+        ]
```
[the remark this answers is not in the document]
hemant10yadav: this pattern is common in our codebase.

ajeety4 (approved):
LGTM.
Just a non blocking query

commcare_connect/templates/opportunity/user_visit_details.html
```diff
@@ -91,6 +91,34 @@
...
+        <div class="flex flex-col gap-2">
+          <audio controls preload="none" class="w-full">
+            <source src="{% url 'opportunity:fetch_audio_attachment' org_slug=request.org.slug opp_id=user_visit.opportunity.opportunity_id pk=audio.pk %}"
+                    type="{{ audio.content_type }}">
+            {% translate "Your browser does not support the audio element." %}
+          </audio>
+          {% if audio.transcript %}
+            <div>
+              <p class="text-sm font-medium text-brand-blue-light">{% translate "Transcript" %}</p>
+              <p class="text-sm font-normal text-brand-deep-purple">{{ audio.transcript }}</p>
```
ajeety4, line 107: Do we have any idea of how long can transcript or translation be ?
Checking in case if it makes the UI too cluttered for long text.

hemant10yadav: That's a good point. I agree this could become a UI issue for longer transcripts or translations. I've made some changes to address this in the subsequent PR, which you can see [here](https://github.com/dimagi/commcare-connect/pull/1340/changes/133505c0af5ea8c09ac48aab4fb7635cb487dd16).

commcare_connect/opportunity/tasks.py
```diff
@@ -376,28 +377,58 @@ def send_push_notification_task(
...
+            defaults={
+                "content_length": blob["length"],
+                "content_type": blob["content_type"],
+            },
+        )
+        if not created:
+            # attachment already exists
+            return
+        content = _fetch_hq_attachment(api_key, url)
+        default_storage.save(str(audio.blob_id), ContentFile(content, name))
```
pxwxnvermx: These functions have similar signatures, the xform_id can be taken from the user_visit. These can be combined into one with the content_type if else statement happening in the combined function.

pxwxnvermx (approved):
Just some thoughts regarding the two save attachment function, rest of the changes look good to me.

commcare_connect/opportunity/models.py
```diff
@@ -993,6 +997,26 @@ class Meta:
         indexes = [models.Index(fields=["blob_id"])]
 
 
+class AudioAttachment(models.Model):
+    user_visit = models.ForeignKey(
+        "UserVisit",
```
zandre-eng: Is there a reason this is using a lazy reference instead passing the class directly?

commcare_connect/opportunity/tasks.py
```diff
@@ -376,28 +377,58 @@ def send_push_notification_task(
...
-                # attachment already exists
-                continue
-            response = httpx.get(
-                url,
-                headers={"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"},
-            )
-            default_storage.save(str(blob_meta.blob_id), ContentFile(response.content, name))
+        content_type = blob.get("content_type") or ""
+        if user_visit is not None and content_type.startswith("audio/"):
+            _save_audio_attachment(api_key, url, name, blob, user_visit)
```
zandre-eng: This will only save audio attachments for newly processed forms. Do we have to consider backfilling for any pre-existing audio attachments that are currently saved as blob attachments?

commcare_connect/opportunity/views.py
```diff
@@ -1146,6 +1147,21 @@ def fetch_attachment(request, org_slug, opp_id, blob_id):
     return FileResponse(attachment, filename=blob_meta.name, content_type=blob_meta.content_type)
 
 
+@org_member_required
+@opportunity_required
+def fetch_audio_attachment(request, org_slug, opp_id, pk):
+    audio = get_object_or_404(AudioAttachment, pk=pk)
+
+    if audio.user_visit.opportunity_id != request.opportunity.id:
+        return HttpResponseNotFound()
```
zandre-eng: Nit: Instead of having a separate check here you could incorporate this into the query above:
```
audio = get_object_or_404(
    AudioAttachment,
     pk=pk,
     user_visit__opportunity=request.opportunity
)
```

zandre-eng (approved):
One question about backfilling, but otherwise minor nits.

mkangia:
Hey @hemant10yadav 

Can the PR highlight why a new model was needed?

commcare_connect/opportunity/models.py
hemant10yadav: Good Catch I have used the class now.

commcare_connect/opportunity/tasks.py
hemant10yadav: No backfill is needed because we didn't support audio previously.

commcare_connect/opportunity/views.py
hemant10yadav: Thanks Added

commcare_connect/opportunity/tasks.py
hemant10yadav: Thanks I have added the change

hemant10yadav:
> Hey @hemant10yadav
> 
> Can the PR highlight why a new model was needed?

Audio attachments need transcript/translation fields for LLOs to fill in later, and routing by content_type at download time keeps it separate from BlobMeta without touching existing attachment handling. For additional context, please refer to this [document](https://docs.google.com/document/d/1dsozUHyEIMFrfg3zXFM6MIAkl-1bp9Y16KkNFSgOATs/edit?tab=t.g3ve57aqbc5o).

zandre-eng (approved):

commcare_connect/opportunity/tasks.py
```diff
@@ -376,28 +377,42 @@ def send_push_notification_task(
...
+    response = httpx.get(
+        url,
+        headers={"Authorization": f"ApiKey {api_key.user.email}:{api_key.api_key}"},
+    )
+    return response.content
+
+
+def _save_attachment(api_key, url, name, blob, xform_id, user_visit=None):
+    content_type = blob.get("content_type") or ""
+    if user_visit is not None and content_type.startswith("audio/"):
```
ctsims, line 398: The flexibility on the content type here is good. I think the content types here are going to be a little funny, btw, because there's a lot of ambiguity / metadata lossage between different layers, so the more flexible we can be the better. 

One of the things we'll need to evaluate that's a little gross is that there's a good chance we'll get some submissions with `.m4a` extensions but are `ogg` containers internally, for instance.

ctsims (approved):
This is looking pretty good to me.

Have we been able to assess the content type inputs from HQ yet on what kinds of files it thinks these are? 

I'll admit that the practice here currently isn't good - we're almost certainly shipping up AAC encoded MP4 files in files with `.mp3` extensions, for example, so if it's pulling content types from filenames it might be problematic

commcare_connect/opportunity/tasks.py
hemant10yadav: That makes sense. Removing the explicit type attribute seems like the right approach for now, especially since MIME type isn't reliable. We can revisit this later if we see playback issues or decide to audit audio MIME types more broadly.

hemant10yadav:
Tested this on staging and it is working fine - 
https://connect-staging.dimagi.com/a/nitins-network-manager/opportunity/6155eec2-d61e-4e80-a3b7-9a76fc2354e2/user_visits/?user=f5778674-9998-47e1-b0f4-216b4a73e504

pxwxnvermx (approved):

PR #1340, opened 2026-07-01
Add transcript and translation editing for audio attachments

hemant10yadav:
## Product Description
Screenshot 2026-07-02 at 12 30 06 AM
Screenshot 2026-07-02 at 12 30 36 AM

## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-2607

Depends on #1339 (audio attachment model and playback), so this diff only shows the transcribe feature on top of it. Program managers are blocked from the transcribe page since transcripts fall outside their scope. Saving redirects back to the worker's filtered visit list rather than the single visit detail page, since that's where reviewers land this feature from.

### Safety story

Tested locally: submitting a transcript/translation saves correctly, the transcribe page 404s for a mismatched opportunity or a program manager, and the existing audio playback tests still pass.

### Automated test coverage

Added tests for rendering the transcribe form, saving it, and the two 404 cases; all passing.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/templates/opportunity/user_visit_details.html
+++ b/commcare_connect/templates/opportunity/user_visit_details.html
@@ -114,6 +114,16 @@
               <p class="text-sm font-normal text-brand-deep-purple">{{ audio.translation }}</p>
             </div>
           {% endif %}
+          {% if not request.is_opportunity_pm %}
+            <a href="{% url 'opportunity:audio_attachment_transcribe' org_slug=request.org.slug opp_id=user_visit.opportunity.opportunity_id pk=audio.pk %}"
+               class="button button-sm outline-style w-fit">
+              {% if audio.transcript or audio.translation %}
+                <i class="fa-regular fa-pen-to-square mr-1"></i>{% translate "Edit Transcript" %}
+              {% else %}
+                <i class="fa-solid fa-plus mr-1"></i>{% translate "Add Transcript" %}
+              {% endif %}
+            </a>
+          {% endif %}
         </div>
       {% endfor %}
     </div>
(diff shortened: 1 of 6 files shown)


hemant10yadav:
@coderabbitai review

commcare_connect/templates/opportunity/user_visit_details.html
```diff
@@ -103,17 +103,45 @@
...
+              <p x-ref="transcriptText"
+                 class="text-sm font-normal text-brand-deep-purple"
+                 :class="expanded ? '' : 'line-clamp-3'">{{ audio.transcript }}</p>
+              <button type="button"
+                      x-show="overflowing"
+                      x-cloak
+                      @click="expanded = !expanded"
+                      class="block w-fit ml-auto text-sm font-medium text-brand-blue-light hover:underline"
+                      x-text="expanded ? '{% translate 'Show less' %}' : '{% translate 'Show more' %}'"></button>
             </div>
```
hemant10yadav, line 118: <video src='https://github.com/user-attachments/assets/828630ef-cdb2-408c-b359-259a31ea4198'>

zandre-eng (dismissed):

ctsims (dismissed):
This looks good to me on a first pass. 

From a spec perspective it definitely won't always be the case that users should be able to edit transcripts even when they are available (that'll be pretty specific to individual programs), but I don't think that's worth worrying about upfront since we don't yet know how we're going to differentiate the datatypes.

hemant10yadav:
> From a spec perspective it definitely won't always be the case that users should be able to edit transcripts even when they are available (that'll be pretty specific to individual programs), but I don't think that's worth worrying about upfront since we don't yet know how we're going to differentiate the datatypes.

That's fine. I think we can address that later by putting the edit functionality behind a feature flag that can be configured at the program, organization, or even specific opportunity level. That should be a relatively small change when we need it.

pxwxnvermx (dismissed):

pxwxnvermx (approved):

PR #1344, opened 2026-07-02
Cs/ccct 2551 ocs oauth

Charl1996:
## Product Description

Groundwork for the Open Chat Studio (OCS) integration. Adds the OAuth2 plumbing so a signed-in user can **connect their OCS account** to CommCare Connect and Connect can call OCS on their behalf (list chatbots, trigger bots). No user-facing feature is wired up yet beyond a reusable "Connect OCS account" prompt component; this PR is the auth + API-client foundation those features will build on.

## Technical Summary
Ticket: https://dimagi.atlassian.net/browse/CCCT-2551
Spec: https://github.com/dimagi/commcare-connect/blob/d9f89841c7deed170a613967d0b013016cf421f1/docs/superpowers/specs/2026-06-19-ocs-conversations-design.md#ocs-api-client

- **New `ocs_provider` app** — an allauth OAuth2 provider (`id="ocs"`) with an adapter whose authorize/token/userinfo URLs derive from a new `OCS_BASE_URL` setting. PKCE (S256) is enabled (OCS requires it); default scopes are `openid chatbots:read chatbots:interact`; the account UID comes from the userinfo `sub` claim.
- **Provider-agnostic token refresh** (`utils/oauth_tokens.py`) — `refresh_access_token(user, provider, token_url)` returns a valid `SocialToken`, refreshing via the `refresh_token` grant when expired. Uses `select_for_update` to serialise concurrent refreshes (rotating single-use refresh tokens), honours the response's `expires_in`, and raises `TokenRefreshError` / `SocialTokenMissingError` on failure. Replaces the dead, provider-unfiltered `refresh_access_token` removed from `commcarehq_api.py` (HQ uses API keys, so it had no callers).
- **OCS API client** (`utils/ocs_api.py`) — `user_has_connected_ocs`, `list_chatbots` (follows cursor pagination), and `trigger_bot`, all fetching a valid bearer token via the refresh util and wrapping HTTP/network failures in `OcsApiError`.
- **Account linking** (`users/adapters.py`) — lets an OCS `connect` flow bypass the SSO "email already exists" guard so an OCS account can be linked to an already-authenticated user; other providers/processes are unaffected.
- **Error handling** (`utils/middleware.py`) — a raised `TokenRefreshError` now redirects the user with a message prompting them to (re)connect their OCS account.

## Safety Assurance
I tested the oauth flow locally (using ngrok) as well as on staging. I used [this button](https://github.com/dimagi/commcare-connect/pull/1347) to trigger the flow from the UI. 

### Safety story

Nothing in the existing HQ auth path changes behaviour — the only edits to shared code are 
(a) removing an unused HQ `refresh_access_token` that had no callers
(b) an adapter guard scoped to `provider == "ocs"` + `AuthProcess.CONNECT`

Login/signup for existing providers is unchanged (covered by tests). OCS endpoints are driven by `OCS_BASE_URL`, defaulting to the public OCS URL and overridable per environment.

### Automated test coverage

Extensive unit tests accompany every module: provider config/PKCE/scope/uid extraction, adapter URL binding and route registration, the connect-prompt template, token-refresh happy paths + concurrency/expiry/error cases + correct-provider isolation, the OCS API client (pagination, bearer header, optional-payload handling incl. falsey values, error mapping), the adapter email-guard bypass, and the middleware redirect.

### QA Plan
QA not happening for this PR.

### Labels & Review


commcare_connect/ocs_provider/provider.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class OcsProvider(OAuth2Provider):
+    id = "ocs"
+    name = "Open Chat Studio"
+    account_class = OcsAccount
+    pkce_enabled_default = True
+
+    def get_default_scope(self):
+        return ["openid", "chatbots:read", "chatbots:interact"]
```
snopoke: you should include `sessions:read`

commcare_connect/utils/middleware.py
```diff
@@ -24,6 +34,10 @@ def process_exception(self, request, exception):
             api_url = "#"  # TODO: make this a real URL
             messages.error(request, mark_safe(API_KEY_ERROR.format(url=api_url)))
             return HttpResponseRedirect(request.headers["referer"])
+        if isinstance(exception, TokenRefreshError):
+            connect_url = f"{reverse('ocs_login')}?{urlencode({'process': 'connect', 'next': request.path})}"
+            messages.error(request, format_html(OCS_CONNECT_ERROR, url=connect_url))
+            return HttpResponseRedirect(request.headers.get("referer", "/"))
```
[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/commcare-connect/pull/1344/commits/3112f8ca238150cc545f2b74a9ac836a0517cbc2

commcare_connect/utils/ocs_api.py
```diff
@@ -0,0 +1,78 @@
...
+def trigger_bot(
+    user,
+    *,
+    identifier,
+    experiment,
+    participant_data=None,
+    start_new_session=None,
+    message_text=None,
+    prompt_text=None,
+    session_data=None,
```
zandre-eng: Nit: Would it be possible to add type hints or descriptions in the docstring for these params? Some of these don't feel completely obvious with what the expected format is, so it would be nice to see that with type hints.

Feel free to push back if you don't think it's necessary.

commcare_connect/ocs_provider/views.py
```diff
@@ -0,0 +1,22 @@
...
+    provider_id = OcsProvider.id
+    access_token_url = f"{settings.OCS_BASE_URL}/o/token/"
+    authorize_url = f"{settings.OCS_BASE_URL}/o/authorize/"
+    profile_url = f"{settings.OCS_BASE_URL}/o/userinfo/"
+    redirect_uri_protocol = "https"
+
+    def complete_login(self, request, app, token, **kwargs):
+        response = httpx.get(self.profile_url, headers={"Authorization": f"Bearer {token}"}, timeout=10)
+        extra_data = response.json()
+        return self.get_provider().sociallogin_from_response(request, extra_data)
```
zandre-eng, line 21: Something like a non-200 response, network blip, or bad response could raise a 500 here. It would be nice to have some error handling to catch and surface these issues a bit more cleanly.

commcare_connect/utils/middleware.py
```diff
@@ -23,7 +34,20 @@ def process_exception(self, request, exception):
         if isinstance(exception, CommCareTokenException):
             api_url = "#"  # TODO: make this a real URL
             messages.error(request, mark_safe(API_KEY_ERROR.format(url=api_url)))
-            return HttpResponseRedirect(request.headers["referer"])
+            return HttpResponseRedirect(_safe_referer(request))
+        if isinstance(exception, TokenRefreshError):
+            connect_url = f"{reverse('ocs_login')}?{urlencode({'process': 'connect', 'next': request.path})}"
+            messages.error(request, format_html(OCS_CONNECT_ERROR, url=connect_url))
```
zandre-eng: `refresh_access_token` was made provider-agnostic, however here we translate any `TokenRefreshError` into an OCS specific message and login redirect. This isn't a problem now since OCS is the only caller, but a future provider on this util would silently send users to the OCS connect flow. Could you add a comment  on `refresh_access_token` noting that its `TokenRefreshError` is handled as OCS specific here?

commcare_connect/utils/oauth_tokens.py
```diff
@@ -0,0 +1,73 @@
...
+    access_token = data.get("access_token")
+    if not access_token:
+        raise TokenRefreshError(f"Token refresh response missing access_token: {response.text}")
+
+    token.token = access_token
+    if data.get("refresh_token"):
+        token.token_secret = data["refresh_token"]
+
+    if data.get("expires_in") is not None:
+        token.expires_at = timezone.now() + datetime.timedelta(seconds=int(data["expires_in"]))
```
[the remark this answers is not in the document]
zandre-eng: This looks like like a valid concern. It might be the case that OCS always returns an `expires_in` but since this is provider-agnostic another provider might omit this value.

commcare_connect/utils/ocs_api.py
```diff
@@ -0,0 +1,78 @@
...
+    url = f"{settings.OCS_BASE_URL}/api/v2/chatbots/"
+    chatbots = []
+    while url:
+        try:
+            response = httpx.get(url, headers=headers, timeout=OCS_HTTP_TIMEOUT)
+            response.raise_for_status()
+        except httpx.HTTPStatusError:
+            raise OcsApiError(f"Failed to list chatbots: {response.text}")
+        except httpx.RequestError as e:
+            raise OcsApiError(f"Failed to list chatbots: {e}")
```
zandre-eng, line 33: Nit: These drop the original traceback, so the full traceback won't get preserved here:
```suggestion
        except httpx.HTTPStatusError as e:
            raise OcsApiError(f"Failed to list chatbots: {response.text}") from e
        except httpx.RequestError as e:
            raise OcsApiError(f"Failed to list chatbots: {e}") from e
```

commcare_connect/utils/middleware.py
```diff
@@ -1,16 +1,27 @@
...
 
 API_KEY_ERROR = """
     Unable to retrieve applications from CommCare HQ.<br>
     Please re-login using CommCare HQ or add a <a href="{url}">CommCare API Key</a>.
 """
 
+OCS_CONNECT_ERROR = (
+    "Your Open Chat Studio account isn't connected (or its session expired).<br>"
+    'Please <a href="{url}">connect your OCS account</a> and try again.'
+)
```
zandre-eng: Minor nit: Translation for this string.

zandre-eng (commented):
Overall generally looks fine. A few comments and questions from my side.

commcare_connect/utils/ocs_api.py
Charl1996: Good suggestion. Some are redundant so I'll remove them anyway.

commcare_connect/ocs_provider/views.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1344/commits/74d96296813cbde89926c6d623708a030246af99

commcare_connect/utils/middleware.py
Charl1996: Excellent point: https://github.com/dimagi/commcare-connect/pull/1344/commits/ad5d953c6bb4f14645a2c7b327fff254e6b1d893

mkangia:
Hey @calellowitz @sravfeyn 

Would appreciate if at least one of you can share a review on this. 
Thanks!

commcare_connect/utils/oauth_tokens.py
```diff
@@ -0,0 +1,74 @@
...
+
+class TokenRefreshError(Exception):
+    """Raised when an OAuth2 access token could not be refreshed."""
+
+
+class SocialTokenMissingError(TokenRefreshError):
+    """Raised when the user has no connected account/token for the provider."""
+
+
+def refresh_access_token(user, provider, token_url, *, force=False):
```
sravfeyn, line 17: nit: I would refactor this and its submethods into a single class. Or even implement this on provider class itself.

commcare_connect/utils/ocs_api.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+        response.raise_for_status()
+    except httpx.HTTPStatusError:
+        raise OcsApiError(f"Failed to trigger bot: {response.text}")
+    except httpx.RequestError as e:
+        raise OcsApiError(f"Failed to trigger bot: {e}")
+    return response.json()
+
+
+def _valid_token(user):
```
sravfeyn: nit: rename to `_get_valid_token`?

commcare_connect/utils/middleware.py
```diff
@@ -1,16 +1,24 @@
...
+from commcare_connect.utils.oauth_tokens import TokenRefreshError
 
 API_KEY_ERROR = """
     Unable to retrieve applications from CommCare HQ.<br>
     Please re-login using CommCare HQ or add a <a href="{url}">CommCare API Key</a>.
 """
 
+TOKEN_REFRESH_ERROR = _(
+    "We couldn't verify one of your connected accounts (it may have been disconnected "
+    "or its session expired). Please reconnect the account and try again."
```
sravfeyn: Reconnect here just means, re-login with social login? And may be `one of your connected accounts` should just be `your (connected) account`.

commcare_connect/utils/middleware.py
```diff
@@ -23,7 +31,19 @@ def process_exception(self, request, exception):
         if isinstance(exception, CommCareTokenException):
             api_url = "#"  # TODO: make this a real URL
             messages.error(request, mark_safe(API_KEY_ERROR.format(url=api_url)))
-            return HttpResponseRedirect(request.headers["referer"])
+            return HttpResponseRedirect(_safe_referer(request))
+        if isinstance(exception, TokenRefreshError):
+            messages.error(request, TOKEN_REFRESH_ERROR)
+            return HttpResponseRedirect(_safe_referer(request))
```
sravfeyn, line 37: Shouldn't this be signin page?

commcare_connect/ocs_provider/tests/test_adapter.py
```diff
@@ -0,0 +1,16 @@
+from django.conf import settings
+from django.urls import reverse
+
+from commcare_connect.ocs_provider.views import OcsOAuth2Adapter
+
+
+def test_adapter_urls_derive_from_ocs_base_url():
```
sravfeyn: Do we really need these two tests?

Charl1996: Probably not of big value, so removing them.

sravfeyn:
May be this is a bit too late, just curious, did we really need OCS based signup aren't we doing it just to connect OCS which could also be accomplished with API tokens?

Charl1996:
> could also be accomplished with API tokens

See the design decisions [here](https://github.com/dimagi/commcare-connect/blob/d9f89841c7deed170a613967d0b013016cf421f1/docs/superpowers/specs/2026-06-19-ocs-conversations-design.md#design-decisions-resolved-during-brainstorming--grilling) specifying that using an API token was explicitly decided against. I guess @pxwxnvermx can give more detail as to why using an API key is less desirable in this case as well.

commcare_connect/utils/middleware.py
Charl1996: No, not exactly, because for an OCS refresh error the user should still be authenticated with Connect, so we shouldn't redirect the user to log in to Connect again for this case.

sravfeyn (approved):

commcare_connect/utils/oauth_tokens.py
```diff
@@ -0,0 +1,74 @@
...
+                "client_id": token.app.client_id,
+                "client_secret": token.app.secret,
+                "refresh_token": token.token_secret,
+            },
+            timeout=10,
+        )
+    except httpx.RequestError as e:
+        raise TokenRefreshError(f"Failed to refresh token: {e}") from e
+    if response.status_code != 200:
+        raise TokenRefreshError(f"Failed to refresh token: {response.text}")
```
pxwxnvermx, line 61: This could be simplified with a `response.raise_for_status()` for non-2xx status codes and then re-raising the TokenRefreshError from the HttpStatusError raised by httpx for the status.

commcare_connect/commcarehq_provider/views.py
```diff
@@ -15,6 +16,8 @@ class CommcareHQOAuth2Adapter(OAuth2Adapter):
 
     def complete_login(self, request, app, token, **kwargs):
         response = httpx.get(self.profile_url, headers={"Authorization": f"Bearer {token}"})
+        if response.status_code != 200:
+            raise OAuth2Error("Failed to fetch profile data from CommCare HQ.")
```
pxwxnvermx, line 20: This should also change to use `raise_for_status` it covers a larger subset of status_code which can lead to the API failing.

commcare_connect/ocs_provider/provider.py
```diff
@@ -0,0 +1,26 @@
...
+    pkce_enabled_default = True
+
+    def get_default_scope(self):
+        return ["openid", "chatbots:read", "chatbots:interact", "sessions:read"]
+
+    def extract_uid(self, data):
+        uid = data.get("sub")
+        if not uid:
+            raise ValueError("OCS userinfo response missing 'sub'")
+        return str(uid)
```
pxwxnvermx, line 23: How do you feel about reducing the code to the suggestion below? I think the KeyError would also highlight the same information as above.

```suggestion
        return str(data["sub"])
```

commcare_connect/ocs_provider/views.py
```diff
@@ -0,0 +1,25 @@
...
+    provider_id = OcsProvider.id
+    access_token_url = f"{settings.OCS_BASE_URL}/o/token/"
+    authorize_url = f"{settings.OCS_BASE_URL}/o/authorize/"
+    profile_url = f"{settings.OCS_BASE_URL}/o/userinfo/"
+    redirect_uri_protocol = "https"
+
+    def complete_login(self, request, app, token, **kwargs):
+        response = httpx.get(self.profile_url, headers={"Authorization": f"Bearer {token}"}, timeout=10)
+        if response.status_code != 200:
+            raise OAuth2Error("Failed to fetch profile data from Open Chat Studio.")
```
pxwxnvermx, line 19: Same comment regarding using `response.raise_for_status()`.

pxwxnvermx (commented):
Just some comments on the handling the non-2xx status codes on httpx response. I think we should prefer using the `response.raise_for_status()` method wherever we have to check for non-2xx. Rest of the changes look good.

Charl1996:
@pxwxnvermx I will address the API changes in [this PR](https://github.com/dimagi/commcare-connect/pull/1359) since that PR covers the usage of the APIs

PR #1345, opened 2026-07-02
Work Area Inaccessibility Request Status

zandre-eng:
## Product Description

Field workers can now raise a new "inaccessible work area" request after a previous request for that work area has been reviewed. Previously, once a request had been denied, the worker was permanently blocked from ever submitting another for the same work area (the request submission would silently fail). The program manager's review flow is unchanged, so they approve or deny as before.

As defensive hardening, if a program manager clicks "Review Inaccessible" for a work area whose request is no longer pending (for example, the map is showing a stale status), the review modal now shows an error toast instead of opening blank:
Screenshot_20260702_153013

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2575)

The inaccessibility request flow was updated in [this PR](https://github.com/dimagi/commcare-connect/pull/1225) to prevent multiple inaccessibility requests per work area being created, preventing a `MultipleObjectsReturned` exception from happening in the review lookup. This fix presented a new problem, because reviewed requests are never deleted, so the lingering row blocked every subsequent request. Rather than delete reviewed requests (losing history), the changes from this PR gives each request a review status (`PENDING`/`APPROVED`/`DENIED`) and enforces a **partial** unique constraint on pending requests only. This preserves the full audit trail while still guaranteeing a single active request per work area, and all request lookups (form-receiver guard and both review views) are scoped to the pending request.

This PR also includes a backfilling migration to correctly apply the status for all pre-existing inaccessibility requests. This should be safe considering the low row counts for inaccessibility requests on both the staging and prod environments.

### Safety story

The change is additive and low-risk: a new status field with a partial unique constraint, plus a data migration that backfills existing rows from their work area's current status (inferring `PENDING`/`APPROVED`/`DENIED`). Because the pre-existing unique constraint already allowed at most one request row per work area, the partial constraint cannot fail to apply, and the backfill also clears orphaned rows that were silently blocking re-requests. The client change only affects an error path (it prevents opening an empty modal) and cannot regress the success path.

### Automated test coverage

Tests cover the end-to-end re-request flow (a worker can re-raise only after the prior request is resolved, and a second request is rejected while one is pending), that approve/deny stamps the correct outcome on the request, and that both the review modal and the review action target the pending request and return 404 rather than acting on a historical one.

### QA Plan

### Labels & Review


sravfeyn (approved):
LGTM, though I would prune out some of non-essential tests

hemant10yadav (approved):

zandre-eng:
> LGTM, though I would prune out some of non-essential tests

@sravfeyn I have refactored the tests. Let me know if you think there are any other that are non-essential.

sravfeyn (approved):
Thanks. LGTM

PR #1353, opened 2026-07-06
Add task mode and OCS fields to TaskType and AssignedTask

mkangia:
## Product Description

No direct user-facing UI change. 
This adds data-model support for OCS (Open Chat Studio) chatbot-based tasks alongside existing in-app relearn tasks, and exposes the task mode/slug through the mobile API so the app can
distinguish how a task should be delivered.

## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2550

Best to review by commit
- a small change to use "mode" instead of "type" as suggested by ticket because `task_type.mode` reads well than `task_type.type` when it comes to the choice of the task being ocs or relearn.
- a small refactor to extract "50" limit spread across models into a constant
- date_modified was also added to the AssignedTask model and serializer as requested by the mobile team.

### Safety story

Inherently low-risk, additive-only change

### Labels & Review


commcare_connect/opportunity/api/serializers/mobile.py
```diff
@@ -299,10 +299,21 @@ class Meta:
...
-        fields = ["assigned_task_id", "task_name", "task_description", "status", "due_date", "date_created"]
+        fields = [
+            "assigned_task_id",
+            "task_name",
+            "task_description",
+            "task_mode",
+            "slug",
+            "status",
+            "due_date",
+            "date_created",
```
shubham1g5, line 315: can we also add a `date_modified` to this model and send it to mobile ? Currently there is no way for mobile to know if a task payload has changed since the last time mobile synced these other than comparing every other field in this model and `date_modified` can solve that rather easily for us.

Charl1996 (approved):

Charl1996:
Happy to reapprove once Shubham's suggestion is in.

commcare_connect/opportunity/api/serializers/mobile.py
mkangia: Sure @shubham1g5 

Added date modified.
For existing tasks it's going to be same as the created timestamp. makes sense?

mkangia:
thanks @Charl1996 
that change was more than a simple commit. Requesting another pass on review please.

commcare_connect/opportunity/tests/test_models.py
```diff
@@ -467,8 +467,8 @@ def test_assigned_task_stores_ocs_fields(self):
         assert task.ocs_session_id == "session-1"
 
     def test_assigned_task_date_modified_updates_on_save(self):
-        created_at = datetime.datetime(2026, 1, 1, tzinfo=datetime.timezone.utc)
-        updated_at = datetime.datetime(2026, 1, 2, tzinfo=datetime.timezone.utc)
+        created_at = datetime.datetime(2026, 1, 1, tzinfo=datetime.UTC)
```
Charl1996, line 470: Is there anything different between datetime.timezone.utc and datetime.UTC?

Charl1996 (approved):

commcare_connect/opportunity/api/serializers/mobile.py
shubham1g5: thanks, that sounds good to me.

commcare_connect/opportunity/tests/test_models.py
mkangia: It was something prek was not happy about. 
https://github.com/dimagi/commcare-connect/actions/runs/28869404208/job/85627865098

Claude agreed
> that's prek's pyupgrade hook. The project targets py311.

So seems like a new standard to be followed for updated version of Python.

PR #1354, opened 2026-07-06
Add opportunity breadcrumbs to microplanning pages

mkangia:
## Product Description

The microplanning **Progress Map** page had no breadcrumb navigation, so there was no way to get back to the opportunity or the opportunities list from it. The related **Progress Tracker** (Coverage Progress) page had a breadcrumb, but it only went back as far as the Progress Map.

This adds the standard opportunity breadcrumb chain to both pages so navigation is consistent:
- **Progress Map**: `Opportunities › [opportunity] › Progress Map`
- **Progress Tracker**: `Opportunities › [opportunity] › Progress Map › Progress Tracker`

**BEFORE**

image

--------------

image

--------------
**AFTER**

image

--------------

image

## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2591

### Safety story

Low risk, presentation-only change. Tested locally for UI changes.

### Labels & Review

### DIFF
--- a/commcare_connect/microplanning/views.py
+++ b/commcare_connect/microplanning/views.py
@@ -159,6 +159,14 @@ def microplanning_home(request, *args, **kwargs):
         "mapbox_api_key": settings.MAPBOX_TOKEN,
         "task_id": request.GET.get("task_id"),
         "opportunity": opportunity,
+        "path": [
+            {"title": _("Opportunities"), "url": reverse("opportunity:list", kwargs={"org_slug": request.org.slug})},
+            {
+                "title": opportunity.name,
+                "url": reverse("opportunity:detail", args=(request.org.slug, opportunity.opportunity_id)),
+            },
+            {"title": _("Progress Map")},
+        ],
         "metrics": get_metrics_for_microplanning(opportunity),
         "tiles_url": tiles_url,
         "visit_tiles_url": visit_tiles_url,
@@ -1002,6 +1010,11 @@ def coverage_progress(request, *args, **kwargs):
         "filter_form": filterset.form,
         "export_hrefs": export_hrefs,
         "path": [
+            {"title": _("Opportunities"), "url": reverse("opportunity:list", kwargs={"org_slug": request.org.slug})},
+            {
+                "title": opportunity.name,
+                "url": reverse("opportunity:detail", args=(request.org.slug, opportunity.opportunity_id)),
+            },
             {
                 "title": _("Progress Map"),
                 "url": reverse(
--- a/commcare_connect/templates/microplanning/home.html
+++ b/commcare_connect/templates/microplanning/home.html
@@ -12,6 +12,7 @@
 {% endblock javascript %}
 {% block content %}
   <div class="min-w-0">
+    {% include 'components/breadcrumbs.html' with path=path %}
     <h2 class="mb-4 text-brand-deep-purple font-medium">{% translate "Select Opportunity Area" %}</h2>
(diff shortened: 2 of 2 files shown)


zandre-eng (approved):

commcare_connect/microplanning/views.py
```diff
@@ -1002,6 +1010,11 @@ def coverage_progress(request, *args, **kwargs):
         "filter_form": filterset.form,
         "export_hrefs": export_hrefs,
         "path": [
+            {"title": _("Opportunities"), "url": reverse("opportunity:list", kwargs={"org_slug": request.org.slug})},
+            {
+                "title": opportunity.name,
+                "url": reverse("opportunity:detail", args=(request.org.slug, opportunity.opportunity_id)),
+            },
```
pxwxnvermx, line 1017: The breadcrumbs should also include Program > [Program Name] as well. The breadcrumbs without the Program information are for Non-Managed Opps.

mkangia: Hey @pxwxnvermx 

Kept the breadcrumbs based on the path taken.
We go from list of opportunities to the specific opportunity and then to Progress Map, from that page to Progress Tracker.
Let me know if that does not seems like the right way to approach it.

hemant10yadav: Agreed with Pawan here. The breadcrumbs should include Program.

mkangia: Good, I would be happy to add it if I can understand how the Program fits in the picture.

If I were to navigate to Progress Map, it would be via the opportunities list and then clicking on an opportunity to views the Progress Map for it. There is no program selection in that path.

May be you are thinking of a logical model hierarchy where each opportunity would be under a program so it should be `Program > Program Name > Opportunities > Opportunity Name > Progress Map` but then should that also include Organization? 
In order to see what should be in the breadcrumbs, I am looking at the navigation.

Open to discussion.

mkangia: I do see that we have both on connect

image

image

image

Is there a reasoning for which page should have it and which should not?

I asked Claude what it thought and it said

image

the interesting point was that the pages that lead to the Progress Map page, don't include Program either, which makes a good point.

pxwxnvermx: > May be you are thinking of a logical model hierarchy where each opportunity would be under a program so it should be Program > Program Name > Opportunities > Opportunity Name > Progress Map but then should that also include Organization?

I think my approach was more based on the style choices we made when we initially added the New UI. All Managed Opportunities got the Program links in the breadcrumbs and all Non-Managed Opportunities only had the Opportunities > Opportunity Name in the breadcrumbs. This was mostly done to make it easy to identify the Program and the Opportunity at a glance on any page.

> Is there a reasoning for which page should have it and which should not?

To me it seems like that some pages might have just deviated from the above reasoning as it was not communicated anywhere either in Code through comments or through the Readme.

mkangia: Thanks for the inputs you both.

I checked with Product and it seems the breadcrumbs might need to change in near future so any decision we make here might undergo a revisit anyway.

> To me it seems like that some pages might have just deviated from the above reasoning as it was not communicated anywhere either in Code through comments or through the Readme.

Looks like this would also get streamlined in the revisit.

But plus, since the pages leading to the relevant pages for this PR do not have the program path, I'd keep that consistent.

I appreciate the feedback here.

PR #1358, opened 2026-07-07
Fix allauth account-flow UX gaps: dead-end pages, redundant copy

hemant10yadav:
## Product Description

Signup and password-reset screens now always give users a way back to sign in, drop copy that implied an invite ("is an e-mail address for user X"), and no longer stack redundant flash messages on pages that already state the outcome.

[page@4e4ac221b26bec7b6affafd41b44d85d.webm](https://github.com/user-attachments/assets/8dab0250-d95a-4caf-9cb7-11e93e89013d)

## Technical Summary

[CI-767](https://dimagi.atlassian.net/browse/CI-767)

- Added a signup-specific confirmation email template so self-signup no longer reuses the invite-style wording from allauth's default template.
- Pages that already state the outcome in their own copy (verification-sent, password-changed) drain the flash messages instead of displaying them, so they don't leak onto the next page like login.
- Removed dead Alpine `submitWithValidation()` handlers left over from an old form setup; those buttons now submit normally.

### Safety story

Tested the full signup, email confirmation, login, logout, and password-reset flow manually against a local dev server, including invalid-password and invalid-reset-link cases.

### Automated test coverage

No automated tests added or changed; this is a template and copy only change.

### QA Plan

### Labels & Review

[CI-767]: https://dimagi.atlassian.net/browse/CI-767?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/templates/account/email/email_confirmation_signup_message.txt
+++ b/commcare_connect/templates/account/email/email_confirmation_signup_message.txt
@@ -1,0 +1,6 @@
+{% extends "account/email/base_message.txt" %}
+{% load i18n %}
+
+{% block content %}{% autoescape off %}{% blocktrans with site_domain=current_site.domain %}You're receiving this e-mail because you signed up for an account on {{ site_domain }}.
+
+To confirm your e-mail address, go to {{ activate_url }}{% endblocktrans %}{% endautoescape %}{% endblock %}
--- a/commcare_connect/templates/account/password_reset_from_key.html
+++ b/commcare_connect/templates/account/password_reset_from_key.html
@@ -50,15 +50,17 @@
           </button>
         </div>
         <div class="text-sm text-gray-400 text-center mt-auto border-t-gray-200 border-t-2 pt-4 ">
-          Don't have an account?
-          <a href="#"
-             class="text-brand-indigo font-medium"
-             @click="submitWithValidation()"
-             @submit.prevent="submitWithValidation()">Signup</a>
+          {% translate "Don't have an account?" %}
+          <a href="{% url 'account_signup' %}"
+             class="text-brand-indigo font-medium">{% translate "Signup" %}</a>
         </div>
       </form>
     {% else %}
-      <p>{% translate 'Your password is now changed.' %}</p>
+      {# Silently drain messages — this page already says it #}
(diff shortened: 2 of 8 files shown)


commcare_connect/templates/account/email/email_confirmation_signup_message.txt
```diff
@@ -0,0 +1,6 @@
+{% extends "account/email/base_message.txt" %}
```
hemant10yadav, line 1: By default, `email_confirmation_signup_message.txt` includes `email_confirmation_message.txt`, which uses wording intended for someone else registering your email. This override provides a signup-specific message ("you signed up for an account") while leaving the shared template unchanged for the add-email and re-verification flows.

commcare_connect/templates/account/password_reset_from_key.html
```diff
@@ -50,15 +50,17 @@ <h6 class="title text-brand-deep-purple">New Password</h6>
...
-             @submit.prevent="submitWithValidation()">Signup</a>
+          {% translate "Don't have an account?" %}
+          <a href="{% url 'account_signup' %}"
+             class="text-brand-indigo font-medium">{% translate "Signup" %}</a>
         </div>
       </form>
     {% else %}
-      <p>{% translate 'Your password is now changed.' %}</p>
+      {# Silently drain messages — this page already says it #}
+      {% for message in messages %}{% endfor %}
```
hemant10yadav, line 60: Django messages are only cleared when a template iterates over them. Since this page already shows the outcome, we consume them here to prevent them from appearing together on later pages.

zandre-eng (approved):

commcare_connect/templates/account/password_reset_from_key.html
Charl1996: So the django message also shows something like "password changed successfully"?

hemant10yadav: Yep

Charl1996 (approved):

PR #1361, opened 2026-07-08
Add token-based invite flow with revoke and pending invites list

hemant10yadav:
## Product Description

Workspace admins can now invite users who are not on the Connect platform. Invited users can sign up using the invitation link.

[final_1x.webm](https://github.com/user-attachments/assets/47f13bb5-a4ff-4c4e-89c8-03d00fc9c094)

## Technical Summary

[CI-768](https://dimagi.atlassian.net/browse/CI-768)

- Replaced the ad-hoc `invite_id` field on membership with a dedicated `OrganizationInvite` model carrying its own token, status, and 7-day expiry.
- The uniqueness constraint only applies while an invite is pending, so a revoked or expired invite can be sent again without hitting a duplicate error.
- The accept flow now branches on whether the invitee already has an account. Existing users get redirected straight into the workspace. Brand-new users land on a password-setup page that completes allauth signup.

### Safety story

Ran the full organization test suite locally and manually tested the invite, accept, and revoke flows against a local server.

### Automated test coverage

Added and updated tests for the model, form, task, and views. All 71 tests in the organization app pass.

### QA Plan

### Labels & Review

[CI-768]: https://dimagi.atlassian.net/browse/CI-768?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/organization/forms.py
```diff
@@ -147,16 +152,30 @@ def __init__(self, *args, **kwargs):
...
-        if not user:
-            raise ValidationError("User with this email does not exist or is already a member")
+        OrganizationInvite.retire_expired(self.organization, email)
+        if OrganizationInvite.objects.filter(
+            organization=self.organization, email=email, status=OrganizationInvite.Status.INVITED
+        ).exists():
+            raise ValidationError(gettext("An invite has already been sent to this email."))
 
-        self.instance.user = user
         return email
```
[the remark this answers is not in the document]
hemant10yadav: Thanks, fixed in 3f15e9e.

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+            days=self.EXPIRY_DAYS
+        )
+
+    @classmethod
+    def retire_expired(cls, organization, email):
+        """Frees up the unique pending-invite slot so a lapsed invite can be re-sent."""
+        cutoff = timezone.now() - timedelta(days=cls.EXPIRY_DAYS)
+        cls.objects.filter(
+            organization=organization, email=email, status=cls.Status.INVITED, date_created__lt=cutoff
+        ).update(status=cls.Status.EXPIRED, modified_by=email)
```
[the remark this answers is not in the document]
hemant10yadav: Thanks, fixed in 3f15e9e.

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+
+    def accept(self, user):
+        membership, _created = UserOrganizationMembership.objects.get_or_create(
+            organization=self.organization, user=user, defaults={"role": self.role}
+        )
+        self.status = self.Status.ACCEPTED
+        self.modified_by = user.email
+        self.save(update_fields=["status", "modified_by", "date_modified"])
+        return membership
```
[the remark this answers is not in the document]
hemant10yadav: Thanks, fixed in 3f15e9e.

commcare_connect/organization/tables.py
```diff
@@ -41,3 +42,28 @@ class Meta:
...
+        return format_html("<div class=' underline underline-offset-4'>{}</div>", value)
+
+    def render_revoke(self, record):
+        return format_html(
+            '<button type="button" class="button button-sm outline-style text-red-600" '
+            'hx-post="{}" hx-target="#pending_invites_container" hx-swap="innerHTML" hx-confirm="{}">{}</button>',
+            record.get_revoke_url(),
+            gettext("Revoke invite for {email}?").format(email=record.email),
+            gettext("Revoke"),
+        )
```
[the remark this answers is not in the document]
hemant10yadav: Thanks, fixed in 3f15e9e.

commcare_connect/organization/views.py
```diff
@@ -96,13 +108,67 @@ def remove_members(request, org_slug):
...
-        request, message=gettext("Accepted invite for joining {org_slug} workspace.").format(org_slug=org_slug)
+def accept_invite(request, org_slug, token):
+    invite = get_object_or_404(OrganizationInvite, token=token, organization__slug=org_slug)
+
+    if invite.status != OrganizationInvite.Status.INVITED or invite.is_expired:
+        if invite.status == OrganizationInvite.Status.INVITED and invite.is_expired:
+            invite.status = OrganizationInvite.Status.EXPIRED
+            invite.save(update_fields=["status", "date_modified"])
+        messages.error(request, gettext("This invitation has expired. Ask an admin to send you a new one."))
+        return redirect("account_login")
```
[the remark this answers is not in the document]
hemant10yadav: Thanks, fixed in 3f15e9e.

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+    @classmethod
+    def retire_expired(cls, organization, email):
+        """Frees up the unique pending-invite slot so a lapsed invite can be re-sent."""
+        cutoff = timezone.now() - timedelta(days=cls.EXPIRY_DAYS)
+        cls.objects.filter(
+            organization=organization, email=email, status=cls.Status.INVITED, date_created__lt=cutoff
+        ).update(status=cls.Status.EXPIRED, modified_by=email)
+
+    def accept(self, user):
+        membership, _created = UserOrganizationMembership.objects.get_or_create(
```
Charl1996: If a UserOrganizationMembership already exists it seems like this path would allow the already invited user to "accept" an invite for a different role on the same group and in that way change their role.

Granted, the invite will have to be triggered by an admin, but just noting the possibility here in case it's unintended.

commcare_connect/organization/forms.py
```diff
@@ -147,16 +152,30 @@ def __init__(self, *args, **kwargs):
         )
 
     def clean_email(self):
-        email = self.cleaned_data["email"]
-        user = User.objects.filter(email=email).exclude(memberships__organization=self.organization).first()
+        email = self.cleaned_data["email"].strip().lower()
```
Charl1996, line 155: 👍

commcare_connect/organization/models.py
Charl1996: I see 2 commits later we do check if the user already exists as a member of this org, so maybe the above is less of a concern.

commcare_connect/organization/forms.py
```diff
@@ -147,16 +152,30 @@ def __init__(self, *args, **kwargs):
...
+
+        if User.objects.filter(email=email, memberships__organization=self.organization).exists():
+            raise ValidationError(gettext("This person is already a member of this workspace."))
 
-        if not user:
-            raise ValidationError("User with this email does not exist or is already a member")
+        OrganizationInvite.retire_expired(self.organization, email)
```
Charl1996: Nit: Would marking previous invites as "revoked" not be more accurate here?

commcare_connect/organization/tables.py
```diff
@@ -41,3 +42,28 @@ class Meta:
...
+
+    class Meta:
+        model = OrganizationInvite
+        fields = ("email", "role", "date_created")
+        sequence = ("index", "email", "role", "date_created", "revoke")
+
+    def render_role(self, value):
+        return format_html("<div class=' underline underline-offset-4'>{}</div>", value)
+
+    def render_revoke(self, record):
```
Charl1996: Instead of manually constructing the template in python we can use [TemplateColum](https://django-tables2.readthedocs.io/en/latest/pages/api-reference.html#django_tables2.columns.TemplateColumn)

commcare_connect/organization/views.py
```diff
@@ -96,13 +108,67 @@ def remove_members(request, org_slug):
     return redirect(redirect_url)
 
 
-@login_required
-def accept_invite(request, org_slug, invite_id):
-    get_object_or_404(UserOrganizationMembership, invite_id=invite_id, organization__slug=org_slug)
-    messages.success(
-        request, message=gettext("Accepted invite for joining {org_slug} workspace.").format(org_slug=org_slug)
+def accept_invite(request, org_slug, token):
```
Charl1996, line 112: Should we rate limit this endpoint since it's open to the public?

Charl1996, line 112: There's a lot of branching going on in this function. Can we refactor this function a bit?

Charl1996:
Overall looking good. Just 1 or 2 queries.

commcare_connect/organization/models.py
hemant10yadav: Good catch — but `clean_email` already blocks inviting someone who's already a member of this org (`memberships__organization=self.organization` check), so by the time `accept()` runs the invitee shouldn't have an existing membership here under normal flow. Matches what you found in the later commit, so I think we're good.

commcare_connect/organization/forms.py
hemant10yadav: `retire_expired` only touches invites where `date_created__lt=cutoff`, i.e. ones that are genuinely past the 7-day window — so "expired" is accurate here. "Revoked" would imply an admin explicitly pulled the invite, which isn't what's happening, it just aged out on its own.

commcare_connect/organization/tables.py
hemant10yadav: Good call, switched `render_revoke` to a `TemplateColumn` — done in 81e020a5.

commcare_connect/organization/views.py
hemant10yadav: Fair instinct — but the endpoint is also token-gated, same threat model as a password-reset link, which isn't rate-limited here either.

hemant10yadav: Agreed, split it into `_reject_invalid_invite`, `_accept_invite_for_authenticated_user`, `_redirect_existing_user_to_login`, and `_accept_invite_for_new_user` — done in 697145f5.

commcare_connect/organization/forms.py
```diff
@@ -147,16 +152,30 @@ def __init__(self, *args, **kwargs):
...
-        email = self.cleaned_data["email"]
-        user = User.objects.filter(email=email).exclude(memberships__organization=self.organization).first()
+        email = self.cleaned_data["email"].strip().lower()
+
+        if User.objects.filter(email__iexact=email, memberships__organization=self.organization).exists():
+            raise ValidationError(gettext("This person is already a member of this workspace."))
 
-        if not user:
-            raise ValidationError("User with this email does not exist or is already a member")
+        OrganizationInvite.retire_expired(self.organization, email)
```
sravfeyn: Do we not need to say "The invite expired" if it actually expired here?

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+        INVITED = "invited", _("Invited")
+        ACCEPTED = "accepted", _("Accepted")
+        REVOKED = "revoked", _("Revoked")
+        EXPIRED = "expired", _("Expired")
+
+    organization = models.ForeignKey(Organization, on_delete=models.CASCADE, related_name="invites")
+    email = models.EmailField()
+    role = models.CharField(max_length=20, choices=UserOrganizationMembership.Role.choices)
+    status = models.CharField(max_length=20, choices=Status.choices, default=Status.INVITED)
+    token = models.CharField(max_length=64, unique=True, default=_generate_invite_token)
```
sravfeyn: I think we can directly do `default=secrets.token_urlsafe`.

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+                fields=["organization", "email"],
+                condition=models.Q(status="invited"),
+                name="unique_pending_org_invite",
+            )
+        ]
+
+    def __str__(self):
+        return f"Invite for {self.email} to {self.organization}"
+
+    def get_revoke_url(self):
```
sravfeyn: `nit`: I wouldn't put this here. Also it doesn't need to be defined as a method, it's getting used only once?

commcare_connect/organization/models.py
```diff
@@ -80,4 +82,62 @@ def is_program_manager(self):
...
+        return self.status == self.Status.INVITED and self.date_created < timezone.now() - timedelta(
+            days=self.EXPIRY_DAYS
+        )
+
+    @classmethod
+    def retire_expired(cls, organization, email):
+        """Frees up the unique pending-invite slot so a lapsed invite can be re-sent."""
+        cutoff = timezone.now() - timedelta(days=cls.EXPIRY_DAYS)
+        cls.objects.filter(
+            organization=organization, email=email, status=cls.Status.INVITED, date_created__lt=cutoff
```
sravfeyn: I think we could model the re-invite workflow to be more simple by using `OrganizationInvite.objects.update_or_create` and remove the need for tracking/deleting stale ones. That would also simplify the unique constraint on the model, and potentially remove the need for `Expired` status as well. (Reinvites can be inferred via pg_history already).

sravfeyn: Though, if it's a bigger lift, I wouldn't worry about it. Worth considering.

commcare_connect/organization/views.py
sravfeyn: `nit` For future. I would use a CBV with separate methods than separate functions since they are logically all tied together, and wouldn't need to pass around request/invite to each method.

commcare_connect/templates/organization/accept_invite.html
```diff
@@ -0,0 +1,72 @@
+{% extends "account/base.html" %}
+{% load i18n %}
+{% block head_title %}
+  {% translate "Join Workspace" %}
+{% endblock %}
+{% block inner %}
+  <div class="w-full min-h-[620px] h-fit flex flex-col gap-4">
+    <form method="post"
+          class="flex flex-col gap-4"
+          x-data="{ password: '', confirmPassword: '', termsChecked: false, isFormValid() { return this.termsChecked && this.password && this.confirmPassword; } }">
```
sravfeyn, line 10: I think these come from `account/signup.html` and `account/password_reset_from_key.html`, should they be extracted into a reusable template component?

commcare_connect/organization/forms.py
Charl1996: Oh, right. I thought the intent was to "invalidate" the existing invite and generate a new one, but seems not.

commcare_connect/organization/forms.py
hemant10yadav: Traced through this — `retire_expired()` runs before the "already sent" check, so any expired invite gets flipped to `EXPIRED` and excluded from the query first. By the time we'd hit the "already sent" error, the invite is genuinely still active, so the message is accurate. No case where it says "already sent" for an actually-expired one.

commcare_connect/organization/models.py
hemant10yadav: Done in [528723a8](https://github.com/dimagi/commcare-connect/commit/528723a88393cf911b2737f5d7e7075487e713f0).

commcare_connect/organization/models.py
hemant10yadav: Agreed, moved it — the revoke URL is now built inline in the `PendingInviteTable` template column via `{% url %}` instead of a model method. [528723a8](https://github.com/dimagi/commcare-connect/commit/528723a88393cf911b2737f5d7e7075487e713f0).

commcare_connect/templates/organization/accept_invite.html
hemant10yadav: Good catch — extracted a shared `account/password_field.html` partial and reused it across `signup.html`, `password_reset_from_key.html`, and `accept_invite.html`. [690f3737](https://github.com/dimagi/commcare-connect/commit/690f373720dad7a8ef6e4d1cae6dfe3513ca92dc).

commcare_connect/organization/models.py
hemant10yadav: Went ahead and did this after all — replaced `retire_expired()`/`EXPIRED` status with `send_invite()` using `update_or_create`, so there's one row per (org, email) that gets reset in place on re-invite. `is_expired` now checks `date_modified` instead of the immutable `date_created`. [24053843](https://github.com/dimagi/commcare-connect/commit/24053843dc7d1763a2582fbd01158c9306715089).

commcare_connect/templates/account/password_field.html
```diff
@@ -0,0 +1,14 @@
+<div class="input-group {% if field.errors %}error{% endif %}">
+  <div class="simple-input">
+    <input type="password"
+           placeholder="{{ placeholder }}"
+           required
+           id="{{ field.auto_id }}"
+           name="{{ field.html_name }}"
+           {% if persist_value and field.value != None %}value="{{ field.value|stringformat:'s' }}"{% endif %}
+           {% if x_model %}x-model="{{ x_model }}"{% endif %}>
```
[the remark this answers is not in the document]
hemant10yadav: Fixed it! I accidentally deleted your reply earlier.

commcare_connect/organization/forms.py
sravfeyn: I am not able to fully follow but I was referring to whether when a user clicks expired invite should we say "The invite expired", similar to other errors you have added here.

sravfeyn: I see it below in the view. 👍

sravfeyn (approved):
LGTM. Thanks!

Charl1996 (approved):

PR #1373, opened 2026-07-13
Validate Payment Unit Budget

zandre-eng:
## Product Description

Payment Units draw each worker's delivery limit from a shared opportunity budget. Previously, an admin could add or enlarge a Payment Unit (or create Payment Units via the automation API) in a way that left the budget unable to fund the full limit for every worker who had *already* claimed — silently reducing some workers' limits. This change validates that the opportunity's total budget can still cover all existing claimants at the full limit, and blocks the change otherwise.

- **Automation API:** rejects the request with `400` and the specific message, e.g. _"The opportunity budget cannot give the full limit to all 3 workers who have already claimed. Increase it to at least 3000 USD."_
- **Web form:** the invalid submission is blocked. Note: until the [companion form re-render PR](https://github.com/dimagi/commcare-connect/pull/1380) merges, the web form surfaces this as a generic "Invalid Data" message on the Payment Units list page rather than the specific message inline (see Technical Summary).

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-741)

The rule lives in one place — `Opportunity.validate_budget_for_payment_units`, which takes the complete effective set of units and raises `ValidationError` — so the web form and the DRF serializer share identical logic and message. Each caller only differs in how it assembles that set: the form uses the proposed/edited unit plus the other existing units, while the batch API uses all proposed units plus all existing ones. The method raises Django's core `ValidationError` rather than returning a message so both entry points work unchanged: the form's `clean()` surfaces it, and DRF's `run_validation` converts it to a `400` (verified to produce output identical to `serializers.ValidationError` for a string message).

One sequencing consequence: the specific, user-facing message on the **web form** depends on the [companion PR](https://github.com/dimagi/commcare-connect/pull/1380) that re-renders the form on error instead of redirecting. Until that merges, the web form still redirects to the list with a generic "Invalid Data" toast; the automation API returns the specific message regardless, since it doesn't go through that view.

### Safety story

The change is fail-closed validation at the write boundary — it only blocks invalid saves and adds no data migrations. It early-returns when the opportunity has no budget or no existing claimants, so opportunities that don't meet those preconditions are entirely unaffected. Fully reversible by reverting the branch.

### Automated test coverage

Tests cover the model rule through both entry points: the form path (reject an over-budget add, accept one within budget, skip when there are no claimants, and reject enlarging an existing unit beyond budget) and the automation API (reject when the budget can't cover existing claimants with nothing persisted, and accept when it can).

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/forms.py
+++ b/commcare_connect/opportunity/forms.py
@@ -1378,33 +1378,15 @@ class PaymentUnitForm(forms.ModelForm):
         """
         Reject the create/edit if the budget can't give every existing claimant the full limit.
         """
-        opportunity = self.opportunity
-        claimant_count = OpportunityClaim.objects.filter(opportunity_access__opportunity=opportunity).count()
-        if not claimant_count or not opportunity.total_budget:
-            return
-
         max_total = cleaned_data.get("max_total")
         amount = cleaned_data.get("amount")
         if max_total is None or amount is None:
             return
-        org_amount = cleaned_data.get("org_amount") or 0
 
-        other_units = opportunity.paymentunit_set.exclude(pk=self.instance.pk)
-        budget_per_user = max_total * (amount + org_amount) + Opportunity.budget_per_user_for_units(other_units)
+        proposed_unit = PaymentUnit(max_total=max_total, amount=amount, org_amount=cleaned_data.get("org_amount") or 0)
+        other_units = self.opportunity.paymentunit_set.exclude(pk=self.instance.pk)
 
-        required_budget = budget_per_user * claimant_count
-        if opportunity.total_budget < required_budget:
-            raise ValidationError(
-                _(
-                    "The opportunity budget cannot give the full limit to all %(claimants)d workers "
-                    "who have already claimed. Increase it to at least %(required)d %(currency)s before saving."
-                )
-                % {
-                    "claimants": claimant_count,
-                    "required": required_budget,
-                    "currency": opportunity.currency_code or "",
-                }
-            )
+        self.opportunity.validate_budget_for_payment_units([proposed_unit, *other_units])
 
 
 class SendMessageMobileUsersForm(forms.Form):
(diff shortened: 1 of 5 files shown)


ajeety4 (approved):
The changes looks good. 

(Out of the scope of this PR)  I noticed we redirect to the list page even in case of errros instead of returning the form, and wondering what could be the reason for it.
We lose the form state, and user need to fill out the form again.  It would be nice to check and fix that too if possible.

commcare_connect/opportunity/forms.py
```diff
@@ -1370,8 +1370,42 @@ def clean(self):
...
+        other_units = opportunity.paymentunit_set.exclude(pk=self.instance.pk)
+        budget_per_user = max_total * (amount + org_amount) + Opportunity.budget_per_user_for_units(other_units)
+
+        required_budget = budget_per_user * claimant_count
+        if opportunity.total_budget < required_budget:
+            raise ValidationError(
+                _(
+                    "The opportunity budget cannot give the full limit to all %(claimants)d workers "
+                    "who have already claimed. Increase it to at least %(required)d %(currency)s before saving."
+                )
```
hemant10yadav: 👍

hemant10yadav (commented):
We should implement the same validation in the automation APIs as well.

zandre-eng:
> We should implement the same validation in the automation APIs as well.

@hemant10yadav Good catch, I've added the same validation for the automation APIs (https://github.com/dimagi/commcare-connect/pull/1373/commits/120429e1b951c05e53d84a219325cbe1846b6014).

zandre-eng:
> (Out of the scope of this PR) I noticed we redirect to the list page even in case of errros instead of returning the form, and wondering what could be the reason for it.
We lose the form state, and user need to fill out the form again. It would be nice to check and fix that too if possible.

@ajeety4 I'll create a follow-up PR to address keeping the user in the form.

ajeety4 (approved):

zandre-eng:
I did a rebase to remove cleanly showing the errors on the payment unit list page. [This follow-up PR](https://github.com/dimagi/commcare-connect/pull/1380) handles showing these errors rather in the payment unit form.

ajeety4 (approved):

hemant10yadav (approved):
LGTM

PR #1378, opened 2026-07-15
CCCT-2528: Track when a visit's review status was last agreed/changed

ajeety4:
## Product Description

No direct UI changes. This adds tracking the datetime when the PM review status changes for a user visit, similar to how we track status today.

## Technical Summary

[CCCT-2528](https://dimagi.atlassian.net/browse/CCCT-2528)

1. Adds `review_status_modified_date` to `UserVisit`, set whenever `review_status` changes. Since `review_status=agree` is a terminal state (_both write paths lock a visit out of further review-status changes once agreed_), this single generic field doubles as "when was this agreed" while also capturing "when was this disagreed" for free, with no extra migration needed if that's useful later.
It is also consistent with how we track the datetime for for visit's `status` changes  today.

2. While implementing this, found and fixed two related pre-existing bugs on the sibling `status_modified_date` field (unrelated to the new feature):
- `approve_visits` and `reject_visits` never persisted `status_modified_date` on their bulk update paths (the value was computed in memory via the model's `__setattr__` hook, but silently dropped since the field wasn't listed in the `bulk_update`/`.update()` call).
- Both `UserVisit` and `CompletedWork`'s custom `__init__` set `self.status_modified_date = now()`, but Django's `Model.__init__` (given kwargs, the normal case) resets every field not present in those kwargs back to its default afterward   which was `None`, since neither field had an explicit `default=`. This silently cleared the timestamp back to `None` right after construction for any object that never has its `status` reassigned away from the default (e.g. a flagged visit that just sits at `pending`). Fixed with `default=now` on both fields.

> Note: `review_status_modified_date` is set explicitly at each write site rather than via a `__setattr__/save()` hook (like `status_modified_date` uses), because writes also go through bulk_update()/queryset.update(), which bypass hooks entirely and the field still needs to be listed explicitly in fields=[...] for a hook-computed value to actually persist. We hit exactly this gap once already: approve_visits's bulk_update() silently dropped `status_modified_date` despite the hook computing it correctly in memory. Explicit-set is no less safe and simpler to reason about.

### Safety story

All changes are additive/forward-looking - no backfill of historical data.
Tested on staging approve flow for NM and PM including import

### Automated test coverage

- New tests directly reproducing each of the three bugs (RED before the fix, GREEN after), per the branch's commit history.
- `status_modified_date`/`review_status_modified_date` assertions added to every existing test that changes `UserVisit.status` or `.review_status` (receiver integration tests, bulk CSV import tests, view tests), so a future regression in either timestamp is caught rather than silently reintroduced.

### QA Plan
None

- PM agree/disagree on a visit (`user_visit_review`) → `review_status_modified_date` updates.
- NM bulk-approve a previously-disagreed visit → both `status_modified_date` and `review_status_modified_date` update.
- NM bulk-reject a visit → `status_modified_date` updates.
- CSV bulk review-status import → `review_status_modified_date` updates for changed rows only, already-agreed (locked) rows untouched.
- Confirm on deploy that existing rows' new `review_status_modified_date` column comes up `NULL`, not backfilled.

### Labels & Review

[CCCT-2528]: https://dimagi.atlassian.net/browse/CCCT-2528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/migrations/0139_uservisit_status_modified_date_and_review_status.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 4.2.30 on 2026-07-15 05:47
+
+from django.db import migrations, models
+import django.utils.timezone
+
+
+class Migration(migrations.Migration):
```
zandre-eng, line 7: Nit: Could this migration be merged with 0139?

zandre-eng (approved):
Just one nit, but otherwise looks good.

commcare_connect/opportunity/migrations/0139_uservisit_status_modified_date_and_review_status.py
ajeety4: Addressed in 249fe266d4a33dd6b37cf0983a304b27f02ecd06

Charl1996 (approved):
Thanks!

PR #1380, opened 2026-07-15
Re-render payment unit form on error instead of redirecting

zandre-eng:
## Product Description

When creating a Payment Unit, submitting the form with invalid input (e.g. an end date earlier than the start date, or a budget too small to cover existing claimants) previously bounced the user back to the Payment Units list page with a generic "Invalid Data" message, discarding everything they had typed. The form now stays on screen with all entered values preserved and the specific validation errors shown inline next to the relevant fields.

**Old:**
Screenshot_20260713_160324

**New:**
New behaviour

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-741)

The `add_payment_unit` view redirected on an invalid POST, which is what threw away form state. The fix simply lets an invalid submission fall through to the existing `render(...)` call — the same pattern `edit_payment_unit` already uses — so Django/crispy re-render the bound form with inline errors. This also removed the now-redundant toast-error helper.

### Safety story

Low risk and easily reversible: the change only alters the error branch of a single view (the success path and all persistence logic are untouched), and it brings `add_payment_unit` in line with the behavior `edit_payment_unit` has always had. No data model or migration changes.

### Automated test coverage

Tests cover the invalid-submission path: the form re-renders with a 200 (rather than redirecting), no Payment Unit is created, the entered values are preserved in the response, and the validation error is shown inline. The existing valid-submission test confirms creation still works.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/views.py
+++ b/commcare_connect/opportunity/views.py
@@ -853,9 +853,15 @@ def add_payment_unit(request, org_slug=None, opp_id=None):
             "opportunity:add_payment_units", org_slug=request.org.slug, opp_id=request.opportunity.opportunity_id
         )
     elif request.POST:
-        messages.error(request, "Invalid Data")
-        return redirect(
-            "opportunity:add_payment_units", org_slug=request.org.slug, opp_id=request.opportunity.opportunity_id
+        return render(
+            request,
+            "opportunity/add_payment_units.html",
+            dict(
+                opportunity=request.opportunity,
+                paymentunit_count=PaymentUnit.objects.filter(opportunity=request.opportunity).count(),
+                form=form,
+                form_title="Payment Unit Create",
+            ),
         )
 
     path = [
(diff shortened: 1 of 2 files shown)


ajeety4 (approved):
Thanks for the change.

commcare_connect/opportunity/views.py
```diff
@@ -853,9 +853,15 @@ def add_payment_unit(request, org_slug=None, opp_id=None):
...
-        return redirect(
-            "opportunity:add_payment_units", org_slug=request.org.slug, opp_id=request.opportunity.opportunity_id
+        return render(
+            request,
+            "opportunity/add_payment_units.html",
+            dict(
+                opportunity=request.opportunity,
+                paymentunit_count=PaymentUnit.objects.filter(opportunity=request.opportunity).count(),
+                form=form,
+                form_title="Payment Unit Create",
```
hemant10yadav: nit: Missing translation

hemant10yadav (approved):

commcare_connect/opportunity/views.py
zandre-eng: Translations added in [624e42b](https://github.com/dimagi/commcare-connect/pull/1380/commits/624e42b994fb0f4b7273b5dc66a089bc0b72d693).

hemant10yadav (approved):
Thanks!

PR #1382, opened 2026-07-16
Remove case sensitivity from Work Area upload csv file

hemant10yadav:
## Product Description

Work Area CSV uploads were rejected if a column header's casing didn't exactly match the expected name (for example "Target population" instead of "Target Population"), even though the data itself was fine. Header casing no longer matters.

## Technical Summary

[CCCT-2592](https://dimagi.atlassian.net/browse/CCCT-2592)

- Uploaded headers are normalized to the canonical casing (case-insensitive lookup) before columns are validated and rows are read, in both the validation pass and the streaming insert pass.
- The downloadable template still uses canonical casing, so nothing changes there.

### Safety story

The change only touches header matching, not row data, so unrecognized or missing headers still fail validation the same way as before. Tested locally with mixed, upper, and lower-case header variants.

### Automated test coverage

Added `test_case_insensitive_headers` covering mixed/lower/upper-case header variants; the full `test_tasks.py` suite (14 tests) passes.

### QA Plan

### Labels & Review

[CCCT-2592]: https://dimagi.atlassian.net/browse/CCCT-2592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -53,6 +53,7 @@ class WorkAreaCSVImporter:
     def _validate_all_rows(self, f):
         f.seek(0)
         reader = csv.DictReader(f)
+        self._normalize_headers(reader)
         if not self._validate_headers(reader):
             return False
 
@@ -66,6 +67,7 @@ class WorkAreaCSVImporter:
     def _stream_and_insert(self, f):
         f.seek(0)
         reader = csv.DictReader(f)
+        self._normalize_headers(reader)
         batch = []
         batch_size = 5000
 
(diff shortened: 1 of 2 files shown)


commcare_connect/microplanning/tasks.py
```diff
@@ -113,6 +115,10 @@ def _result(self):
             return {"errors": self.errors}
         return {"created": self.created_count}
 
+    def _normalize_headers(self, reader):
+        canonical_by_lower = {header.lower(): header for header in self.HEADERS.values()}
+        reader.fieldnames = [canonical_by_lower.get((h or "").lower(), h) for h in (reader.fieldnames or [])]
```
zandre-eng, line 120: Minor nit: I think the `(h or "")` guard can be dropped as I believe fieldnames are always returned as `str` and never `None`.

zandre-eng (approved):
Looks good, just a minor nit.

ajeety4 (approved):

commcare_connect/microplanning/tasks.py
hemant10yadav: Just an explicit safety guard that I added. I noticed some strange behaviour on different devices, so I'm thinking of keeping it.

PR #1384, opened 2026-07-16
Reject delivery visits while a task is pending

zandre-eng:
## Product Description

Field workers can no longer have delivery visits silently approved while they still have an incomplete assigned task. The mobile app already blocks delivery with "Complete assigned tasks to continue delivering services," but until now the backend accepted and auto-approved any visit that reached it anyway (e.g. via an app bug or out-of-order form sync). Such a visit is now rejected and flagged as **Pending Task**, and its associated completed work is rejected so it is not paid or invoiced.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-760)

### Safety story

The change only tightens an existing decision path — it can move a visit that would otherwise be `approved`/`pending` to `rejected`, and never the reverse — so it cannot loosen validation or over-approve. It affects only visits submitted while a same-app task is `ASSIGNED`, a state the mobile app already prevents, so the expected production blast radius is small. No data migration is involved; the new `pending_task` flag is stored in the existing `flag_reason` JSON field.

### Automated test coverage

Tests cover the delivery gate end to end through the form receiver: a same-app pending task rejects and flags the visit (with the completed work rejected) despite `auto_approve_visits` being on, while a pending task for an unrelated app leaves delivery approved. Existing delivery, auto-approve, flagging, and suspension tests continue to pass unchanged.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -530,7 +530,8 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         if access.suspended:
             flags.append(["user_suspended", "This user is suspended from the opportunity."])
             user_visit.status = VisitValidationStatus.rejected
-            completed_work.status = CompletedWorkStatus.rejected
+            if completed_work is not None:
+                completed_work.status = CompletedWorkStatus.rejected
         if _has_blocking_pending_task(access, app):
             flags.append(["pending_task", "Worker has an incomplete assigned task."])
             user_visit.status = VisitValidationStatus.rejected
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -534,8 +534,9 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         if _has_blocking_pending_task(access, app):
             flags.append(["pending_task", "Worker has an incomplete assigned task."])
             user_visit.status = VisitValidationStatus.rejected
-            completed_work.status = CompletedWorkStatus.rejected
-            completed_work_needs_save = True
+            if completed_work is not None:
+                completed_work.status = CompletedWorkStatus.rejected
+                completed_work_needs_save = True
         if flags:
             user_visit.flagged = True
             user_visit.flag_reason = {"flags": flags}
--- a/commcare_connect/form_receiver/processor.py
+++ b/commcare_connect/form_receiver/processor.py
@@ -592,6 +592,8 @@ def _has_blocking_pending_task(access: OpportunityAccess, app: CommCareApp) -> b
         opportunity_access=access,
         status=AssignedTaskStatus.ASSIGNED,
         task_type__app=app,
+        task_type__archived__isnull=True,
+        task_type__is_active=True,
     ).exists()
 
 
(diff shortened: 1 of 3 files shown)


commcare_connect/form_receiver/processor.py
```diff
@@ -580,6 +586,14 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
     transaction.on_commit(partial(download_user_visit_attachments.delay, user_visit.id))
 
 
+def _has_blocking_pending_task(access: OpportunityAccess, app: CommCareApp) -> bool:
+    return AssignedTask.objects.filter(
+        opportunity_access=access,
+        status=AssignedTaskStatus.ASSIGNED,
+        task_type__app=app,
+    ).exists()
```
hemant10yadav, line 598: Should we also check that the task type is active and not archived?

commcare_connect/form_receiver/processor.py
```diff
@@ -530,6 +531,11 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
             flags.append(["user_suspended", "This user is suspended from the opportunity."])
             user_visit.status = VisitValidationStatus.rejected
             completed_work.status = CompletedWorkStatus.rejected
+        if _has_blocking_pending_task(access, app):
+            flags.append(["pending_task", "Worker has an incomplete assigned task."])
+            user_visit.status = VisitValidationStatus.rejected
+            completed_work.status = CompletedWorkStatus.rejected
+            completed_work_needs_save = True
```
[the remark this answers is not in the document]
zandre-eng: Addressed in [7a2fe65](https://github.com/dimagi/commcare-connect/pull/1384/commits/7a2fe65fbf22aa7a809edf29c14fa238f8d00fcc).

commcare_connect/form_receiver/processor.py
zandre-eng: That sounds right. Addressed in [0675f54](https://github.com/dimagi/commcare-connect/pull/1384/commits/0675f54f4daab0957d24f902e9a77b669e951833).

hemant10yadav (approved):

commcare_connect/form_receiver/processor.py
```diff
@@ -534,8 +534,9 @@ def process_deliver_unit(user, xform: XForm, app: CommCareApp, opportunity: Oppo
         if _has_blocking_pending_task(access, app):
             flags.append(["pending_task", "Worker has an incomplete assigned task."])
             user_visit.status = VisitValidationStatus.rejected
-            completed_work.status = CompletedWorkStatus.rejected
-            completed_work_needs_save = True
+            if completed_work is not None:
```
ajeety4, line 538: :+1: 
Wondering if this check should have been added for above if block (`if access.suspended`) as well ?

ajeety4 (approved):

commcare_connect/form_receiver/processor.py
zandre-eng: Good point, given that `CompletedWork` can be done this check would be good to add there too.

Addressed in [6e28284](https://github.com/dimagi/commcare-connect/pull/1384/commits/6e282846b447fc43b693c8ee39d50e384f1a1fba).

ajeety4 (approved):
Thanks for the additional change

PR #1385, opened 2026-07-16
Fall back to email when Assigned By user has a blank name

hemant10yadav:
## Product Description
Screenshot 2026-07-16 at 5 07 52 PM
Screenshot 2026-07-16 at 5 07 36 PM

## Technical Summary

[CCCT-2570](https://dimagi.atlassian.net/browse/CCCT-2570)

- The column read the user's name field directly, so a blank name showed nothing instead of falling back to another identifier. This is the same bug already fixed for the audit report's Reviewer column.
- Sorting now runs through a Coalesce expression evaluated at query time, so ordering by this column still works when some names are blank.
- Extracted the shared columns and this render/sort logic into a `BaseAssignedTaskTable` mixin, so `WorkerCompletedTaskTable` gets the same fix instead of duplicating it.

### Safety story

This only changes how existing task table columns render and sort. No models, migrations, or view querysets change.

### Automated test coverage

No new tests added.

### QA Plan

No QA ticket needed

### Labels & Review

[CCCT-2570]: https://dimagi.atlassian.net/browse/CCCT-2570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/tables.py
+++ b/commcare_connect/opportunity/tables.py
@@ -3,6 +3,8 @@ from urllib.parse import urlencode
 
 import django_tables2 as tables
 from django.contrib.humanize.templatetags.humanize import intcomma
+from django.db.models import CharField, Value
+from django.db.models.functions import Coalesce, NullIf
 from django.template.loader import render_to_string
 from django.urls import reverse
 from django.utils.html import format_html
@@ -1849,7 +1851,7 @@ class AssignedTaskListTable(OpportunityContextTable):
     due_date = DMYTColumn(verbose_name=gettext_lazy("Due Date"), accessor="due_date")
     assigned_by = tables.Column(
         verbose_name=gettext_lazy("Assigned By"),
-        accessor="assigned_by__name",
+        accessor="assigned_by",
         empty_values=(None,),
         default=gettext_lazy("Deleted user"),
     )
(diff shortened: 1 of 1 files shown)


commcare_connect/opportunity/tables.py
```diff
@@ -1840,32 +1842,36 @@ def value_total_amount_usd(self, record):
...
     assigned_by = tables.Column(
         verbose_name=gettext_lazy("Assigned By"),
-        accessor="assigned_by__name",
+        accessor="assigned_by",
         empty_values=(None,),
         default=gettext_lazy("Deleted user"),
     )
+
+    def render_assigned_by(self, value):
+        return value.name or value.email
```
zandre-eng, line 1858: There are a lot of users that also don't have an `email` value. Should we add `username` as the last resort fallback here? Since all users should have a username.

zandre-eng (approved):
Looks good. Just one question about using emails as a fallback.

commcare_connect/opportunity/tables.py
hemant10yadav: I believe that this can only happen in the case of a mobile worker, and I don't think we expect mobile workers to use our web platform.

ajeety4 (approved):

PR #1387, opened 2026-07-17
Work Area Group centroid

pxwxnvermx:
## Product Description
This PR adds a new centroid field on the Work Area Group. This is calculated from a full boundary around all the work areas inside a work area group, using the centroid function from the shapely library.

These actions lead to a centroid update for Work Area Group

- Work Area edit.
- Work Area getting marked as Inaccessible
- Work Area getting marked as Excluded

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2645)

### Safety story
- Tested Locally.

### Automated test coverage
- Added Automated Tests and Updated tests for changed views and helpers.

### QA Plan
- Will link once ticket is created.

### Labels & Review


commcare_connect/microplanning/migrations/0016_workareagroup_centroid.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.30 on 2026-07-17 05:21
+
+import django.contrib.gis.db.models.fields
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
zandre-eng, line 7: Do we need to do any backfilling for existing groups (i.e. calculate and sync their new centroid to HQ)?

commcare_connect/microplanning/tests/test_helpers.py
```diff
@@ -215,6 +221,119 @@ def test_one_failed_chunk_does_not_block_other_chunks(self, mock_bulk_hq, org_us
...
+
+        # The excluded area is dropped from the calculation, so the centroid is now wa_keep's alone.
+        group.refresh_from_db()
+        assert group.centroid != old_centroid
+        assert group.centroid.x == 79.5
+        assert group.centroid.y == 30.5
+        assert wa_keep.work_area_group_id == group.id  # kept area still belongs to the group
+
+    @patch("commcare_connect.microplanning.helpers.bulk_create_or_update_cases")
+    def test_excluding_all_areas_sets_group_centroid_to_none(self, mock_bulk_hq, org_user_admin, opportunity):
```
zandre-eng, line 263: Nit: I think this test can be removed entirely as `test_exclusion_recomputes_each_affected_group_independently` is testing something very similar.

commcare_connect/microplanning/tests/test_views.py
```diff
@@ -347,6 +352,66 @@ def test_expected_visit_count_change_reevaluates_status(
...
+
+        assert group_1.centroid.x == 77.5
+        assert group_1.centroid.y == 28.5
+        group_1.refresh_from_db()
+        assert group_1.centroid is None
+
+        assert group_2.centroid is None
+        group_2.refresh_from_db()
+        assert group_2.centroid.x == 77.5
+        assert group_2.centroid.y == 28.5
```
zandre-eng, line 424: Minor nit: DRY
```
for group in [group_1, group_2]:
    assert group.centroid.x == 77.5
    assert group.centroid.y == 28.5
    group.refresh_from_db()
    assert group.centroid is None
```

commcare_connect/microplanning/serializers.py
```diff
@@ -29,6 +29,11 @@ def get_properties(self, obj: WorkArea) -> dict:
         if obj.boundary:
             lat_lon_strings = [_coords_to_lat_lon_string(coords) for coords in list(obj.boundary.shell.coords)]
             bounding_box_lat_lon = " ".join(lat_lon_strings)
+
+        wag_centroid = ""
+        if obj.work_area_group and obj.work_area_group.centroid:
```
zandre-eng, line 34: if `obj.work_area_group.centroid` does not exist here should we try and calculate it using `update_centroid`?

zandre-eng (approved):
Overall looks good. Just a few questions and some nits from my side.

commcare_connect/microplanning/tests/test_views.py
pxwxnvermx: I think this logic would not work. The above is testing Work Area assignment from Group 1 to Group 2. Which leaves Group 1 with no WA's and a None centroid. This test is confirming that the update_centroid runs for both Group 1 and Group 2, and after update_centroid one group has a centroid and other does not.

zandre-eng: That's a fair point, thanks for explaining your reasoning.

commcare_connect/microplanning/migrations/0016_workareagroup_centroid.py
pxwxnvermx: Yeah, we might need to backfill the centroid for existing groups and sync them to HQ. I was thinking of running a simple django-shell script to do an `update_centroid` for all WAG's with `centroid` set to Null, and then update the Work Area Cases on HQ as well.

sravfeyn:
Is there a reason to not keep this as a dynamic property calculated as needed than save it on model? It could be cached if performance is concern within sessions. Keeping that property in sync with actions across the product is going to be challenging and could result in inconsistency bugs easily.

pxwxnvermx:
> Is there a reason to not keep this as a dynamic property calculated as needed than save it on model? It could be cached if performance is concern within sessions. Keeping that property in sync with actions across the product is going to be challenging and could result in inconsistency bugs easily.

The only concern would be the performance implications whenever the Work Area's are serialized. We can cache it but we would also need to invalidate the cache every time one of the above action is performed on the Work Areas, to keep the centroid up to date.

@sravfeyn

commcare_connect/microplanning/serializers.py
pxwxnvermx: I think with the backfill this issue should not be a problem.

zandre-eng: Agree it's more of an edge case (e.g. the group gets created programatically via a script). Fine to skip, but just thought I'd flag in case.

sravfeyn (approved):

PR #1392, opened 2026-07-20
Permission decorator refector cleanup 

sravfeyn:
## Product Description

Ticket https://dimagi.atlassian.net/browse/CCCT-2502

No product changes, only cleanup refactor related to permission decorators.

## Technical Summary

- Removes unused methods
- Rename permission decorators for consistency

### Automated test coverage

Exists

### QA Plan

Smoke test TBD
### Labels & Review

### DIFF
--- a/commcare_connect/organization/decorators.py
+++ b/commcare_connect/organization/decorators.py
@@ -57,7 +57,7 @@ def _request_user_is_admin(request):
     ) or request.user.has_perm(ALL_ORG_ACCESS)
 
 
-def request_user_is_program_manager(request):
+def is_org_pm_or_all_access(request):
     return (
         request.org and request.org_membership and request.org_membership.is_admin and request.org.program_manager
     ) or request.user.has_perm(ALL_ORG_ACCESS)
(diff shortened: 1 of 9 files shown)


commcare_connect/organization/decorators.py
```diff
@@ -57,7 +57,7 @@ def _request_user_is_admin(request):
     ) or request.user.has_perm(ALL_ORG_ACCESS)
 
 
-def request_user_is_program_manager(request):
+def is_org_pm_or_all_access(request):
```
mkangia, line 60: Looks like we need more to this naming since its also checking for admin?

commcare_connect/program/utils.py
```diff
@@ -10,17 +10,15 @@ def get_managed_opp(opp_id) -> Opportunity | None:
     return queryset.filter(opportunity_id=opp_id).first()
 
 
-def is_program_manager(request):
+def is_org_pm(request):
```
mkangia, line 13: does this naming change also mention superuser?

commcare_connect/program/utils.py
```diff
@@ -10,17 +10,15 @@ def get_managed_opp(opp_id) -> Opportunity | None:
...
     return request.org.program_manager and (
         (request.org_membership != None and request.org_membership.is_admin) or request.user.is_superuser  # noqa: E711
     )
 
 
-def is_program_manager_of_opportunity(request, opp_id) -> bool:
+def is_opportunity_pm(request, opp_id) -> bool:
```
mkangia, line 19: is this specific for managed opps? that naming "managed" should be redundant now but I believe its outside the scope here.

though we can name this as "is_managed_opp_pm"?

mkangia (approved):
Mostly non blocking since this is better than what it was but left a couple of naming suggestions.

zandre-eng (approved):

commcare_connect/program/utils.py
sravfeyn: Yeah, we don't have the concept of managed opps anymore, so this is for all opps.

commcare_connect/organization/decorators.py
sravfeyn: Yeah, it could become more verbose. Let me know if you have any strong suggestion.

PR #1393, opened 2026-07-21
Work Area Assignment + HQ Sync Chunking

zandre-eng:
## Product Description

No visible UI changes. This fixes a data-integrity bug in work-area (re)assignment: previously, when a program manager assigned a large batch of work areas and the sync to CommCare HQ failed partway through, Connect would silently revert *every* assignment in the batch while HQ kept the ones that had already gone through — leaving the two systems disagreeing about who owns each area (and causing completed visits to show as "Not Visited"). Now a partial failure keeps the assignments that succeeded on both systems, and the program manager gets an error reporting how many areas failed so they can safely retry.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-797)

`save_assignment` performed a dual write (Connect DB + HQ) inside a single `ATOMIC_REQUESTS` transaction, pushing owners to HQ in 100-case chunks. Because the HQ POSTs are non-transactional external side effects, a later chunk failing triggered `set_rollback(True)`, which rolled back the whole Connect transaction but could not un-commit the HQ chunks that already succeeded, causing a divergence. The fix extracts `assign_work_areas_and_sync_to_hq`, which commits each chunk in its own savepoint (DB `bulk_update` + a single HQ POST) and drops `set_rollback` so successful chunks persist. The chunk size is deliberately pinned at or below HQ's bulk-POST size so one savepoint maps to exactly one HQ POST — that 1:1 alignment is what guarantees a chunk either fully commits to both systems or fully rolls back from both. On failure the view returns 502 with the failed count; retries are safe because committed chunks re-UPSERT as no-ops.

### Safety story

The change is confined to the work-area assignment path and mirrors the existing, already-proven per-chunk pattern used by the unassign flow. It is strictly safer than the prior behavior: instead of an all-or-nothing rollback that could diverge from HQ, each chunk is now atomic against its single HQ POST, and failures are logged (previously the failure reached neither Sentry nor the app logs). No migrations and no data backfill; existing assignments are untouched.

### Automated test coverage

Tests cover the successful sync path, a total HQ failure (assignment rolled back, 502 returned), and — the key regression — a partial failure where one chunk commits and another fails, asserting the committed chunk persists, the failed chunk rolls back, and only the committed assignee is notified. Existing assignment and notification tests were re-pointed to the new call site and continue to pass.

### QA Plan

### Labels & Review


ajeety4 (approved):

commcare_connect/microplanning/tests/test_views.py
```diff
@@ -1162,6 +1162,56 @@ def test_hq_failure_rolls_back_db(
         wa.refresh_from_db()
         assert wa.opportunity_access_id is None
 
+    @patch("commcare_connect.microplanning.helpers.HQ_ASSIGN_BULK_CHUNK_SIZE", 1)
+    @patch("commcare_connect.microplanning.helpers.bulk_create_or_update_cases_by_work_areas")
+    def test_partial_hq_failure_commits_good_chunks_only(
```
sravfeyn: `nit`: A unit test on `assign_work_areas_and_sync_to_hq` would suffice than testing the full view and would avoid mocking.

sravfeyn (approved):

commcare_connect/microplanning/tests/test_views.py
zandre-eng: I have refactored the tests to rather do a unit test on `assign_work_areas_and_sync_to_hq`, however I still needed to add mocking since a call to HQ is made within this function.

Addressed in [7a60cce](https://github.com/dimagi/commcare-connect/pull/1393/commits/7a60cce8fa02b1a89d1415219a7847e59985712e).

hemant10yadav (approved):

PR #1394, opened 2026-07-21
Script to reconcile work-area owners from HQ (No Merge)

zandre-eng:
## Product Description

No user-facing UI changes. This adds a one-off operational script used to repair work areas whose Connect assignee had drifted out of sync with their CommCare HQ case owner. After it runs, the affected areas show the correct assignee and their already-recorded visits are counted again (areas flip from "Not Visited" to Visited / Expected Visit Reached in the coverage views), matching what the field worker actually did on HQ.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CI-797)

A partially-applied assignment sync left HQ holding the new case owner while Connect rolled its assignee change back, so `WorkArea.update_status` (which counts visits by the work area's *current* `opportunity_access`) stopped counting the true owner's visits. The script treats **HQ as the source of truth** and only aligns Connect to it — it deliberately does **not** push to HQ. It joins HQ cases to work areas by `case_id` (the same stable key the form receiver uses) rather than external_id, resolves each HQ `owner_id` to a Connect `OpportunityAccess` via `ConnectIDUserLink`, and fetches all cases in a single request against HQ's max page size to keep it to one round-trip. Scope is limited to a pure re-point of already-assigned areas; currently-unassigned areas and terminal states (EXCLUDED/INACCESSIBLE/REQUEST_FOR_INACCESSIBLE) are left untouched, since assigning from scratch is a separate operation.

**Please note**, this PR exists for review **only** and is not intended to be merged.

### Safety story

The script is dry-run by default (`APPLY = False`) and prints a `old -> new: count` summary for review before anything is written; it is Connect-side only (no writes to HQ), runs its writes in a single `transaction.atomic()` under a `pghistory.context` for an audit trail, and is idempotent — already-aligned areas are no-ops, so it is safe to re-run. `update_status` only ever advances status on visit counts, so it cannot downgrade an area. It was validated end-to-end against a live local CommCare HQ and a real PostGIS database: seeding HQ cases, running the dry-run, applying, and confirming the assignee re-point, the status recompute, idempotency on re-run, and that unassigned areas were left alone.

### Automated test coverage

No automated tests — this is a one-off remediation script run manually via the Django shell, not application code on a request path. Confidence comes from the end-to-end validation described above against live HQ and a real database.

### QA Plan

### Labels & Review

### DIFF
--- a/scripts/reconcile_work_area_owners_from_hq.py
+++ b/scripts/reconcile_work_area_owners_from_hq.py
@@ -1,0 +1,105 @@
+"""Reconcile Connect ``WorkArea.opportunity_access`` with CommCare HQ case ``owner_id``.
+
+One-off remediation for divergence caused by a partially-applied assignment sync:
+``save_assignment`` writes the new assignee to Connect, then pushes owner changes to HQ in
+100-case chunks. If a later chunk fails, the whole Connect transaction is rolled back
+(``set_rollback``) while the HQ chunks that already committed stay put -- so HQ shows the new
+owner and Connect reverts to the old assignee, with no audit trail. Visits by the true (HQ)
+owner are then linked but never counted, because ``WorkArea.update_status`` only counts
+visits belonging to the work area's *current* ``opportunity_access``.
+
+This script treats **HQ as the source of truth**, aligns Connect's ``opportunity_access`` to
+match, and recomputes work-area status so already-recorded visits are counted (areas flip to
+VISITED / EXPECTED_VISIT_REACHED). Connect-side only -- it does **not** push to HQ (HQ is
+already correct). Idempotent, so it's safe to re-run.
+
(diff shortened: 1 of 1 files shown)


Charl1996:
> A partially-applied assignment sync

Do we know more about this like what caused it to be "partial"?

zandre-eng:
> Do we know more about this like what caused it to be "partial"?

@Charl1996 I'll use the below as an example:

Let's say we have 100 work areas split across batch A and B. The code will make the DB updates in Connect for reassignment and then push the new owner change to HQ for batch A. Then, for batch B, the DB update goes fine but a snag is hit in the HQ request and it fails. With the code currently, this will cause a rollback in the Connect DB for both batch A and B, but the HQ changes for A have already been made and are persisted. This ultimately causes a partial update where HQ holds some of the new changes, and Connect still has all the old owners.

PR #1395, opened 2026-07-22
Add optional Work Area Group Name column to CSV import

hemant10yadav:
## Product Description

Work Area CSV uploads now accept an optional "Work Area Group Name" column, letting users assign pre-defined groups directly instead of relying on automatic clustering afterward.

## Technical Summary

[CCCT-2646](https://dimagi.atlassian.net/browse/CCCT-2646)

- The group column is all-or-none: every row must have a group name, or none can. A partial mix aborts the import with a single error.
- When groups are assigned this way, the existing clustering endpoint already blocks re-clustering (it errors out if groups already exist for the opportunity), so no extra logic was needed to skip clustering.
- The import batches group creation and centroid updates, one query per unique group set instead of one per row, to keep large imports fast.

### Safety story

Tested manually on local: uploaded CSVs through the actual upload view and Celery task, then checked the database and map rendering for correct group assignment and centroids.

### Automated test coverage

Added importer and view tests covering grouped imports, ungrouped imports, the mixed-column error, and group reuse. All passing.

### QA Plan

### Labels & Review

[CCCT-2646]: https://dimagi.atlassian.net/browse/CCCT-2646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/microplanning/tasks.py
+++ b/commcare_connect/microplanning/tasks.py
@@ -6,6 +6,7 @@ from collections import defaultdict
 from django.contrib.gis.geos import GEOSException, GEOSGeometry
 from django.core.cache import cache
 from django.core.files.storage import default_storage
+from django.db import transaction
 from django.utils.html import strip_tags
 from django.utils.translation import gettext as _
 
@@ -16,7 +17,7 @@ from config import celery_app
 
 from .clustering import WorkAreaGrouper
 from .const import DEFAULT_BUILDING_COUNT
-from .models import SRID, WorkArea
+from .models import SRID, WorkArea, WorkAreaGroup
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +43,10 @@ class WorkAreaCSVImporter:
         "state": "State",
     }
 
+    # Optional column, not part of HEADERS: only "Work Area Group Name" is exempt from
+    # the required-columns check, since a sheet is allowed to omit it entirely.
+    GROUP_NAME_HEADER = "Work Area Group Name"
+
     def __init__(self, opp_id, csv_source):
         self.opp_id = opp_id
         self.csv_source = csv_source
@@ -49,6 +54,9 @@ class WorkAreaCSVImporter:
         self.seen_slugs = set()
         self.created_count = 0
         self.existing_slugs = set()
+        self.has_group_names = False
+        # Ward to use if a given group name doesn't exist yet: the ward of its first-seen row.
+        self.group_wards = {}
 
     def _validate_all_rows(self, f):
         f.seek(0)
@@ -58,18 +66,34 @@ class WorkAreaCSVImporter:
             return False
 
         self.existing_slugs.update(WorkArea.objects.filter(opportunity_id=self.opp_id).values_list("slug", flat=True))
+        total_rows = 0
+        rows_with_group = 0
         for line_num, row in enumerate(reader, start=2):
             self._process_row(line_num, row)
+            total_rows += 1
+            group_name = self.get_group_name(row)
+            if group_name:
+                rows_with_group += 1
+                self.group_wards.setdefault(group_name, self.get_ward(row))
+
+        self.has_group_names = total_rows > 0 and rows_with_group == total_rows
+        if 0 < rows_with_group < total_rows:
+            self._add_error(
+                1,
+                _("Work Area Group Name is required for all Work Areas, or must be omitted for all."),
+            )
 
         self.existing_slugs.clear()
         return len(self.errors) == 0
 
+    @transaction.atomic
     def _stream_and_insert(self, f):
         f.seek(0)
         reader = csv.DictReader(f)
         self._normalize_headers(reader)
         batch = []
         batch_size = 5000
+        group_cache = self._resolve_work_area_groups() if self.has_group_names else {}
 
         for row in reader:
             buildings, visits = self.get_building_and_visit(row)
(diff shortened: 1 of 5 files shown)


commcare_connect/microplanning/tasks.py
```diff
@@ -58,18 +66,34 @@ def _validate_all_rows(self, f):
...
         self.existing_slugs.update(WorkArea.objects.filter(opportunity_id=self.opp_id).values_list("slug", flat=True))
+        total_rows = 0
+        rows_with_group = 0
         for line_num, row in enumerate(reader, start=2):
             self._process_row(line_num, row)
+            total_rows += 1
+            group_name = self.get_group_name(row)
+            if group_name:
+                rows_with_group += 1
+                self.group_wards.setdefault(group_name, self.get_ward(row))
```
ajeety4, line 77: Is ward mandatory for work area group creation ?
If yes, I think we should throw error if ward is empty.  (Ignore if error is already thrown elsewhere)

commcare_connect/microplanning/tasks.py
```diff
@@ -58,18 +66,34 @@ def _validate_all_rows(self, f):
...
+        if 0 < rows_with_group < total_rows:
+            self._add_error(
+                1,
+                _("Work Area Group Name is required for all Work Areas, or must be omitted for all."),
+            )
 
         self.existing_slugs.clear()
         return len(self.errors) == 0
 
+    @transaction.atomic
```
ajeety4, line 89: :+1:

commcare_connect/templates/microplanning/import_work_area_modal.html
```diff
@@ -96,6 +96,9 @@ <h3 class="text-sm font-semibold text-gray-900 flex items-center gap-2 mb-3">
                   <li>{% translate "Expected Visit Count – positive integer" %}</li>
                   <li>{% translate "LGA – name of the LGA the work area is in" %}</li>
                   <li>{% translate "State – name of the state the work area is in" %}</li>
+                  <li>
+                    {% translate "Work Area Group Name (optional): Enter a group name for every row to assign work areas directly and skip automatic clustering. Leave this column blank for every row to use automatic clustering. Do not mix blank and non-blank values." %}
+                  </li>
```
ajeety4, line 101: supernit: `blank and filled values`

commcare_connect/microplanning/tests/test_tasks.py
```diff
@@ -216,6 +216,74 @@ def test_missing_extra_properties(self, opportunity):
...
+        assert groups["group-a"].centroid is not None
+        assert groups["group-b"].centroid is not None
+
+    def test_import_without_group_column_leaves_work_areas_ungrouped(self, opportunity):
+        rows = [self.build_row("area-1"), self.build_row("area-2")]
+        result = WorkAreaCSVImporter(opportunity.id, self.build_csv(rows)).run()
+
+        assert result["created"] == 2
+        assert not WorkAreaGroup.objects.filter(opportunity=opportunity).exists()
+        assert WorkArea.objects.get(slug="area-1").work_area_group is None
```
ajeety4, line 263: nit: can add same for "area-2" as well. 
`assert WorkArea.objects.get(slug="area-2").work_area_group is None`

ajeety4 (commented):
LGTM
Just a single query.

commcare_connect/templates/microplanning/import_work_area_modal.html
hemant10yadav: Fixed in 45975b13.

commcare_connect/microplanning/tests/test_tasks.py
hemant10yadav: Fixed in 45975b13.

commcare_connect/microplanning/tasks.py
hemant10yadav: Yes, ward is mandatory — it's already validated in `_validate_ward` (tasks.py:245-250), which throws "Ward is required." if empty. No code change needed here.

ajeety4: Great ! Thanks for confirming.

ajeety4 (approved):

ajeety4 (approved):

commcare_connect/microplanning/tasks.py
```diff
@@ -58,18 +66,34 @@ def _validate_all_rows(self, f):
...
+            group_name = self.get_group_name(row)
+            if group_name:
+                rows_with_group += 1
+                self.group_wards.setdefault(group_name, self.get_ward(row))
+
+        self.has_group_names = total_rows > 0 and rows_with_group == total_rows
+        if 0 < rows_with_group < total_rows:
+            self._add_error(
+                1,
+                _("Work Area Group Name is required for all Work Areas, or must be omitted for all."),
```
zandre-eng, line 83: Should we include line numbers of which rows are missing group names? I'm thinking this could make it easier for the user if there are a lot of rows.

zandre-eng (approved):

commcare_connect/microplanning/tasks.py
hemant10yadav: Yeah, that's what I was planning in the beginning, but it comes at the cost of additional tracking. We'd need to keep track of every value in that column and store them in memory, which could become a memory issue in the long run.

I also intentionally didn't use that because it's a column requirement. It's always easier to spot an empty column than to find an incorrect value in a column that's been filled in.

zandre-eng: Those are fair points, thanks for sharing your reasoning. I agree given the memory risk it's probably not worth adding per-row errors.

PR #1407, opened 2026-07-28
Remove Unused Dependencies

zandre-eng:
## Product Description

No user-facing changes. This is an internal dependency cleanup that removes twelve unused packages from the Python and front-end manifests. Rendered pages, stylesheets and bundled JavaScript are unchanged.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2473)

Each package was confirmed dead before removal, not merely unreferenced by imports: `django-autocomplete-light` is additionally absent from `INSTALLED_APPS`, which `dal` requires in order to function at all; `pixrem` and `postcss-preset-env` were never registered in the PostCSS plugin list in `webpack/base.config.js`, so they contributed nothing to the build despite being installed; and `pylint-celery`/`pylint-django` were plugins for a linter that is not a declared dependency and is not run by prek or CI, so the accompanying `[tool.pylint.*]` config block was removed with them. `pixrem` is split into its own commit because it was the CVE-driven motivation for the ticket — it carries a moderate advisory with no upstream fix and is effectively unmaintained; no replacement plugin is needed, since a px fallback for browsers without `rem` support is an IE8/IE9-era concern and our browserslist target is "last 2 versions".

Removed:

- **Runtime:** `django-autocomplete-light`, `django-model-utils`
- **Dev:** `pylint-celery`, `pylint-django`, `werkzeug[watchdog]`, `watchfiles`
- **Front-end:** `pixrem`, `postcss-preset-env`, `gulp-concat`, `node-sass-tilde-importer`, `concurrently`, `webpack-dev-server`

The commits are split one logical removal per commit, each with its own lockfile update, so any individual removal can be reverted in isolation.

### Safety story

This change touches only dependency manifests and lockfiles — no application code, templates, settings or migrations, so there is no data impact and nothing to feature-flag. Confidence comes primarily from a build-output comparison: production bundles built from `main` with all original dependencies are **byte-identical** to bundles built from this branch across all four CSS files and every JS bundle except `vendors-bundle.js`, whose only difference is a 74-byte embedded absolute build path from being built in a separate worktree. Identical CSS bytes demonstrate that none of the removed PostCSS or Sass packages was contributing to the stylesheet pipeline. I also verified that no retained package changed version while the trees were pruned (0 version changes across the 493 remaining npm entries, 299 removed), so nothing was silently upgraded alongside the deletions. Reverting is a lockfile-level operation with no cleanup required.

Secondary benefit: `npm audit` drops from 13 findings (1 critical, 6 high, 5 moderate, 1 low) to 6 (5 high, 1 moderate) — the critical came out with `webpack-dev-server`'s dependency tree.

### Automated test coverage

No new tests were added, as this change removes dependencies rather than altering behaviour. The existing suite passes in full (1579 tests) against the pruned environment, which exercises the Django apps that would have failed on a missing runtime import. `npm ci` confirms the lockfile and `package.json` agree, both the production and development webpack builds compile, and all prek hooks pass.

### QA Plan

### Labels & Review


zandre-eng:
The `npm-audit` and `pip-audit` failing checks are pre-existing. They are getting run as this PR introduces dependency changes, and are related to vulnerabilities that exist in our dependencies. These are out of scope for this PR and will be addressed in separate PRs.

Claude comment regarding big final diff on `package-lock.json` file:

**The `package-lock.json` diff is smaller than it looks.** ~2970 insertions / 7361 deletions, but parsed as JSON it's 293 package entries removed, **0 added**, and one edited `devDependencies` list. 100% of the inserted lines exist verbatim among the deleted ones — Myers-diff realignment over repetitive boilerplate, not real change. Crucially, **0 version changes across the 493 retained entries**, so nothing was silently upgraded while the trees were pruned. Reviewable as "293 dead entries deleted, no upgrades."

pxwxnvermx (approved):

hemant10yadav (approved):

pxwxnvermx (approved):

PR #1409, opened 2026-07-29
Add periodic task to clear expired OAuth tokens

hemant10yadav:
## Technical Summary
https://dimagi.atlassian.net/browse/CCCT-2472

- Adds a Celery task that runs django-oauth-toolkit's built-in [`cleartokens`](https://django-oauth-toolkit.readthedocs.io/en/latest/management_commands.html)  management command to purge expired OAuth tokens.

Checked prod to see what this would actually clean up:

| Query | Count |
|---|---|
| Total `AccessToken` rows | 4,455,011 |
| Expired | 4,415,465 |
| Expired with no refresh token | 4,414,662 |

99% of access tokens in prod are already expired, and almost all of those have no refresh token attached, so `cleartokens` will remove them outright rather than leaving orphaned refresh tokens behind. No duplicate tokens were found, so there's no risk of the cleanup colliding with anything else.

### Safety story

The task only calls django-oauth-toolkit's own [`cleartokens`](https://django-oauth-toolkit.readthedocs.io/en/latest/management_commands.html) command, which deletes expired token rows, so there's no new logic to get wrong. The migration is reversible and idempotent.

### Automated test coverage

No new tests added; the task is a thin wrapper around an already-tested library command.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/users/tasks.py
+++ b/commcare_connect/users/tasks.py
@@ -14,4 +14,5 @@ def get_users_count():
 
 @celery_app.task()
 def clear_expired_oauth_tokens():
+    # https://django-oauth-toolkit.readthedocs.io/en/latest/management_commands.html
     call_command("cleartokens")
--- a/commcare_connect/users/tasks.py
+++ b/commcare_connect/users/tasks.py
@@ -1,4 +1,5 @@
 from django.contrib.auth import get_user_model
+from django.core.management import call_command
 
 from config import celery_app
 
@@ -9,3 +10,8 @@ User = get_user_model()
 def get_users_count():
     """A pointless Celery task to demonstrate usage."""
     return User.objects.count()
+
+
+@celery_app.task()
+def clear_expired_oauth_tokens():
+    call_command("cleartokens")
--- a/commcare_connect/users/migrations/0028_create_clear_expired_oauth_tokens_periodic_task.py
+++ b/commcare_connect/users/migrations/0028_create_clear_expired_oauth_tokens_periodic_task.py
@@ -1,0 +1,42 @@
+from django.db import migrations
+from django_celery_beat.models import CrontabSchedule, PeriodicTask
+
+
+def create_periodic_task(apps, schema_editor):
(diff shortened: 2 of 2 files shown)


pxwxnvermx (dismissed):

commcare_connect/users/tasks.py
```diff
@@ -9,3 +10,8 @@
 def get_users_count():
     """A pointless Celery task to demonstrate usage."""
     return User.objects.count()
+
+
+@celery_app.task()
+def clear_expired_oauth_tokens():
+    call_command("cleartokens")
```
ajeety4, line 18: Nit: It would be nice to add a comment about the source of cleartokens (django-oauth-toolkit's built-in command) , since it's not immediately obvious where it comes from reading the code.

ajeety4 (dismissed):
LGTM, assuming we've tested all the authentication flows used in Connect.

One small note: could we also update the PR title and description to mention that the djangoauth-toolkit version is being upgraded as part of this change?

ajeety4:
> LGTM, assuming we've tested all the authentication flows used in Connect.
> 
> One small note: could we also update the PR title and description to mention that the djangoauth-toolkit version is being upgraded as part of this change?

nvm. Just realised you have a separate PR for the upgrade. Would be nice to remove commit 1 then.

hemant10yadav:
> nvm. Just realised you have a separate PR for the upgrade. Would be nice to remove commit 1 then.

I accidentally created the PR from the oauth-toolkit branch, but I've fixed it now.
@ajeety4

commcare_connect/users/tasks.py
hemant10yadav: Sure - [4780788](https://github.com/dimagi/commcare-connect/pull/1409/commits/4780788fac1068be733a3caae931b9ffbb3cf727)

ajeety4 (approved):
Thanks for the fix

zandre-eng (approved):

PR #1413, opened 2026-07-29
Increase timeout for send_message

Charl1996:
## Technical Summary
Up the `send_message` request timeout to 15 seconds from the previous 5 seconds due to some timeout errors I saw in the celery logs. I don't see any errors on ConnectID side pertaining to this endpoint though, but I know Twilio has had some issues recently so I'm wondering if the holdup might be on their side. That said, I'm not sure if this is the right solution, but it's something we _can_ do.

### Safety story
Simple change

### Automated test coverage
N.A

### QA Plan
N.A

### Labels & Review
- [x ] The set of people pinged as reviewers is appropriate for the level of risk of the change

### DIFF
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -93,7 +93,7 @@ def get_user_otp(phone_number):
         raise
 
 
-def _make_request(method, path, params=None, json=None, timeout=5, auth=None) -> Response:
+def _make_request(method, path, params=None, json=None, timeout=10, auth=None) -> Response:
     if json and not method == "POST":
         raise ValueError("json can only be used with POST requests")
     if not auth:
--- a/commcare_connect/connect_id_client/main.py
+++ b/commcare_connect/connect_id_client/main.py
@@ -29,7 +29,7 @@ def fetch_demo_user_tokens() -> list[DemoUser]:
 
 def send_message(message: Message):
     """Send a push notification to a user."""
-    response = _make_request(POST, "/messaging/send/", json=message.asdict())
+    response = _make_request(POST, "/messaging/send/", json=message.asdict(), timeout=15)
     data = response.json()
     return MessagingResponse.build(**data)
 


pxwxnvermx (approved):

commcare_connect/connect_id_client/main.py
```diff
@@ -29,7 +29,7 @@ def fetch_demo_user_tokens() -> list[DemoUser]:
 
 def send_message(message: Message):
     """Send a push notification to a user."""
-    response = _make_request(POST, "/messaging/send/", json=message.asdict())
+    response = _make_request(POST, "/messaging/send/", json=message.asdict(), timeout=15)
```
sravfeyn, line 32: Seems fine to do it, but did you look up on connectid whether this endooint is actually taking more time than usual?

Also does the region migration affect this?

sravfeyn (approved):

commcare_connect/connect_id_client/main.py
zandre-eng, line 32: Should the same timeout be applied to `send_message_bulk`?

zandre-eng (approved):
LGTM, with just one non-blocking query.

ajeety4 (approved):
I see a sentry error which is timeout for "/users/fetch_users". 
Wondering if we should also update this as it is part of adding connect users ?

hemant10yadav:
> I see a sentry error which is timeout for "/users/fetch_users". Wondering if we should also update this as it is part of adding connect users ?

I also noticed that several Connect ID-related tasks on Connect have been failing based on the logs. I ran a simple query against the logs, and over the last 48 hours the following tasks have failed:

* `commcare_connect.opportunity.tasks.invite_user` — **129** failures
* `commcare_connect.opportunity.tasks.notify_user_for_scored_assessment` — **35** failures
* `commcare_connect.opportunity.tasks.add_connect_users` — **6** failures
* `commcare_connect.opportunity.tasks.send_push_notification_task` — **3** failures
* `commcare_connect.opportunity.tasks.send_notification_inactive_users` — **2** failures

zandre-eng (approved):

commcare_connect/connect_id_client/main.py
Charl1996: > Also does the region migration affect this?

I'm suspecting it might play a role, yes.

> did you look up on connectid whether this endooint is actually taking more time than usual?

Still to do this, but it also seems like a more general issue since other endpoints are also timing out. I'm not yet 100% sure if it's just the cross-region factor or if there's a more general issue to address.

hemant10yadav (approved):

ajeety4 (approved):

PR #1415, opened 2026-07-29
Allow Export of presigned references for attachments

ctsims:
## Product Description
Extends the export endpoints to support the creation of pre-authenticated URL's rather than byte responses.

## Technical Summary

[Functional and Tech Spec](https://docs.google.com/document/d/1xb_ncJKN8vMQOZVLN-bMNQE9u61hLirpn9I39MUN02o/edit?tab=t.0) - Includes overview of the reasoning for this model of approach.

This PR is split into two parts. The first commit cleanly refactors the current source of authorization for the `export/.../image` endpoint to ensure that the new endpoint uses the exact same authorization semantics. The second commit creates the new endpoint and response semantics. 

### Safety story

Discussion of the use of pre-authed url's for programmatic access to bulk attachments is in the original specification. As they are still part of the `export` structure here and won't be accessed non-programmatically, providing a pre-signed url isn't a meaningful escalation of access or revocability compared to a direct download. 

Since this can't be tested locally due to the native reliance on S3, it was tested in Staging against test data submitted along the normal path. 

Testing confirmed:
* The existing image endpoints still work correctly
* The existing image endpoints still require authentication
* The existing image endpoints still only work for authorized users (fetches begin failing after removal from project)
* New endpoints work to create links to the attachments which do not require additional auth
* Pre-authed URL's expire after the correct time period (10 minutes)
* New endpoints require authentication
* New endpoints only work for authorized users (fetches begin failing after removal from project)

### Automated test coverage

New tests added to confirm the request semantics, but since the underlying feature is dependent on S3 (for pre-signed url availability) the ultimate fetch can't be tested live, same as the existing image byte fetch endpoint. 

### QA Plan

Since the export API's aren't part of the UX, QA shouldn't be necessary based on the limited scope of changes. 

### Labels & Review

### DIFF
--- a/commcare_connect/data_export/views.py
+++ b/commcare_connect/data_export/views.py
@@ -1,3 +1,4 @@
+import copy
 import csv
 import uuid
 from collections import defaultdict
@@ -512,9 +513,10 @@ def _get_attachment_signed_url(blob_id, expire=ATTACHMENT_SIGNED_URL_EXPIRY):
 
     Caller must guard with ``_default_storage_supports_signed_urls()`` first.
     """
-    # AWS_QUERYSTRING_AUTH is globally False, so build a signing-enabled storage of the
-    # same class (preserving its ``location`` prefix) and request a GET-only URL.
-    signed_storage = type(storages["default"])(querystring_auth=True)
+    # Create a storage handler which allows pre-authed url's, retaining the existing
+    # config of the default handler.
+    signed_storage = copy.copy(storages["default"])
+    signed_storage.querystring_auth = True
     return signed_storage.url(blob_id, expire=expire, http_method="GET")
 
 
--- a/commcare_connect/data_export/views.py
+++ b/commcare_connect/data_export/views.py
@@ -490,6 +490,42 @@ class ImageView(OpportunityDataExportView):
         return FileResponse(attachment, filename=blob_meta.name, content_type=blob_meta.content_type)
 
 
+# Signed URLs are consumed by a follow-up request made immediately after issuance.
+ATTACHMENT_SIGNED_URL_EXPIRY = 60 * 10  # seconds (10 minutes)
+
+
+def _default_storage_supports_signed_urls():
+    """True when the default storage is S3-backed and can produce a portable signed URL.
+
+    django-storages is a production-only dependency, so it may be absent entirely; when it
+    is, there is no S3 backend and therefore no portable URL.
+    """
+    try:
+        from storages.backends.s3boto3 import S3Boto3Storage
+    except ImportError:
+        return False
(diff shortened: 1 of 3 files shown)


commcare_connect/data_export/views.py
```diff
@@ -166,6 +166,16 @@ def _get_opportunity_or_404(user, opp_id):
...
 
 
+def _get_scoped_blob_meta(request):
+    """Resolve the ``blob_id`` query param to its BlobMeta, enforcing that the requesting
+    user has access to the opportunity that owns the blob."""
+    blob_id = request.query_params["blob_id"]
+    blob_meta = BlobMeta.objects.get(blob_id=blob_id)
+    form = UserVisit.objects.get(xform_id=blob_meta.parent_id)
+    _get_opportunity_or_404(request.user, form.opportunity_id)
+    return blob_meta
```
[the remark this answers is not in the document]
ctsims: Leaving this unaddressed, as the semantics here are unchanged from the original code (just moved), and refactoring here would require some decisions about the correct response shape that are out of scope.

ctsims:
PR Feedback addressed (one suggestion taken, one deferred).

commcare_connect/data_export/views.py
```diff
@@ -166,6 +167,16 @@ def _get_opportunity_or_404(user, opp_id):
         raise NotFound()
 
 
+def _get_scoped_blob_meta(request):
+    """Resolve the ``blob_id`` query param to its BlobMeta, enforcing that the requesting
+    user has access to the opportunity that owns the blob."""
+    blob_id = request.query_params["blob_id"]
+    blob_meta = BlobMeta.objects.get(blob_id=blob_id)
+    form = UserVisit.objects.get(xform_id=blob_meta.parent_id)
+    _get_opportunity_or_404(request.user, form.opportunity_id)
```
pxwxnvermx, line 176: nit: This check for opportunity is redundant for views that inherit from OpportunityDataExportView. This can be replaced to check if the UserVisit belongs to the Opportunity already linked on the view parsed from the opp_id param.
```suggestion
    get_object_or_404(UserVisit, xform_id=blob_meta.parent_id, opportunity=self.opportunity)
```

pxwxnvermx (approved):
Small nit. Rest looks good.

commcare_connect/data_export/views.py
ctsims: I was leaving the auth block untouched in this rev as it's time sensitive, but this is a good change and also prevents a different issue that CodeRabbit caught. 

I'm going to seal and merge this PR to make sure it gets out next deploy and open another for this refactor.

PR #1416, opened 2026-07-30
Add ability to update task case property if not assigned

Charl1996:
## Product Description
Add the ability to update a task type case property after the task type has been created. This is **_only_** possible if there exists no `AssignedTask` records with the given task type. 

When as task with task_type has not been assigned yet - is able to update case property: 
image

After a task with task_type has been assigned - is not able to update case property: 
image

## Technical Summary
Ticket: https://dimagi.atlassian.net/browse/CI-848

We allow the editing of a task type case property when the task type is unused, but lock the case property editing when the task type is in use. The "locking" mechanism is an easy win to ensure the user can still correct the case_property in case a spelling mistake or so was during setup, without dealing with the complexities of hitting HQ to rectify the  change. This change is thus only effective during setup phase.  

### Safety story
Tested locally

### Automated test coverage
A couple of unit tests have been added covering some scenarios.

### QA Plan
No QA planned.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/forms.py
+++ b/commcare_connect/opportunity/forms.py
@@ -2104,15 +2104,33 @@ class EditTaskTypeForm(forms.ModelForm):
 
     class Meta:
         model = TaskType
-        fields = ["name", "description"]
+        fields = ["name", "case_property", "description"]
         widgets = {"description": forms.Textarea(attrs={"rows": 2})}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.instance and self.instance.archived:
             self.fields["is_archived"].initial = True
+        if self._has_assigned_tasks():
+            self._lock_case_property()
         self.helper = FormHelper(self)
         self.helper.form_tag = False
+        self.helper.layout = Layout(
+            Field("name"),
+            Field("case_property"),
+            Field("description"),
+            Field("is_archived"),
+        )
+
+    def _has_assigned_tasks(self):
+        return bool(self.instance.pk) and AssignedTask.objects.filter(task_type=self.instance).exists()
+
+    def _lock_case_property(self):
+        field = self.fields["case_property"]
+        field.disabled = True
+        field.help_text = _(
+            "This cannot be changed because the task type has already been assigned to one or more workers."
+        )
 
     def save(self, commit=True):
         instance = super().save(commit=False)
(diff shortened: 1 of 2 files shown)


commcare_connect/opportunity/forms.py
```diff
@@ -2104,15 +2104,33 @@ class EditTaskTypeForm(forms.ModelForm):
...
-        fields = ["name", "description"]
+        fields = ["name", "case_property", "description"]
         widgets = {"description": forms.Textarea(attrs={"rows": 2})}
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.instance and self.instance.archived:
             self.fields["is_archived"].initial = True
+        if self._has_assigned_tasks():
+            self._lock_case_property()
```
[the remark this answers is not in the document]
Charl1996: Good point.

pxwxnvermx (approved):

commcare_connect/opportunity/forms.py
```diff
@@ -2104,15 +2104,33 @@ class EditTaskTypeForm(forms.ModelForm):
...
+            Field("description"),
+            Field("is_archived"),
+        )
+
+    def _has_assigned_tasks(self):
+        return bool(self.instance.pk) and AssignedTask.objects.filter(task_type=self.instance).exists()
+
+    def _lock_case_property(self):
+        field = self.fields["case_property"]
+        field.disabled = True
```
zandre-eng, line 2130: Minor nit: Do we need to add any validation on the back-end once an edit form has been submitted? A user can go into dev tools and remove the disabled param to be able to submit a case property change where it shouldn't be allowed.

The above is a very minor edge case, as I assume most users won't know how to do this. Still, this could be a way for a bad actor to cause unintended side effects through changing the case property on assigned tasks.

Feel free to ignore the above if you think it is not worth defending against.

zandre-eng (approved):
One minor nit, but overall LGTM.

commcare_connect/opportunity/forms.py
Charl1996: > Do we need to add any validation on the back-end once an edit form has been submitted?

Yes, I'm currently working on that part.

Charl1996: Actually, I think it's an edge case so going to do that as a followup PR instead of in this PR.

PR #1417, opened 2026-07-30
CI-859: Redesign opportunity detail header with consumption metric strip

mmmary314:
## Product Description

Redesigns the opportunity detail page header (between the breadcrumb and the page tabs) — replaces the old two-tone header (unbounded description, static max-only "ceiling" tiles) with a single fixed-shape layout:

- Lifecycle (Active/Inactive) and Test badges moved into one cluster next to the title
- The old max-only ceilings strip is now a **consumption strip**: actual vs. cap for Connect Workers, Service Deliveries, and Budget, each with a thin progress bar, plus the delivery window's elapsed time and closed state
- An "Ended <date>" note appears once an opportunity's window has passed
- Learn App / Deliver App / Payment Units tiles are restyled to match, kept clickable to open the existing modal

### Screenshots — real running site

Captured against a live Django server (Postgres/Redis via Docker) with real data, logged in as an org admin — not the design-mockup renders from the first draft of this PR.

**Live opportunity, active:**
live-active

**Test opportunity, active** (note the automatic 5→3 column reflow — this browser window is narrower):
test-active

**Inactive / closed** — Edit button hidden, bars desaturated, "Ended" note, delivery window reads "Closed":
inactive

## Technical Summary

No ticket link — this started from a Claude-generated design handoff, not a Jira ticket.

Architecture, in short:
- `opportunity_header.py`: one function, `get_opportunity_header_context`, builds the actual-vs-cap metrics (workers accepted, approved deliveries, accrued budget — one combined `OpportunityAccess` aggregate query for the first and third) and the delivery-window math (`_delivery_window`/`_pct` helpers).
- `templates/opportunity/opportunity_header.html`: the partial. Theming is **not** resolved in Python — the template applies plain modifier classes straight from `object.is_test`/`object.is_active` (`opp-header--test`, `opp-header--inactive`), and `tailwind.css` defines the color variants as CSS custom properties scoped to those classes. This was a deliberate revision after review feedback that Python shouldn't hand the template a pre-resolved "theme" dict.
- `tailwind.css` gained a set of `.opp-header-*` structural component classes (badge/button/counter/metric-cell layout) instead of repeating raw Tailwind utility strings at each call site, per this repo's convention of predefined style classes.

### Safety story

- Pure presentation change: one new context-building module, one new template partial, no migrations, no schema changes, no new writes (the new query is a read-only aggregate).
- `get_opportunity_header_context` receives an `Opportunity` instance that's already org-scoped by `OpportunityDashboard`'s existing `get_opportunity_or_404` — no new cross-tenant surface.
- Verified end-to-end against a real local instance: ran migrations against a fresh Postgres, logged in through the actual login flow, and walked live-active / Test-active / inactive-closed opportunities plus the narrow-viewport reflow (screenshots above).
- Went through a five-lens self-review (design/quality/security/smells/maintainability) against the full diff before this push; two real issues came out of it and are fixed in the last commit (a timezone inconsistency in the delivery-window date, and a couple of hardcoded color literals that should have referenced existing CSS variables).

### Automated test coverage

- `commcare_connect/opportunity/tests/test_opportunity_header.py` (new): unit tests for `_pct`/`_delivery_window` boundary cases (missing dates, not-yet-started, closed, in-progress, ends-today, months-left floor), plus an integration test against real factories (`test_get_opportunity_header_context`) that pins down the actual/cap/pct values, including the floor-before-multiply behavior for the Service Deliveries cap when the budget doesn't divide evenly across workers.
- Ran the full existing `commcare_connect/opportunity` test suite locally against this branch: **617 passed**, no regressions.

### QA Plan

- [ ] Load `/a/<org_slug>/opportunity/<opp_id>/` for a live opportunity, a Test opportunity, and an inactive/ended opportunity — confirm the header matches the screenshots above
- [ ] Confirm the Edit button and overflow menu still respect viewer/PM permissions
- [ ] Confirm Learn App / Deliver App / Payment Units tiles still open the existing modal
- [ ] Resize the browser to check the metric strip reflow at a narrow viewport with the sidebar expanded

### Labels & Review


mmmary314:
Addressing both nitpicks from the review:

1. **Unit tests for `_pct`/`_delivery_window`** — added in `768671ac`: covers missing dates, not-yet-started, closed, in-progress, the ends-today boundary, and the months-left floor. Writing the ends-today case actually surfaced that the `total_days <= 0` guard in `_delivery_window` was unreachable dead code (the two preceding branches already exclude every case where the window could be zero-length or inverted), so that's been simplified too.

2. **Raw Tailwind utility classes instead of predefined style classes** — partially addressed in `2fa3dcfc`: the repeated *structural* patterns (badge pill, action buttons, resource counters, metric cell/label/value/bar) are now `.opp-header-*` component classes in `tailwind.css`. What's left in `_THEME` (`theme.hero`, `theme.title`, `theme.well`, etc.) are plain color/opacity utility classes like `bg-brand-deep-purple` and `text-white/72` — the same kind of utility Tailwind class already used unwrapped everywhere else in this codebase (e.g. `text-brand-sky`, `bg-brand-indigo`). Wrapping each of the 4 theme × state combinations into named variant classes (`.opp-hero--live-active`, `.opp-hero--test-inactive`, etc.) would mean ~80 single-property classes duplicating what the color utilities already express directly, for a component that only renders in one place. Keeping colors as utilities and only extracting the genuinely repeated *structural* classes seemed like the better tradeoff here — happy to revisit if reviewers feel otherwise.

@coderabbitai review

calellowitz (changes requested):
Hey @mmmary314 Thanks for submitting this.

To make review on this, a little bit easier, I have a couple requests.

1) Are you able to use our PR template? Filling that out gives a lot of great context on the change that makes reviewing it easier. The safety story is especially helpful, especially if you include the types of testing you have done on this change. I know this PR includes a test plan but its not clear if anyone has run through it.

2) The screenshots look like they came from claude design rather than the running site (since they include only the header floating). Are you able to include actual screenshots of what the new opportunity dashboard will look like running, both to verify the fidelity of the implementation to the desired state, and to help see how it all fits together.

commcare_connect/opportunity/opportunity_header.py
```diff
@@ -0,0 +1,170 @@
...
+            "bar_fill_budget": "bg-slate-300",
+        },
+    },
+}
+
+
+def get_opportunity_header_context(opportunity: Opportunity) -> dict:
+    is_inactive = not opportunity.is_active
+    surface = "test" if opportunity.is_test else "live"
+    theme = {**_THEME[surface]["active"], **_THEME[surface]["inactive"]} if is_inactive else _THEME[surface]["active"]
```
Charl1996: It seems like the theme is changed based on the `active` condition of the opportunity, but I would suggest letting the template decide how to render the data given to it instead of telling the template how to render (i.e. handing it a theme).

mmmary314:
Thanks both — addressed:

**@calellowitz**
1. Filled out the org PR template (Product Description / Technical Summary / Safety Assurance) in the description above.
2. Replaced the design-mockup screenshots with real screenshots from a running instance — logged in through the actual login flow against a local Postgres/Redis stack, walking live-active, Test-active, and inactive/closed opportunities.

**@Charl1996** — agreed, and changed it: the header no longer computes a Python "theme" dict and hands it to the template. The template now applies plain modifier classes straight from the raw data (`object.is_test`, `object.is_active` → `opp-header--test`, `opp-header--inactive`), and `tailwind.css` defines the color variants as CSS custom properties scoped to those classes. Python only supplies the metric data now (actual/cap/pct, resources, delivery window) — no rendering decisions.

Also ran a five-lens self-review (design/quality/security/smells/maintainability) against the full diff before this last push, which caught and fixed a timezone inconsistency (the delivery-window calc used the naive system clock instead of Django's TZ-aware `now()`, same pattern `Opportunity.is_active`/`has_ended` already use) and a couple of hardcoded color literals that should have referenced existing CSS variables.

Full `commcare_connect/opportunity` test suite passes locally (617 tests, no regressions).

sravfeyn:
Thanks, the dashboard looks great in terms of design!

While the code is functionally correct, it is a bit inconsistent with our coding standards.

For e.g.
1 ‘theme’ is written as a dict in Python, I think this is now refactored into tailwind, but it should probably be its own dashboard.css (included in tailwind.css) and should try to reuse the existing established patterns in tailwind.css or define it in a reusable way.
2 Its is introducing its own custom date calculations in _delivery_window and introducing some non-essential tests for it, instead of reusing the ones from utils/datetime.py.
3 It is making direct use of text-[nx] without styling the classes in tailwind or making use of font classes.
4. The way the context is generated and is switched into to a new file needs to be refactored.

I think this is because AI tends to solve for local requirements and tends to use shortcuts. This risks establishing anti-patterns that AI might adapt later in the codebase, which makes it much harder to add newer features without causing bugs.

I feel given the size of the PR, iterating via code review could take more of your time understanding our coding standards than necessary. To save time, should we hand this off to a developer to adapt the code to our internal quality standards?

mmmary314:
Thanks for the detailed feedback — took all four points seriously and made real structural changes, not just cosmetic ones. Pushed two commits:

**1. Theme as a Python dict** — already addressed in an earlier commit: the header no longer computes a theme dict in Python at all. The template applies plain modifier classes from raw data (`object.is_test`/`object.is_active` → `opp-header--test`/`opp-header--inactive`), and `tailwind.css` defines the color variants as CSS custom properties scoped to those classes.

**2. Custom date calculations instead of reusing `utils/datetime.py`** — moved the delivery-window math out of the opportunity-specific code entirely. `get_elapsed_percent()` and `get_months_remaining()` now live in `commcare_connect/utils/datetime.py`, built on `dateutil.relativedelta` (already a dependency of that module, used by its other functions) instead of hand-rolled year/month arithmetic. `OpportunityDashboard` just calls them.

**3. Raw `text-[Npx]` instead of styled classes** — extended the Tailwind `@theme` scale with two named tokens (`--text-2xs` = 11px, `--text-xs-plus` = 13px), the same way brand colors are already tokenized, and replaced every arbitrary bracket value in the header template and its component classes. Confirmed zero `text-[` usages remain in either file.

**4. Context generation living in its own new file** — deleted `opportunity_header.py`. The aggregate query moved to `get_opportunity_header_stats()` in `helpers.py`, next to the existing `get_opportunity_worker_progress()`/`get_opportunity_delivery_progress()` it mirrors. Context assembly moved to `get_header_context()`/`get_delivery_window_context()` methods directly on the `OpportunityDashboard` view — the same pattern `WorkerPageView.get_worker_kpi_context()` already uses elsewhere in this file.

Before pushing, I ran this whole set of changes through a five-lens self-review (design/quality/security/smells/maintainability) specifically checking whether the refactor actually delivered on these points rather than just moving code around. That caught a few of its own issues along the way — a benign but real double-clock-read in the delivery-window calc, an explanatory comment that got dropped in the move, a dropped unit test for the percent-clamp helper, and an inconsistent CSS token name — all fixed in the second commit.

Full `opportunity` + `utils` test suites pass locally (688 tests). Happy to keep iterating if there's more that doesn't match the codebase's conventions — genuinely want this to land in a shape the team would have written themselves, not just something that looks right on the surface.

tailwind/tailwind.css
```diff
@@ -852,3 +857,194 @@ select:disabled,
...
 }
+
+/*--- Opportunity header -------------------------
+   State lives on the root `.opp-header` element as plain modifier classes
+   (`--test`, `--inactive`) driven straight from the raw `is_test`/`is_active`
+   data; every themed descendant just reads the CSS variables that modifier
+   sets, so the template/CSS decide how to render rather than being handed a
+   pre-resolved theme object. ------------------------- */
+@layer components {
+  .opp-header {
```
sravfeyn, line 871: It looks like based on my earlier review to use abstractions and modularity and tailwind.css utilities, this code changes have gone to the over-abstraction side.

This has resulted in 21 variables for the root component that are getting overridden 16, 10 and 9 times for each theme state latter (for test, inactive, and test+inactive) 

This makes it difficult to infer the code (for e.g. to infer color of `.opp-header-metric-value` it would need a chain of lookups), it would also make the composition more difficult than needed, for e.g.
```
<div class="opp-header {% if object.is_test %}opp-header--test{% endif %} 
                       {% if not object.is_active %}opp-header--inactive{% endif %}">
```

Instead of components with variables like this - since we have four themes (normal, in_active, test, inactive+test) we can have four separate group of classes/components for each theme while reusing common between them.

Then the composition would be simpler
```
<div class="opp-header opp-header--{{ theme }}">
```

and in view

```
if opportunity.is_test:
    theme = 'inactive-test' if not opportunity.is_active else 'active-test'
else:
    theme = 'inactive' if not opportunity.is_active else 'active'
```

sravfeyn:
Thanks @mmmary314 for the updates, sorry for the delay on review here. I have added [one comment](https://github.com/dimagi/commcare-connect/pull/1417/changes#r3842559874) that should hopefully help Claude refactor this in a better way. Let me know if this is urgent, happy to approve and then the refactor improvements could be a separate dev ticket.

mmmary314:
Implemented exactly as suggested — pushed.

`OpportunityDashboard.get_header_context` now computes one theme string (`"active"`/`"inactive"`/`"active-test"`/`"inactive-test"`) from `is_test`/`is_active`:

```python
if opportunity.is_test:
    theme = "inactive-test" if not opportunity.is_active else "active-test"
else:
    theme = "inactive" if not opportunity.is_active else "active"
```

and the template just does `class="opp-header opp-header--{{ header.theme }}"`. `tailwind.css`'s "Opportunity header" section is now four flat, self-contained rule sets — every themed element's color for a given theme is one direct selector away (`.opp-header--<theme> .opp-header-<part> { color: ... }`), no variables to resolve. Rules identical across theme pairs are grouped with a comma selector (that's the "reuse what's common" from the original scheme, just without the indirection).

Ran this through a five-lens self-review before pushing, which caught two subtle CSS issues worth mentioning: a same-specificity override that only worked because of rule order (now split into two unambiguous rules), and a dropped `border: 1px solid transparent` on the live theme that the Test theme's real border didn't have an equivalent for (restored, for box-model parity across all four themes). Also added test coverage for the theme computation itself, which had none.

Full `opportunity` + `utils` test suites pass (692). Let me know if this shape works, or if there's anything else that doesn't sit right — happy to keep going.

PR #1419, opened 2026-08-03
CCCT-2525: CompletedWorkInvoice snapshot table + invoiced_approved_count  (schema only)(1/4)

ajeety4:
## Product Description

No user-facing change. Additive schema only: introduces the `CompletedWorkInvoice` snapshot table and a `CompletedWork.invoiced_approved_count` watermark for the upcoming forward-delta invoicing work. Nothing reads or writes them yet.

## Technical Summary

Part of the forward-delta invoicing effort (tech spec: *[Invoice line-item locking & post-invoice completed-work changes](https://github.com/dimagi/commcare-connect/blob/36b3c25d462cf594b0e472936cbc273526c059a7/docs/superpowers/specs/2026-06-08-invoice-line-item-locking-design.md#good-to-have--legacy-drift-report)*; ticket **[CCCT-2525](https://dimagi.atlassian.net/browse/CCCT-2525)** — schema + backfill (_currently blocked_)). This PR is the **schema slice only**:

- **New `CompletedWorkInvoice`** — one row per `(invoice, completed_work)`, a frozen per-work billing snapshot carrying **FLW pay** (`flw_amount_local` / `flw_amount_usd`) and **org pay** (`org_amount_local` / `org_amount_usd`), a `billed_count`, a `month` (first-of-month period attribution), and `exchange_rate` stored as a **Decimal value** (`1` for USD, not an FK). Both FKs `on_delete=PROTECT`; `unique_together (invoice, completed_work)`.
- **New `CompletedWork.invoiced_approved_count`** (`IntegerField`, default `0`) — the billing watermark.
- A `CompletedWorkInvoiceFactory` for downstream tests.

> Note: Backfill of existing invoices and the read/write cutover ship in follow-up PRs.

### Safety story

Purely additive, backwards-compatible schema. Nothing reads or writes the new table/field yet, so there is **no runtime behavior change**. The migration is metadata-only: a `CreateModel` for a new table plus an `AddField` with a constant `default=0` (fast metadata-only column add on PostgreSQL 11+ — no table rewrite, no long lock). Existing `CompletedWork` rows get `invoiced_approved_count = 0`, which is the correct initial watermark. No data migration and no `RunPython`, so no multidb hints are required.

### Automated test coverage

Schema-only change, verified locally via `./manage.py migrate`, `./manage.py makemigrations --check` (no missing migrations), and `./manage.py check`. Behavioral tests land with the follow-up backfill and cutover PRs that wire the table into invoicing.

### QA Plan

No user-facing behavior in this PR to QA. QA is covered by the follow-up backfill and cutover PRs.

### Labels & Review

[CCCT-2525]: https://dimagi.atlassian.net/browse/CCCT-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/opportunity/models.py
```diff
@@ -941,6 +945,46 @@ def completion_date(self):
...
+    flw_amount_local = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    flw_amount_usd = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    org_amount_local = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    org_amount_usd = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    exchange_rate = models.DecimalField(
+        max_digits=10,
+        decimal_places=6,
+        default=1,
+        help_text=gettext_lazy("Local-currency-per-USD rate used for this row's month; 1 for USD."),
+    )
```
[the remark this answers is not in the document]
ajeety4: Intentional .This mirrors ExchangeRate.rate (also max_digits=10, decimal_places=6), which is the source of the value. The field is populated from get_exchange_rate() → either ExchangeRate.rate or a literal 1 (USD), so it can never receive a value that ExchangeRate.rate couldn't already store; matching precision guarantees a lossless round-trip and keeps the two in sync. Widening only this snapshot field would let it claim to support rates the source can't produce.

commcare_connect/opportunity/models.py
```diff
@@ -941,6 +945,46 @@ def completion_date(self):
...
+    billed_count = models.PositiveIntegerField(
+        validators=[MinValueValidator(1)],
+        help_text=gettext_lazy("Approved units billed on THIS invoice (delta)."),
+    )
+    flw_amount_local = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    flw_amount_usd = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    org_amount_local = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    org_amount_usd = models.DecimalField(max_digits=10, decimal_places=2, default=0)
+    exchange_rate = models.DecimalField(
```
zandre-eng: I can't remember if there was a discussion around this, but is there a reason we are saving the exchange rate raw value instead of doing a FK to an `ExchangeRate` row?

zandre-eng (approved):
Looks good, just one query around the exchange rate field.

commcare_connect/opportunity/models.py
ajeety4: Since we're freezing amounts and this is financial data, it felt safer to store the applied exchange rate rather than an FK to a row that could be edited or deleted. We also don't have an ExchangeRate row for USD, so an FK would need to be nullable. 

Happy to switch to an FK if you think the tradeoffs are worth it.

zandre-eng: Thanks for explaining your thought process. I'm happy with keeping it a decimal, as the only real value I could see with using a FK is being able to view the exchange rate date (not sure if we'd care about that here). I just wanted to confirm the approach here as I believe everywhere else we use a FK to reference the exchange rate.

ajeety4: > using a FK is being able to view the exchange rate date 

I did some rethinking about this, and I think the FK could be useful for knowing which exchange rate date was actually used.

The row month usually captures the same date, but there is an edge case where the lookup falls back to a previous month's exchange rate if the current month doesn't have one.

So I'm leaning towards going back to the FK approach and can consider a follow-up ticket to make ExchangeRate immutable once it has been used for invoicing (enforced at the application layer) to address the concern about historical records changing.

There is an alternative of storing a rate_date column, but that feels like additional overhead for this use case.

Happy to hear any thoughts on this.

zandre-eng: > So I'm leaning towards going back to the FK approach and can consider a follow-up ticket to make ExchangeRate immutable

That sounds good, I'm happy with the approach of using a FK instead.

ajeety4: Thanks. Addressed in 52efb8919d27efb0400ceab1dfadc3673a10fd5f

commcare_connect/opportunity/models.py
```diff
@@ -941,6 +945,46 @@ def completion_date(self):
...
+    invoices) so invoice line items are read from frozen rows instead of recomputed from
+    CompletedWork.saved_* on every view. Stores both FLW pay and org (workspace) pay.
+    """
+
+    invoice = models.ForeignKey(
+        PaymentInvoice,
+        on_delete=models.PROTECT,
+        related_name="work_items",
+    )
+    completed_work = models.ForeignKey(
```
sravfeyn, line 961: What happens to `invoice` field on completed_work, post this?

sravfeyn (approved):

commcare_connect/opportunity/models.py
ajeety4: It will be removed as a follow up work

commcare_connect/opportunity/migrations/0143_completedwork_invoiced_approved_count_and_more.py
```diff
@@ -0,0 +1,42 @@
...
+        ),
+        migrations.CreateModel(
+            name='CompletedWorkInvoice',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('created_by', models.CharField(max_length=255)),
+                ('modified_by', models.CharField(max_length=255)),
+                ('date_created', models.DateTimeField(auto_now_add=True)),
+                ('date_modified', models.DateTimeField(auto_now=True)),
+                ('month', models.DateField(help_text='First of the month the billed units are attributed to (period attribution).')),
```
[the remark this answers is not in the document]
ajeety4: Same reasoning as above.
month is set only through _billed_month() (invoice_line_items.py), which always returns get_month_start_date(...) — and that function unconditionally does value.replace(day=1) (commcare_connect/utils/datetime.py). Every CompletedWorkInvoice row is created through this path, so month.day == 1 is guaranteed by the code that constructs the value, not by MinValueValidator/full_clean().

commcare_connect/opportunity/migrations/0143_completedwork_invoiced_approved_count_and_more.py
```diff
@@ -0,0 +1,42 @@
...
+
+    dependencies = [
+        ('opportunity', '0140_opportunity_supervising_organization'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='completedwork',
+            name='invoiced_approved_count',
+            field=models.IntegerField(default=0, help_text='Approved units already billed on a live invoice.'),
```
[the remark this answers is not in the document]
ajeety4: Both invariants are already enforced at the write paths (future PRs):

billed_count >= 1 is guaranteed by saved_approved_count__gt=F("invoiced_approved_count") before creating invoice lines.
invoiced_approved_count >= 0 is enforced by rollback_invoice_line_items clamping with max(0, ...).
We previously decided against DB CheckConstraints to stay consistent with the codebase, where invariants are enforced at the write layer.

zandre-eng (approved):

hemant10yadav (approved):

commcare_connect/opportunity/migrations/0143_completedwork_invoiced_approved_count_and_more.py
```diff
@@ -0,0 +1,42 @@
...
+                ('date_modified', models.DateTimeField(auto_now=True)),
+                ('month', models.DateField(help_text='First of the month the billed units are attributed to (period attribution).')),
+                ('billed_count', models.PositiveIntegerField(help_text='Approved units billed on THIS invoice (delta).', validators=[django.core.validators.MinValueValidator(1)])),
+                ('flw_amount_local', models.DecimalField(decimal_places=2, default=0, max_digits=10)),
+                ('flw_amount_usd', models.DecimalField(decimal_places=2, default=0, max_digits=10)),
+                ('org_amount_local', models.DecimalField(decimal_places=2, default=0, max_digits=10)),
+                ('org_amount_usd', models.DecimalField(decimal_places=2, default=0, max_digits=10)),
+                ('completed_work', models.ForeignKey(on_delete=django.db.models.deletion.PROTECT, related_name='invoice_items', to='opportunity.completedwork')),
+                ('exchange_rate', models.ForeignKey(help_text="ExchangeRate row used for this row's month.", on_delete=django.db.models.deletion.PROTECT, related_name='completed_work_invoices', to='opportunity.exchangerate')),
```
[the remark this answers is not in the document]
ajeety4: See the comment [here](https://github.com/dimagi/commcare-connect/pull/1419#discussion_r3712287831)
The exchange rate value is immutable

PR #1425, opened 2026-08-04
CCCT-2526: Invoice line-item module + freeze on write (2/4)

ajeety4:
Stacked on #1419.

## Product Description

Right now an invoice can change after it has been issued. The line items are worked out fresh from the delivery data every time someone opens the invoice, so if a visit gets approved, rejected or flagged as a duplicate later, the invoice quietly starts saying something different from what was actually billed. This happens on invoices that have already been paid too.

There is a second problem. A delivery can only be attached to one invoice. So if more visits are approved after that delivery was billed, the extra units can never be billed at all — that work simply never gets paid for.

The fix is to record what was billed instead of recalculating it. Each delivery now keeps a count of how many units have already been billed. When an invoice is raised we save its line items into a table and update that count. Invoices are then read back from the saved rows, so they can't change any more. Anything approved later still counts as unbilled, so it goes onto the next invoice.

This PR on its own doesn't change anything a user sees. It is part 1 of 4, and the stack has to merge together.

### What this PR does

Adds the new module, and switches both places that create invoices over to it.

- New `utils/invoice_line_items.py`. It finds the deliveries with unbilled units, prices them (worker pay and org pay, in local currency and USD), and groups them into line items by month and payment unit.
- `bill_invoice` is the only thing that writes. It locks the rows it is about to bill, saves them, and takes the invoice amount from those same rows, so the total and the line items can't disagree. If there is nothing to bill, it writes nothing at all.
- The manual invoice form and the monthly task both go through it now.
- Deleting an opportunity now clears the saved rows first. Both foreign keys are `PROTECT`, so without this the delete fails as soon as any invoice exists.

Worth reviewing commit by commit. There are six, and each message explains why.

## Technical Summary

### Safety story

No database changes in this PR — those are in #1419. The old `CompletedWork.invoice` column is left alone and keeps its values, so if we have to roll back, the old code still finds what it expects. Dropping that column is a separate ticket ([CCCT-2556](https://dimagi.atlassian.net/browse/CCCT-2556)).

This PR is not deployable on its own. The stack goes out after #1419 and the backfill.

### Automated test coverage

`pytest commcare_connect/opportunity` — 632 passed. `prek run -a` clean.

The new `test_invoice_line_items.py` covers billing only the unbilled units, pricing including org pay, which month a late approval gets billed under, saving the rows and updating the count, and the case where there is nothing to bill. `test_deletion.py` covers deleting an opportunity that has saved rows, which failed before this fix.

### QA Plan

Nothing to QA on its own. QA is on the merged stack — acceptance criteria and risk areas shared separately.

### Labels & Review


commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,352 @@
...
+)
+from commcare_connect.utils.datetime import get_month_start_date
+
+CENTS = Decimal("0.01")
+
+
+def get_billable_completed_works_qs(opportunity, start_date, end_date):
+    """Approved works that still have unbilled units.
+
+    The watermark decides what is billable; the dates only *scope* it:
```
zandre-eng: Minor nit: Would it be possible to replace watermark with a different term? I think this may be confusing for devs reading this that are unfamiliar with how this term relates to the logic.

Feel free to ignore if you think watermark is fine.

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,352 @@
...
+def _build_billable_rows(opportunity, start_date, end_date, for_update=False):
+    """Price each billable work's delta into a BillableRow.
+
+    `end_date` is the billing cutoff — the invoice's end date — and carries two roles: it bounds
+    which first-billing works are selected, and it is the month a late delta is attributed to (a
+    late delta bypasses the bound but still needs a month to be billed under).
+
+    for_update=True is only for writes. Read-only callers must leave it False.
+    """
+    from commcare_connect.opportunity.visit_import import get_exchange_rate
```
zandre-eng: Minor nit: Could this import be moved to the top, or is it blocked by a circular import exception?

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,352 @@
...
+    delta takes the billing month, because its `status_modified_date` is frozen at the original
+    approval (it only moves on a *status* change, and a late duplicate stays `approved`).
+    """
+    if work.invoiced_approved_count == 0 and work.status_modified_date:
+        return get_month_start_date(work.status_modified_date)
+    return get_month_start_date(end_date)
+
+
+def _to_usd(amount_local: Decimal, exchange_rate: int | Decimal) -> Decimal:
+    """Round each row to cents before summing so `invoice.amount_usd` always equals
```
zandre-eng: Minor nit: I think some of these docstrings (such as this one), can be simplified or removed outright. For example the code in `_to_usd` feels self-explanatory enough that a docstring isn't needed.

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -147,6 +153,60 @@ def get_billable_line_items(opportunity, start_date, end_date):
...
+            CompletedWorkInvoice(
+                invoice=invoice,
+                completed_work=row.completed_work,
+                month=row.month,
+                billed_count=row.billed_count,
+                flw_amount_local=row.flw_amount_local,
+                flw_amount_usd=row.flw_amount_usd,
+                org_amount_local=row.org_amount_local,
+                org_amount_usd=row.org_amount_usd,
+                exchange_rate=row.exchange_rate,
```
zandre-eng, line 208: Nit: This could be compacted to the following:
```
CompletedWorkInvoice.objects.bulk_create(
    [CompletedWorkInvoice(invoice=invoice, **asdict(row)) for row in rows]
)
```

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,352 @@
...
+        "date_approved": completed_work.status_modified_date,
+        "approved_count": billed_count,
+        "flw_amount_local": flw_amount_local,
+        "org_amount_local": org_amount_local,
+        "total_amount_local": total_amount_local,
+        "total_amount_usd": total_amount_usd,
+    }
+
+
+def _line_item(
```
zandre-eng: Since this function essentially just passes most of the params into a dict with a few minor calculations, how do you feel about changing this into a dataclass instead?
```
@dataclass(frozen=True)
class LineItem:
    ...

    @property
    def total_amount_local(self):
        return self.flw_amount_local + self.org_amount_local

    @property
    def total_amount_usd(self):
        return self.flw_amount_usd + self.org_amount_usd
```

commcare_connect/opportunity/tests/test_invoice_line_items.py
```diff
@@ -0,0 +1,723 @@
+import datetime
```
zandre-eng, line 1: There are a lot of tests added in this file. I think it might be a good opportunity to see which can be merged, parameterized, or removed due to redundancy (perhaps a good task for Claude). For example, I think all the tests in `TestBillableSelection` can be parametrized together.

zandre-eng (commented):
I took an initial pass, and it generally looks good with a few minor comments for now. I will still take another pass as it's quite a big PR.

If possible, it could be nice to streamline this PR a little to make it easier for review. There are a few fix/refactor commits which could be rebased, or the PR could be split into multiple PRs using [stacked PRs](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/). Feel free to push back if you think the PR structure doesn't have much flexibility for refactoring.

ajeety4:
> I took an initial pass, and it generally looks good with a few minor comments for now. I will still take another pass as it's quite a big PR.
> 
> If possible, it could be nice to streamline this PR a little to make it easier for review. There are a few fix/refactor commits which could be rebased, or the PR could be split into multiple PRs using [stacked PRs](https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/). Feel free to push back if you think the PR structure doesn't have much flexibility for refactoring.

Thanks, Zandre, for taking an initial pass.

I did consider splitting this into multiple PRs, but decided to keep it as a single PR since it's one feature, and having everything in one place felt easier to follow.

I can definitely use separate PRs for follow-up fixes, and I'll also take another look at whether there's a sensible way to split up the core functionality.

ajeety4:
Note for reviewers: I will be breaking this PR into smaller ones. This will be ready for review post that.

ajeety4:
@zandre-eng Thanks for the suggestion. I have split this PR into smaller PRs and created a stack.

zandre-eng (approved):
Thanks for the PR refactor, this is very helpful. Overall looks good, considering that this is all going to QA. Just still my previous comments, but none blocking.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: I considered using a dataclass here. Since this is just the final display/output object and it gets converted to a dict anyway, I wasn't sure it added much value. We'd still pass the same fields into the constructor and then serialize it immediately. Happy to switch if we prefer using dataclasses here.

commcare_connect/opportunity/tests/test_invoice_line_items.py
ajeety4: Addressed in https://github.com/dimagi/commcare-connect/pull/1425/commits/353ae5bd0a09c11260a8f220bec22d438b1db0a5

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Good idea. Addressed in https://github.com/dimagi/commcare-connect/commit/31b05a25416ea6779c160c45c4c8b2265187bc67

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Yes it is blocked by a circular import exception

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Good find.
I have gone ahead and got rid of this term as it is not really needed. It represents a field 'invoiced_approved_count' so simpled replaced it. - https://github.com/dimagi/commcare-connect/pull/1425/commits/93017dd26570b4338d369385099bbce270b51d80

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,253 @@
...
+                "flw_amount_usd": Decimal(0),
+                "org_amount_usd": Decimal(0),
+                "exchange_rate": row.exchange_rate,
+            },
+        )
+        group["number_approved"] += row.billed_count
+        group["flw_amount_local"] += row.flw_amount_local
+        group["org_amount_local"] += row.org_amount_local
+        group["flw_amount_usd"] += row.flw_amount_usd
+        group["org_amount_usd"] += row.org_amount_usd
```
sravfeyn: We are converting various amounts to usd and are adding amounts in local and usd and are needing to track them separately. If we introduce `Money` class, we will be able to get rid of the this complexity and eliminate methods `total_amount_local`,  `total_amount_usd`, `_to_usd`. 

And `__add__` combined with  `usd` will help in summing up Money values directly, will avoid in needing to sum up local and usd currencies separately, and BillableRow can drop usd/local fields and use `Money` field directly. Further _line_item could become a dataclass and can drop separate usd tracking fields as well.

```
@dataclass(frozen=True)
class Money:
    local: Decimal
    usd: Decimal

    @classmethod
    def zero(cls) -> "Money":
        return cls(Decimal(0), Decimal(0))

    @classmethod
    def from_local(cls, local: Decimal, exchange_rate: Decimal | int) -> "Money":
        return cls(local=local, usd=(local / exchange_rate).quantize(Decimal("0.01")))

    def __add__(self, other: "Money") -> "Money":
        return Money(self.local + other.local, self.usd + other.usd)
```
```
@dataclass(frozen=True)
class BillableRow:
    completed_work: CompletedWork
    billed_count: int
    month: datetime.date
    flw_pay: Money
    org_pay: Money
    exchange_rate: Decimal | int

    @property
    def total(self) -> Money:
        return self.flw_pay + self.org_pay
```

```
@dataclass(frozen=True)
class LineItem:
    month: datetime.date
    payment_unit_name: str
    number_approved: int
    flw_pay: Money
    org_pay: Money
    exchange_rate: Decimal | int
    currency: str

    @property
    def total(self) -> Money:
        return self.flw_pay + self.org_pay
```

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Thanks for pointing this out. I was being somewhat conservative here since invoicing code can be difficult to follow. I made some updates and cleanup in the latest commit: https://github.com/dimagi/commcare-connect/pull/1425/commits/31b05a25416ea6779c160c45c4c8b2265187bc67

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _line_item(
+    month,
+    payment_unit_name,
+    number_approved,
+    flw_amount_local,
+    org_amount_local,
+    flw_amount_usd,
+    org_amount_usd,
```
sravfeyn: [Above suggestion](https://github.com/dimagi/commcare-connect/pull/1425/changes#r3734842514) should reduce these four fields to two. And also the total field

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+def _to_usd(amount_local: Decimal, exchange_rate: int | Decimal) -> Decimal:
+    """Round each row to cents before summing so `invoice.amount_usd` always equals
+    the sum of the stored rows. Although `DecimalField(decimal_places=2)` rounds on
+    save, leaving values at full precision in memory would allow invoice totals to
+    drift from persisted line items.
+
+    Keep the default `ROUND_HALF_EVEN` to match Django's `DecimalField`
+    quantization and existing backfilled rows.
```
sravfeyn: `nit`: While docstrings are great, I think for a simple function like this its a bit too verbose and unnecessary. I am not sure if we can tune claude not to be too verbose like this. Which makes me feel like the value of other docstrings in other functions here.

commcare_connect/opportunity/tests/test_invoice_line_items.py
sravfeyn: Plus one.

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -0,0 +1,220 @@
...
+    @property
+    def total_amount_local(self):
+        return self.flw_amount_local + self.org_amount_local
+
+    @property
+    def total_amount_usd(self):
+        return self.flw_amount_usd + self.org_amount_usd
+
+
+def _build_billable_rows(opportunity, start_date, end_date, for_update=False):
```
sravfeyn: There's an opportunity to refactor this to avoid the `for_update` flag on
`_build_billable_rows` — which also separates "get billable items" from
"invoice them."

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Great idea. Thanks
https://github.com/dimagi/commcare-connect/pull/1425/commits/e0d1e61c3424e610f933c116600337a4b4449730

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Addressed in https://github.com/dimagi/commcare-connect/pull/1425/commits/58e9fedd0eecaa2615e86dcb91cc9d6a5af01663

zandre-eng (approved):
Thanks for the changes Ajeet, looks good.

hemant10yadav (approved):
Looks Good!

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Agreed.  This was addressed in https://github.com/dimagi/commcare-connect/pull/1425/commits/31b05a25416ea6779c160c45c4c8b2265187bc67

zandre-eng (approved):

PR #1427, opened 2026-08-04
Worker Name Sync

zandre-eng:
## Product Description

When a worker changes their name in the mobile app's Manage Profile section, Connect web now
picks the new name up instead of showing the old one until the worker is re-invited to an
opportunity.

Names that already drifted before this change are repaired by a one-off management command,
so no worker has to wait for organic activity to get their name corrected. There is no new
UI and this is a back-end only change.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/QA-8605)

Connect keeps its own copy of the worker name in `users.User.name`, and until now only the invite flow ever wrote it. This means a worker name change on mobile was invisible in Connect until the worker's next invite.

This PR adds the receiving half of a push from PersonalID: a
client-authenticated `POST /users/update_profile/` endpoint. It reuses
`ClientProtectedResourceMixin`, the same OAuth2 client-credentials mechanism PersonalID already
uses for `/users/toggles/` and `/users/resend_invites/`, so no new credentials or settings are
needed on either side. Validation and the write live in a plain `update_user_profile` function
rather than in the view, keeping them unit-testable without HTTP and leaving the view to do
nothing but map the outcome to a status code. A blank or missing field returns 400 rather than
500 specifically so PersonalID's Celery task treats it as a permanent failure and stops
retrying it.

The `refresh_worker_names` command is deliberately a separate, one-shot repair rather than a
scheduled job: with the push in place there is nothing for a recurring run to find. It is
kept for reruns if pushes are ever known to have been lost.

This is one half of a cross-repo change — the [PersonalID side](https://github.com/dimagi/connect-id/pull/279) that calls this endpoint is a
separate PR. This one is inert until that ships, so it is safe to merge first.

### Safety story

The endpoint is client-authenticated and writes a single column for a single already-existing
user via a filtered `.update()` — it never creates users, never touches any other field, and
cannot fall back to a broad update. There is no schema change and no migration. Blank and
whitespace-only names are rejected at the boundary, so an empty upstream name cannot blank out
a name Connect is currently displaying. An unknown username is a no-op returning
`{"updated": false}` rather than an error. The repair command defaults to acting only on
workers who have at least one `OpportunityAccess`, supports `--dry-run`, skips names that
already match, and isolates lookup failures per batch so one bad batch reports and continues
rather than aborting the run — a rerun picks up whatever was missed. Reverting is a matter of
removing the URL; nothing about the data model changes.

### Automated test coverage

Tests cover the profile-update function for the successful rename path including whitespace
stripping, and the endpoint for the update, unknown-user and bad-request outcomes. The
management command is covered for updating only names that actually differ, skipping workers
without a phone number, and leaving the local name untouched when PersonalID returns nothing,
returns a blank name, or the lookup fails outright. Full `commcare_connect/users/` suite passes
(127 tests).

### QA Plan

This comes out of QA testing, so QA will be verifying this once both parts have been merged and deployed.

### Labels & Review


commcare_connect/users/views.py
```diff
@@ -306,6 +306,18 @@ def post(self, request, *args, **kwargs):
         return HttpResponse(status=200)
 
 
+@method_decorator(csrf_exempt, name="dispatch")
+class UpdateUserProfileView(ClientProtectedResourceMixin, View):
```
pxwxnvermx, line 310: This view can be changed to an API view with the TokenAuthentication, we can directly use the access tokens generated on PersonalID to request this view and update the user's name. That removes the need for having the server Client Credentials authentication and also keeps this action a user action instead of being a server action.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+    def _refresh_user(self, user, connectid_user, totals, dry_run):
+        if user.name == connectid_user.name:
+            totals["unchanged"] += 1
+            return
+        if dry_run:
+            self.stdout.write(f"{user.username}: {user.name!r} -> {connectid_user.name!r}")
+            totals["updated"] += 1
+            return
+        try:
+            update_user_profile(connectid_user.username, connectid_user.name)
```
ajeety4: Wanted to confirm the performance aspect here since we're updating one user at a time. I assume this shouldn't be a problem because the total number of users is relatively low. Does that sound reasonable?

pxwxnvermx: The user can be updated here directly, without calling the update_user_profile helper. It would also remove the refetch for the user which is already available here.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+        for batch in batched(phone_numbers, batch_size):
+            self._refresh_batch(batch, names_by_phone, totals, dry_run)
+
+        self._report(totals, without_phone, dry_run)
+
+    def _refresh_batch(self, batch, users_by_phone, totals, dry_run):
+        try:
+            found_users = fetch_users(batch)
+        except Exception as e:
+            # One bad batch must not abort the run; a rerun picks up whatever was missed.
```
ajeety4, line 89: Curious how the rerun picks up missed batches ? I am assuming it will still fetch for all users, but update only the missed ones . Is that right ?

ajeety4 (commented):
Looks good.
Just a minor query around the skipped updates.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+
+        phone_numbers = sorted(names_by_phone)
+        totals = {"checked": len(phone_numbers), "updated": 0, "unchanged": 0, "skipped": 0, "not_found": 0}
+        for batch in batched(phone_numbers, batch_size):
+            self._refresh_batch(batch, names_by_phone, totals, dry_run)
+
+        self._report(totals, without_phone, dry_run)
+
+    def _refresh_batch(self, batch, users_by_phone, totals, dry_run):
```
pxwxnvermx: The users_by_phone, totals and dry_run parameters here can be instantiated on the class and then used inside the functions. This will remove the overhead of passing them down the function chains.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+            return
+
+        found_phone_numbers = set()
+        for connectid_user in found_users:
+            found_phone_numbers.add(connectid_user.phone_number)
+            for user in users_by_phone.get(connectid_user.phone_number, []):
+                self._refresh_user(user, connectid_user, totals, dry_run)
+        totals["not_found"] += len(set(batch) - found_phone_numbers)
+
+    def _refresh_user(self, user, connectid_user, totals, dry_run):
```
pxwxnvermx: The totals and dry_run parameters can be moved to the Class instance and then used with the `self.totals` and `self.dry_run` variables.

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: That is correct. This command is idempotent so it can be re-run without risk if needed. For batches that have already updated we'll need to fetch from PersonalID again to compare the name there to the one in Connect, but if the two names are the same then no DB write will be done for that row in Connect.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+        if user.name == connectid_user.name:
```
hemant10yadav: We should skip this iteration if user.username doesn't match connectid_user.username. A local user can have multiple records with the same phone number, while ConnectID allows only one active account per phone number. Since we already know the username don't match, we can avoid the unnecessary update_user_profile() query and keep the updated count accurate.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+
+    def handle(self, *args, **options):
+        batch_size = options["batch_size"]
+        if batch_size < 1:
+            raise CommandError("--batch-size must be at least 1.")
+
+        dry_run = options["dry_run"]
+        users = User.objects.filter(opportunityaccess__isnull=False).distinct()
+        names_by_phone = {}
```
hemant10yadav: Could we move the phone number filter into the queryset instead of filtering in Python? That avoids loading web users that we already know won't be processed.

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: This is a good point. I've gone ahead and added batching to the command as a safety precaution. Even if there aren't a lot of updates it doesn't hurt to add batching so that we don't unnecessarily bombard the DB with individual writes.

Addressed in [29bc36f](https://github.com/dimagi/commcare-connect/pull/1427/commits/29bc36f080d7f0d7663796f132125105acd74d30).

ajeety4 (approved):

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,126 @@
...
+        to_update = []
+        for connectid_user in found_users:
+            found_phone_numbers.add(connectid_user.phone_number)
+            for user in users_by_phone.get(connectid_user.phone_number, []):
+                if self._stage_update(user, connectid_user.name, totals, dry_run):
+                    to_update.append(user)
+        totals["not_found"] += len(set(batch) - found_phone_numbers)
+
+        if to_update:
+            User.objects.bulk_update(to_update, ["name"], batch_size=UPDATE_BATCH_SIZE)
```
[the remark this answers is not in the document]
zandre-eng: Not fixing. This is real, but negligible. The only lossy window would be between a batch's `fetch_users` and its `bulk_update`, which practically should be a sub-second.

hemant10yadav (commented):
Looks good to me. Just a few minor suggestions.

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: Good catch. Added a validation check for username in [59bea5a](https://github.com/dimagi/commcare-connect/pull/1427/commits/59bea5ae3dfe1dcfc33139b453e96658e2741007).

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: Sure, I've changed the count to use a DB query instead in [59bea5a](https://github.com/dimagi/commcare-connect/pull/1427/commits/59bea5ae3dfe1dcfc33139b453e96658e2741007).

pxwxnvermx (commented):
I left a comment about changing the authentication that is used for the UpdateUserProfileView to be more user focused than server focused. As we use Token Introspection the access tokens generated on PersonalID work on Connect, so we directly do API requests from PersonalID using the users access token to update the profile as well.

commcare_connect/users/views.py
zandre-eng: Currently, the PersonalID related `update_profile` endpoint is being passed basic auth from mobile, so we can't use a token directly from the user's auth. We'll need to first create a short-lived token for the user which we can then use to auth with Connect. 

@pxwxnvermx Is this what you had in mind?

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: This seems to be outdated code. I have since changed the approach to use a `bulk_update` for all the users that have updated names ([29bc36f](https://github.com/dimagi/commcare-connect/pull/1427/commits/29bc36f080d7f0d7663796f132125105acd74d30)).

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: Addressed in [45aecec](https://github.com/dimagi/commcare-connect/pull/1427/commits/45aecec0fb6e39aaf5f02b2bf01b86d3efd4c888).

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: Addressed in [45aecec](https://github.com/dimagi/commcare-connect/pull/1427/commits/45aecec0fb6e39aaf5f02b2bf01b86d3efd4c888).

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def _stage_update(self, user, new_name):
+        """Set ``user.name`` in memory, returning whether it needs saving."""
+        new_name = (new_name or "").strip()
+        if not new_name:
+            # ConnectID allows a blank name; leave the existing local one alone.
+            self.stderr.write(f"Skipping {user.username}: name is blank in ConnectID")
```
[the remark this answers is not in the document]
zandre-eng: This is a once-off repair script. The naming is fine given the code's expected lifetime.

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,142 @@
...
+            self.totals["skipped"] += 1
+            return False
+        if len(new_name) > NAME_MAX_LENGTH:
+            # ConnectID stores the name in a TextField, so it can hold more than this column accepts.
```
sravfeyn: Should we not make connect one also a textfield as part of this work?

commcare_connect/users/management/commands/refresh_worker_names.py
```diff
@@ -0,0 +1,142 @@
...
+from commcare_connect.users.models import User
+from commcare_connect.utils.itertools import batched
+
+DEFAULT_BATCH_SIZE = 100
+# --batch-size sizes the PersonalID lookup; cap the UPDATE separately so a large lookup batch
+# doesn't turn into one enormous CASE WHEN statement.
+UPDATE_BATCH_SIZE = 500
+
+
+class Command(BaseCommand):
```
sravfeyn, line 37: Do you plan to cleanup the command and tests post the run?

commcare_connect/users/helpers.py
```diff
@@ -2,10 +2,29 @@
...
 
 
+class InvalidProfileUpdate(Exception):
+    pass
+
+
+NAME_MAX_LENGTH = User._meta.get_field("name").max_length
+
+
+def update_user_profile(username, name):
```
sravfeyn: `nit`: This is a one time use method, so it best belongs on the view as a sub method (could even be a static method)

commcare_connect/users/views.py
```diff
@@ -306,6 +306,18 @@ def post(self, request, *args, **kwargs):
...
+    """Called by PersonalID when a user changes their profile."""
+
+    def post(self, request, *args, **kwargs):
+        try:
+            updated = update_user_profile(request.POST.get("username"), request.POST.get("name"))
+        except InvalidProfileUpdate:
```
sravfeyn: `nit` I think this can be managed without introducing a new exception class by refacotoring the above logic

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: That was my intent, yes. After the initial run this command shouldn't be necessary anymore as the auto sync between PersonalID and Connect should take care of any new name updates. 

This command could be kept around in case we need to ever do a manual sync (e.g. the auto sync fails), but I feel this will be keeping around unnecessary code for an unlikely scenario.

commcare_connect/users/management/commands/refresh_worker_names.py
zandre-eng: After thinking through this some more, I agree we should make the change here to bring consistency between Connect and PersonalID.

Addressed in [a9c683b](https://github.com/dimagi/commcare-connect/pull/1427/commits/a9c683bcbf4f07a4d54a2b544dbc4c95edda6535).

commcare_connect/users/helpers.py
zandre-eng: I've completely removed the helper method and the logic now happens in the view function itself.

Addressed in [534c4d2](https://github.com/dimagi/commcare-connect/pull/1427/commits/534c4d287924f334c465818736e40f0bfdbcec79).

commcare_connect/users/views.py
zandre-eng: I have refactored the logic, and so exception classes are not used anymore.

Addressed in [534c4d2](https://github.com/dimagi/commcare-connect/pull/1427/commits/534c4d287924f334c465818736e40f0bfdbcec79).

commcare_connect/users/migrations/0029_alter_user_name.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ('users', '0028_create_clear_expired_oauth_tokens_periodic_task'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
+            name='name',
+            field=models.TextField(blank=True, verbose_name='Name of User'),
+        ),
```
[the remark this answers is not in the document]
zandre-eng: Skipping. This change brings consistency between PersonalID and Connect, so any rollback strategy will also need to consider the name lengths stored on PersonalID.

commcare_connect/users/views.py
pxwxnvermx: My initial though was that the update_profile endpoint used TokenAuthentication that would mean we can directly use that token to also do a request to Connect. But if that's not the case we can skip this and keep the current auth.

pxwxnvermx (approved):

PR #1431, opened 2026-08-06
CCCT-2526: Read invoice line items from the saved rows (3/4)

ajeety4:
Stacked on #1425.

## Product Description

See #1425 for the problem and the approach.

Part 1 saves the line items when an invoice is raised. This part reads them back, which is what actually stops an issued invoice from changing. Part 2 of 4, and the stack has to merge together.

## What this PR does

Part 1 saves the line items. This PR starts reading them back.

- The invoice review page and the PDF now read the saved rows instead of recalculating from the delivery data. An approval that lands after the invoice was issued can no longer move its line items.
- The line-item preview on the create form now uses the same code as billing, so what the NM sees before saving is what gets billed.
- The deliveries CSV picks its source: saved rows for an issued invoice, live figures for a preview. Without this the invoice would stop drifting while the export behind it carried on drifting.
- Deletes the old helpers that read through the invoice link, and moves their last tests over to the new module.

Cancelling and rejecting still clears the old invoice link, which nothing writes any more — so on this branch a cancel would release nothing. Part 3 fixes that.

Worth reviewing commit by commit. There are four, and each message explains why.

## Technical Summary

### Safety story

No database changes. The old `CompletedWork.invoice` column is left alone and keeps its values, so a rollback lands on data the old read path still understands.

Invoices issued before this goes out have no saved rows yet, which is what the backfill is for. That is why the stack can only be deployed after #1419 and the backfill, in that order.

This PR is not deployable on its own.

### Automated test coverage

`pytest commcare_connect/opportunity` — 635 passed. `prek run -a` clean.

Covers reading the saved rows grouped by month and payment unit, an invoice not changing when the delivery counts are recalculated afterwards, an invoice with no rows returning nothing, and the CSV rows for both an issued invoice and a preview. The tests that used to drive the deleted helpers now drive the new code and keep the same coverage of multi-month grouping and per-month exchange rates.

### QA Plan

Nothing to QA on its own. QA is on the merged stack — acceptance criteria and risk areas shared separately.

### Labels & Review


commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -164,6 +164,78 @@ def get_billable_line_items(opportunity, start_date, end_date):
     return group_line_items(_build_billable_rows(works, opportunity.currency_code, end_date))
 
 
+def get_invoice_line_items(invoice):
+    records = (
+        invoice.work_items.values("completed_work__payment_unit", "month")
```
hemant10yadav, line 170: Should we include exchange_rate__rate in the grouping instead of using Max() here? Since the rate is expected to be consistent for a (month, payment_unit) group, grouping by it would make that assumption explicit and avoid silently picking one rate if the data is inconsistent.

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -164,6 +164,78 @@ def get_billable_line_items(opportunity, start_date, end_date):
...
+            flw_pay=Money(record["flw_local"], record["flw_usd"]),
+            org_pay=Money(record["org_local"], record["org_usd"]),
+            exchange_rate=record["rate"],
+        )
+        for record in records
+    ]
+
+
+@dataclass(frozen=True)
+class DeliveryRow:
```
hemant10yadav: **Nit:** `BillableRow` and `DeliveryRow` can be merged.

`DeliveryRow` is just `BillableRow` minus `exchange_rate`. Nothing that consumes a delivery row uses that field, so it seems safe to drop the second class.

`get_billable_delivery_rows()` then only exists to convert `BillableRow → DeliveryRow`. With a single type, it could be simplified to:

```python
def get_billable_delivery_rows(opportunity, start_date, end_date):
    works = get_billable_completed_works_qs(opportunity, start_date, end_date)
    return _build_billable_rows(works, opportunity.currency_code, end_date)
```

Similarly, `get_invoice_delivery_rows()` could return `BillableRow` directly. It already has `item.exchange_rate` available, we'd just need to add it to the `select_related`.

 Let me know if I'm missing something.

hemant10yadav (commented):
Overall, looks good. I just have one question.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Thanks. Good point.
Addressed in https://github.com/dimagi/commcare-connect/pull/1431/commits/e38f754649f7a13c70c3d467002fe8ecb50604f0 .
FYI, I also renamed the class to be generic and caller agnostic.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: I agree that grouping by `exchange_rate__rate` would make the consistency assumption explicit, but it would also change the grouping/cardinality to (month, payment_unit, rate). If there were inconsistent rates, we'd get multiple rows instead of the intended one row per (month, payment_unit).

Validation would be good to have in principle, but there is currently only one write path, and it inserts the exchange rate month-wise, so the invariant is enforced at the point where the data is created. Given that, I think keeping Max() here is reasonable rather than adding validation to the read path.

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -164,6 +164,57 @@ def get_billable_line_items(opportunity, start_date, end_date):
...
+            number_approved=record["number_approved"],
+            flw_pay=Money(record["flw_local"], record["flw_usd"]),
+            org_pay=Money(record["org_local"], record["org_usd"]),
+            exchange_rate=record["rate"],
+        )
+        for record in records
+    ]
+
+
+def get_invoice_delivery_rows(invoice):
```
zandre-eng: Nit: What do you think about renaming this to something like `get_invoice_delivery_rows_for_export`? Initially, looking at the changes here I was a little unsure what the difference in use case was between this function and `get_invoice_line_items`, as they both fetch line items and return slightly different data.

zandre-eng (approved):
Overall looks good, just one minor nit query.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Good point. Addressed in https://github.com/dimagi/commcare-connect/pull/1431/commits/ec53f481d68222f06594877a509e2691a9946876

zandre-eng (approved):
Thanks for the rename, looks good!

Charl1996 (approved):

hemant10yadav (approved):

PR #1432, opened 2026-08-06
CCCT-2526: Release line items when an invoice is cancelled or rejected (4/4)

ajeety4:
Stacked on #1431.

## Product Description

See #1425 for the problem and the approach.

This part makes cancelling and rejecting work again, and completes the core change — from here the stack is functionally whole. Part 4 of 4.

## What this PR does

Cancelling or rejecting a service-delivery invoice has to give its deliveries back, otherwise they stay marked as billed and can never be invoiced again.

- Cancelling or rejecting now deletes the invoice's saved rows and reduces each delivery's billed count by what *that* invoice billed.
- Replaces the old code that cleared the invoice link. 
- Adds a test that nothing writes the old invoice link, on both creating and cancelling. (good to have)

Worth reviewing commit by commit. 

## Technical Summary

### Decisions

1. Add an `is_delta field ` to the CompletedWorkInvoice snapshot row to identify whether a delivery is initial billing or an additional (delta) billing.
    
      **Desired Rule**
      The first delivery for a work should always use the delivery month as the billing month.
      Additional deliveries that comes after fist billing should use the billing month in which the delta is processed.
      
      **Reason**
      - Consider a work completed in January with an approved count of 1. The January invoice bills the first delivery. In February, a new delivery arrives, increasing the count to 2, and the February invoice bills the additional delivery as a delta.
      - If the January invoice is later cancelled and recreated for January–March, the system will incorrectly treat the January delivery as a delta and assign March as its billing month.

      Storing is_delta lets us distinguish the February delivery as a delta and correctly identify the January delivery as the initial billing, preserving January as its billing month.

2.  The model change and migration are included in this PR because the edge case requiring this field arises from the rollback/cancellation functionality.

### Safety story

No database changes. The old `CompletedWork.invoice` column is left alone and keeps its values, so a rollback lands on data the old read path still understands.

Deploy order for the stack is unchanged: #1419, then the backfill, then this.

### Automated test coverage

`pytest commcare_connect/opportunity` — 639 passed. `prek run -a` clean.

Covers the billed count going back down by the right amount, an earlier invoice covering the same delivery being left alone, the delivery becoming billable again afterwards, and the old invoice link never being written on either create or cancel.

### QA Plan

Nothing to QA on its own. QA is on the merged stack — acceptance criteria and risk areas shared separately. The cancel and re-raise cases are the ones that exercise this PR.

### Labels & Review

[CCCT-2530]: https://dimagi.atlassian.net/browse/CCCT-2530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Charl1996:
FYI @ajeety4 The description says this is "Part 3 of 4", but the title says "4/4".

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -13,27 +14,31 @@
...
 
-    `invoiced_approved_count` decides what is billable; the dates only *scope* it:
+    `saved_approved_count > invoiced_approved_count` decides what is billable.
 
-    - first-billing works (`invoiced_approved_count == 0`) are scoped by the window, where their
-      `status_modified_date` is meaningful;
-    - late deltas (already partly billed) bypass the window entirely. A late duplicate keeps the
-      work at `approved`, so its `status_modified_date` never moves off the original approval;
-      windowing that stale date would silently defer a delta that must bill now.
+    - a work with no first billing is scoped by the window, where its `status_modified_date` is
```
Charl1996: Nit: I find this docstring slightly hard to comprehend. Any chance you can dumb it down?

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -234,6 +258,31 @@ def bill_invoice(invoice, start_date, end_date):
...
 
+def rollback_invoice_line_items(invoice):
+    with transaction.atomic():
+        billed_by_work = dict(invoice.work_items.values_list("completed_work_id", "billed_count"))
+        if not billed_by_work:
+            return
+
+        works = []
+        for work in CompletedWork.objects.select_for_update(of=("self",)).filter(id__in=billed_by_work):
+            released = billed_by_work[work.id]
```
Charl1996: nit: rename `released` to `billed_count_released`?

commcare_connect/opportunity/models.py
```diff
@@ -977,6 +977,9 @@ class CompletedWorkInvoice(BaseModel):
         related_name="completed_work_invoices",
         help_text=gettext_lazy("ExchangeRate row used for this row's month."),
     )
+    # True when this billing represents additional approved work
+    # obtained after the initial billing.
+    is_delta = models.BooleanField(default=False)
```
Charl1996, line 982: I can't say I feel confident as to why this field is needed?

The PR description talks about this field being needed in case an invoice is cancelled and recreated, but when it's cancelled don't we delete the CompletedWorkInvoice records?

(maybe there's something obvious I'm missing)

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Good point. 
Addressed in https://github.com/dimagi/commcare-connect/pull/1432/commits/9316ae7e1505a4b248d8191512feb58b05b17a5c.  Hope this is much clearer.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1432/commits/9316ae7e1505a4b248d8191512feb58b05b17a5c

commcare_connect/opportunity/models.py
ajeety4: This is required for a specific edge case mentioned in the PR description, specifically in the **Reason** section of point 1 of the Decision.
Let me add a bit more context _(actually all)_ to explain the full situation.

**The key point** is that the first billing captures the approval date, while later duplicates (late deltas in context of billing) do not and only update the approval count.

**Desired Rule**: The first billing is scoped to the invoice window, while a late delta is billed in the next invoice.

**How do we identify a delta?**
invoiced approved count > 0 currently indicates that the work was billed before. However, this fails in the following case:

**The Edge case:**  

- January: Work added for first time. approved count = 1 → January invoice bills it → invoiced count = 1.
- February: New duplicate delivery arrived. approved count = 2 → February invoice bills the additional delivery → invoiced count = 2.
- January invoice is cancelled → its CompletedWorkInvoice records are deleted and invoiced count is reduced by 1 → approved count = 2, invoiced count = 1.
- A January–April invoice is now created. Since approved count > invoiced count, there is work to bill. But since invoiced count > 0, the system incorrectly treats it as a late delta and bills the original January work in April.

So, invoiced count alone cannot tell us whether the first billing has actually happened.
This is why is_delta was added:

```
Delta billing → is_delta = true
If there is no CompletedWorkInvoice with is_delta = false, the first billing has not happened yet.
```

In short, invoiced count tells us whether something was invoiced, while is_delta tells us whether the first billing has happened.

Charl1996: Ok, that makes sense. Thanks

Charl1996 (approved):
Overall looks good.

Charl1996 (approved):

PR #1433, opened 2026-08-06
Invoice create/review form fixes

ajeety4:
## Product Description
Follow up PR to the invoice line items work

Fixes to the invoice form and the invoice page, found while working on the line-item change in #1425. Most are pre-existing bugs(_minor nothing critical_) rather than anything the new billing code introduced.

## What this PR does

Fixes:
- A failed line-item request now shows an error banner with a Retry button, clears the stale amount and table, and disables Submit.
image

- An invalid submit always shows why. The view was building and validating a second copy of the form while rendering, so the errors shown came from a second database read — and if the data changed back in between, no error was shown at all.
- Rejects an end date that is today or in the future. Billing can only include completed days, so yesterday is the latest allowed date.
- Shows a 0 amount properly instead of a blank field for invoices with zero amounts _(today the system allows creating it if there are no deliveries).
-  One task fix: if a single opportunity fails during the monthly invoicing run, the run now carries on. Before, one failure stopped everything, and the invoices that had already been created were never announced.

Improvements:
- Rejects a total that no longer matches the period. If the deliveries moved between opening the form and submitting it, the NM gets a clear error and the corrected figures rather than an invoice for a number they never saw.
- Replaces the empty line-item table with a message saying which case it is: released because the invoice was cancelled or rejected, or nothing was billable for that period. This allows to differentiate between the two scenarios, instead of user assuming something.

 **In case the invoice was cancelled**
image

**No deliveries found**
image

Worth reviewing commit by commit.

## Technical Summary

### Safety story

No database changes. Everything here is either a form validation, a template change or an error-handling change, so a defect is a code revert.

The riskiest one is rejecting a total that no longer matches the period, because it can block a submit. It only triggers when the deliveries genuinely changed since the form was opened, and the page re-fetches so the NM can submit the corrected figures straight away.

Two of these need the earlier parts of the stack (the total check and the monthly task fix). The rest apply on their own.

### Automated test coverage

`pytest commcare_connect/opportunity` — 652 passed. `prek run -a` clean.

Covers the window rules (period must be finished, end after start, no future end), the posted total never being trusted, a stale total being rejected, an invalid submit rendering its own errors once, a zero invoice showing 0 rather than a blank, the empty-table message for both cases, and one opportunity failing without stopping the monthly run.

### QA Plan
QA to be considered as it has some UI changes.

### Deployment

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/forms.py
+++ b/commcare_connect/opportunity/forms.py
@@ -1992,64 +1992,9 @@ class AutomatedPaymentInvoiceForm(forms.ModelForm):
 
     @property
     def line_items(self):
-        if self.line_items_table:
-            table = HTML(
-                """
-                {% load django_tables2 %}
-                <div class="overflow-x-auto mb-4">
-                    {% render_table form.line_items_table %}
-                </div>
-                """
-            )
-        else:
-            table = HTML(
-                """
-                {% load i18n %}
-                <div x-show="lineItemsError"
-                  x-cloak
-                  role="alert"
-                  class="flex items-center justify-between p-4 mb-4 rounded-lg border-[0.5px]
-                         bg-message-error text-message-error-text border-message-error-border">
-                  <div class="flex items-center gap-2">
-                    <i class="fa-solid fa-circle-exclamation"></i>
-                    <span>
-                      {% translate "Could not load the deliveries for this period. Please try again, and contact support if the problem persists." %}
-                    </span>
-                  </div>
-                  <button type="button"
-                          class="button button-md outline-style ml-4"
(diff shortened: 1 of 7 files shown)


Charl1996:
> A failed line-item request now shows an error banner with a Retry button

What did it do before?

Charl1996 (approved):
Looks good overall. Thanks for improvements, they seem really useful.

ajeety4:
> > A failed line-item request now shows an error banner with a Retry button
> 
> What did it do before?

No action. The table was empty and invoice could still be submitted.

PR #1437, opened 2026-08-06
Prefetch claim limits in the opportunity user data export

Charl1996:
## Product Description

No user-facing change. The opportunity user data export returns the same payload with far fewer queries.

## Technical Summary

`OpportunityUserDataSerializer.get_claim_limits` ran one query per user row, plus one per claim limit (`payment_unit_id` uses `source="payment_unit.payment_unit_id"`, which lazy-loads the FK). The serializer is built per object on both export paths — the v1.0 CSV stream and the v2.0 page — so this was an N×M+N.

Now the view prefetches the claim limits with their payment units, and the serializer reads that cache. The claim itself is `select_related` — it's already joined for the `date_claimed` annotation, so that's free and saves the prefetch a round trip. A 3 user × 2 payment unit export goes from **10 queries to 2**, flat in both dimensions.

Note: the shared `OpportunityClaimLimitSerializer(many=True)` has to live at module level — as a class attribute, DRF's metaclass would collect it as a declared field.

### Safety story

Read-only change: no writes, no migrations, no API contract change. Output is identical; only the queries change.

### Automated test coverage

New `TestOpportunityUserDataView` covers query counts on both the v2.0 and CSV paths (the CSV test also guards the `chunk_size` that `.iterator()` needs to keep prefetches), plus output shape and the no-claim case. The query-count test fails against the old code with `Expected to perform 2 queries but 10 were done`. Full `pytest commcare_connect/data_export`: 80 passed.

### QA Plan

Covered by automated tests; no separate QA ticket.

### Labels & Review

### DIFF
--- a/commcare_connect/data_export/serializer.py
+++ b/commcare_connect/data_export/serializer.py
@@ -18,7 +18,6 @@ from commcare_connect.opportunity.models import (
     CompletedWork,
     LabsRecord,
     Opportunity,
-    OpportunityClaimLimit,
     Payment,
     PaymentInvoice,
     TaskType,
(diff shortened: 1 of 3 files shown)


zandre-eng (approved):

pxwxnvermx (approved):

Charl1996:
Closing as this PR is not needed anymore.

PR #1439, opened 2026-08-06
Add write APIs for work areas, groups, and implementation areas

hemant10yadav:
## Product Description

No User facing chnages.

## Technical Summary

[CCCT-2650](https://dimagi.atlassian.net/browse/CCCT-2650)

- Bulk-create endpoints convert the JSON payload into an in-memory CSV and hand it to the same importer used by the existing CSV upload flow, so validation lives in one place instead of being duplicated.
- All new endpoints are gated behind the microplanning waffle flag, scoped per opportunity.

### Safety story

Ran the full data_export, microplanning, and organization test suites locally (452 tests, all passing) after each change. The bulk-create endpoints reuse the same importer and validation code already exercised by the CSV upload flow, so there's no new, untested path for parsing or persisting these models.

### Automated test coverage

Added tests for every new endpoint covering permission checks, field validation, HQ sync/rollback, and the assign/unassign flows; all passing.

### QA Plan

No QA ticket.

### Labels & Review

[CCCT-2650]: https://dimagi.atlassian.net/browse/CCCT-2650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/data_export/serializer.py
```diff
@@ -435,3 +475,141 @@ class Meta:
...
+            elif has_access_key:
+                # Explicit `opportunity_access: null` — unassign rather than silently ignore.
+                to_unassign.append(instance)
+            elif item and instance.opportunity_access_id:
+                needs_hq_resync.append(instance)
+
+            instances.append(instance)
+
+        if touched_fields:
+            WorkArea.objects.bulk_update(instances, fields=list(touched_fields))
```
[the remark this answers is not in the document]
hemant10yadav: Thanks for flagging this. It's a real race, but re-writing a field to its already-current value is harmless — no corruption, just a redundant write — and every change is already tracked via the audit trail. Splitting bulk_update into N calls per distinct field-set would add ongoing complexity for a narrow, low-probability window on an admin-only endpoint

commcare_connect/data_export/views.py
```diff
@@ -71,20 +82,68 @@
...
+
+class BulkListMixin:
+    """Forces list input and caps batch size, so an unbounded payload can't blow up memory
+    or DB pressure. DRF's ListSerializer already supports `max_length` natively. Shared by
+    bulk-create and bulk-update views."""
+
+    def get_serializer(self, *args, **kwargs):
+        kwargs["many"] = True
+        kwargs["max_length"] = BULK_MAX_ITEMS
+        return super().get_serializer(*args, **kwargs)
```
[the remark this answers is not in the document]
hemant10yadav: Checked this against the actual 3.14.0— ListSerializer already supports max_length/min_length in 3.14.0 No version bump needed.

commcare_connect/data_export/views.py
```diff
@@ -643,3 +702,153 @@ def check_permissions(self, request):
...
+                {
+                    "results": serializer.data,
+                    "unassign_skipped": unassign_result["skipped"],
+                    "unassign_failed_ids": unassign_result["failed_ids"],
+                }
+            )
+        return Response(serializer.data)
+
+
+class CSVImporterBulkCreateView(APIView):
```
hemant10yadav, line 778: Reused the existing `CSVImporter` by serializing the JSON payload to CSV rather than introducing a separate JSON import path. This keeps all validation and post-import logic in one place and avoids the two paths drifting over time. Although the ticket expects payloads of around 100 rows, I've set the limit to 500 since the importer should comfortably handle that as well.

commcare_connect/organization/decorators.py
```diff
@@ -35,6 +35,27 @@ def user_is_program_manager_by_slug(user, org_slug):
     ).exists()
 
 
+def user_is_opportunity_admin(user, opportunity):
```
ajeety4, line 38: Checking if superuser should also be considered here  as  we used to do it earlier?

commcare_connect/data_export/serializer.py
```diff
@@ -492,3 +495,105 @@ def validate_name(self, value):
...
+            "boundary",
+            "case_properties",
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        opportunity = self.context["view"].opportunity
+
+        self.fields["id"].queryset = WorkArea.objects.filter(opportunity=opportunity).select_related(
```
ajeety4: nit: Checking if we remove could the select_related("opportunity", "work_area_group") here? In the create() flow, opportunity is taken from self.context["view"].opportunity, and work_area_group is accessed via work_area_group_id, so neither related object appears to be accessed from the WorkArea instance.

commcare_connect/data_export/views.py
```diff
@@ -748,8 +748,13 @@ def patch(self, request, *args, **kwargs):
                 status=status.HTTP_502_BAD_GATEWAY,
             )
 
-        response_data = serializer.data
         unassign_result = getattr(serializer, "unassign_result", None)
         if unassign_result:
-            response_data = {"results": response_data, "unassign_result": unassign_result}
-        return Response(response_data)
+            return Response(
```
ajeety4, line 768: :+1:

ajeety4 (approved):
Looks good to me. :+1:

commcare_connect/organization/decorators.py
hemant10yadav: `has_perm()` checks `self.is_superuser` first, so it works for superusers as well.

commcare_connect/data_export/serializer.py
hemant10yadav: Good catch on `opportunity` — it's unused (not in the output fields, and create() pulls it from context). Kept `work_area_group` though, since it's a PrimaryKeyRelatedField in the output and dropping it would cause an N+1 when the response serializes each instance's work_area_group. Fixed in 64924b8.

pxwxnvermx (approved):

commcare_connect/organization/decorators.py
ajeet2040: Ah right. Thanks

PR #1448, opened 2026-08-12
django 5.2 LTS

hemant10yadav:
## Product Description

No user-facing change.

## Technical Summary

[CCCT-2471](https://dimagi.atlassian.net/browse/CCCT-2471)

- Django, django-allauth, django-tables2, and django-redis/redis move together in this PR because django-tables2 and allauth can't support both the old and new Django versions at once.

### Safety story

Ran the full test suite after the upgrade. Also tested the signup/login flow, including social login since there was a related fix. Tested the query_string replacement as well, and everything is working fine.

### Automated test coverage

Added tests for the OAuth2 login flow, which had none before, updated a few existing tests for allauth's stricter checks, and the full suite passes.

### QA Plan

No QA ticket, tested manually on local and on staging

### Labels & Review

[CCCT-2471]: https://dimagi.atlassian.net/browse/CCCT-2471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/settings/production.py
```diff
@@ -62,12 +62,8 @@
...
 AWS_S3_CUSTOM_DOMAIN = env("DJANGO_AWS_S3_CUSTOM_DOMAIN", default=None)
 aws_s3_domain = AWS_S3_CUSTOM_DOMAIN or f"{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com"
-# STATIC
-# ------------------------
-STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 # MEDIA
 # ------------------------------------------------------------------------------
-DEFAULT_FILE_STORAGE = "commcare_connect.utils.storages.MediaRootS3Boto3Storage"
 MEDIA_URL = f"https://{aws_s3_domain}/media/"
 STORAGES["default"]["BACKEND"] = "commcare_connect.utils.storages.MediaRootS3Boto3Storage"  # noqa: F405
```
hemant10yadav, line 68: `STATICFILES_STORAGE` and `DEFAULT_FILE_STORAGE` were removed in Django 5.1, not just deprecated: [Django 5.1 release notes → "Features removed in 5.1"](https://docs.djangoproject.com/en/5.1/releases/5.1/#features-removed-in-5-1). Django now uses `STORAGES['default']['BACKEND']`, which is already configured here, so removing the legacy settings.

config/settings/base.py
```diff
@@ -306,9 +307,8 @@
 # django-allauth
 # ------------------------------------------------------------------------------
 ACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_ACCOUNT_ALLOW_REGISTRATION", True)
-ACCOUNT_AUTHENTICATION_METHOD = "username_email"
-ACCOUNT_EMAIL_REQUIRED = True
-ACCOUNT_USERNAME_REQUIRED = False
+ACCOUNT_LOGIN_METHODS = {"username", "email"}
```
hemant10yadav, line 310: This is a 1:1 replacement, not a policy change. The previous configuration:

```python
ACCOUNT_AUTHENTICATION_METHOD = "username_email"
ACCOUNT_EMAIL_REQUIRED = True
ACCOUNT_USERNAME_REQUIRED = False
```

is equivalent to the new settings. Per [allauth 65.4.0 release notes](https://docs.allauth.org/en/latest/release-notes/2025.html#id38), `ACCOUNT_AUTHENTICATION_METHOD` was replaced by `ACCOUNT_LOGIN_METHODS` in a backwards-compatible change. See the [full settings reference](https://docs.allauth.org/en/latest/account/configuration.html)

commcare_connect/templates/organization/organization_home.html
```diff
@@ -58,7 +58,7 @@ <h2 class="text-xl font-semibold text-brand-deep-purple mb-4">{% trans 'Workspac
       <div x-show="selectedTab === 'members'"
            id="invoice_members_container"
            class="w-full"
-           hx-get="{% url 'organization:org_member_table' request.org.slug %}?{% querystring %}"
+           hx-get="{% url 'organization:org_member_table' request.org.slug %}{% querystring_replace %}"
```
hemant10yadav, line 61: Two changes here:

1. **`{% querystring %}` → `{% querystring_replace %}`** — `django-tables2` 3.0.0 removed the old tag entirely, so it would cause a template render error. The same rename is applied to the other 3 templates in this PR. See the [3.0.0 changelog](https://github.com/jieter/django-tables2/blob/master/CHANGELOG.md)
2. **Removed the literal `?`** — `querystring_replace` already includes `?`, so the old code was generating `??` in the URL. This fixes that pre-existing issue as well.

config/settings/base.py
```diff
@@ -159,6 +159,7 @@
     "django_htmx.middleware.HtmxMiddleware",
     "django.middleware.csrf.CsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
+    "allauth.account.middleware.AccountMiddleware",
```
hemant10yadav, line 162: Now mandatory — allauth raises `ImproperlyConfigured` on startup unless `allauth.account.middleware.AccountMiddleware` is added to `settings.MIDDLEWARE`. This requirement was introduced in allauth 0.56.0. See the [allauth 0.56.0 release notes](https://docs.allauth.org/en/latest/release-notes/2023.html#id15)

commcare_connect/commcarehq_provider/views.py
```diff
@@ -3,19 +3,24 @@
...
+        # Deferred import avoids a circular import with provider.py, which imports this class.
+        from .provider import CommcareHQProvider
+
+        self.provider_id = CommcareHQProvider.id
+        super().__init__(request)
+
     def complete_login(self, request, app, token, **kwargs):
-        response = httpx.get(self.profile_url, headers={"Authorization": f"Bearer {token}"})
+        # allauth passes a SocialToken instance here (not the raw string) since 65.x.
+        response = httpx.get(self.profile_url, headers={"Authorization": f"Bearer {token.token}"})
```
hemant10yadav, line 23: This is the important one: the allauth 65.x change to `complete_login()` now passes a `SocialToken` instead of the raw token string. Using `str(token)` produces `"social application token (None)"`, causing HQ login to send an invalid `Authorization` header and return 401.

Fixed by using `token.token` and added a regression test in `commcarehq_provider/tests/test_views.py` to verify the actual token is sent. See the [allauth source](https://github.com/pennersr/django-allauth/blob/65.19.0/allauth/socialaccount/providers/oauth2/views.py#L54)

commcare_connect/users/adapters.py
```diff
@@ -48,7 +47,11 @@ def _next_url_from_state(request):
         return next_url if is_safe_redirect_url(request, next_url) else None
 
     def pre_social_login(self, request: HttpRequest, sociallogin: SocialLogin):
-        if sociallogin.is_existing:
+        if sociallogin.account.pk:
+            # This exact SocialAccount has already signed in before; allow reauthentication.
+            # (sociallogin.is_existing checks the resolved *user*, which can also become
+            # true for a brand-new social account matched to an existing user by email --
+            # exactly the case _reject_if_email_already_registered below needs to catch.)
```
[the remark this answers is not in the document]
hemant10yadav: Verified — the check order predates this PR. I only changed the condition of the first check (`is_existing` → `account.pk`); the order itself (existing-account check before the `CONNECT`/provider checks) is unchanged since the method was first written, well before this migration. So if this is a real gap, it's pre-existing behavior, not a regression introduced here — reordering the checks now would be a genuine security/policy change bundled into a dependency-upgrade PR. Worth its own ticket if we want to lock OCS logins to connect-only in all cases, but out of scope here. Not applying in this PR.

config/settings/base.py
```diff
@@ -306,9 +307,8 @@
 # django-allauth
 # ------------------------------------------------------------------------------
 ACCOUNT_ALLOW_REGISTRATION = env.bool("DJANGO_ACCOUNT_ALLOW_REGISTRATION", True)
-ACCOUNT_AUTHENTICATION_METHOD = "username_email"
-ACCOUNT_EMAIL_REQUIRED = True
-ACCOUNT_USERNAME_REQUIRED = False
+ACCOUNT_LOGIN_METHODS = {"username", "email"}
+ACCOUNT_SIGNUP_FIELDS = ["username", "email*", "password1*", "password2*"]
```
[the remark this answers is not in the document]
hemant10yadav: Not applying this one — it'd be a regression, not a fix. The setting this replaced was `ACCOUNT_USERNAME_REQUIRED = False`, and `git log -p` on this line shows it's been `False` since the earliest commit in this file's history — username has never been required at signup. `"username"` without the asterisk is the correct 1:1 translation (allauth's own docs: no asterisk = field present but optional). Adding `"username*"` would newly force a field the app has never required.

commcare_connect/users/adapters.py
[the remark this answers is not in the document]
hemant10yadav: We only provide the option to log in using HQ only, so I think it’s fine for now.

config/settings/production.py
ajeety4: Shoud we also be adding STATICFILES_STORAGE equivalent under `STORAGES` ?
The [default](https://docs.djangoproject.com/en/5.1/ref/settings/#storages) for it seems to be different that what we have currently.

commcare_connect/users/views.py
```diff
@@ -1,14 +1,14 @@
...
 
-from allauth.account.models import transaction
 from django.contrib import messages
 from django.contrib.auth import get_user_model
 from django.contrib.auth.decorators import login_required, permission_required
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.messages.views import SuccessMessageMixin
+from django.db import transaction
```
ajeety4, line 11: :+1:

ajeety4 (commented):
Looks good to me.
Just a single query

config/settings/production.py
hemant10yadav: Good catch to double check, but we're covered — `STORAGES["staticfiles"]` is already set in [base.py](https://github.com/dimagi/commcare-connect/blob/666afdc46dbe449b3ca34618163aa5eed105eed1/config/settings/base.py#L180), and `production.py` inherits it via `from .base import *`. We only override `STORAGES["default"]` here for S3 media. The removed `STATICFILES_STORAGE` line was redundant with that, not missing an equivalent.

ajeety4: Thanks

ajeety4 (approved):

commcare_connect/opportunity/tables.py
```diff
@@ -85,6 +83,11 @@ class UserVisitTable(OrgContextTable):
     flag_reason = columns.Column("Flags", accessor="flag_reason", empty_values=({}, None))
     details = columns.Column(verbose_name="", empty_values=())
 
+    def value_visit_date_export(self, value):
+        # DateTimeColumn.value() hardcodes isoformat(sep=" ") for exports, ignoring `format`.
```
pxwxnvermx: nit: The comments are a little extra verbose for some changes. I think they can be removed.

commcare_connect/opportunity/tests/test_views.py
```diff
@@ -2679,6 +2688,7 @@ def test_create_task_ocs_failure_shows_error(self, client, program_manager_org_u
         assert any("chatbot" in str(m).lower() for m in msgs)
 
     def test_create_task_invalid_form(self, client, program_manager_org_user_admin, opportunity):
+        _create_ocs_social_app()
```
pxwxnvermx: nit: This can reuse the `_create_social_app` function that exists, instead of creating a new one for  this use case.

commcare_connect/ocs_provider/tests/test_views.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@pytest.mark.django_db
+def test_complete_login_sends_the_actual_access_token_string(httpx_mock, rf):
+    # allauth passes a SocialToken instance (not the raw string) to complete_login() since 65.x.
+    # str(SocialToken(...)) is "social application token (None)", not the access token -- if
+    # complete_login() ever formats `token` directly instead of `token.token`, this request never
+    # matches and httpx_mock raises, catching the regression.
+    app = SocialApp.objects.create(provider="ocs", name="ocs", client_id="ocs-client", secret="ocs-secret")
+    app.sites.add(Site.objects.get_current())
```
pxwxnvermx: nit: This can be replaced with the `_create_social_app` function.

commcare_connect/ocs_provider/tests/test_connect_prompt.py
```diff
@@ -1,9 +1,13 @@
 import pytest
+from allauth.socialaccount.models import SocialApp
+from django.contrib.sites.models import Site
 from django.template.loader import render_to_string
 
 
 @pytest.mark.django_db
 def test_connect_prompt_renders_ocs_connect_link(rf, user):
+    app = SocialApp.objects.create(provider="ocs", name="ocs", client_id="ocs-client", secret="ocs-secret")
+    app.sites.add(Site.objects.get_current())
```
pxwxnvermx: Same comment as below (`_create_social_app`)

commcare_connect/ocs_provider/tests/test_provider.py
```diff
@@ -18,7 +21,8 @@ def test_extract_uid_uses_sub():
 
 def test_pkce_is_enabled_with_s256_challenge():
     # OCS requires PKCE; allauth only emits a challenge when PKCE is enabled on the provider.
-    provider = OcsProvider.__new__(OcsProvider)
+    app = SocialApp(provider="ocs", name="ocs", client_id="ocs-client", secret="ocs-secret")
```
pxwxnvermx, line 24: nit: Same comment as below (`_create_social_app`).

pxwxnvermx (approved):
Changes look good to me. Some minor nits about reusing a test function across files and the verbosity of comments.

zandre-eng (approved):
Overall looks good. I agree with the comment that some of the code comments could potentially be reduced or completely removed.

commcare_connect/opportunity/tables.py
hemant10yadav: Removed. Fixed in 9377f001.

commcare_connect/opportunity/tests/test_views.py
hemant10yadav: Good catch, thanks — dropped the duplicate helper and reused _create_social_app. Fixed in 9377f001.

commcare_connect/ocs_provider/tests/test_views.py
hemant10yadav: Fixed in 9377f001.

commcare_connect/ocs_provider/tests/test_connect_prompt.py
hemant10yadav: Fixed in 9377f001.

commcare_connect/ocs_provider/tests/test_provider.py
hemant10yadav: Left this one as-is — it builds an unsaved SocialApp instance with no DB (test_pkce_is_enabled_with_s256_challenge isn't django_db-marked), so it's not actually the same duplicate pattern as the other three. Reusing _create_social_app there would mean adding a DB dependency to what's currently a fast, pure unit test — didn't think that trade was worth it for a 1-line dedup. Happy to reconsider if you feel differently.

pxwxnvermx (approved):
Looks good. Thanks

PR #1451, opened 2026-08-13
Open Org Creation

zandre-eng:
## Product Description

Any signed-up user can now create a workspace. Previously creation required an internal permission, so a user who signed up without an invite had no way forward.

Signing in without belonging to any workspace now lands on a dedicated page that explains the situation and offers two next steps — ask an administrator for an invite, or create a workspace — instead of the old dead-end message:

**Old:**
Screenshot_20260813_132619

**New:**
Screenshot_20260813_144045

From there, the workspace creation page is reachable by everyone. For a user with no workspaces the existing-workspace dropdown is empty, so the page is create-only and the "Switch to Selected Workspace" button does not appear:
Screenshot_20260813_144131

Choosing a workspace name that is already taken now gives an actionable message rather than telling the user to pick the existing workspace from a list they cannot see:
Screenshot_20260813_144205

Users who already belong to a workspace see no change: they are routed to their workspace as before, and the "Create Workspace" entry in the header remains limited to internal users.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2565)

Removing the permission gate could not be done on its own: the create form's workspace field spanned every `Organization`, and the data feeding its TomSelect dropdowns embedded every workspace's name and slug in the page, so opening the page would have exposed the full customer list to anyone who could sign up. Both the field queryset and that JSON are now scoped to the workspaces a user belongs to, which also stops a crafted POST from selecting a workspace the user cannot join. Because the field's built-in duplicate check derives from the same queryset, `clean_org` was added so duplicate names stay blocked platform-wide rather than only among visible workspaces. The old `/home/` route was retired instead of repurposed: its authenticated half is what became the new page, its anonymous half duplicated the prelogin landing page at `/`, and its name collided with prelogin's own `home` route, silently shadowing it.

Note for reviewers: this branch targets `pkv/ccct-2561-merge-lloentity-org-models`, which supplies the `Organization.verified` field this change relies on.

### Safety story

No model, manager, or migration changes, and no existing data is touched or rewritten — the change is limited to view access, form querysets, one route, and one template. The visibility scoping only narrows what is shown to users who previously received a 403 on this page; permission holders and superusers see exactly what they saw before. Members are redirected off the new landing page, so no one who already has a workspace can be shown a "you have no workspace" state.

### Automated test coverage

Tests cover the visibility helper across every privilege level, and verify that the create form both hides workspaces a user cannot see and rejects attempts to select or duplicate them. View tests confirm a non-privileged user can create a workspace that comes out unverified with its creator as admin, that a membership-less user reaches the new page and is offered workspace creation, and that a member is redirected away from it. The full suite passes.

### QA Plan

QA is planned.

### Labels & Review


commcare_connect/organization/helpers.py
```diff
@@ -0,0 +1,14 @@
+from commcare_connect.organization.models import Organization
+from commcare_connect.utils.permission_const import WORKSPACE_ENTITY_MANAGEMENT_ACCESS
+
+
+def orgs_visible_to(user):
```
hemant10yadav: Would `@classmethod` on the `Organization` model be a better approach here?

commcare_connect/organization/forms.py
```diff
@@ -237,11 +238,20 @@ class OrganizationSelectOrCreateForm(forms.Form):
         ),
     )
 
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user")
+        super().__init__(*args, **kwargs)
+        self.fields["org"].queryset = self.visible_orgs.order_by("name")
```
hemant10yadav: Should we also scope the `llo_entity queryset` based on the user's visibility?

hemant10yadav (commented):
Overall, looks good to me. I’ve left one suggestion and one question for consideration. Let me know what you think.

commcare_connect/organization/helpers.py
zandre-eng: I originally thought about doing this, but decided against it since this function is only getting called in one place. But after some more thought, I agree it does feel like the more natural place.

Addressed in [7967f66](https://github.com/dimagi/commcare-connect/pull/1451/commits/7967f661fdf9c42a2a9e39b18ea21bfcdcb5ce4b).

commcare_connect/organization/forms.py
zandre-eng: Good question, let me follow up on this.

zandre-eng: @hemant10yadav I have confirmed that we should apply the same scope to the LLO Entity queryset.

Addressed in [ed6277b](https://github.com/dimagi/commcare-connect/pull/1451/commits/ed6277bbb8df738bb1bf21629d2ddfb1eeab0071).

Charl1996:
I wonder about the message exposing that a workspace with a particular name already exists - I don't think it's particularly a security risk, but it does expose data that I think is unnecessary.

Can we make the message more generic to say something like: "The entered workspace name is invalid"?

zandre-eng:
> I wonder about the message exposing that a workspace with a particular name already exists - I don't think it's particularly a security risk, but it does expose data that I think is unnecessary.
Can we make the message more generic to say something like: "The entered workspace name is invalid"?

@Charl1996 Great point. I was thinking about this for a bit, and initially I did want to go for something generic. I eventually decided against it as a generic message felt like it could make the user experience frustrating. 
Receiving a "name is invalid" doesn't tell the user anything about why it's invalid, so the user would have to essentially guess by reformatting their chosen name slightly to figure out what we're validating against.

With the above said, I'm happy to go with your suggestion (or think through a possible alternative) if you think we should rather lean towards not revealing any data to the user. Let me know your thoughts?

Charl1996:
> it could make the user experience frustrating.

That is true. 

> Receiving a "name is invalid" doesn't tell the user anything about why it's invalid,

How about "Cannot choose that name - choose a different name"? This way the user at least knows its not invalid because of characters or so, but due to the name itself. That said, a clever bad actor could still guess that this means the name is already taken. I'm not sure if we need to be too pedantic about this, so OK to keep it as is.

mkangia:
@Charl1996 @zandre-eng Great discussion.

Do we have a different slug and display name?
We could take some inspiration from CommCareHQ about what it does with domain names. 

Or we can ask product what they would want in this situation? I am surprised this is new, can't user's already create organizations today?

zandre-eng:
> Do we have a different slug and display name?

@mkangia They generally have a very similar display name and slug (e.g. "My Org" and "my-org").

> I am surprised this is new, can't user's already create organizations today?

Previously, only users with the flag enabled could create new orgs. This would typically be Dimagi staff, and the create form would list all the LLO Entities and Organizations within Connect, so a user would be able to easily see which already exist.

After looking at what HQ does it seems they simply state "Project name already taken - please try another". Given this, and the discussion with Charl I've decided to go ahead with a small refactor to make it a bit more generic ([d6b649f](https://github.com/dimagi/commcare-connect/pull/1451/commits/d6b649f577191cefd80022a2615caf114ad17592)).

FYI @Charl1996

commcare_connect/organization/forms.py
```diff
@@ -255,6 +262,28 @@ def get_entity_wise_orgs(self):
...
+        return llo_entity
+
+    def clean_org(self):
+        """Block duplicate workspace names platform-wide, not just among visible workspaces.
+
+        The field's own duplicate check only covers workspaces the user can see and select,
+        so on its own it would let a user create a workspace named after one they cannot see.
+        """
+        org = self.cleaned_data["org"]
+        if org and not org.pk and Organization.objects.filter(name__iexact=org.name).exists():
```
Charl1996, line 283: Does this need to be case sensitive?

Charl1996 (approved):
Looks good to me

commcare_connect/organization/forms.py
zandre-eng: Good question. I think a case insensitive check here is the right approach. Let's say for example "Org A" exists with a slug "org-a". If someone tries to add a new org called "org A", a case sensitive check would let this go through fine and the latter org would have a slug "org-a-1", which could end up creating confusing URLs for the user.

With that said, I'm happy to drop this if you think it's fine given the above example.

hemant10yadav (approved):

commcare_connect/organization/forms.py
Charl1996: No I think it's fine as is - thanks for explaining.

mkangia: @zandre-eng can you help me here.

If someone names the org "org A" while we already have "Org A" (so org-a already exists), this would go through and we would end up with org-a--1 which is confusing but that is okay?

mkangia: I do understand that this is going to be rare case and we might just see a lot of slugs like `test-N` so this happening with real project names is going to be rare.

zandre-eng: > If someone names the org "org A" while we already have "Org A" (so org-a already exists), this would go through and we would end up with org-a--1

@mkangia That is correct, at least that's how the behaviour was before this PR's changes. Now, with the case insensitive validation a user won't be able to do "Org A" vs "org A", and so we won't end up silently creating very similar slugs like "org-a-1".

> I do understand that this is going to be rare case and we might just see a lot of slugs like test-N so this happening with real project names is going to be rare.

I agree it's probably a rare case, but it does keep things explicit for the user such that their org's slug name will always match the name they gave. I'm happy to chat through this more if you believe there's valid reasons for us to keep the old behaviour.

mkangia: I am aligned with this. I believe users creating test organizations will get into issue rather than real projects which seems fine to me.

PR #1453, opened 2026-08-13
CCCT-2558: Role-aware program and opportunity permissions

ajeety4:
## Product Description

- Today, access is primarily determined by two things:
  * The organization associated with the request (Program/PM org or NM/Delivery org)
  * The user's role within that organization

- With the introduction of new organization types, we're expanding this model to support additional relationships to programs and opportunities.
- The existing access for PM org users and NM org users remains unchanged.
- `ALL_ORG_ACCESS` still overrides everything.

**User Impact** 
With this change, existing entities (NM org, PM org) will continue to have the same level of access and new entities will gain access to the resources they are allowed. See [this section](https://github.com/dimagi/commcare-connect/blob/167e0d8329e8a14fe77e5a99d3a707d57159be78/docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md#permission--access-model) for their allowed access.

### Permission / access model
Please see [this section](https://github.com/dimagi/commcare-connect/blob/167e0d8329e8a14fe77e5a99d3a707d57159be78/docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md#permission--access-model) for revised access model.

## Technical Summary

[CCCT-2558](https://dimagi.atlassian.net/browse/CCCT-2558)
[Spec](https://github.com/dimagi/commcare-connect/blob/167e0d8329e8a14fe77e5a99d3a707d57159be78/docs/superpowers/specs/2026-06-17-program-opportunity-roles-design.md)

- As mentioned above, access is currently determined based on the organization type (PM or NM) and the user's role.The same access decorators are currently used across organization, program, and opportunity views.
With this change, access checks are now exposed through scope-based decorators. Views only need to specify the scope they are accessing (e.g. program, organization, or opportunity), while the access layer abstracts the underlying organization and user relationships used to determine the effective access level.

- Rather than simply replacing the existing methods, new access methods were added for organizations and programs (the opportunity access decorators were updated in place).This felt cleaner and, by using consistent handlers across the different scopes, makes the access logic more robust and easier to understand.

### Access decorators

| Scope | Decorator | Details |
|---|---|---|
| **Organization** | `org_view_access_required` | |
| | `org_standard_access_required` |  |
| | `org_manage_access_required` |  |
| **Program** | `program_view_access_required` |  |
| | `program_standard_access_required` |  |
| | `program_manage_access_required` |  |
| **Opportunity** | `opp_view_access_required` |  |
| | `opp_standard_access_required` |  |
| | `opp_manage_access_required` |  |
| | `opportunity_pm_required` | user is an admin of an organization that can act as the PM |
| | `opportunity_nm_required` | user is an admin of the Delivery/NM organization|

The changes roughly cover:

* Introducing the access-level layer
* Adding program access methods and applying program-level scoping to related views
* Adding the equivalent organization and opportunity access methods
* Moving organization/role-specific logic behind the access layer
* Some views do not expose the relevant organization/program/opportunity ID or UUID in the URL. These use an **inline access check** instead.

- Also removes `get_managed_opp` and its 24-hour quickcache - it had no invalidation path and saved at most one indexed lookup per request.

**Three pre-existing bugs fixed,**

- `manage_application` never checked that the caller's org was related to the application's program, so any program-manager org admin could accept or reject **any** program's applications by enumerating the sequential integer id.
- `ManagedOpportunityViewMixin` took its program from the URL `pk` without validating it against the opportunity, letting a foreign program's currency, country and applicant list be written onto an opportunity.

### Safety story

No schema or data migrations, and no changes to opportunity execution or payments. The work is permission logic plus two narrowing bugfixes; the only widening is the intended one (funder and supervising org gaining access). Existing program-org behaviour is preserved and pinned by a parametrized matrix. 

QA to be performed.

### Automated test coverage
Added

### QA Plan
Yes, QA to be performed.

### Deployment
- [x]  This PR can be deployed: any required configuration, commands, or other manual steps required before this PR can be deployed are done.

### Labels & Review

[CCCT-2558]: https://dimagi.atlassian.net/browse/CCCT-2558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mkangia:
Hey @ajeety4 

Just couple of quick thoughts from reading the description

> request_can_manage_program / request_can_view_program / request_supervises_opportunity

how about request_can_supervise_opportunity to follow the same pattern?

> Deliberate gaps (tracked, not oversights):

I was not able to fully understand both these points. May be its just me. Can you help me understand these by rephrasing or expanding?

commcare_connect/program/utils.py
```diff
@@ -1,24 +1,163 @@
...
+        return False
+    if request.user.has_perm(ALL_ORG_ACCESS):
+        return True
+    # Below ALL_ORG_ACCESS: its holders normally have no membership in the org.
+    if organization_role_level(request.org_membership) is not AccessLevel.MANAGE:
+        return False
+    return request.org.id in opportunity_manage_org_ids(opportunity)
+
+
+def opportunity_manage_org_ids(opportunity) -> set[int]:
```
mkangia: nit: how about `opportunity_managing_org_ids`?

commcare_connect/opportunity/views.py
```diff
@@ -319,7 +320,7 @@ def get_table_data(self):
         return OpportunityData(org, is_program_manager, self.get_filter_values()).get_data()
 
 
-class OpportunityInit(OrgPMRequiredMixin, CreateView):
+class OpportunityInit(OpportunityObjectMixin, ProgramManageAccessMixin, CreateView):
```
ajeety4, line 334: Any user with a program manage access should be able to create opportunity.

commcare_connect/program/views.py
```diff
@@ -164,15 +163,15 @@ def get_form_kwargs(self):
         return kwargs
 
 
-class ManagedOpportunityInit(ManagedOpportunityViewMixin, ProgramManagerMixin, OpportunityInit):
+class ManagedOpportunityInit(ManagedOpportunityViewMixin, OpportunityInit):
```
ajeety4, line 167: The inherited class `OpportunityInit` takes care of the access

commcare_connect/organization/views.py
```diff
@@ -48,7 +48,7 @@ def no_organization(request):
     return render(request, "organization/no_organization.html")
 
 
-@org_admin_required
+@org_manage_access_required
```
ajeety4, line 50: Anyone with manage access to org under the new access requirements acts similar to `org_admin_required` and similarly for `org_viewer_required` to `org_view_access_required`.
The name was updated instead of using existing names to keep things consistent with other access based decorators.

commcare_connect/program/utils.py
```diff
@@ -66,23 +95,24 @@ def _base_access_level(request) -> AccessLevel | None:
     return None
 
 
-@quickcache(vary_on=["opp_id"], timeout=60 * 60 * 24)
-def get_managed_opp(opp_id) -> Opportunity | None:
```
ajeety4, line 6: Opportunity is no longer fixed, things like supervisor can change. Plus this cache was never invalidated.

commcare_connect/microplanning/views.py
```diff
@@ -223,7 +223,7 @@ def microplanning_home(request, *args, **kwargs):
         kwargs={"org_slug": request.org.slug, "opp_id": opportunity.opportunity_id},
     )
 
-    is_program_manager = is_org_pm_or_all_access(request)
+    is_program_manager = is_opportunity_pm(request, opportunity)
```
ajeety4, line 231: We want to provide access to anyone who can acts as PM of opportunity based on new rules. 
Same reason for changing `org_pm_required` to `opp_pm_required`

commcare_connect/program/utils.py
ajeety4: No longer present, as the changes  in draft PR were updated.

ajeety4:
> Hey @ajeety4
> 
> Just couple of quick thoughts from reading the description
> 
> > request_can_manage_program / request_can_view_program / request_supervises_opportunity
> 
> how about request_can_supervise_opportunity to follow the same pattern?
> 
> > Deliberate gaps (tracked, not oversights):
> 
> I was not able to fully understand both these points. May be its just me. Can you help me understand these by rephrasing or expanding?

The changes were updated, and these are no longer present.

commcare_connect/program/utils.py
```diff
@@ -16,9 +113,18 @@ def is_org_pm(request):
...
+    return _can_manage_opportunity(request, opportunity) and request.org.id == opportunity.organization_id
+
+
+def is_opportunity_pm(request, opportunity) -> bool:
+    """Anyone else who can manage the opportunity reaches it from the program side."""
+    return _can_manage_opportunity(request, opportunity) and request.org.id != opportunity.organization_id
+
+
+def _can_manage_opportunity(request, opportunity) -> bool:
+    return opportunity_access_level_from_request(request, opportunity) is AccessLevel.MANAGE
```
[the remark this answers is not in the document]
ajeety4: https://github.com/dimagi/commcare-connect/pull/1453/commits/fd066d037b37a6a010b86871a6777a3cdf679013

commcare_connect/opportunity/views.py
```diff
@@ -826,15 +838,17 @@ def payment_import(request, org_slug=None, opp_id=None):
...
 def render_payment_import_progress(request, org_slug, task_id):
     """Renders the payment import modal: a spinner while the import runs, then the row errors
     that stopped it. An import that finishes without row errors refreshes the page instead, so
     its outcome shows up as a standard banner."""
 
     def ownership_check(request, task_meta):
-        get_opportunity_or_404(org_slug=org_slug, pk=task_meta.get("args")[0])
+        opportunity = get_opportunity_or_404(task_meta.get("args")[0])
+        if opportunity_access_level_from_request(request, opportunity) < AccessLevel.STANDARD:
+            raise Http404()
```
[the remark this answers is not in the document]
ajeety4: https://github.com/dimagi/commcare-connect/pull/1453/commits/fd066d037b37a6a010b86871a6777a3cdf679013

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
+from enum import IntEnum
+
 from commcare_connect.cache import quickcache
 from commcare_connect.opportunity.models import Opportunity
 
 
+class AccessLevel(IntEnum):
```
mkangia, line 10: nit: Can we add documentation for what these stand for and also for the function "effective", primarily the arguments because its not clear off hand what is org_level and user_level?

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
...
+    @staticmethod
+    def effective(org_level: "AccessLevel", user_level: "AccessLevel") -> "AccessLevel":
+        return min(org_level, user_level)
+
+
+def user_org_access(membership) -> AccessLevel:
+    if not membership:
+        return AccessLevel.NONE
+    if membership.is_admin:
+        return AccessLevel.MANAGE
```
mkangia, line 35: What would the difference be, between "Manage" and "Standard"? It is hard to predict, specially when there is also a "View" level.

I was thinking we could rename "Manage" to "Admin" but it makes sense to have the access level entirely independent of the role.

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
...
+    if membership.is_member:
+        return AccessLevel.STANDARD
+    if membership.is_viewer:
+        return AccessLevel.VIEW
+    return AccessLevel.NONE
+
+
+def org_program_access(org, program) -> AccessLevel:
```
mkangia: nit: since an organization also has a program, I got a bit confused here. I though the org and program were related.

I'd suggest a name like "org_access_for_program"

commcare_connect/program/utils.py
```diff
@@ -1,13 +1,112 @@
...
+
+
+def opportunity_access_level_from_request(request, opportunity) -> AccessLevel:
+    if not opportunity:
+        return AccessLevel.NONE
+    base_access = _base_access_level(request)
+    if base_access is not None:
+        return base_access
 
+    org_level = org_opportunity_access(request.org, opportunity)
```
hemant10yadav: These two functions look almost identical, so we can refactor the shared logic into a common helper function:

```python
def _resource_access_level(request, resource, org_access_fn) -> AccessLevel:
    if not resource:
        return AccessLevel.NONE

    base_access = _base_access_level(request)
    if base_access is not None:
        return base_access

    org_level = org_access_fn(request.org, resource)
    user_level = user_org_access(request.org_membership)

    return AccessLevel.effective(org_level, user_level)
```

Then both `program_access_level_from_request()` and `opportunity_access_level_from_request()` can use this shared function, while passing their respective organization access function.

commcare_connect/organization/decorators.py
```diff
@@ -166,9 +197,33 @@ def dispatch(self, request, *args, **kwargs):
...
+class ProgramViewAccessMixin:
+    @method_decorator(program_view_access_required)
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+
+class OrgManageAccessMixin:
+    @method_decorator(org_manage_access_required)
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
```
hemant10yadav, line 221: Could we apply these decorators directly to the CBVs using `@method_decorator(..., name="dispatch")` instead of creating permission-only mixins?

This keeps the permission logic centralized, avoids unnecessary permission-only mixins, and makes the required permission explicit on the view. Do you see any issues with this approach?

commcare_connect/organization/decorators.py
```diff
@@ -23,89 +33,107 @@ def user_is_org_admin(user, organization):
...
+    return decorator
 
 
-def org_pm_required(view_func):
-    return _get_decorated_function(view_func, is_org_pm_or_all_access)
+def _opportunity_gate(has_required_access, opp_id_kwarg="opp_id"):
+    def decorator(view_func):
+        def permission_test(request, *args, **kwargs):
+            opp_id = kwargs.get(opp_id_kwarg)
+            opportunity = opportunity_by_id(opp_id) if opp_id else None
```
hemant10yadav: Could we check `opportunity = getattr(request, "opportunity", None)` first here and only fetch it if it’s not already present?

This would avoid an extra DB query when `opp_standard_access_required` and `opportunity_required` are used together. The same optimization could apply to `opportunity_required` as well.

hemant10yadav (commented):
Did a first pass — overall, this looks solid. I’ve left a few suggestions around simplifying the implementation and avoiding some unnecessary DB queries.

commcare_connect/organization/decorators.py
ajeety4: I like this idea. It makes the required permission explicit on the view and is easier to remember/use than tracking down which mixin does what. 

However, this kind of change means checking the current MRO for the viewsso switching only some mixins to decorators does not quietly change check ordering. 

There are ~22 view classes across 8 permission mixins that would need this per-view check, which is more scope than I want to add to this PR. I think it's a good candidate for a follow-up ticket/PR instead . Does that sound good to you?

commcare_connect/organization/decorators.py
ajeety4: Thanks.  I missed this.

https://github.com/dimagi/commcare-connect/pull/1453/commits/206a7ac5357d6771cdf41359cbd8c47e4569b71a

commcare_connect/program/utils.py
ajeety4: Good point. I mainly went with separate functions to keep readibility better compared to modest benefit provided by dedupe. However a  case could be made for keeping access logic in a single function.
Considering that, refactored in https://github.com/dimagi/commcare-connect/pull/1453/commits/206a7ac5357d6771cdf41359cbd8c47e4569b71a

commcare_connect/data_export/views.py
```diff
@@ -97,6 +97,28 @@ class BaseDataWriteView(APIView):
     required_scopes = ["write"]
 
 
+def user_is_opportunity_admin(user, opportunity):
```
mkangia, line 100: nit: I'd suggest renaming this to "user_has_admin_access". Since we allow some users admin access even when they are not admin, this naming will throw someone off when it is being used.

```
if user_is_opportunity_admin(user, opp):
     # ...
```
without looking at the function definition, one would assume we are just checking for the admin role on the user.

commcare_connect/data_export/views.py
```diff
@@ -97,6 +97,28 @@ class BaseDataWriteView(APIView):
...
+    """Admin of any org that can manage this opportunity, from either the delivery or program side."""
```
mkangia: what do we mean by "the delivery or program side"?

commcare_connect/data_export/views.py
```diff
@@ -97,6 +97,28 @@ class BaseDataWriteView(APIView):
...
+    if user.has_perm(ALL_ORG_ACCESS):
+        return True
+    return UserOrganizationMembership.objects.filter(
+        user=user,
+        organization_id__in=opportunity_managing_org_ids(opportunity),
+        role=UserOrganizationMembership.Role.ADMIN,
+    ).exists()
+
+
+def user_is_opportunity_pm(user, opportunity):
```
mkangia, line 111: nit: I'd suggest a similar renaming here.

commcare_connect/data_export/views.py
```diff
@@ -97,6 +97,28 @@ class BaseDataWriteView(APIView):
...
+    """Admin of an org that manages this opportunity from the program side, i.e. not the delivery org."""
```
mkangia: it would be good for the function name to also include the fact about it not being the delivery org. I believe the better term is owning organization.

commcare_connect/microplanning/views.py
```diff
@@ -118,7 +118,7 @@
 
 
 @require_GET
-@org_admin_required
+@opp_manage_access_required
```
mkangia, line 121: Just confirming that this still means that only admins have access?

commcare_connect/microplanning/views.py
mkangia, line 231: nit: another naming suggestion here.

If this is keeping the same access as before, that means those that have all access can still access the opportunity and not necessarily just PMs.

mkangia (commented):
Still taking a pass, but submitting what I have so far.

The naming needs a reconsideration here, I have flagged the ones I thought were concerning. Its very important that the names correctly highlight what the functions are doing or checking for.

commcare_connect/program/utils.py
ajeety4: My initial thought was that the AccessLevel class provides enough context here, and org_level and user_level are fairly self-explanatory as parameters. Since effective() is simply returning the lower of the two access levels, I wasn't sure additional parameter documentation would add much value.

I have kept these generic to keep effective() independent of the caller and avoid coupling it to any specific context for why the two access levels are being combined.
Also added some documentation to make things clear.

ajeety4: https://github.com/dimagi/commcare-connect/pull/1453/commits/a12efffa437b20d04587ff38365e8274d50865be

commcare_connect/program/utils.py
ajeety4: This is a good point.

> What would the difference be between "Manage" and "Standard"?

These names have been taken from the spec. I feel like Manage is fairly clear here — similar to the admin-level access you mentioned. Standard, on the other hand, is not very clear. This level is intended for users who are members of the organization, since certain functionality requires the user to be a member.

I'm open to suggestions for a better naming scheme here. I also added some documentation for better clarity.

commcare_connect/program/utils.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1453/commits/a12efffa437b20d04587ff38365e8274d50865be

commcare_connect/data_export/views.py
ajeety4: I’m not sure the suggested rename would be the best fit here, since this is specifically checking for admin access to an opportunity, rather than checking whether the user has an admin role.

I feel `user_is_opportunity_admin` makes the scope clear and indicates that we’re checking whether the user is an admin of the opportunity.

commcare_connect/data_export/views.py
ajeety4: By "delivery side", I mean the organization that works on and manages deliveries for the opportunity.
By "program side", I mean any organization that has access to the program that the opportunity belongs to.

commcare_connect/data_export/views.py
ajeety4: I actually prefer the current name, as this terminology is widely used across the codebase and is also used when determining access to various functionalities at requirements time.

commcare_connect/data_export/views.py
ajeety4: Similar reasoning as [above](https://github.com/dimagi/commcare-connect/pull/1453#discussion_r3912732475).
Also, PM by default implies that the delivery org is excluded. The intention of the existing comment is to provide additional context and make the behavior here easier to understand.

commcare_connect/microplanning/views.py
ajeety4: Yes admins of all eligible organizations (owner org, delivery org, funder org, supervisor org)

commcare_connect/microplanning/views.py
ajeety4: This is not the same as `is_opportunity_pm`. This refers to the organization that is marked as the Program Manager, which allows their admin users to create programs.
I understand this can be confusing since the PM concept is being used for both the organization and the opportunity. I’m going to remove this and use `user_is_org_pm` instead, since it serves the same purpose.

commcare_connect/organization/decorators.py
hemant10yadav: Yep, sounds good to me. Let’s track it in a follow-up ticket/PR.

commcare_connect/program/utils.py
hemant10yadav: Thanks

hemant10yadav (approved):

commcare_connect/microplanning/views.py
mkangia: oh lord, ya we need a better name for saying that an organization has program manager access enabled.
Saying that an opportunity is pm is wrong anyway since its not the opportunity but the organization that is has PM access.
Anyway, not to be addressed in this PR.

commcare_connect/microplanning/views.py
mkangia: thanks, we need to get used to the new access levels.

commcare_connect/program/utils.py
mkangia: > org_level and user_level are fairly self-explanatory as parameters

As of now, I really can't tell what these mean. How does one have two different access levels? I believe one would need to know how these are assigned. And a question one could wonder, the access levels could be scoped to an organization/opportunity/program but what does user_level access mean?

> I have kept these generic to keep effective() independent of the caller and avoid coupling it to any specific context for why the two access levels are being combined.

This was a great call :+1:

commcare_connect/program/utils.py
mkangia: I don't have great suggestion here for now.
May be?
`Read Only, Manage, Full`

So, Manage could be "Full"?

I think its just getting used to, documentation is helpful yes. Can ask broader team for more suggestions?

commcare_connect/data_export/views.py
mkangia: > Since this is specifically checking for admin access to an opportunity, rather than checking whether the user has an admin role

Right. That was my point. `user_is_opportunity_admin` makes me read if user is opportunity pm rather than thinking of access.

commcare_connect/data_export/views.py
mkangia: are these terms usually used in code or are getting introduced by this PR?
I would vote to use existing terminology wherever possible.

commcare_connect/organization/decorators.py
```diff
@@ -23,89 +33,103 @@ def user_is_org_admin(user, organization):
     ).exists()
 
 
-def user_is_program_manager_by_slug(user, org_slug):
-    """Check if user is admin of a program-manager org identified by slug, or has ALL_ORG_ACCESS."""
+def user_is_org_pm(user, organization):
+    """Creating programs takes two things: an org that is a program manager, and a caller who is its admin."""
```
mkangia: how is this comment related to this function?

commcare_connect/organization/decorators.py
```diff
@@ -23,89 +33,103 @@ def user_is_org_admin(user, organization):
     ).exists()
 
 
-def user_is_program_manager_by_slug(user, org_slug):
-    """Check if user is admin of a program-manager org identified by slug, or has ALL_ORG_ACCESS."""
+def user_is_org_pm(user, organization):
```
mkangia, line 36: nit: not going to push on this here but this naming could be improved upon. We can take it as a follow up later.

commcare_connect/organization/decorators.py
```diff
@@ -23,89 +33,103 @@ def user_is_org_admin(user, organization):
...
 
-def user_is_opportunity_admin(user, opportunity):
-    if user.has_perm(ALL_ORG_ACCESS):
-        return True
-    return UserOrganizationMembership.objects.filter(
-        user=user,
-        organization_id__in=[opportunity.organization_id, opportunity.program.organization_id],
-        role=UserOrganizationMembership.Role.ADMIN,
-    ).exists()
+def org_is_program_manager(organization):
```
mkangia, line 42: this is an example of naming that sits well with me. 💯

commcare_connect/program/utils.py
```diff
@@ -1,13 +1,123 @@
...
+def org_opportunity_access(org, opportunity) -> AccessLevel:
+    """The one delivering it, supervising it and the one running its program. Watcher has view access."""
+    if not org or not opportunity:
+        return AccessLevel.NONE
+    if org.id in (opportunity.organization_id, opportunity.supervising_organization_id):
+        return AccessLevel.MANAGE
+    return org_access_for_program(org, opportunity.program)
+
+
+def opportunity_managing_org_ids(opportunity) -> set:
```
mkangia: nit: `orgs_ids_with_manage_access_to_opportunity`

commcare_connect/program/utils.py
```diff
@@ -16,9 +126,18 @@ def is_org_pm(request):
...
-def is_opportunity_pm(request, opp_id) -> bool:
-    managed_opp = get_managed_opp(opp_id)
-    return bool(managed_opp and managed_opp.program.organization.slug == request.org.slug and is_org_pm(request))
+def is_opportunity_nm(request, opportunity) -> bool:
+    """The network manager is the org delivering the opportunity."""
+    return _can_manage_opportunity(request, opportunity) and request.org.id == opportunity.organization_id
+
+
+def is_opportunity_pm(request, opportunity) -> bool:
+    """Anyone else who can manage the opportunity reaches it from the program side."""
```
mkangia: I am confused that the comment says "reaches it from the program side" but then checks for the request org not being equal to opportunity org. I am not sure how those 2 things map to each other.

nit: the naming seems to not match the description or the implementation.

commcare_connect/opportunity/views.py
```diff
@@ -252,34 +257,36 @@
 PAYMENT_IMPORT_CLAIMED_SESSION_KEY = "shown_payment_import"
 
 
-def get_opportunity_or_404(pk, org_slug):
-    opp = get_object_by_uuid_or_int(Opportunity.objects.all(), str(pk), uuid_field="opportunity_id")
-
-    if opp.organization.slug == org_slug or opp.program.organization.slug == org_slug:
-        return opp
+def get_opportunity_or_404(opp_id):
+    opportunity = opportunity_by_id(opp_id)
```
mkangia, line 265: how come we skipped the check for slugs?

commcare_connect/opportunity/views.py
```diff
@@ -252,34 +257,36 @@
...
 
-    raise Http404("Opportunity not found.")
+    if not opportunity:
+        raise Http404("Opportunity not found.")
```
mkangia: nit: translation?

commcare_connect/opportunity/views.py
```diff
@@ -588,14 +601,16 @@ def review_visit_export(request, org_slug, opp_id):
     return redirect(f"{redirect_url}?export_task_id={result.id}")
 
 
-@org_member_required
+@login_required
```
mkangia, line 608: how is this same as checking for standard access? is that the default access?

mkangia: how about if `opp_standard_access_required` is decorated with login_required or we still add `opp_standard_access_required` for uniformity?

commcare_connect/organization/decorators.py
```diff
@@ -23,89 +33,103 @@ def user_is_org_admin(user, organization):
...
+        return _get_decorated_function(view_func, has_required_access)
 
+    return decorator
 
-def _request_user_is_viewer(request):
-    return (request.org and request.org_membership) or request.user.has_perm(ALL_ORG_ACCESS)
 
+def _opportunity_gate(has_required_access, opp_id_kwarg="opp_id"):
+    def decorator(view_func):
+        def permission_test(request, *args, **kwargs):
```
mkangia: nit: `permission_check` just to avoid confusions with tests.

mkangia (commented):
I tried reviewing all at once since there are quite a few commits now but that went bad.
I will go commit by commit.

I am trying to focus now on the feature and ignore naming as much as possible. I understand the naming was bad in the existing code but if we can try to avoid making it worse, that would be nice.

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
...
+    VIEW = 1
+    STANDARD = 2
+    MANAGE = 3
+
+    @staticmethod
+    def effective(org_level: "AccessLevel", user_level: "AccessLevel") -> "AccessLevel":
+        return min(org_level, user_level)
+
+
+def user_org_access(membership) -> AccessLevel:
```
mkangia: nit: `membership_access_level`

This function standalone, is being passed a membership and has nothing to do with a user so naming this `membership_access_level` seems like the cleanest and most clear name for this.

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
...
+    if not membership:
```
mkangia, line 32: nit: should it instead be an error if membership is not passed instead of short circuiting to None?

commcare_connect/program/utils.py
```diff
@@ -1,7 +1,42 @@
...
+        return AccessLevel.MANAGE
+    if membership.is_member:
+        return AccessLevel.STANDARD
+    if membership.is_viewer:
+        return AccessLevel.VIEW
+    return AccessLevel.NONE
+
+
+def org_program_access(org, program) -> AccessLevel:
+    if not org or not program:
```
mkangia, line 45: nit: should it instead be an error if org and program are not passed instead of short circuiting to None?

commcare_connect/organization/decorators.py
```diff
@@ -66,56 +68,76 @@ def has_permission(self, request, view):
...
-def opportunity_pm_required(view_func):
+def opportunity_pm_required(view_func, *args, **kwargs):
     return _get_decorated_function(view_func, lambda request: request.is_opportunity_pm)
 
 
+def _program_access_level_gate(minimum, program_id_kwarg="pk"):
+    def decorator(view_func):
+        def has_required_access(request, *args, **kwargs):
+            program_id = kwargs.get(program_id_kwarg)
+            program = Program.objects.filter(program_id=program_id).first()
```
mkangia: do we want to worry about if there is no program returned?

commcare_connect/opportunity/views.py
ajeety4: Good question.

The earlier function was doing two things:

* Finding the opportunity by opportunity ID.
* Checking whether the organization in the request was allowed to access the opportunity.

This has now been simplified so that the function is responsible only for fetching the opportunity, following the Single Responsibility Principle. The access function handles the authorization separately, including checking the organization in the request.

commcare_connect/opportunity/views.py
ajeety4: As part of these changes, access for this function (and a couple of other functions - see [commit](https://github.com/dimagi/commcare-connect/pull/1453/commits/729e62028a43ff4b4c96d9a92a7f2f81886993d6)) is checked inline because the access check requires the opportunity ID.

So, instead of using the access decorator, the function uses the [direct access](https://github.com/dimagi/commcare-connect/pull/1453/files#diff-1914cafdbf59040f7f77df353660e2634bcf3ec8e789b018d9864bc6ee7fded2R612) check. `login_required` is still used to handle authentication, while access to the opportunity is checked inline using the opportunity ID.

commcare_connect/program/utils.py
ajeety4: I agree that `membership_access_level` describes the immediate input/output of the function, but I don't think it's necessarily clearer in this context. The membership here specifically represents a user's membership in an organization, and this function calculates the user's access level to that organization from the membership.

So I intentionally used `user_org_access` _(renamed later to `user_access_for_org`)_ to capture what the value represents rather than just what object is passed into the function.

commcare_connect/program/utils.py
ajeety4: `request.org_membership` can legitimately be None when the current user doesn't have a `UserOrganizationMembership` row for request.org (e.g. a superuser/ALL_ORG_ACCESS user, or a user viewing an org they don't belong to). This is a normal "no role in this org" case, so I think it should resolve to NONE rather than raise an error.

commcare_connect/program/utils.py
ajeety4: I would prefer to keep this the same shape as `user_access_for_org` for consistency.

commcare_connect/organization/decorators.py
ajeety4: In case the program is `None`, `program_access_level_from_request(request, None`) reaches _resource_access_level's i
```
if not resource:
 return AccessLevel.NONE,
 
```
  so NONE >= minimum is False and access is correctly denied rather than raising an error, so I don't think an extra guard is needed here.

commcare_connect/organization/decorators.py
ajeety4: Good catch. Thanks. Removed in af0910d497d1bf837fc0dbcee42bfdbc5f7ad5d2

commcare_connect/program/utils.py
ajeety4: af0910d497d1bf837fc0dbcee42bfdbc5f7ad5d2

commcare_connect/program/utils.py
ajeety4: I have removed the terms : `program side` and `delivery org` considering that they are causing confusion and not adding much value.

ajeety4:
> I tried reviewing all at once since there are quite a few commits now but that went bad. I will go commit by commit.
> 
> I am trying to focus now on the feature and ignore naming as much as possible. I understand the naming was bad in the existing code but if we can try to avoid making it worse, that would be nice.

Thanks, Manish, for the naming suggestions. They were quite helpful, and I’ve implemented some of them.

I do disagree with the characterization that the changes in this PR are making the naming worse, though. My intention with these changes has been to make the access-related concepts clearer, and I feel the overall clarity is better than what was there before. I’m happy to continue improving the naming where we can.

commcare_connect/program/utils.py
ajeety4: Good idea. Asked for suggestions

commcare_connect/opportunity/views.py
mkangia: okay, thanks for the details :+1:

commcare_connect/opportunity/views.py
mkangia: Cool, thank you for sharing the context.

commcare_connect/program/utils.py
mkangia: well put, that makes sense, the naming is based on what the function returns :+1: 

`user_access_for_org` naming change definitely makes this better.

commcare_connect/program/utils.py
mkangia: Cool.

mkangia:
> I do disagree with the characterization that the changes in this PR are making the naming worse, though. My intention with these changes has been to make the access-related concepts clearer, and I feel the overall clarity is better than what was there before. I’m happy to continue improving the naming where we can.

Okay @ajeety4 :+1: Thanks for considering the naming suggestions

commcare_connect/program/utils.py
mkangia: Looks like the generic renaming got skipped for tests.

commcare_connect/program/utils.py
```diff
@@ -37,6 +38,24 @@ def org_program_access(org, program) -> AccessLevel:
...
+
+    return AccessLevel.effective(org_level, user_org_access(request.org_membership))
+
+
+def _base_access_level(request) -> AccessLevel | None:
+    if not request.org:
+        return AccessLevel.NONE
+    if request.user.has_perm(ALL_ORG_ACCESS):
+        return AccessLevel.MANAGE
+    return None
```
mkangia, line 120: nit: I believe this is not needed since a function automatically returns None but okay if you want to keep it explicit here.

commcare_connect/opportunity/views.py
mkangia: nit: just checking how this works with `OpportunityObjectMixin` now added which expects an opportunity to be present but an opportunity is yet to be added in this view right?

mkangia: also can you help me understand one more thing.
We initially were checking if the organization had program manager access or not with `OrgPMRequiredMixin` but now we are just checking if the user has manage access with no check on the org? So is that change now exposing this view to non program manager orgs as well?

commcare_connect/organization/tests/test_permissions.py
```diff
@@ -15,24 +20,24 @@ class TestAllOrgAccessPermission:
...
             return HttpResponse("OK")
 
-        @org_viewer_required
+        @org_view_access_required
         def dummy_viewer_view(request, org_slug):
             return HttpResponse("OK")
 
-        # Dummy class-based vie
-        class DummyOrgViewerView(OrgPMRequiredMixin, View):
+        class DummyAdminMixinView(OrgManageAccessMixin, View):
```
mkangia: just confirming that the name said "ViewerView" but this was actually for admins so we are now naming this "AdminMixinView"?

commcare_connect/organization/decorators.py
```diff
@@ -29,16 +29,15 @@ def user_is_org_admin(user, organization):
     ).exists()
 
 
-def user_is_program_manager_by_slug(user, org_slug):
-    """Check if user is admin of a program-manager org identified by slug, or has ALL_ORG_ACCESS."""
+def user_is_org_pm(user, organization):
```
mkangia: nit: `is_user_pm_org_admin`, just a naming suggestion.

commcare_connect/organization/decorators.py
```diff
@@ -29,16 +29,15 @@ def user_is_org_admin(user, organization):
...
-    return UserOrganizationMembership.objects.filter(
-        user=user,
-        organization__slug=org_slug,
-        organization__program_manager=True,
-        role=UserOrganizationMembership.Role.ADMIN,
-    ).exists()
+    return org_is_program_manager(organization) and user_is_org_admin(user, organization)
+
+
+def org_is_program_manager(organization):
```
mkangia, line 42: 👍 

What I like about this is that it is correctly stating that an org is program manager and not the user.

commcare_connect/organization/decorators.py
```diff
@@ -62,14 +61,14 @@ def user_is_opportunity_pm(user, opportunity):
...
-    """DRF permission: user must be admin of a program-manager organization."""
+class IsProgramManagerOrgAdmin(BasePermission):
+    """DRF twin of org_pm_required. There is no request.org here, so the acting org comes from the slug."""
 
     def has_permission(self, request, view):
         org_slug = request.data.get("organization") or view.kwargs.get("org_slug")
         if not org_slug:
             return False
-        return user_is_program_manager_by_slug(request.user, org_slug)
+        return user_is_org_pm(request.user, Organization.objects.filter(slug=org_slug).first())
```
mkangia: nit: Might be worth caching a look for org by slug considering this will fire frequently for views.

commcare_connect/organization/decorators.py
```diff
@@ -85,9 +84,11 @@ def _request_user_is_admin(request, *args, **kwargs):
...
 
 def is_org_pm_or_all_access(request, *args, **kwargs):
-    return (
-        request.org and request.org_membership and request.org_membership.is_admin and request.org.program_manager
-    ) or request.user.has_perm(ALL_ORG_ACCESS)
+    """Same rule as user_is_org_pm."""
+    if request.user.has_perm(ALL_ORG_ACCESS):
+        return True
+    membership = request.org_membership
+    return org_is_program_manager(request.org) and bool(membership and membership.is_admin)
```
mkangia: nice 👍

commcare_connect/organization/decorators.py
```diff
@@ -148,6 +168,10 @@ def has_required_access(request, *args, **kwargs):
 org_standard_access_required = _org_access_level_gate(AccessLevel.STANDARD)
 org_manage_access_required = _org_access_level_gate(AccessLevel.MANAGE)
 
+# The party gates ask which side of the opportunity the org sits on, not how much access it has.
```
mkangia: nit: what are "party" gates?

commcare_connect/program/utils.py
```diff
@@ -48,6 +50,33 @@ def program_access_level_from_request(request, program) -> AccessLevel:
...
+    if base_access is not None:
+        return base_access
+
+    org_level = org_opportunity_access(request.org, opportunity)
+
+    return AccessLevel.effective(org_level, user_org_access(request.org_membership))
+
+
+def org_opportunity_access(org, opportunity) -> AccessLevel:
+    """The one delivering it, supervising it and the one running its program. Watcher has view access."""
```
mkangia: nit: is this comment trying to clarify when this function would be applicable or used?

commcare_connect/program/utils.py
```diff
@@ -48,6 +50,33 @@ def program_access_level_from_request(request, program) -> AccessLevel:
...
+    if not org or not opportunity:
+        return AccessLevel.NONE
+    if org.id in (opportunity.organization_id, opportunity.supervising_organization_id):
+        return AccessLevel.MANAGE
+    return org_program_access(org, opportunity.program)
```
mkangia: I wonder if this should be `None`? Not sure for now why the fallback is to org program access.

commcare_connect/program/utils.py
```diff
@@ -66,23 +95,24 @@ def _base_access_level(request) -> AccessLevel | None:
...
         (request.org_membership != None and request.org_membership.is_admin) or request.user.is_superuser  # noqa: E711
     )
 
 
-def is_opportunity_pm(request, opp_id) -> bool:
-    managed_opp = get_managed_opp(opp_id)
-    return bool(managed_opp and managed_opp.program.organization.slug == request.org.slug and is_org_pm(request))
+def is_opportunity_nm(request, opportunity) -> bool:
+    """The network manager is the org delivering the opportunity."""
+    return _can_manage_opportunity(request, opportunity) and request.org.id == opportunity.organization_id
```
mkangia, line 145: this would also be true for the organization that owns the opportunity as well right?

mkangia: hmm, I also see `is_opp_pm` which is checking for the opposite, I am not sure how these two functions are working.

commcare_connect/organization/decorators.py
```diff
@@ -99,15 +99,19 @@ def _request_user_is_viewer(request, *args, **kwargs):
     return (request.org and request.org_membership) or request.user.has_perm(ALL_ORG_ACCESS)
 
 
-def org_member_required(view_func, *args, **kwargs):
+# TODO: These 3 are widely used in the opportunity app, and we are renaming it to opportunity access decorators
+# in this commit for easier review. The next commit points at them at opportunity_access_level_from_request.
```
mkangia: 👍

commcare_connect/opportunity/views.py
ajeety4: > checking how this works with OpportunityObjectMixin

Thanks for catching this. This was a noop. Removed it.

>  So is that change now exposing this view to non program manager orgs as well?

Yes, this is intended, but only for Funder orgs, since they can also create opportunities for programs they have access to.

The `ProgramManageAccessMixin` takes care of allowing both the Program Manager org and the Funder org for the specific program.

The change also makes it easier to think in terms of the resource itself: anyone with manage access to a program can create an opportunity for that program. We don't need to worry about the specific org or user at this layer, since that's the abstraction the access layer provides.

commcare_connect/organization/tests/test_permissions.py
ajeety4: The [viewer](https://github.com/dimagi/commcare-connect/pull/1453/files/3d2713039198f7f9108e5ad927aabf93824ac2af#diff-0b97bee7078af181844427feb6677f89b84bd8413c4ffb95eec1437d59b56797L35) one is still present. 
This was added in place of earlier [DummyOrgMemberView](https://github.com/dimagi/commcare-connect/pull/1453/files/3d2713039198f7f9108e5ad927aabf93824ac2af#diff-0b97bee7078af181844427feb6677f89b84bd8413c4ffb95eec1437d59b56797L35) . I ’m guessing the reorganization of the classes made the change appear differently in the diff.

Also, there’s no test for the member class-based view here, since we don’t have a mixin for it. It isn’t needed anywhere at the moment.

commcare_connect/organization/decorators.py
ajeety4: Agreed.

This has been a confusing aspect for me as well. However, I didn’t consider renaming it at this point, since the concept of “user is PM of <resource>”, especially for opportunities, is also used in discussions and tickets, and not just in the code.

commcare_connect/program/utils.py
ajeety4: I think the intent was to clarify who has access to the opportunity, but I agree it doesn't add much value here, so I'll remove it.

commcare_connect/program/utils.py
ajeety4: So this is how the opportunity access works:
- The NM org and the supervisor has full access to the opportunity directly.
- The PM org , the funder org and the watcher org access to the opportunity depends on their access level to the program of that opportunity. The function `org_program_access` takes care of that.

commcare_connect/program/utils.py
ajeety4: > this would also be true for the organization that owns the opportunity as well right?

Not really. This specifically checks whether the organization is the NM org i.e. the organization running the opportunity and handling the day-to-day delivery operations.

In short:
- `is_opportunity_pm`: refers to any org that has full access to the opportunity. The Funder and Supervisor admins have full access similar to the PM org, so they effectively act as a “PM” for the purpose of this check.
    ( _This is also the confusing name we've been discussing in this PR._)

- `is_opportunity_nm`: refers specifically to the org that runs the opportunity.

> I also see is_opp_pm which is checking for the opposite, I am not sure how these two functions are working.

Good question.

The opposite of `is_opportunity_pm` means a user of any org that does not act as a PM. This does include the NM org, but it can also include other orgs that don't have any access to the opportunity.

That's why `is_opportunity_nm` is useful - it makes the NM check explicit rather than relying on the inverse of is_opportunity_pm.

commcare_connect/program/utils.py
ajeety4: I think the intent was to clarify who has access to the opportunity, but I agree it doesn't add much value here, so I'll remove it.
https://github.com/dimagi/commcare-connect/pull/1453/commits/ddd94ad4c93fe005b680f6accb0cfaf1596f8f30

commcare_connect/organization/decorators.py
ajeety4: This only backs ProgramCreateView.post (program creation which is low-traffic), and slug is a unique indexed lookup, so I feel caching here isn't worth the complexity.

commcare_connect/organization/decorators.py
ajeety4: As this comment didn't add much value, removed. https://github.com/dimagi/commcare-connect/pull/1453/commits/ddd94ad4c93fe005b680f6accb0cfaf1596f8f30

commcare_connect/organization/decorators.py
ajeety4: https://github.com/dimagi/commcare-connect/pull/1453/commits/ddd94ad4c93fe005b680f6accb0cfaf1596f8f30

commcare_connect/opportunity/views.py
mkangia: Cool, thanks.

Is essentially the program manager, now just limited to creating a program? it seems like everything else is now available to both PM and NM?

commcare_connect/organization/decorators.py
mkangia: > However, I didn’t consider renaming it at this point

Not sure I understand. I just wanted to highlighted a name that I felt was good with no intention of any renaming.

commcare_connect/opportunity/views.py
ajeety4: > Is essentially the program manager, now just limited to creating a program? 

Yes this is correct.

>  it seems like everything else is now available to both PM and NM?

This is true for opportunity, however there are few functionalities divided among PM and NM.
e.g only NM is allowed to create invoice and only PM is allowed to approved an invoice

commcare_connect/organization/decorators.py
ajeety4: My bad.  I should have been specific.
I meant to say for other names that are associated with PM . for e,g `is_user_opportunity_pm`.

commcare_connect/program/utils.py
mkangia: Not requesting for any change here but this is a great example of what I have been trying to highlight in this PR, so hopefully it would be a useful feedback, if it makes sense to you.

If I were to write this function, it could be written as 

```
def org_access_level_to_opportunity(org, opportunity):
    """
       org access level if it has direct access to the opportunity
       or
       org access level if it has access to the program of the opportunity
    """
    if not org or not opportunity:
        return AccessLevel.NONE
    if org.id in (opportunity.organization_id, opportunity.supervising_organization_id):
        return AccessLevel.MANAGE
    else:
        return org_program_access(org, opportunity.program)
```

This could be a personal preference but its nice to read a code that flows well and does not need a follow up thought chain to understand it.
The fact that you had to explain to me, tells that maybe its a bit too complex to unfurl or is using context that isn't available when reading this function.

commcare_connect/program/utils.py
mkangia: > so they effectively act as a “PM” for the purpose of this check.

Right. When I see PM I think of the "Program Mananger" check on an Organization and you are referring to the user role which is a product concept that does not exist in code. That sort of mixup could lead to confusion.

Possibly not related, but reminds me of [something](https://github.com/dimagi/commcare-connect/pull/1505/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R85) I read on a PR earlier.

mkangia: Offlined.

Opp org is NM org always. NM is a concept of the product that seems to have a minor presence in code.

If this is to be made better we could have,

```
def is_opportunity_nm_with_manage_access(request, opportunity):
    return is_opportunity_nm(request.org.id, opportunity) and _can_manage_opportunity(request, opportunity)


# this function could also live in opp utils
def is_opportunity_nm(org_id, opportunity) -> bool:
    """The network manager is the org delivering the opportunity."""
    return org_id == opportunity.organization_id
```

mkangia (approved):
I think this is good to go now. 

I see there are some existing issues with code that this PR brought to highlight which we can address as a followup to improve in the entire codebase.

commcare_connect/opportunity/views.py
mkangia: >  e.g only NM is allowed to create invoice and only PM is allowed to approved an invoice

Can a funder ~~NM~~ non-program-manager-org or a supervising ~~NM~~ non-program-manager-org, approve invoices?

commcare_connect/organization/decorators.py
mkangia: okay :+1:

commcare_connect/opportunity/views.py
ajeety4: Yes, as they have same level of access as the PM org.

mkangia (approved):
offlined that renaming was discussed with team and we agreed on changing access level manage to admin.

commcare_connect/opportunity/views.py
```diff
@@ -1828,7 +1827,7 @@ def invoice_list(request, org_slug, opp_id):
     )
 
 
-class InvoiceCreateView(OppNMRequiredMixin, OpportunityObjectMixin, CreateView):
+class InvoiceCreateView(OppStandardAccessMixin, OpportunityObjectMixin, CreateView):
```
mkangia, line 1830: The code change looks like we removed the check for NM and added Standard Access check but then limited the response to NM org.

The commit says  "for only nm org standard users".
What did the code change functionally?

ajeety4: The functional change is that `OppNMRequiredMixin` required the user to have **admin access** to the opportunity and the org to be an NM org.

This change replaces the admin-access requirement with **standard access** via `OppStandardAccessMixin`, while still requiring the org to be an NM org.
(I removed `OppNMRequiredMixin` because of the confusion around the term and moved the NM-org check inline)

So functionally, it changes the requirement from **NM org + admin access** to **NM org + standard access**.

mkangia (approved):

mkangia:
:tada:

PR #1457, opened 2026-08-17
Repoint LLO Entity Surfaces to Organization

pxwxnvermx:
## Product Description

- Removed LLO Entity model, and tests related to that model.
- Created new Organization Creation flow with profile details.
- Updated the report filters to use organization to filter.
- Removed CreateableModelChoice Field utility and tests. This was not getting used anywhere.
- Removed the validation on Year Of Establishment field on the model. It was validating from 1800 to current year + 1 and this validation would need a migration every year to keep it up to date. This has been moved to the form now instead.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2562)

### Safety story
Tested Locally

### Automated test coverage
- Automated Tests have been updated.

### QA Plan
- None

### Labels & Review


commcare_connect/organization/migrations/0014_remove_organization_llo_entity_delete_lloentity.py
```diff
@@ -0,0 +1,20 @@
...
+        ("organization", "0013_backfill_organization_profile_from_entity"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name="organization",
+            name="llo_entity",
+        ),
+        migrations.DeleteModel(
+            name="LLOEntity",
```
mkangia, line 18: wont this impact requests during the deploy once the table is removed but still referenced in the live code?
So, I believe we should drop the table as a follow up.

commcare_connect/users/models.py
```diff
@@ -79,6 +79,7 @@ class Meta:
             ("product_features_access", "Can access and manage product features (flags and switches)"),
             ("manage_internal_permissions", "Can manage internal permissions for users"),
             ("invoice_report_access", "Access the Invoice reports dashboard"),
+            ("llo_entity_internal_access", "Can access the internal LLO organization profile export"),
```
mkangia, line 82: nit: as I am reviewing this PR, going commit wise because this seems a big PR, I see this permission has been added but I can't understand why this is needed and how/where this will be used. This is a feedback on PR structure where this commit would look better if one could also where the permission would be used, or a ToDo comment sharing how/where this will be used to better understand the intention behind this permission.

commcare_connect/reports/helpers.py
```diff
@@ -103,14 +103,16 @@ def get_table_data_for_year_month(
...
+    organization=None,
     opportunity=None,
     country=None,
     period=None,
 ):
     """Return KPI report rows aggregated by month or quarter.
 
     Args:
+        organization: Rollups are per-workspace until the duplicate workspaces that used
+                      to share an LLO entity are merged.
```
mkangia, line 115: can you clarify what this means?
what happens for two organizations that used to have the same LLO?

commcare_connect/users/admin.py
```diff
@@ -72,20 +73,31 @@ class UserOrganizationMembershipInline(admin.TabularInline):
...
-    list_filter = ["program_manager", "funder"]
+    list_filter = ["program_manager", "funder", "verified", "has_used_connect"]
+    filter_horizontal = ["countries"]
+
+    def get_queryset(self, request):
+        return super().get_queryset(request).annotate(_member_count=Count("members", distinct=True))
+
+    @admin.display(description=_("Members"), ordering="_member_count")
+    def member_count(self, obj):
+        return obj._member_count
```
mkangia, line 97: why are these changes being made to the Organization on admin?
I could not find these on the ticket or the PR description.

commcare_connect/organization/models.py
```diff
@@ -34,11 +29,7 @@ class Organization(BaseModel):
     funder = models.BooleanField(default=False)
     short_name = models.CharField(max_length=40, null=True, blank=True)
     has_used_connect = models.BooleanField(default=False)
-    year_of_establishment = models.PositiveSmallIntegerField(
-        null=True,
-        blank=True,
-        validators=[MinValueValidator(1800), MaxValueValidator(_current_year() + 1)],
```
mkangia, line 40: why was this not needed anymore?

mkangia (changes requested):
I'd request that this PR is simplified for review. 

There is a lot happening,  so 
1. keeping scope limited
2. using smaller commits with helpful commit message to understand things better

would be good.
I could review up till half and then it was a bit too much to understand and review.

commcare_connect/organization/models.py
pxwxnvermx: This validator is incorrect, it fixes our year validations between 1800 to Current Year + 1 (2027), so every year we would need to create a new migration to fix the validation. I have moved this validation from the database level to be on the form level.

commcare_connect/reports/helpers.py
pxwxnvermx: Currently nothing happens, both organizations would share the Short Name field from the LLO Entity.

commcare_connect/organization/migrations/0014_remove_organization_llo_entity_delete_lloentity.py
pxwxnvermx: Thanks, I forgot to consider that. I will create a seperate PR for removing the LLO Entity model.

commcare_connect/users/admin.py
pxwxnvermx: Happy to remove these, mostly just utility based changes for the admin.

commcare_connect/users/models.py
pxwxnvermx: The commits are sequential for this change, the permission was added right before it is getting used in the next commit for the data serializer.

commcare_connect/organization/models.py
mkangia: Thank you.

I feel this is useful context. Having this on the commit message or in the tech summary of the PR would have been helpful for review.

commcare_connect/users/admin.py
mkangia: I don't want to discourage from making improvements and adding useful changes.

My concern was that it was not mentioned in the ticket or the PR description ("also made some utility changes to the org admin which will make it easier to ...").
Additionally, keeping these in a separate commit message helps isolate the requirements from other additional benefits during review.

So, I won't ask you to remove it. If you think its useful, you could keep it in a separate commit and share in the PR description why this would be useful.

commcare_connect/users/models.py
mkangia: > The commits are sequential for this change, the permission was added right before it is getting used in the next commit for the data serializer.

Right. I noticed when I moved to the next commit and that is definitely better than a random order.

During review, seeing functionally how the code would sit together is really helpful and is mostly feasible.
For example in this case,

```
def serializer(*args, **kwargs):
    #...
    if _user_has_permission(user):
        # allow
    else:
        # reject

def _user_has_permission(user):
    # ToDo: add permission check
    return True
```
and in the following commit add definition for `_user_has_permission`

When reviewing I find this pattern really useful & I have also seen being used by other devs. Seeing code that is written but not used, leads to a review of the code in isolation. Here, in the snippet above, `_user_has_permission` will be added in isolation in the following commit but it's need & signature are clear and defined beforehand which helps link it to the rest of the code. I'd recommend using the above pattern and giving it a try.

commcare_connect/organization/models.py
pxwxnvermx: Added info to the decsription.

pxwxnvermx:
I am going to close this PR and create two new PRs.

1. For the changes to repoint all surfaces from LLO Entity to Organization
2. For removing the LLO Entity model.

There were a few changes merged for the Org Creation form recently which are also in conflict with my current changes, having a new PR makes it easier to resolve those as well. I can base my changes on top of those changes.

FYI @mkangia

pxwxnvermx:
Updated PRs here

1. https://github.com/dimagi/commcare-connect/pull/1466 
2. https://github.com/dimagi/commcare-connect/pull/1467

mkangia:
Thanks for the flexibility @pxwxnvermx

PR #1460, opened 2026-08-17
Program & Opportunity Access Redesign UI

mkangia:
## Product Description

Three fields already existed in the data model with no user interface. Two live on Program, one on Opportunity; all three point at an Organization.

These are now visible on the UI to be set, all behind the `enable_program_access_redesign` switch:

| Field                      | On                        |Set on |
| -------------------------| -----------------------|-----------------------|
| Funder                  | Program             |Create only        |
| Watchers              | Program             |Create & Edit     |
| Supervising Org  | Opportunity      |Create & Edit      |

funder and supervising_organization are single selects; watchers is a multi-select.

- Program funder — organizations marked as funders, excluding the program's own.
- Program watchers — all organizations except the program's own and its funder.
- Opportunity supervising organization — the program's organization, its funder, or any organization with an accepted ProgramApplication for that program. Defaults to the program's organization, reproducing today's implicit behaviour.

Changes to these three fields is audited.

Two supporting changes added to this PR to make the above usable:

- Organization.funder is now settable in the Django admin (this was missed in the earlier ticket that was to add this)
- Changes to all three fields are exposed on the Django admin for reference

With the switch off, none of the three fields render though the audits will happen since they are not gated behind a switch.

Screenshots pending — see QA Plan.

## Technical Summary

Design doc: docs/superpowers/specs/2026-08-17-program-access-redesign-design.md (included in this PR).

No model changes. Program.funder, Program.watchers (migration 0016) and Opportunity.supervising_organization (migration 0140) all already exist and are already read by the permission layer.

This PR is purely the missing UI, so saving through relevant forms takes effect against the existing permission code.
Only migrations added are for the adding audit tables & triggers.

The test suite shared a Redis cache with the local dev server, so toggling any waffle switch in the dev admin could fail unrelated tests. config/settings/test.py now uses LocMemCache.

Safety Assurance

Safety story

Everything is behind enable_program_access_redesign, which defaults inactive. With it off, all three fields are absent from self.fields, so they can be neither rendered nor submitted, and existing behaviour is bit-for-bit unchanged — including Opportunity.save() remaining the only thing that sets the supervising organization.

One pre-existing gap this PR does not close, flagged for the reviewer: opportunity_manage_org_ids returns supervising_organization_id without re-checking eligibility, and nothing resets the column when an application is rejected (program/views.py, program/api/views.py only write the status). So an organization set as supervisor and later rejected retains oversight until reassigned. This predates the change and isn't widened by it — the dropdown stops offering such an organization, so the only way to be in that state is to already be in it. Closing it means either re-checking eligibility in opportunity_manage_org_ids or resetting the column on rejection; both are permission-layer changes that belong in their own ticket.

## Automated test coverage

Tests added

## QA Plan

Needs a QA ticket. 
Suggested script: create a program with funder + watchers, reopen to confirm the funder is locked and its organization is absent from the watcher list; then invite and accept an organization, create a managed opportunity, confirm the supervisor defaults to the program's organization and offers exactly the program org / funder / accepted applicant; reassign it and confirm the previous organization loses access to that opportunity.

Labels & Review


mkangia:
Redone https://github.com/dimagi/commcare-connect/pull/1477

PR #1289, opened 2026-06-18
add org model update spec

calellowitz:
## Technical Summary

Tech Spec 2 of 3 for the permission and org structure updates [design doc](https://docs.google.com/document/d/1oORKTbPcjOUn2yr7jK5dbpvKEU8qNedBSdGpCl4QZh0/edit?tab=t.5bl3pzocqv7l)

This spec covers the removal of the LLO Entity model. We are hoping to move to a world where Organization refers to a real world organization without the need for a model above that. LLO entity was a nice try to unify these, but added more complexity than it helped and we no longer believe individual orgs need multiple "workspaces."

PR #1461, opened 2026-08-17
feat(organization): merge workspaces from the Django admin

Charl1996:
## Product Description

This PR adds a **superuser-only** `Merge selected workspaces` action to the Organizations changelist in the Django admin: tick exactly two, pick which one survives, review a side-by-side comparison, confirm.

The confirmation page is the point of the change — it shows the two profiles, a row count for every relation the merge would repoint, and an explicit "what gets discarded" block (the source's feature flag names, the membership-role conflict rule).

#### New action once orgs are selected from the organizations list
image

#### Details page
image
image
image
image

### Note
The "What gets discarded" section above shows this:
> Pending invites are carried over, but their emailed links die with the deleted workspace's slug.

Only the invite records are carried over from the source organization, but a user that is still to accept the invite will not be able to do so after the merge won't be able to and the new (merged) org's admin will have to resend invites.

We can choose to improve this process but I kept it out of the scope of this PR for now.

## Technical Summary

[CCCT-2563](https://dimagi.atlassian.net/browse/CCCT-2563)

`commcare_connect/organization/merge.py` is the service; the admin is a thin two-phase action over it. Decisions worth knowing before reading the diff:

- **The target wins** on every profile field, membership role, pending invite and `program_manager`/`funder` status.
- **Blocking merges** If a source org is a `program_manager` and the target org is not, the merge will be blocked. If a source is a `funder` and specified as a `funder` on program records but the target is not a `funder`, the merge is blocked. If a source and target has program applications for the same program but the applications have conflicting statuses, the merge is blocked. 
- **A CommCare app both workspaces are connected to refuses the merge**, before anything changes. `OpportunityInitForm` and `ManagedOpportunityCreateSerializer` both call `get_or_create(cc_app_id, cc_domain, organization, hq_server)` — a key the database does not enforce — so moving the source's apps across would raise `MultipleObjectsReturned` on every later opportunity create or edit for the survivor. Reconciling automatically would mean repointing six FKs including per-app `LearnModule` / `DeliverUnit` / `TaskType` rows, so a human picks the real app instead.
- **Feature flags are not transferred.** The source is explicitly removed from its own flags first, so `m2m_changed` fires and waffle's `flag:<name>:organizations` cache is flushed — a plain cascade delete would leave the source's pk sitting in that cache.
- **`HANDLED_RELATIONS` is a manifest of every relation pointing at `Organization`**, and a test asserts it matches the models. Adding a new FK to `Organization` without handling it in the merge fails CI rather than silently going unmigrated. The confirmation page's counts are generated from that same manifest, so the preview can't under-report what the merge would touch.

`OrganizationAdmin` moves from `users/admin.py` into the app that owns the model. Admin URL names derive from the model's `app_label`, so they're unchanged.

### Safety story

Blast radius is limited to Django admin superusers; nothing in the user-facing app calls the service. The whole merge runs in one transaction, so any error leaves the database untouched — there's a test that injects a failure mid-merge and asserts every relation is still pointing at the source afterwards. `MergeNotAllowed` is raised before any write, and surfaces in the admin as a red message rather than a 500.

### Automated test coverage

- `commcare_connect/organization/tests/test_merge.py` (53 tests) covers the service.
- `commcare_connect/organization/tests/test_admin.py` (12 tests) covers the action.

### QA Plan

QA needed

### Deployment

### Labels & Review

[CCCT-2563]: https://dimagi.atlassian.net/browse/CCCT-2563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+    watched = list(source.watched_programs.all())
+    target.watched_programs.add(*watched)
+    source.watched_programs.clear()
+    return len(watched)
+
+
+def _merge_program_applications(source: Organization, target: Organization) -> tuple[int, int]:
+    """Move applications, keeping one row per program at the most advanced status.
+
+    ``(program, organization)`` is not unique in the database but ``invite_organization`` relies on it, calling
```
zandre-eng: Should we enforce this unique constraint on the DB level? If our code already assumes this state it would make sense to ensure our DB stays in an expected state.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def get_actions(self, request):
+        actions = super().get_actions(request)
+        if not request.user.is_superuser:
+            actions.pop(MERGE_ACTION, None)
+        return actions
+
+    @admin.action(description="Merge selected workspaces (irreversible)")
+    def merge_workspaces(self, request, queryset):
+        if not request.user.is_superuser:
```
zandre-eng, line 95: What do you think about adding a regression test to validate this auth that only superusers can merge orgs?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+    OrganizationReassignment("opportunity", "Opportunity", "supervising_organization"),
+    OrganizationReassignment("opportunity", "Payment", "organization"),
+    OrganizationReassignment("opportunity", "LabsRecord", "organization"),
+    OrganizationReassignment("program", "Program", "organization"),
+    OrganizationReassignment("program", "Program", "funder"),
+)
+
+# Every relation pointing at Organization. test_merge.py asserts this matches the models, so adding a new foreign
+# key to Organization without handling it here fails CI.
+HANDLED_RELATIONS = frozenset(
```
zandre-eng, line 54: This asserts that all the relations for `Organization` are present in this set, however nothing validates that the relation is actually getting repointed. 

We could probably assume if a dev is adding a new entry here that they would make the necessary changes to handle the repoint as well, but if a dev forgets to do that then CI will be green and the new relation will silently be deleted instead of repointed.

I feel this is probably fine, but just thought I'd bring it up in case we want to be strict here.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+            "opportunity creation. Remove the redundant app from one workspace first."
+        )
+
+
+def _reassign_simple_relations(source: Organization, target: Organization) -> dict[str, int]:
+    reassigned = {}
+    for relation in SIMPLE_REASSIGNMENTS:
+        count = relation.reassign(source, target)
+        reassigned[relation.label] = count
+        logger.info("Reassigned %s %s rows from %s to %s", count, relation.label, source.slug, target.slug)
```
zandre-eng, line 228: Minor nit: This is doing an individual DB update for each reassignment. Given how small the list is this is fine, but this could also be refactored to just update all the fields first and then do a single DB save at the end.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+    }
+)
+
+APPLICATION_STATUS_PRECEDENCE = [
+    ProgramApplicationStatus.REJECTED,
+    ProgramApplicationStatus.DECLINED,
+    ProgramApplicationStatus.INVITED,
+    ProgramApplicationStatus.APPLIED,
+    ProgramApplicationStatus.ACCEPTED,
+]
```
zandre-eng: Would it make sense to move this to `program/models.py` where `ProgramApplicationStatus` gets defined? I'm just thinking if `ProgramApplicationStatus` ever gets changed then it might not be obvious that this list will need to be updated as well.

commcare_connect/templates/admin/organization/merge_confirmation.html
```diff
@@ -0,0 +1,126 @@
...
+              </td>
+            {% endfor %}
+          </tr>
+        </tbody>
+      </table>
+      <ul class="help">
+        <li>
+          Feature flags are <strong>not</strong> transferred. The surviving workspace keeps only its own.
+        </li>
+        <li>Where a user belongs to both workspaces, the surviving workspace's role wins.</li>
```
zandre-eng: Nit: It might be worth making explicit here that if the user does not belong to both workspaces, then they will be granted source-side roles over the survivor's data.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+    moved = 0
+    discarded = 0
+    for invite in source.invites.all():
+        is_live = invite.status == OrganizationInvite.Status.INVITED and not invite.is_expired
+        if not is_live or invite.email.lower() in taken_emails:
+            discarded += 1
+            continue
+
+        invite.organization = target
+        invite.save(update_fields=["organization"])
```
zandre-eng: Minor nit: Instead of doing individual saves here you could add the invites that need updating to a list and then do a single bulk update at the end of the loop.

zandre-eng (commented):
I took an initial pass, and the changes generally looks good. Just a few questions and nits from my side.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,164 @@
...
+            for label, field in PROFILE_FIELDS
+        ],
+        "count_rows": [(label, [count[label] for count in counts]) for label in sorted(HANDLED_RELATIONS)],
+        "flag_names": [sorted(organization.flag_set.values_list("name", flat=True)) for organization in organizations],
+        "hidden_programs": _hidden_programs_per_survivor(organizations),
+    }
+
+
+def _hidden_programs_per_survivor(organizations: list[Organization]) -> list[list[str]]:
+    """One cell per column, answering "which programs go dark if I keep this workspace?".
```
mkangia: not sure I understand this.

and, can an opportunity have more than one program? I am confused why this says "programs"

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,164 @@
...
+    merge_organizations,
+    programs_hidden_by_merge,
+    relation_counts,
+)
+from commcare_connect.organization.models import Organization, UserOrganizationMembership
+from commcare_connect.users.forms import OrganizationCreationForm
+
+MERGE_ACTION = "merge_workspaces"
+
+# Fields the target keeps and the source loses, so an admin can see what is about to be discarded.
```
mkangia: nit: I find it strange that the comment says that these are the fields that will be retained but the name of the constant holding the values is called "PROFILE_FIELDS".

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,164 @@
...
+        return actions
+
+    @admin.action(description="Merge selected workspaces (irreversible)")
+    def merge_workspaces(self, request, queryset):
+        if not request.user.is_superuser:
+            raise PermissionDenied
+
+        selected = queryset.order_by("pk")
+        if selected.count() != 2:
+            self.message_user(request, "Select exactly two workspaces to merge.", messages.WARNING)
```
mkangia, line 100: nit: Should not this be an error instead of warning, since we are blocking on it and not letting proceed?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+    OrganizationReassignment("opportunity", "Opportunity", "organization"),
+    OrganizationReassignment("opportunity", "Opportunity", "supervising_organization"),
+    OrganizationReassignment("opportunity", "Payment", "organization"),
+    OrganizationReassignment("opportunity", "LabsRecord", "organization"),
+    OrganizationReassignment("program", "Program", "organization"),
+    OrganizationReassignment("program", "Program", "funder"),
+)
+
+# Every relation pointing at Organization. test_merge.py asserts this matches the models, so adding a new foreign
+# key to Organization without handling it here fails CI.
```
mkangia, line 53: very cool 🆒

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,275 @@
...
+        "organization.OrganizationInvite.organization",
+        "organization.UserOrganizationMembership.organization",
+        "program.Program.funder",
+        "program.Program.organization",
+        "program.Program.watchers",
+        "program.ProgramApplication.organization",
+    }
+)
+
+APPLICATION_STATUS_PRECEDENCE = [
```
mkangia: nit: I am not confident that this belongs in this file and not with the model ProgramApplicationStatus.

mkangia (commented):
took a quick pass but will try to do another better one

commcare_connect/organization/admin.py
Charl1996: Do you prefer a rename to something like "FIELDS_TO_KEEP" or so?

commcare_connect/organization/admin.py
Charl1996: I'll update the comment. I realize sometimes some of these docstrings are "high context" making it difficult for reviewers to really understand.

commcare_connect/organization/merge.py
Charl1996: Fair point - I thought since this is merge-specific it makes sense to put it here, but since both you and Manish have raised this I'll move it.

commcare_connect/organization/admin.py
Charl1996: I'll update the comment.

> can an opportunity have more than one program

No - I'll make it more clear in the comment

commcare_connect/organization/admin.py
mkangia: I think the name for the constant in itself is okay.
Its combination of it with the comment that is confusing.

If the fields are defined specifically to be used for things to retain, then renaming the constant and dropping the comment entirely seems good.

If we just want a mapping of the fields with their display / humanized value, then the way it is currently seems good but the comment would need to rephrased or be skipped. If one wants to find out the fields that are retained, they would look at the method that has the retaining logic. I believe even now one would look at the method first and then trace to this mapping, making the comment redundant.

Charl1996: Oh, I think I get you. I'll remove the comment if you feel like it does not add the intended value.

commcare_connect/organization/merge.py
Charl1996: Good point. I'll think about this one a bit.

commcare_connect/templates/admin/organization/merge_confirmation.html
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/commits/02d0e586d8efe06ef2deb64dce10f5c07ff86eed

commcare_connect/organization/merge.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/changes/e6afb63f52861038ca94c0d22f4f3e394f5d11db

commcare_connect/organization/admin.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/commits/007199f0707c6e04e6fb278ed965ac89dc48f76a

commcare_connect/organization/merge.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/changes/c497c6f6516d70c6d697e3573a55e1a6b0d47fac

commcare_connect/organization/merge.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/changes/c497c6f6516d70c6d697e3573a55e1a6b0d47fac

commcare_connect/organization/merge.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/commits/0c737c3646e71c4f3527fec36335c903bc0be195

commcare_connect/organization/admin.py
Charl1996: I feel like warning is good here since, although we block someone, the result is not due to an error that happened but rather a "misuse" of the feature.

mkangia:
Hey @Charl1996 

(non blocking but worth a follow up if needed)
Do we see any possible issues if we delete the organization but its slug is left lingering somewhere in tables and then the same slug is reused later? Or we don't use slug for any relation?
Just thinking if its safer to archive rather than delete.

mkangia:
Hey @Charl1996 

> Feature flags are not transferred. The source is explicitly removed from its own flags first

Can you share the  intention for doing this? I am not able to understand why this would be needed instead of just leaving the assigned flags.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    """Merge ``source`` into ``target``, then delete ``source``.
+
+    Raises ``MergeNotAllowed`` before any change is made; any other exception rolls the whole merge back.
+    """
+    _reject_invalid_merge(source, target)
+    stale_opportunities = _opportunities_cached_against(source)
+
+    with transaction.atomic():
+        reassigned = _reassign_simple_relations(source, target)
+        watched = _move_program_watchers(source, target)
```
mkangia: nit: considering the naming around, this could be "moved". I'd recommend `moved_watchers_count`.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+            applications_moved=apps_moved,
+            applications_deduped=apps_deduped,
+            self_applications_removed=self_apps,
+            memberships_moved=members_moved,
+            memberships_discarded=members_dropped,
+            invites_moved=invites_moved,
+            invites_discarded=invites_dropped,
+            flags_cleared=flags_cleared,
+        )
+        source.delete()
```
mkangia, line 126: what happens if this delete fails?

Charl1996:
@mkangia 

> > Feature flags are not transferred. The source is explicitly removed from its own flags first
> 
> Can you share the intention for doing this? I am not able to understand why this would be needed instead of just leaving the assigned flags.

This is a [very specific requirement](https://github.com/dimagi/commcare-connect/blob/5b63b8210b2073af274d2d08e1004f84e3da045d/docs/superpowers/specs/2026-06-17-llo-entity-merge-and-org-merging-design.md#:~:text=Feature%20flags%3A%20the%20target%20org%27s%20Flag%20memberships%20are%20deliberately%20not%20modified%20by%20the%20merge.%20The%20source%20org%27s%20flag%20rows%20vanish%20with%20its%20deletion.%20Opportunities%20reassigned%20from%20the%20source%20fall%20under%20whatever%20flags%20the%20target%20org%20already%20has%20%E2%80%94%20no%20per%2Dopportunity%20flag%20handling%20needed.) from the spec.

I think the question is probably better asked there.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+
+    logger.info("Merged workspace %s into %s: %s", summary.source_slug, summary.target_slug, summary)
+    return summary
+
```
mkangia, line 130: Looks like the ordering of functions is off. I was expecting to see `_reject_invalid_merge` here.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+def relation_counts(organization: Organization) -> dict[str, int]:
+    counts = {}
+    for label in HANDLED_RELATIONS:
+        app_label, model_name, field_name = label.split(".")
+        model = apps.get_model(app_label, model_name)
+        counts[label] = model._default_manager.filter(**{field_name: organization}).count()
+    return counts
+
+
+def programs_hidden_by_merge(source: Organization, target: Organization) -> list[str]:
```
mkangia: Does it even make sense to merge a PM into NM?
Why not block it at the start itself? so that the user decides if they want to upgrade the target to PM.
Or this is acceptable end result as of now?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    if source.pk is None or target.pk is None:
+        raise MergeNotAllowed("Both organizations must be saved before they can be merged.")
+    if source.pk == target.pk:
+        raise MergeNotAllowed("An organization cannot be merged into itself.")
+    _reject_shared_commcare_apps(source, target)
+
+
+def _reject_shared_commcare_apps(source: Organization, target: Organization) -> None:
+    """Refuse a merge that would leave the target with two rows for one HQ app."""
+    app_key = ("cc_app_id", "cc_domain", "hq_server_id")
```
mkangia, line 145: nit: If you control the ordering in this app_key, I would suggest using the same ordering as you need elsewhere.

Visually its easy to miss that later the domain is displayed first.

Charl1996:
> Do we see any possible issues if we delete the organization but its slug is left lingering somewhere in tables and then the same slug is reused later? Or we don't use slug for any relation? Just thinking if its safer to archive rather than delete.

Good question! We don't store the org slug on any other tables so we don't have to worry about unintentional data access when recycling org slugs.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    shared = set(source.apps.values_list(*app_key)) & set(target.apps.values_list(*app_key))
+    if shared:
+        conflicts = sorted({f"{cc_domain}/{cc_app_id}" for cc_app_id, cc_domain, _ in shared})
+        raise MergeNotAllowed(
+            f"Both workspaces are connected to the same CommCare app(s): {', '.join(conflicts)}. "
+            "Merging would leave the surviving workspace with duplicates that break "
+            "opportunity creation. Remove the redundant app from one workspace first."
```
mkangia, line 152: How does one remove the application? Additionally, it is a safe operation for the user to do or there are side effects of it that we should communicate.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    return len(watched)
+
+
+def _merge_program_applications(source: Organization, target: Organization) -> tuple[int, int]:
+    """Move applications, keeping one row per program at the most advanced status."""
+    target_applications = {app.program_id: app for app in target.programapplication_set.all()}
+
+    moved = 0
+    deduped = 0
+    for application in source.programapplication_set.all():
```
mkangia: nit: I'd suggest naming this to program_application instead of application since we also have CommCare Application (this has confused me once already in existing code (not this PR) when reading)

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+        existing = target_applications.get(application.program_id)
+        if existing is None:
+            application.organization = target
+            application.save(update_fields=["organization"])
```
mkangia: just confirming that this fires any related calls or signals? There might be none today but might be added in future.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+            existing.status = surviving_status
+            existing.save(update_fields=["status"])
+        application.delete()
+        deduped += 1
+
+    return moved, deduped
+
+
+def _most_advanced_status(*statuses: str) -> str:
+    return max(statuses, key=APPLICATION_STATUS_PRECEDENCE.index)
```
mkangia: if a program application was rejected in source but accepted in target, we would now have it as accepted?

Similarly, if a program application was accepted in source but rejected in target, we would now have it as accepted?

Either or both of these look concerning to me since it either gives unintended access or revokes access.

I wonder if program applications statuses don't have a logical "sequence". 

Option would be to block the merge if there are applications to the same program in both orgs.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+def _remove_self_applications(target: Organization) -> int:
+    """Drop applications an organization holds to a program it now owns.
```
mkangia: nit: so this is remove program applications and not applications. This bit me again.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+
+    Repointing ``Program.organization`` can leave the target holding an application to its own program, a row
+    ``invite_organization`` refuses to create in the first place.
+    """
+    deleted, _ = ProgramApplication.objects.filter(organization=target, program__organization=target).delete()
+    return deleted
+
+
+def _merge_memberships(source: Organization, target: Organization) -> tuple[int, int]:
```
mkangia, line 269: nit: can we add a documentation to this function for what it does functionally to achieve the merge?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    target_user_ids = set(target.memberships.values_list("user_id", flat=True))
+    source_memberships = source.memberships.all()
+
+    discarded, _ = source_memberships.filter(user_id__in=target_user_ids).delete()
```
mkangia, line 277: I believe we need to also consider the role the user had so we don't give a stronger membership?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+
+def _move_pending_invites(source: Organization, target: Organization) -> tuple[int, int]:
+    """Move the live invites the target has no equivalent for"""
+    taken_emails = {email.lower() for email in target.invites.values_list("email", flat=True)}
+    taken_emails |= {email.lower() for email in target.memberships.values_list("user__email", flat=True) if email}
+
+    moving = []
+    discarded = 0
+    for invite in source.invites.all():
+        is_live = invite.status == OrganizationInvite.Status.INVITED and not invite.is_expired
```
mkangia, line 291: nit: 
easier to read & also less error prone
```
is_live = (invite.status == OrganizationInvite.Status.INVITED and not invite.is_expired)
```

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+        if not is_live or invite.email.lower() in taken_emails:
+            discarded += 1
+            continue
+
+        moving.append(invite.pk)
```
mkangia: nit: just a naming suggestion, `to_be_moved`.

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,289 @@
...
+    flags = list(source.flag_set.all())
+    for flag in flags:
+        flag.organizations.remove(source)
+    return len(flags)
+
+
+def _opportunities_cached_against(source: Organization) -> list[Opportunity]:
+    """Opportunities whose ``get_managed_opp`` entry holds a reference to the source.
+
+    That cache stores the opportunity with its program and the program's organization attached, so a merge
```
mkangia: what cache are we referring to here?

mkangia: okay I understand it now, `get_managed_opp` is cached. May be this can be simplified

```
opportunities_linked_to_source_before_move = _opportunities_linked_to_source(source)


_clear_opportunity_caches(opportunities_linked_to_source_before_move)
```

```
def _opportunities_linked_to_source(source: Organization):
    return list(
        Opportunity.objects.filter(
            Q(organization=source)
            | Q(supervising_organization=source)
            | Q(program__organization=source)
            | Q(program__funder=source)
        ).only("id", "opportunity_id")
    )


def _clear_opportunity_caches(opportunities: Sequence[Opportunity]) -> None:
    # clear cached get_managed_opp for relevant opportunities that still point to the source

    for opportunity in opportunities:
        clear_managed_opp_cache(opportunity)
```

I would think of this being implemented like the above, feel free to take it exactly or parts of it.

The function that fetches the linked orgs does not need to talk about why it should fetch them.
I believe that context is relevant and clear either by the function calls sequence and naming in the function that is orchestrating the merge.

mkangia (commented):
Have reviewed the merge script which is the core logic, on to the rest now.

mkangia:
Hey @calellowitz 
Can you help understand this?

https://github.com/dimagi/commcare-connect/pull/1461#issuecomment-5393555401

Primarily to understand why the flags need to be removed from the source before deletion.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,161 @@
...
+    ("Created", "date_created"),
+    ("Created by", "created_by"),
+]
+
+
+def workspace_label(organization: Organization) -> str:
+    return f"{organization.name} ({organization.slug})"
+
+
+def merge_preview(organizations: list[Organization]) -> dict:
```
mkangia: nit: its always just two organizations right?
Might want to consider keeping that clear with a signature that has source, target like rest of the feature since we are anyway never merging more than 2.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,161 @@
...
+        ],
+        "count_rows": [(label, [count[label] for count in counts]) for label in sorted(HANDLED_RELATIONS)],
+        "flag_names": [sorted(organization.flag_set.values_list("name", flat=True)) for organization in organizations],
+        "hidden_programs": _hidden_programs_per_survivor(organizations),
+    }
+
+
+def _hidden_programs_per_survivor(organizations: list[Organization]) -> list[list[str]]:
+    """
+    Returns which programs will be hidden a given organization is to be elected as the surviving org.
```
mkangia: nit: not sure I understand this statement.

is it "will be hidden for a given organization"?

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,161 @@
...
+            (label, [getattr(organization, field) for organization in organizations])
+            for label, field in PROFILE_FIELDS
+        ],
+        "count_rows": [(label, [count[label] for count in counts]) for label in sorted(HANDLED_RELATIONS)],
+        "flag_names": [sorted(organization.flag_set.values_list("name", flat=True)) for organization in organizations],
+        "hidden_programs": _hidden_programs_per_survivor(organizations),
+    }
+
+
+def _hidden_programs_per_survivor(organizations: list[Organization]) -> list[list[str]]:
```
mkangia: nit: I wonder if this logic is better suited for merge.py rather than here since its relevant specifically to the operation

mkangia: Similar comment about limiting to `source, target` instead of a list of organizations.

Plus, introducing a new concept of a "survivor" does add additional complexity. Sticking to "source" and "target" would be ideal.

commcare_connect/program/migrations/0018_programapplication_unique_program_application_per_organization.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("program", "0017_delete_managedopportunity"),
+    ]
+
+    operations = [
+        migrations.AddConstraint(
+            model_name="programapplication",
+            constraint=models.UniqueConstraint(
+                fields=("program", "organization"),
```
mkangia, line 15: Can an organization re-apply?

commcare_connect/templates/admin/organization/merge_confirmation.html
```diff
@@ -0,0 +1,128 @@
+{% extends "admin/base_site.html" %}
+{% block breadcrumbs %}
```
mkangia, line 2: nit: if there a better way to do this without overriding the entire block?
or maybe skip it? I wonder if this is unnecessary maintenance if this block is modified in future django versions.

commcare_connect/templates/admin/organization/merge_confirmation.html
```diff
@@ -0,0 +1,128 @@
...
+{% endblock %}
+{% block content %}
+  <p class="errornote">
+    Merging is irreversible. Every relation below is repointed at the workspace you keep, and the other workspace is
+    deleted. Run it once and check the result &mdash; there is no undo.
+  </p>
+  <form method="post">
+    {% csrf_token %}
+    <input type="hidden" name="action" value="{{ action }}">
+    {% for pk in selected_pks %}<input type="hidden" name="_selected_action" value="{{ pk }}">{% endfor %}
```
mkangia, line 18: nit: is this intended as expected? I wonder if the indentation was reviewed by the relevant library.

commcare_connect/templates/admin/organization/merge_confirmation.html
```diff
@@ -0,0 +1,128 @@
...
+            <tr>
+              <th scope="row">{{ label }}</th>
+              {% for value in values %}<td>{{ value|default_if_none:"&mdash;" }}</td>{% endfor %}
+            </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+      <p class="help">
+        Only the surviving workspace's profile is kept. Nothing above is carried over from the workspace you delete
+        &mdash; including <code>program_manager</code> and <code>funder</code>.
```
mkangia, line 46: not sure I understand the bit about things being "carried over". What do you mean by not carrying over program_manager and funder?

commcare_connect/templates/admin/organization/merge_confirmation.html
```diff
@@ -0,0 +1,128 @@
...
+      </table>
+      <ul class="help">
+        <li>
+          Feature flags are <strong>not</strong> transferred. The surviving workspace keeps only its own.
+        </li>
+        <li>
+          Where a user belongs to both workspaces, the surviving workspace's role wins, otherwise the workspace which the user has membership on.
+        </li>
+        <li>
+          A survivor that is not a <strong>Program manager</strong> inherits the other workspace's programs but hides
```
mkangia: nit I would strongly suggest sticking to source, target and merge.
I see the use of "inherits" which conceptually is also different.

I suggest "An organization that is a Program Manager merges into another workspace that is not a Program Manager, results into Programs being migrated but hidden, along with the relevant visits, until the target workspace is updated to be have Program Manager access."

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,161 @@
...
+            program
+            for other in organizations
+            if other.pk != survivor.pk
+            for program in programs_hidden_by_merge(other, survivor)
+        )
+        for survivor in organizations
+    ]
+
+
+class OrganizationMergeForm(forms.Form):
```
mkangia, line 49: Can you help understand how the selection works here? I am having a hard time understanding this from the code. We have just one target but somehow still have two selected orgs?
It would be ideal for someone to understand this without needing to look at the screenshots in the relevant PR or seeing things on a UI.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,161 @@
...
+        if not request.user.is_superuser:
+            raise PermissionDenied
+
+        selected = queryset.order_by("pk")
+        if selected.count() != 2:
+            self.message_user(request, "Select exactly two workspaces to merge.", messages.WARNING)
+            return None
+
+        confirming = "confirm" in request.POST
+        form = OrganizationMergeForm(request.POST if confirming else None, selected=selected)
```
mkangia, line 104: in what case the selection will still be present but confirming would be False?

mkangia (commented):
Completed pass of all code except tests which I can see once the changes have been finalized.

It's a complex one so i took the opportunity to also share some PR design level feedback, hope that helps.

commcare_connect/organization/merge.py
Charl1996: All the changes are rolled back. The whole merge happens as an atomic transaction.

commcare_connect/organization/merge.py
Charl1996: I'm curious to get @calellowitz 's input here. 

> Does it even make sense to merge a PM into NM?

I think the answer is probably no and I'm open to making the call now to block this use-case, but Cal might have more insights into how this will be used in the future.

commcare_connect/organization/merge.py
Charl1996: I'm not sure I follow?

commcare_connect/organization/merge.py
Charl1996: > How does one remove the application?

I don't think there's a standard way today, but I know we don't allow the same app on the same domain so we should definitely block here.

commcare_connect/organization/merge.py
Charl1996: Yes, signals would still fire.

commcare_connect/organization/merge.py
Charl1996: > I wonder if program applications statuses don't have a logical "sequence".

What do you mean here? Isn't `APPLICATION_STATUS_PRECEDENCE` doing that?

commcare_connect/organization/merge.py
Charl1996: The user keeps the target role here, so no stronger membership can be given if the user has a stronger one in `source`.

commcare_connect/organization/admin.py
Charl1996: > its always just two organizations right?

Correct.

Good point.

commcare_connect/organization/admin.py
Charl1996: My mistake - I edited this comment and clearly didn't reread it carefully.

commcare_connect/program/migrations/0018_programapplication_unique_program_application_per_organization.py
Charl1996: I don't think we have a flow for that today.

commcare_connect/templates/admin/organization/merge_confirmation.html
Charl1996: "Carried over" as in "moved from source to target". 

> What do you mean by not carrying over program_manager and funder?

This refers to the status an org can have, so if an org is a `program_manager` and is merged into another org the resultant org won't be `program_manager` just because this one was (we don't even look at these flags on the source org) - same for `funder`.

commcare_connect/organization/admin.py
Charl1996: This form is used on the confirmation page where the user needs to select which select org will survive (should be the "target").

commcare_connect/organization/admin.py
Charl1996: The `confirm` flag exists because a Django admin action has no URL of its own but is dispatched from the changelist's POST handler. If we want a confirmation page (e.g. this one that shows the summary) we need to render one from the POST handler and make sure this confirmation page POSTs back to the same endpoint. 

The problem however is that the function needs something in the POST body to tell "the user just picked this action" from "the user has now confirmed it" - that's the `confirm` flag in the POST.

This is how django's delete action also works (i.e. no dedicated URL to show the confirmation) so I've followed this same pattern. However, I would be fine to split out the changelist page and the confirmation page to sit at 2 different endpoints if you think that would be helpful?

commcare_connect/program/migrations/0018_programapplication_unique_program_application_per_organization.py
mkangia: Cool, thanks for confirming.

commcare_connect/organization/merge.py
```diff
@@ -230,6 +230,10 @@ def _remove_self_program_applications(target: Organization) -> int:
 
 
 def _merge_memberships(source: Organization, target: Organization) -> tuple[int, int]:
+    """Move the source's memberships, dropping the ones for users the target already has.
```
mkangia, line 270: sorry can you expand on why the target membership is dropped?

commcare_connect/organization/admin.py
mkangia: Thanks. This is all very useful context. So the POST is being passed to clarify the target & to confirm whereas the selection is for the first step (and may be for second as well i.e the confirmation?). 

Looking at this in the UI was helpful. I think without that, just understanding from code is tricky but I leave that to you if you want to make it easier to understand.

mkangia (commented):
Thanks for addressing the feedback.

Just one pending query here, (and one new comment on the PR)

for the organization being merged in, if it is supervisor for a program, we are giving that access to the target organization. Just confirming that is what we need? Its an indirect access outside the organization being merged in, so just want to be sure.

also, similarly if the organization was a funder to a program, we are making the target a funder. But I think we are not marking the target as a funder by setting the funder flag on the organization which will conflict? I am assuming my understanding here is correct since on Django Admin I see ` program.Program.funder`  under related records but you have mentioned we won't modify the flags: program manager and funder.

commcare_connect/organization/merge.py
Charl1996: It's actually the other way around - we're dropping the `source` membership for a user if the user already have a `target` membership.

Here's the line of code that does that:
```
discarded, _ = source_memberships.filter(user_id__in=target_user_ids).delete()
```

Charl1996:
> for the organization being merged in, if it is supervisor for a program, we are giving that access to the target organization. Just confirming that is what we need?

In short the answer is yes. Currently we're not granting access or anything based on `supervising_organization` (except reading it in the data export), but I think the point of the merge is to "move the responsibilities of one org to another".

> if the organization was a funder to a program, we are making the target a funder. But I think we are not marking the target as a funder by setting the funder flag on the organization

That is true - good catch! It does seem to conflict conceptually if an org designated as a program funder has `organization.funder` set to `False`. 

@calellowitz the spec says we should *not* transfer `funder` status, so does that extend to `program.funder`?

commcare_connect/organization/merge.py
mkangia: Aha, understood.

mkangia:
Hey @Charl1996 

> In short the answer is yes. Currently we're not granting access or anything based on supervising_organization (except reading it in the data export)

A supervising orgranization will have a view access to the organization right? not just limited to exports? Either way, it is additional access beyond the source org, which I wanted to check about.

calellowitz:
> @calellowitz the spec says we should not transfer funder status, so does that extend to program.funder?

I suspect this is a rare edge case, so maybe good to fail with an explicit error and the person running the command can decide whether they need to mark the target org as a funder before doing the copy. I think the program role should definitely transfer if both orgs are designated funders.

commcare_connect/organization/merge.py
calellowitz: Similar to the funder answer, I think failing with a useful error message and letting the person running the command decide whether to make the target a PM makes the most sense to me.

calellowitz:
> Primarily to understand why the flags need to be removed from the source before deletion.

I don't think that is what the spec is saying. The spec says that deleting the org will delete the flag rows (through cascading deletes of foreign keys). As far as I understand, there is no need to delete the flags before the merge, but it is correct that they will not transfer over.

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,159 @@
...
+from commcare_connect.users.forms import OrganizationCreationForm
+
+MERGE_ACTION = "merge_workspaces"
+
+PROFILE_FIELDS = [
+    ("Name", "name"),
+    ("Slug", "slug"),
+    ("Program manager", "program_manager"),
+    ("Funder", "funder"),
+    ("LLO entity", "llo_entity"),
```
hemant10yadav: I don’t think this model exists on main anymore, right?

commcare_connect/organization/merge.py
```diff
@@ -0,0 +1,297 @@
...
+    logger.info("Merged workspace %s into %s: %s", summary.source_slug, summary.target_slug, summary)
+    return summary
+
+
+def _reject_invalid_merge(source: Organization, target: Organization) -> None:
+    if source.pk is None or target.pk is None:
+        raise MergeNotAllowed("Both organizations must be saved before they can be merged.")
+    if source.pk == target.pk:
+        raise MergeNotAllowed("An organization cannot be merged into itself.")
+    _reject_shared_commcare_apps(source, target)
```
hemant10yadav, line 137: Are we comfortable allowing an NM org and PM org to be merged? I don’t think anything technically breaks, but there are some subtle side effects worth considering.

If the NM org is deleted and its opportunities move to the PM org, the surviving org could end up approving and paying its own invoices, and vice versa. The same applies to visit reviews, where members act as NM and admins as PM within the same org. Notifications could also become confusing, with PM members being notified about NM actions and vice versa.

Also, `Payment.organization` represents the org that was paid as the NM, so after a PM↔NM merge, historical payments could appear as the org paying itself.

Nothing here seems harmful, but I wanted to make sure we’ve explicitly considered this case before allowing the merge.

hemant10yadav (commented):
Looks good overall. I left two comments for consideration: one regarding a model that no longer exist on main, and another around the NM/PM org merge edge case.

commcare_connect/organization/admin.py
Charl1996: You're right. That was recent changes and I've not updated with `main` since then. I'll update

Charl1996:
@mkangia @calellowitz Please note [this commit](https://github.com/dimagi/commcare-connect/pull/1461/commits/3803accece4525a389b5a76345627bb010763cfe) where we will now block the merge if the target org is not a funder and the source org was a `program.funder`. Note however that if the `target_org` is a `funder` the merge would still allow the `program.funder` to be updated from `source_org` -> `target_org`.

commcare_connect/organization/merge.py
Charl1996: I think the short answer is "yes" in light of [this spec](https://github.com/dimagi/commcare-connect/pull/1289/changes).

commcare_connect/organization/merge.py
```diff
@@ -150,6 +151,24 @@ def _reject_shared_commcare_apps(source: Organization, target: Organization) ->
         )
 
 
+def _reject_funded_programs_a_non_funder_would_inherit(source: Organization, target: Organization) -> None:
+    """Refuse a merge that would make a workspace fund programs without being designated a funder.
+
+    ``Program.funder`` is repointed like any other relation, but ``Organization.funder`` is a profile field the
```
mkangia, line 171: what do we mean by "repointed"? We don't change the Program funder right?

commcare_connect/organization/merge.py
mkangia: @calellowitz  just tagging you so you are aware of this decision making here.

mkangia: Actually @Charl1996 

I see this response from Cal earlier, https://github.com/dimagi/commcare-connect/pull/1461#discussion_r3917374214 and I think he is suggesting to avoid NM-PM merge.

commcare_connect/organization/merge.py
Charl1996: > We don't change the Program funder right?

Not if the target is not a `funder`, no. If the target is a `funder` organization we do.

mkangia: aha, is this talking about the funder for the workspace?! I read it as changing funder attribute of a Program which got me confused and hence worried.

mkangia:
I believe we are mostly ready with this, just waiting for @Charl1996 to confirm if any changes are needed for https://github.com/dimagi/commcare-connect/pull/1461#discussion_r3933848705

commcare_connect/organization/merge.py
Charl1996: Thanks for flagging that comment! I completely missed it

Charl1996: https://github.com/dimagi/commcare-connect/pull/1461/commits/04237b12029733172b4ad17224b39bb29e48ce15

hemant10yadav (approved):

commcare_connect/organization/admin.py
```diff
@@ -44,21 +43,9 @@ def merge_preview(source: Organization, target: Organization) -> dict:
         ],
         "count_rows": [(label, [count[label] for count in counts]) for label in sorted(HANDLED_RELATIONS)],
         "flag_names": [sorted(organization.flag_set.values_list("name", flat=True)) for organization in organizations],
-        "hidden_programs": _hidden_programs_per_candidate(source, target),
```
mkangia: I like it when things are simplified 🥇

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,145 @@
...
+    list_display = ["name", "created_by", "program_manager", "funder"]
+    search_fields = ["name"]
+    ordering = ["name"]
+    inlines = [UserOrganizationMembershipInline]
+    list_filter = ["program_manager", "funder"]
+    actions = [MERGE_ACTION]
+
+    def get_actions(self, request):
+        actions = super().get_actions(request)
+        if not request.user.is_superuser:
```
mkangia, line 89: nit: just curious but not against the double check here, the Admin end is visible only to superusers right?

commcare_connect/organization/admin.py
```diff
@@ -0,0 +1,145 @@
...
+        source_pk, source_repr = source.pk, str(source)
+        try:
+            summary = merge_organizations(source, target)
+        except MergeNotAllowed as error:
+            self.message_user(request, str(error), messages.ERROR)
+            return None
+
+        LogEntry.objects.log_action(
+            user_id=request.user.pk,
+            content_type_id=get_content_type_for_model(target).pk,
```
mkangia, line 120: I believe you want to use "source" here?

commcare_connect/organization/merge.py
mkangia: nit: Can you confirm if users actually can delete the linked commcare application? Just wanted to check that they can actually handle this conflict if it occurs or it would be the admin who would need to assist

commcare_connect/organization/merge.py
mkangia: My concern is with that logically the statuses are not in an order.

>  if a program application was rejected in source but accepted in target, we would now have it as accepted?

This example is what I am referring to.

commcare_connect/organization/admin.py
Charl1996: That's true. This is maybe slightly redundant, but given the impact of the action I like the extra pre-caution.

mkangia (commented):
Just one thing about moving program applications to considered. Rest looks good.

commcare_connect/organization/admin.py
```diff
@@ -77,7 +77,7 @@ class UserOrganizationMembershipInline(admin.TabularInline):
 @admin.register(Organization)
 class OrganizationAdmin(admin.ModelAdmin):
     form = OrganizationCreationForm
-    list_display = ["name", "created_by", "program_manager", "funder"]
+    list_display = ["name", "short_name", "slug", "created_by", "program_manager", "funder", "verified"]
```
mkangia: so we replaced `has_used_connect` with `slug`. No worries, we can add it back if someone needs it.

commcare_connect/organization/admin.py
Charl1996: I don't think it matters, since both content_types will be the same for `source` and `target` (i.e. `Organization`)

Charl1996:
> Just wanted to check that they can actually handle this conflict if it occurs or it would be the admin who would need to assist

As far as I know it definitely needs an admin to assist.

Charl1996:
> if a program application was rejected in source but accepted in target, we would now have it as accepted?

This is true - the target's status wins.

commcare_connect/organization/admin.py
mkangia: it's just confusing to see "target" in a logic that is being run entirely for the source.

mkangia:
> As far as I know it definitely needs an admin to assist.

sounds good. We can see how this would be done once we get there.

mkangia:
> This is true - the target's status wins.

Offlined. @Charl1996 is working on a change based on offline discussions.

mkangia (approved):
Thanks for the follow up @Charl1996

PR #1463, opened 2026-08-17
Centralize HQ Case Create Flag

zandre-eng:
## Product Description

No user-facing changes. This is an internal refactor of the CommCare HQ bulk case-update
helpers; the requests Connect sends to HQ are byte-for-byte identical to before.

## Technical Summary

[Link to ticket here](https://dimagi.atlassian.net/browse/CCCT-2511)

Follow-up on the remaining open review feedback from
[#1174](https://github.com/dimagi/commcare-connect/pull/1174).

HQ's case API treats `create` as a per-row flag, so every caller of
`bulk_create_or_update_cases` was hand-stamping it onto each case dict it assembled. That
put a correctness-critical decision (`None` = upsert keyed on `external_id` vs `False` =
update-only) in four places where it was easy to omit on a new row and invisible at the
call site. It is now a single `create` parameter on the helper, injected into every row of
every chunk, so each call site states its intent once.

`bulk_update_usercases` derives the domain, API key and HQ server from the first entry and
applies them to the whole batch. That contract was documented but unenforced, so a
mixed-opportunity batch would have posted the remainder to the wrong domain. It now raises
rather than silently mis-routing, and the docstring spells out the expected `updates` shape.

### Safety story

The refactor is behaviour-preserving at the HQ boundary — work-area upserts still send
`create: None` and every other path `create: False` — so no existing case data is affected.
The one behavioural change is a new `ValueError` guard that fires only on an input shape no
current caller produces, converting a silent wrong-domain write into a loud failure. Both
commits are independently revertible and neither touches models or migrations.

### Automated test coverage

Tests assert the `create` flag is injected into every row of the outgoing HQ payload across
all three flag values, that the work-area upsert path still requests `create=None`, and that
the usercase and unassign/exclude paths send the payload shape HQ expects. A new test covers
the mixed-opportunity guard, verifying it raises before any HQ request is issued.

### QA Plan

### Labels & Review

### DIFF
--- a/commcare_connect/commcarehq/api.py
+++ b/commcare_connect/commcarehq/api.py
@@ -184,11 +184,20 @@ def create_or_update_case(
 def bulk_update_usercases(updates: dict[OpportunityAccess, dict[str, Any]]) -> None:
     """Update usercase properties on CommCare HQ for multiple users in a single bulk request.
 
-    All entries in `updates` must belong to the same opportunity. The domain, API key, and
-    HQ server are derived from the first entry and applied to the entire batch.
+    `updates` maps each worker's `OpportunityAccess` to the case update to apply to their
+    usercase, in the shape HQ's case API expects, e.g.::
+
+        {access: {"properties": {"task_case_id": ""}}}
+
+    All entries must belong to the same opportunity, since the domain, API key, and HQ server
+    are derived from the first entry and applied to the entire batch.
     """
     if not updates:
         return
+
+    opportunity_ids = {access.opportunity_id for access in updates}
+    if len(opportunity_ids) > 1:
+        raise ValueError(f"All updates must belong to the same opportunity, got {sorted(opportunity_ids)}.")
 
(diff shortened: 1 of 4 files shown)


hemant10yadav (approved):

commcare_connect/commcarehq/api.py
```diff
@@ -136,13 +136,17 @@ def bulk_create_or_update_cases(
     api_key: HQApiKey,
     domain: str,
     cases_data: list[dict[str, Any]],
+    create: bool | None = False,
```
Charl1996, line 139: If `create` is `False` and the case does not exist, will HQ fail?

zandre-eng: That is right, the HQ request will fail and a `CommCareHQAPIException` exception will be raised in `bulk_create_or_update_cases`.

Charl1996 (approved):

PR #1466, opened 2026-08-21
Repoint LLO Entity Surfaces to Organization

pxwxnvermx:
## Product Description

- LLO Entity filter refactored to use Organization on Reports
- LLO Entity field removed from the OrganizationChangeForm on Organization Home page.
- Added new Organization Profile Form with new fields for the Organziation model.
- Removed unused form utils.
- Removed the validation on Year Of Establishment field on the model. It was validating from 1800 to current year + 1 and this validation would need a migration every year to keep it up to date. This has been moved to the form now instead.

The LLO Entity model will be removed after this PR is deployed.

## Technical Summary
[Ticket Link](https://dimagi.atlassian.net/browse/CCCT-2562)

### Safety story
Tested Locally

### Automated test coverage
- Automated Tests have been updated.

### QA Plan
- Will request QA.

### Labels & Review


commcare_connect/templates/layouts/header.html
```diff
@@ -16,8 +16,8 @@ <h1 class="text-lg text-brand-deep-purple font-medium">{{ header_title|default:"
             <div class="text-left">
               <div x-show="hasSelectedOrg" x-transition.opacity x-cloak>
                 <p class="text-sm font-medium text-brand-deep-purple cursor-pointer">{{ request.org.name }}</p>
-                {% if request.org.llo_entity %}
-                  <p class="text-gray-500 text-sm font-medium">{{ request.org.llo_entity }}</p>
+                {% if request.org.short_name %}
+                  <p class="text-gray-500 text-sm font-medium">{{ request.org.short_name }}</p>
```
zandre-eng, line 20: This field is nullable on the DB level. Will this ever be true in practice? If so, we might want to add a fallback here to the normal `org.name` field.

commcare_connect/organization/models.py
```diff
@@ -55,11 +50,7 @@ class Organization(BaseModel):
     llo_entity = models.ForeignKey(LLOEntity, on_delete=models.SET_NULL, null=True)
     short_name = models.CharField(max_length=40, null=True, blank=True)
     has_used_connect = models.BooleanField(default=False)
-    year_of_establishment = models.PositiveSmallIntegerField(
-        null=True,
-        blank=True,
-        validators=[MinValueValidator(1800), MaxValueValidator(_current_year() + 1)],
```
zandre-eng, line 61: Was there a specific reason for removing this validation from the DB?

commcare_connect/organization/forms.py
```diff
@@ -57,21 +51,164 @@ def __init__(self, *args, **kwargs):
...
+        x_cloak=True,
+    )
+
+
+def validate_year_of_establishment(year):
+    """Bounds the year against the calendar at call time, so the ceiling moves without a migration."""
+    if year is None:
+        return year
+    current = timezone.now().year
+    if year < 1800 or year > current:
```
zandre-eng: Minor nit: Could you replace the `1800` magic number with a variable const? It's pretty obvious what the magic number means in this context, but a const var does help with better self-documentation.

zandre-eng (approved):
The changes generally look fine, just some non-blocking feedback.

I see this isn't marked for QA. Given that this PR doesn't just remove references but also changes the org create form, do you think QA might be beneficial here?

commcare_connect/organization/models.py
pxwxnvermx: Yep, It was validating from 1800 to current year + 1 and this validation would need a migration every year to keep it up to date, so I removed it from the database level and moved it to the form level.

commcare_connect/templates/layouts/header.html
pxwxnvermx: The org name is already visible in the org selector, the short_name was added here to identify if there was a certain short way of mentioning the current LLO.

pxwxnvermx:
> Given that this PR doesn't just remove references but also changes the org create form, do you think QA might be beneficial here?

I have done local testing for this. But happy to send this for QA.

commcare_connect/organization/forms.py
pxwxnvermx: Yep, created a constant for this. https://github.com/dimagi/commcare-connect/pull/1466/commits/5f2bec602d66a92e69ed1941dc7084c22f531d77

pxwxnvermx:
@coderabbitai review

commcare_connect/organization/models.py
zandre-eng: Ah, that's a fair point. Thanks for the explanation.

Charl1996 (approved):

zandre-eng (approved):

commcare_connect/reports/views.py
```diff
@@ -58,9 +58,9 @@ class DeliveryReportFilters(django_filters.FilterSet):
         queryset=Program.objects.all(),
         label="Program",
     )
-    llo = django_filters.ModelChoiceFilter(
-        queryset=LLOEntity.objects.all().order_by("name"),
-        label="LLO",
+    organization = django_filters.ModelChoiceFilter(
+        queryset=Organization.objects.all().order_by("name"),
```
mkangia, line 62: nit: is this visible to a PM or NM or both?

commcare_connect/users/models.py
```diff
@@ -79,6 +79,7 @@ class Meta:
             ("product_features_access", "Can access and manage product features (flags and switches)"),
             ("manage_internal_permissions", "Can manage internal permissions for users"),
             ("invoice_report_access", "Access the Invoice reports dashboard"),
+            ("llo_entity_internal_access", "Can access the internal LLO organization profile export"),
```
mkangia, line 82: are we still keeping LLOEntity but internally? I am unable to understand why this permission was added. Could not find any mention of it in the PR description.

commcare_connect/templates/layouts/header.html
```diff
@@ -16,8 +16,8 @@ <h1 class="text-lg text-brand-deep-purple font-medium">{{ header_title|default:"
             <div class="text-left">
               <div x-show="hasSelectedOrg" x-transition.opacity x-cloak>
                 <p class="text-sm font-medium text-brand-deep-purple cursor-pointer">{{ request.org.name }}</p>
-                {% if request.org.llo_entity %}
-                  <p class="text-gray-500 text-sm font-medium">{{ request.org.llo_entity }}</p>
+                {% if request.org.short_name %}
```
mkangia, line 19: Why can't we remove this entirely?
If the organization full name is present, how come we need the short name under the full name?

commcare_connect/templates/organization/organization_create.html
```diff
@@ -18,6 +18,7 @@
   <div class="container md:max-w-screen-lg mx-auto">
     <form action="{% url 'organization_create' %}"
           method="post"
+          novalidate
```
mkangia, line 21: nit: can you share why this was added?

mkangia (commented):
Hey @pxwxnvermx 

Did we change entirely how an organization is created in this PR that was meant to change the references for LLOEntity to Organization?

Also, this commit seems to be doing a lot
https://github.com/dimagi/commcare-connect/pull/1466/changes/36b17fbdead8235ed9070aaed3718021eb1823cd

Would be worth breaking it down.

commcare_connect/templates/layouts/header.html
pxwxnvermx: This keeps the org header similar to the way it was before, the LLO Entity earlier also rendered the same thing.

commcare_connect/users/models.py
pxwxnvermx: This permission is mostly to gate the `llo_entity` data export API. Other Dimagi Systems are still referring to that API and the internal access mostly indicates that the access is for Dimagi.

commcare_connect/reports/views.py
pxwxnvermx: The reports are visible to Superusers, these reports are across multiple opportunities, organizations and programs so the PM and NM access boundaries are not applicable here.

commcare_connect/templates/layouts/header.html
mkangia: okay, I believe we had just the organization name before and the LLO was specifically added.
https://github.com/dimagi/commcare-connect/commit/79dbd5c19ec8e9f95191297e8b26b11f351573a6

So, was wondering why we can't go back to that itself.

commcare_connect/users/models.py
mkangia: that is useful context.

commcare_connect/reports/views.py
mkangia: thanks!

I recently read that we want to limit where we show a list of all organizations. I believe PMs already have access to it but NMs don't. So, just wanted to confirm.

pxwxnvermx:
Hi @mkangia 

> Did we change entirely how an organization is created in this PR that was meant to change the references for LLOEntity to Organization?

Yeah, the organization creation was changed entirely based on the changes proposed in the ticket description. There are more fields that contain details that need to be filled by a user while creating an organization.

> Also, this commit seems to be doing a lot
https://github.com/dimagi/commcare-connect/commit/36b17fbdead8235ed9070aaed3718021eb1823cd
Would be worth breaking it down.

Cool. I will break that commit into smaller chunks.

commcare_connect/templates/organization/organization_create.html
pxwxnvermx: This was one of the changes proposed by coderabbit. https://github.com/dimagi/commcare-connect/pull/1466#discussion_r3829604158

The validation blocks user from submitting the form if an error occurs in the earlier form fields. Submitting incorrect data would lead to validation errors, that will be automatically highlighted, so having the novalidate removes the default validation on the HTML form and triggers it when the button is pressed using our custom handlers.

mkangia: Cool. Our custom handlers in Javascript?

pxwxnvermx: Nope our handlers actually trigger the HTML form validation, so we don't need to trigger it by default.

mkangia:
> Cool. I will break that commit into smaller chunks.

Just noting that @pxwxnvermx and I offlined to agree to discard splitting the commit.

PR #1475, opened 2026-08-25
Count duplicate visits in total/verified deliveries counts

ajeety4:
## Product Description

Opportunity Dashboard List Page: Columns `Total Deliveries` and `Verified Deliveries` were not counting the duplicate approved deliveries. Same for `Services Delivered` in the single opportunity detail page.
This fixes them to include the duplicate counts, making it match the `Delivered` and `Approved` counts in the worker delivery table.

This makes it consistent across all the pages that displays it.

## Technical Summary

[Ticket](https://dimagi.atlassian.net/browse/CI-913)

`saved_completed_count`/`saved_approved_count` on `CompletedWork` already account for duplicates per entity. Switched the relevant aggregations (`OpportunityData.get_data_qs`, `get_deliveries_count_subquery` and its callers) from `Count(distinct CompletedWork.id)` to summing these fields, matching how the worker delivery table already computes its totals. Also removed the now-inaccurate "excluding duplicates" tooltip.

### Safety story

Read-only aggregation change on existing stored fields; no schema or write-path changes. No existing data is impacted, only how counts are computed for display.

### Automated test coverage

Updated/added tests in `test_helpers.py` and `test_views.py` covering `get_opportunity_delivery_progress`, `get_opportunity_worker_progress`, and `OpportunityData.get_data_qs` with a duplicate-approved-visit scenario, asserting totals include the duplicate.

### QA Plan

- Verify Total/Verified Deliveries on the program manager opportunity list and single-opportunity delivery stats/worker progress tiles match the worker delivery table's Delivered/Approved totals for an opportunity with duplicate approved visits.

### Labels & Review

### DIFF
--- a/commcare_connect/opportunity/helpers.py
+++ b/commcare_connect/opportunity/helpers.py
@@ -61,7 +61,8 @@ def inactive_workers_subquery(days_ago):
     return Coalesce(Subquery(subquery, output_field=IntegerField()), Value(0))
 
 
-def get_deliveries_count_subquery(status=None):
+def get_deliveries_count_subquery(status=None, sum_field=None):
+    agg = Sum(sum_field) if sum_field else Count("id", distinct=True)
     filters = {"opportunity_access__opportunity_id": OuterRef("pk")}
     if status is not None:
         filters["status"] = status
@@ -70,8 +71,8 @@ def get_deliveries_count_subquery(status=None):
         Subquery(
             CompletedWork.objects.filter(**filters)
             .values("opportunity_access__opportunity_id")
-            .annotate(count=Count("id", distinct=True))
-            .values("count"),
+            .annotate(count=agg)
+            .values("count")[:1],
             output_field=IntegerField(),
         ),
         0,
@@ -509,8 +510,10 @@ class OpportunityData:
                 CompletedWork.objects.filter(opportunity_access__opportunity_id__in=opp_ids)
                 .values("opportunity_access__opportunity_id")
                 .annotate(
-                    total_deliveries=Count("id", distinct=True),
-                    verified_deliveries=Count("id", filter=Q(status=CompletedWorkStatus.approved), distinct=True),
+                    total_deliveries=Coalesce(Sum("saved_completed_count"), Value(0)),
+                    verified_deliveries=Coalesce(
+                        Sum("saved_approved_count", filter=Q(status=CompletedWorkStatus.approved)), Value(0)
+                    ),
                 )
             )
             deliveries_by_opp = {
@@ -712,7 +715,7 @@ def get_opportunity_delivery_progress(opp_id):
         deliveries_from_yesterday=deliveries_from_yesterday_sq(),
         accrued_since_yesterday=accrued_since_yesterday_sq,
         most_recent_delivery=most_recent_delivery_sq,
-        total_deliveries=get_deliveries_count_subquery(),
+        total_deliveries=get_deliveries_count_subquery(sum_field="saved_completed_count"),
         recent_payment=recent_payment_sq,
         workers_invited=workers_invited_subquery(),
         pending_invites=pending_invites_subquery(),
@@ -729,8 +732,10 @@ def get_opportunity_worker_progress(opp_id):
     return (
         Opportunity.objects.filter(id=opp_id)
         .annotate(
-            total_deliveries=get_deliveries_count_subquery(),
-            approved_deliveries=get_deliveries_count_subquery(CompletedWorkStatus.approved),
+            total_deliveries=get_deliveries_count_subquery(sum_field="saved_completed_count"),
+            approved_deliveries=get_deliveries_count_subquery(
+                CompletedWorkStatus.approved, sum_field="saved_approved_count"
+            ),
             rejected_deliveries=get_deliveries_count_subquery(CompletedWorkStatus.rejected),
             total_accrued=total_accrued_sq(),
             total_paid=total_paid_sq(),
--- a/commcare_connect/opportunity/tests/test_helpers.py
+++ b/commcare_connect/opportunity/tests/test_helpers.py
@@ -327,7 +327,12 @@ def test_opportunity_delivery_stats(opportunity):
     payment_due = total_accrued - total_paid
 
     # total deliveries=4 deliveries_from_yesterday=3
-    cw = CompletedWorkFactory(opportunity_access=oa1, status_modified_date=now(), status=CompletedWorkStatus.pending)
+    cw = CompletedWorkFactory(
+        opportunity_access=oa1,
+        status_modified_date=now(),
+        status=CompletedWorkStatus.pending,
+        saved_completed_count=1,
+    )
     UserVisitFactory.create(
         opportunity=opportunity,
         opportunity_access=oa1,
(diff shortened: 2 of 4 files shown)


mkangia:
> Opportunity Dashboard counts for Total Deliveries and Verified Deliveries, as well as the single-opportunity page's Services Delivered and Approved counts, were previously undercounting when there were duplicate approved visits. They now count each approved visit separately, matching the worker delivery table.

I assume this was AI generated. But anyway, I could not read this well.
It seems we were trying to tell what the Opp dashboard counts but then the jump to mention the issue went off, mainly sentence framing I believe.

ajeety4:
> > Opportunity Dashboard counts for Total Deliveries and Verified Deliveries, as well as the single-opportunity page's Services Delivered and Approved counts, were previously undercounting when there were duplicate approved visits. They now count each approved visit separately, matching the worker delivery table.
> 
> I assume this was AI generated. But anyway, I could not read this well. It seems we were trying to tell what the Opp dashboard counts but then the jump to mention the issue went off, mainly sentence framing I believe.

Thanks for pointing.

Opportunity ~~Dashboard~~ List Page: Columns `Total Deliveries` and `Verified Deliveries` were not counting the duplicate approved deliveries.  Same for `Services Delivered` in the single opportunity detail page.

This fixes them to include the duplicate counts, making it match the `Delivered` and `Approved` counts in the worker delivery table.

Updated the description.

commcare_connect/opportunity/helpers.py
```diff
@@ -712,7 +715,7 @@ def get_opportunity_delivery_progress(opp_id):
         deliveries_from_yesterday=deliveries_from_yesterday_sq(),
         accrued_since_yesterday=accrued_since_yesterday_sq,
         most_recent_delivery=most_recent_delivery_sq,
-        total_deliveries=get_deliveries_count_subquery(),
+        total_deliveries=get_deliveries_count_subquery(sum_field="saved_completed_count"),
```
mkangia, line 718: Is this something we need to worry about?

https://github.com/dimagi/commcare-connect/blob/b9c6023d283432eac3b1fe9df2be2007ca1f2a1a/commcare_connect/opportunity/models.py#L828-L833

commcare_connect/opportunity/helpers.py
```diff
@@ -509,8 +510,10 @@ def get_data_qs(opp_ids, program_manager=False):
                 CompletedWork.objects.filter(opportunity_access__opportunity_id__in=opp_ids)
                 .values("opportunity_access__opportunity_id")
                 .annotate(
-                    total_deliveries=Count("id", distinct=True),
-                    verified_deliveries=Count("id", filter=Q(status=CompletedWorkStatus.approved), distinct=True),
+                    total_deliveries=Coalesce(Sum("saved_completed_count"), Value(0)),
```
mkangia, line 513: Apologies for the extra info request here, For those not familiar with these fields `saved_completed_count` and `saved_approved_count` can you help understand why using these solves the issue?

You mentioned they already account for duplicates. Can you share the code where this happens for reference?

mkangia (commented):
Hey @ajeety4 

one question on the approach.

For those opportunities that don't worry about duplicates, this change is probably what we need, but what about the ones that still mark visits as duplicate and won't want them counted? is that not a scenario we are to consider?

Code wise the changes are all okay, just want some more info on the code/decisions.

commcare_connect/opportunity/helpers.py
ajeety4: Good question.

I believe we don't need to worry about this.  Looking at the code, they are updated every time the completed work status or count changes. 
Also they are a source of truth for invoicing, reporting and data export. Also used in  a view where we display a count. 

That being said, I will create a ticket to double verify and clean this up.

commcare_connect/opportunity/helpers.py
ajeety4: No worries. These are fair questions.

The `CompletedWork` model represents a delivery for a given entity ID, delivery unit, and worker. When the exact same delivery arrives again, instead of creating another `CompletedWork` row, the `saved_completed_count` field on the existing model is incremented. The same applies to `saved_approved_count` but it is incremented after approval.

So, importantly, `saved_completed_count > 1` means there are duplicate deliveries for that particular work.

> Can you help me understand why using these solves the issue?

The existing logic was only counting `CompletedWork` rows. This meant that when `saved_completed_count > 1`, the additional deliveries represented by that count were not included, resulting in an undercount.

The Worker Deliver page, on the other hand, uses `saved_completed_count`, so it included those additional deliveries and showed the correct higher count. This discrepancy between the two counts was what caused confusion for Sarvesh in this case.

Using `saved_completed_count` (and `saved_approved_count` where applicable) ensures that those duplicate deliveries are also included in the count, bringing the two views into agreement.

ajeety4:
> Hey @ajeety4
> 
> one question on the approach.
> 
> For those opportunities that don't worry about duplicates, this change is probably what we need, but what about the ones that still mark visits as duplicate and won't want them counted? is that not a scenario we are to consider?
> 
> Code wise the changes are all okay, just want some more info on the code/decisions.

> but what about the ones that still mark visits as duplicate and won't want them counted? is that not a scenario we are to consider?

Good point. 
For such opps, these were already being counted in the `Worker Delivery table` , which is used as a source of truth by the end users and also used for payments.
However, it does not hurt to check once with product on this.
Thanks for pointing this out.

commcare_connect/opportunity/helpers.py
Charl1996: This is very useful context

Charl1996:
> However, it does not hurt to check once with product on this.

Code changes looks good, but will wait to hear back on this point.

commcare_connect/opportunity/tests/test_helpers.py
```diff
@@ -459,11 +500,19 @@ def test_opportunity_worker_progress_stats(opportunity):
         visit_date=yesterday,
     )
 
+    # A CompletedWork with 2 duplicate approved visits should add 2 to total/approved deliveries, not 1.
```
mkangia, line 503: Ideally this is one duplicate that was approved as well right?

mkangia (approved):
Code changes are okay.

I leave it to you and the product to make the call if this change is safe or not.

commcare_connect/opportunity/views.py
```diff
@@ -3392,7 +3392,7 @@ def opportunity_delivery_stats(request, org_slug, opp_id):
             "icon": "fa-clipboard-list",
             "name": _("Services Delivered"),
             "status": _("Total"),
-            "value": header_with_tooltip(stats.total_deliveries, _("Total delivered so far excluding duplicates")),
+            "value": header_with_tooltip(stats.total_deliveries, _("Total delivered so far")),
```
mkangia, line 3395: one question though,
if this tooltip said this was excluding duplicates, why was it confusing?

ajeety4: Good question.

Based on my discussion with Sarvesh, and thinking about how an end user would view this:

- The Workers Deliveries page, which is linked from this page, includes duplicate deliveries and therefore showed a different count. Since users could not easily determine the number of duplicates from the Workers Deliveries page, the difference in counts added to the confusion.
- Also, the concept of duplicate deliveries is no longer applicable to newer opportunities with automatic visit verification enabled.

mkangia: > Also, the concept of duplicate deliveries is no longer applicable to newer opportunities

That is useful context to make decisions here.

ajeety4:
Received a go ahead from this on Product, so merging this.

PR #1476, opened 2026-08-25
Refresh pending org invites and template the invite email

Charl1996:
## Product Description

Org admins can now re-send an invite to an address that already has a pending one — previously that was rejected with "An invite has already been sent to this email.", so there was no way to resend or correct a mistaken role without waiting out the 7-day window. Re-inviting issues a new token and restarts the expiry.

The invite email also now names the assigned role and the expiry date, and ships an HTML alternative alongside the plain-text body.

Send invite to user:
image

After cooldown period:
image

Clicking resend:
image

## Technical Summary

[CCCT-2566](https://dimagi.atlassian.net/browse/CCCT-2566). Part of the signup/invite overhaul spec in #1290.

Note: most of the items required in the ticket has already been done so this PR only does what is left:
1. Reinvite flow
2. Adding dedicated template files for the emails.
3. Show action result (reinvite / revoke)

The Resend button, the Expires on column and the cooldown go slightly beyond the ticket — they're what makes the reinvite flow reachable and safe from the UI.

Implementation notes:

- `resend_invite` delegates to `OrganizationInvite.send_invite`, so token rotation and the expiry restart stay in one place; the row's role carries over untouched (changing a role stays the form's job).
- `RESEND_COOLDOWN` / `is_in_resend_cooldown` derive from `date_modified`, the same way `expiry_date` does — no new field, no migration.
- **Shared template change.** `base.html`'s message markup moved to `components/messages.html`, wrapped in `<div id="messages">`; `components/messages_oob.html` re-includes it under `hx-swap-oob="beforeend:#messages"` so an htmx response can append to the banner area. This is the only change in the diff that touches every page. `beforeend` rather than replacing `#messages` is deliberate — `#pending_invites_container` reloads on the `loadMembers` event, and an `outerHTML` swap would wipe a banner already on screen any time a render had nothing queued.
- The OOB include sits *above* `{% if table.data %}` in `pending_invites_table.html`: revoking the last pending invite renders an empty table, so a fragment inside the guard would swallow its own message.
- The row's action buttons moved out of inline `template_code` into `organization/pending_invite_actions.html`.

### Safety story

No schema change and no migration, so existing invites are unaffected.

One behaviour changes worth flagging:
- A resend **rotates the token**, so an invite link already in flight stops working.  so the recipient gets a 404 rather than the friendly "this invitation has expired"
page.

### Automated test coverage

Reinvite and expiry:
- `test_reinvite_refreshes_existing_pending_invite` — re-posting a pending invite witreuses the row, and issues a new token and expiry.
- `test_expiry_date_is_expiry_days_after_last_update`, plus the existing `is_expired`

Resend endpoint:
- `test_admin_resend_issues_a_new_token_and_expiry` — rotates the token, bumps the exhe email.
- `test_resend_is_refused_during_the_cooldown` — no token change, no email.
- `test_only_pending_invites_can_be_resent` — accepted and revoked invites 404.
- `test_member_cannot_resend_invite` — non-admins 404.

Cooldown:
- `test_resend_cooldown_lapses_after_the_cooldown_window` — the window boundary on th
- `test_reinvite_rejected_while_the_invite_is_in_resend_cooldown` — the add-member fo

Table:
- `test_table_shows_the_expiry_date` — the Expires on column renders the right date.
- `test_resend_button_is_disabled_during_the_cooldown` — the button carries `disabled` only inside the window.

Action feedback:
- `test_resend_reports_success_in_a_message`, `test_refused_resend_reports_the_cooldoe_reports_success_in_a_message` — each outcome reaches the response as an out-of-band
`#messages` append.

Email templating:
- `test_message_includes_role_and_expiry_date` and `test_sends_html_alternative` — th, expiry and token URL, and an HTML part is sent.

### QA Plan

QA planned.

### Labels & Review

- [x] The set of people pinged as reviewers is appropriate for the level of risk of t

[CCCT-2566]: https://dimagi.atlassian.net/browse/CCCT-2566?atlOrigin=eyJpIjoiNWRkNTljTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Charl1996:
### Code review

Found 1 issue:

1. The new plain-text invite email HTML-escapes user-supplied names. Django autoescapes `.txt` templates too, so an org named `Sam's Clinic & Co` invited by `O'Brien` renders in the plain-text body as `O&#x27;Brien has invited you to join Sam&#x27;s Clinic &amp; Co`. The previous f-string in `organization/tasks.py` did not escape, so this is a regression introduced by the move to templates. Fix by wrapping the block in `{% autoescape off %}` — precedent at [`account/email/email_confirmation_signup_message.txt#L4`](https://github.com/dimagi/commcare-connect/blob/b959f1ff60b940fdc1c245e452dcbfc629d57bc4/commcare_connect/templates/account/email/email_confirmation_signup_message.txt#L3-L6). Note the same pattern exists in the other `.txt` email templates, so a fix in `email/base.txt` would cover them all.

https://github.com/dimagi/commcare-connect/blob/b959f1ff60b940fdc1c245e452dcbfc629d57bc4/commcare_connect/templates/organization/email/invite.txt#L3-L5

Charl1996:
> ### Code review
> Found 1 issue:
> 
> 1. The new plain-text invite email HTML-escapes user-supplied names. Django autoescapes `.txt` templates too, so an org named `Sam's Clinic & Co` invited by `O'Brien` renders in the plain-text body as `O&#x27;Brien has invited you to join Sam&#x27;s Clinic &amp; Co`. The previous f-string in `organization/tasks.py` did not escape, so this is a regression introduced by the move to templates. Fix by wrapping the block in `{% autoescape off %}` — precedent at [`account/email/email_confirmation_signup_message.txt#L4`](https://github.com/dimagi/commcare-connect/blob/b959f1ff60b940fdc1c245e452dcbfc629d57bc4/commcare_connect/templates/account/email/email_confirmation_signup_message.txt#L3-L6). Note the same pattern exists in the other `.txt` email templates, so a fix in `email/base.txt` would cover them all.
> 
> https://github.com/dimagi/commcare-connect/blob/b959f1ff60b940fdc1c245e452dcbfc629d57bc4/commcare_connect/templates/organization/email/invite.txt#L3-L5
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)

I don't feel comfortable turning the autoescaping mechanism off, so will ignore this.

commcare_connect/organization/forms.py
```diff
@@ -159,8 +159,8 @@ def clean_email(self):
             raise ValidationError(gettext("This person is already a member of this workspace."))
 
         existing = OrganizationInvite.objects.filter(organization=self.organization, email=email).first()
-        if existing and existing.status == OrganizationInvite.Status.INVITED and not existing.is_expired:
-            raise ValidationError(gettext("An invite has already been sent to this email."))
+        if existing and existing.status == OrganizationInvite.Status.INVITED and existing.is_in_resend_cooldown:
+            raise ValidationError(gettext("An invite was just sent to this address. Try again in a few minutes."))
```
hemant10yadav, line 255: Small edge case: the form and send_invite() have slightly different cooldown checks. A revoked invite can pass validation but then no-op, while still showing a success message.

commcare_connect/templates/components/messages_oob.html
```diff
@@ -0,0 +1 @@
+<div hx-swap-oob="beforeend:#messages">{% include "components/messages.html" %}</div>
```
hemant10yadav: Since this wrapper has a single caller (`pending_invites_table.html`), could we inline it there and drop the file?

hemant10yadav (commented):
Looks good! Just a couple of minor suggestions.

mkangia:
Hey @Charl1996 

Before I jump in the review, just a question on the approach

1. is there a reason why we are regenerating the token instead of just reusing the existing one. I skimmed through the spec but could not find something.
2. as a user, I am not confident it is clear that "Resend" would also change the token. If we want to do that, we should make it explicit. I see that we are using both terms "Reinvite" and "Resend" for that action. I would think of "resend" as simply sending the same exact email again. Whereas "Reinvite" could mean more like resetting the token.

Emails do tend to clutter. I am thinking that if the invitation is sent again, user could end up clicking on the old email itself, using the old token. I guess an edge case but seems possible. But if there a reason to refresh the token, sure, but it would be better to have it explicit.

ajeety4 (approved):
LGTM

Charl1996:
@mkangia 

The spec mentions this:
```
Validation: reject if the email already belongs to a current member of this org; if a pending (non-expired, non-accepted) invite already exists for this email+org, **refresh it** (new token + new `expires_at`) rather than erroring.
```
So it seems we need a new token.

commcare_connect/organization/forms.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1476/commits/3bf32518f21c06bd5662e51d5afa96699f8a34a6

Charl1996:
@mkangia I've renamed the misleading wording [here](https://github.com/dimagi/commcare-connect/pull/1476/changes/bca7dd8d304f9ddf7ce2afd2de579342fb0edb12)

mkangia:
Hey @Charl1996 

> The spec mentions this:

Right. Its unclear to me why that is needed. ~~Additionally, if we agree to doing it, we should reconsider the name for the button.~~

I see the text change to "Reinvite" which makes sense now, thank you. Though if you can share the reason for regenerating token instead of just re-sending.

Charl1996:
> Though if you can share the reason for regenerating token instead of just re-sending.

I mostly did it because the spec suggested it, but also thought it's probably a bit safer to have a fresh invite token instead of constantly extending the existing token and potentially making an invite token valid for weeks (if the user keeps resending).

hemant10yadav (approved):

commcare_connect/templates/organization/email/invite.html
```diff
@@ -0,0 +1,11 @@
+{% extends "email/base.html" %}
+{% block content %}
+  <p>Hi,</p>
+  <p>
+    {{ inviter }} has invited you to join "{{ invite.organization.name }}" on Connect as
+    {{ invite.get_role_display }}.
+  </p>
+  <p>
+    <a href="{{ invite_url }}">Accept the invitation</a>. This invite expires on {{ invite.expiry_date|date:"F j, Y" }}.
```
ajeety4: One thing to consider: `expiry_date` includes a time, but the current format only displays the date. This could be slightly misleading, as users may assume the invite remains valid until the end of that day.
If we go with showing the time as well, it would be good to demote that it is UTC time.

commcare_connect/organization/views.py
```diff
@@ -187,6 +187,23 @@ def _accept_invite_for_new_user(request, invite, org_slug):
     return render(request, "organization/accept_invite.html", {"form": form, "invite": invite})
 
 
+@api_view(["POST"])
+@org_admin_required
+def resend_invite(request, org_slug, invite_id):
+    invite = get_object_or_404(
+        OrganizationInvite, pk=invite_id, organization__slug=org_slug, status=OrganizationInvite.Status.INVITED
+    )
+    if not invite.is_in_resend_cooldown:
```
ajeety4: what happens in case an invitation is revoked and then resent (in case it is possible ) ? I am guessing it stills need the cooldown period - and if yes is it intentional ?

ajeety4: I see this was already taken care of.

ajeety4 (commented):
Looks good.
Just a single query around displaying expiry date.

commcare_connect/templates/organization/email/invite.html
Charl1996: I'll implement this as part of [this PR](https://github.com/dimagi/commcare-connect/pull/1481), which is a display update of the page.

ajeety4: Thanks

ajeety4 (approved):
Thanks for the change

hemant10yadav (approved):

PR #1477, opened 2026-08-25
Program & Opportunity Access Redesign UI

mkangia:
## Product Description

Three fields already existed in the data model with no user interface. Two live on Program, one on Opportunity; all three point at an Organization.

These are now visible on the UI to be set, all behind the `enable_program_access_redesign` switch:

| Field                      | On                        |Set on |
| -------------------------| -----------------------|-----------------------|
| Funder                  | Program             |Create only        |
| Watchers              | Program             |Create & Edit     |
| Supervising Org  | Opportunity      |Create & Edit      |

funder and supervising_organization are single selects; watchers is a multi-select.

- Program funder — organizations marked as funders, excluding the program's own.
- Program watchers — all organizations except the program's own and its funder.
- Opportunity supervising organization — the program's organization, its funder, or any organization with an accepted ProgramApplication for that program. Defaults to the program's organization, reproducing today's implicit behaviour.

Changes to these three fields is audited.

Two supporting changes added to this PR to make the above usable:

- Organization.funder is now settable in the Django admin (this was missed in the earlier ticket that was to add this)
- Changes to all three fields are exposed on the Django admin for reference

With the switch off, none of the three fields render though the audits will happen since they are not gated behind a switch.

Sample references:
[Opportunity](https://connect-staging.dimagi.com/a/mk-test-1/opportunity/0fed8509-e244-455a-a08f-cd1c99df8d8b/edit)
[Program](https://connect-staging.dimagi.com/a/mk-test-1/program/) - Click edit on "New funded program"

### Screenshots

------
- Setting funders & watchers for a program

image

------

- Setting Supervising Organization on a *new* Opportunity

image

-----

- Setting Supervising Organization on an *existing* Opportunity

image

-----

## Technical Summary

https://dimagi.atlassian.net/browse/CCCT-2559

Recommended review by commit.

Design doc: docs/superpowers/specs/2026-08-17-program-access-redesign-design.md (included in this PR).

No model changes. Program.funder, Program.watchers (migration 0016) and Opportunity.supervising_organization (migration 0140) all already exist and are already read by the permission layer.

This PR is purely the missing UI, so saving through relevant forms takes effect against the existing permission code.
Only migrations added are for the adding audit tables & triggers.

The test suite shared a Redis cache with the local dev server, so toggling any waffle switch in the dev admin could fail unrelated tests. config/settings/test.py now uses LocMemCache.

### Safety story

Everything is behind enable_program_access_redesign, which defaults inactive. With it off, all three fields are absent from self.fields, so they can be neither rendered nor submitted, and existing behaviour is bit-for-bit unchanged — including Opportunity.save() remaining the only thing that sets the supervising organization.

One pre-existing gap this PR does not close, flagged for the reviewer: org_opportunity_access returns MANAGE for supervising_organization_id without re-checking that the organization is still eligible, and nothing resets the column when an application is rejected. So an organization set as supervisor and later rejected retains oversight until reassigned. This predates the change and isn't widened by it — the dropdown stops offering such an organization, so the only way to be in that state is to already be in it. Closing it belongs in its own ticket.

### Automated test coverage

Tests added.

### QA Plan

To go through QA.

### Deployment

- [X] This PR can be deployed: any required configuration, commands, or other manual steps required before this PR can be deployed are done.

### Labels & Review


docs/superpowers/specs/2026-08-17-program-access-redesign-design.md
```diff
@@ -0,0 +1,605 @@
+# Program access redesign: funder, watchers, and supervising organization
```
ajeety4, line 1: Just checking if you intend to add the design doc here ? Guessing it will be removed ?

commcare_connect/program/forms.py
```diff
@@ -21,13 +26,21 @@ class ProgramForm(forms.ModelForm):
         widget=forms.Select(attrs={"data-tomselect": "1"}),
         empty_label="Select a country",
```
ajeety4: Optional nit: This is not related to the PR, but since we are touching the form, it would be nice to add translations for this and the fields above as well. Please feel free to ignore this.

commcare_connect/opportunity/models.py
```diff
@@ -84,6 +84,25 @@ def __str__(self):
...
+# Both labels, "supervising_organization_insert" and "supervising_organization_update", are
+# given explicitly. A label must be unique among a model's trackers, and the
+# fields=["active"] tracker below already holds the default "update", so leaving this
+# tracker's UpdateEvent to default raises ValueError. Only that update label is strictly
+# required; "supervising_organization_insert" is named to match rather than left as the
+# default "insert", because the label is stored as pgh_label on every event row and a bare
+# "insert" beside "supervising_organization_update" would read inconsistently. It also
+# keeps the default "insert" free, which would otherwise collide if another field on this
+# model is tracked on insert later.
+@pghistory.track(
```
ajeety4, line 101: What are your thoughts on using a custom model name for the event model generated by pghistory? This would give it a stable name that won't change if we add or modify the tracked fields, and we could then simply add `supervising_organization` to the existing list of tracked fields.

Is there a requirement to determine which field caused the change from the individual event row itself? If not, we should be able to infer that by comparing the current and previous event rows, since pghistory stores the tracked field values in each event.

config/settings/test.py
```diff
@@ -19,6 +20,23 @@
 )
 TEST_RUNNER = "django.test.runner.DiscoverRunner"
 
+# CACHES
+# ------------------------------------------------------------------------------
+# Give tests their own Redis database. Sharing one with a local dev server leaked waffle
```
ajeety4: Nit: Do we need to keep the full explanation of the Waffle/cache interaction here? I think the comment could be shortened to explain the main reason for using a separate Redis DB, while the details about `WAFFLE_CREATE_MISSING_SWITCHES` and `transaction.on_commit` could go in the PR description.

Something like:

> Use a separate Redis database for tests so cached Waffle switches do not leak between the test suite and a local development server. We keep django-redis because code under test uses `cache.lock()`, which LocMemCache does not provide.

commcare_connect/opportunity/forms.py
```diff
@@ -333,6 +342,29 @@ def clean_users(self):
                 raise ValidationError(gettext("This opportunity has ended. You cannot invite more workers."))
         return self._validate_and_parse_users(user_data)
 
+    def _can_edit_supervising_organization(self):
+        """Only an org with PM-level oversight may reassign it.
```
ajeety4, line 347: :+1:

ajeety4 (approved):
Looks good.
A non blocking query

commcare_connect/opportunity/models.py
mkangia: Hey @ajeety4 

Good point & definitely something I briefly thought but quickly dropped it since this changes how we have done things so far & thought of leaving it out for an architectural change to be considered for the entire project.

But now, I spent some time with Claude since you also raised it as well.

Things to consider:
1. pghistory tracks the value for all tracked fields in the audit row irrespective of which all have changed. This makes it tricky when we are showing the changes to a specific field.
2. when we add a new field to be tracked, we essentially add a new column to the table and populate that as empty for existing rows. Nothing unusual but something that makes the table bigger.
3. when inserting rows to tracked model, it is beneficial to have less inserts/triggers being fired.

My initial thought is, the fields that don't change often should be kept separate but related fields to be tracked under one tracked model like the profile of a user: first name, last name, phone number etc.
So since supervising orgs and funder are not expected to change often, it seems fine to keep them separate even with the findings.

config/settings/test.py
mkangia: Sure,

https://github.com/dimagi/commcare-connect/pull/1477/commits/8144c6eabe1abe6e4248f5e8aec941b00277d719

commcare_connect/program/forms.py
mkangia: https://github.com/dimagi/commcare-connect/pull/1477/commits/7e8ba0a84d439ca610e334408221cffd6e6a0e4f

docs/superpowers/specs/2026-08-17-program-access-redesign-design.md
mkangia: I did intend to leave it. I believe our policy is that the design doc is optional to add/review?
Since I worked with claude on the approach and we had quite some back and forth, I thought it was worth leaving some design decisions here if needed for reference.

ajeety4: Yes, there is no such restriction. Just wanted to double check.

commcare_connect/opportunity/models.py
ajeety4: That makes sense. Thanks for the rationale behind it. I don't there are any concerns with it.
Agreed this can be taken as an architectural change along with broader discussion with the team.

ajeety4 (dismissed):

mkangia:
Hey @ajeety4 

requesting for a follow up review pass on this.

mkangia:
Hey @Charl1996 

Just tagging you to confirm the following change looks okay to you
https://github.com/dimagi/commcare-connect/pull/1477/commits/34a59289db765c0c4fc86cc9dc03ef6eccf1a880

This is related to the merge feature

commcare_connect/organization/merge.py
```diff
@@ -65,6 +66,19 @@ def reassign(self, source: Organization, target: Organization) -> int:
...
         "program.ProgramApplication.organization",
+        "program.ProgramWatcher.organization",
+    }
+)
+
+# Audit rows record what was true when they were written, so a merge must not rewrite them: "this
+# organization became the funder" stays a fact about the merged-away organization. The pghistory
+# foreign keys are declared db_constraint=False for exactly this reason, and the audit admin renders
+# a dash once the organization is gone, so these keep pointing at the source after it is deleted.
+SKIPPED_RELATIONS = frozenset(
```
ajeety4, line 77: :+1:

ajeety4 (commented):
The changes looks good to me.
Will approve once the failed test is fixed.

ajeety4 (approved):

mkangia:
Thanks for flagging the test failure @ajeety4 
Fortunately just a test failure due to access level name change.

PR #1481, opened 2026-08-27
Restyle the org members page and fix the empty members tab on refresh

Charl1996:
## Product Description

Brings the organization members page in line with the [Claude Design mockup](https://claude.ai/design/p/12be6925-0c17-45c0-9580-ecdc1a7e2d61?file=Organization+Members.dc.html) and fixes a bug where the members tab came up empty on a refresh.

image

image

## Technical Summary

Builds on `cs/CCCT-2566-invitation-model` (CCCT-2566).

In addition to the styling updates this PR also fixes a bug where a user on the "Members" tab would get a blank page when refreshing the page from this tab.

### Safety story
Tested locally

### Automated test coverage
No new tests, only updated a test relating to table date formatting.

### QA Plan
Not planned

### Labels & Review


commcare_connect/templates/organization/organization_home.html
```diff
@@ -60,13 +73,13 @@ <h2 class="text-xl font-semibold text-brand-deep-purple mb-4">{% trans 'Workspac
            class="w-full"
            hx-get="{% url 'organization:org_member_table' request.org.slug %}{% querystring_replace %}"
            hx-swap="innerHTML"
-           hx-trigger="loadMembers from:body"></div>
+           hx-trigger="{% if members_tab_active %}load, {% endif %}loadMembers from:body"></div>
```
hemant10yadav: Since these containers use `x-show` and remain in the DOM, could we consider HTMX’s `revealed once` trigger here? It would fetch the content when the tab first becomes visible, whether on initial load or a later switch, and could let us remove `isHistoryAvailable`, and the `loadTabContent()`/`loadMembers` wiring. Might be worth exploring as a simpler approach.

commcare_connect/organization/tables.py
```diff
@@ -1,64 +1,44 @@
...
-                "@click.stop": "",  # used to stop click propagation
-                "name": "row_select",
-                "type": "checkbox",
-                "class": "checkbox",
-                "value": lambda record: record.pk,
-                "id": lambda record: f"row_checkbox_{record.pk}",
-                ":disabled": lambda record: f"currentUserMembershipId === '{record.pk}'",
-            },
-        },
+    select = select_column(
```
hemant10yadav, line 13: Nice!

hemant10yadav (approved):
One small suggestion otherwise, looks good to me!

commcare_connect/organization/tables.py
```diff
@@ -1,64 +1,44 @@
 import django_tables2 as tables
-from django.utils.html import format_html
+from django.utils.translation import gettext_lazy as _l
```
mkangia: nit: is `_l` the new and better way to do this? I have always been used to using just `_` which I understand collides with python.

commcare_connect/templates/organization/organization_home.html
```diff
@@ -79,7 +92,9 @@ <h2 class="text-xl font-semibold text-brand-deep-purple mb-4">{% trans 'Workspac
     Alpine.data('orgTabs', () => ({
       selectedTab: '{{ request.GET.active_tab|default:"details" }}',
       showAddMemberModal: false,
-      isHistoryAvailable: false,
+      // The `load` trigger on the containers has already fetched the tables when we
+      // land on the members tab, so a later switch back must not refetch them.
+      isHistoryAvailable: {{ members_tab_active|yesno:"true,false" }},
```
mkangia: nit: not specific to this PR but it seems more and more logic is now in HTML.
Is there a reason this needs to be in HTML file?

commcare_connect/organization/tables.py
Charl1996: I updated the code. (I force pushed since I rebased as well)

commcare_connect/templates/organization/organization_home.html
Charl1996: This has been removed in favour of a new approach.

commcare_connect/templates/organization/organization_home.html
Charl1996: This has been removed in favour of a new appraoch

commcare_connect/web/templatetags/sort_link.py
```diff
@@ -52,7 +52,7 @@ def update_query_params(context, **kwargs):
 
 
 @register.simple_tag(takes_context=True)
-def sortable_header(context, field, label, use_view_url=True):
+def sortable_header(context, field, label, use_view_url=True, order_by_field="sort"):
```
hemant10yadav, line 55: I’m guessing this is to avoid sorting issues caused by multiple tables on the same page?

commcare_connect/templates/organization/organization_home.html
```diff
@@ -70,6 +88,7 @@ <h2 class="text-xl font-semibold text-brand-deep-purple mb-4">{% trans 'Workspac
     Alpine.data('orgTabs', () => ({
       selectedTab: '{{ request.GET.active_tab|default:"details" }}',
       showAddMemberModal: false,
+      isNewEntity: false,
```
hemant10yadav: Is this variable used anywhere?

commcare_connect/templates/organization/organization_home.html
```diff
@@ -51,13 +69,13 @@ <h2 class="text-xl font-semibold text-brand-deep-purple mb-4">{% trans 'Workspac
...
+           hx-trigger="{% if members_tab_active %}load, {% endif %}loadMembers from:body"></div>
       <div x-show="selectedTab === 'members'"
            id="pending_invites_container"
            class="w-full"
-           hx-get="{% url 'organization:pending_invites_table' request.org.slug %}"
+           hx-get="{% url 'organization:pending_invites_table' request.org.slug %}{% querystring_replace %}"
            hx-swap="innerHTML"
-           hx-trigger="loadMembers from:body"></div>
+           hx-trigger="{% if members_tab_active %}load, {% endif %}loadMembers from:body"></div>
     </div>
```
hemant10yadav, line 74: I still think we can get rid of some of the custom JS and let `hx-trigger="intersect once"` handle the tab loading. This would remove `loadTabContent`, `isHistoryAvailable`, and the tab-related `if`s.

commcare_connect/web/templatetags/sort_link.py
Charl1996: Correct

commcare_connect/templates/organization/organization_home.html
Charl1996: No! Good catch.

commcare_connect/templates/organization/organization_home.html
Charl1996: https://github.com/dimagi/commcare-connect/pull/1481/commits/68ef7ac3d6959432d6e4bb3e3bc5ba38ee4d3391

commcare_connect/templates/organization/organization_home.html
Charl1996: https://github.com/dimagi/commcare-connect/pull/1481/commits/2d97fb8dfe8ac1db9260bd7892b24ae4a066630e

hemant10yadav (approved):
Thanks LGTM!

PR #1487, opened 2026-08-28
Auto-zoom the microplanning map to the filtered work areas

snopoke:
[CCCT-2749](https://dimagi.atlassian.net/browse/CCCT-2749)

## Product Description

The microplanning map now moves to what was filtered. Clearing a filter widens back out. If a filter combination matches nothing the viewport doesn't change.

## Technical Summary

Work areas are loaded as vector tiles which don't contain geometry so each filter change requires an API call to fetch the new bounds.

A new `workareas_bounds` endpoint takes the same query params as `WorkAreaTileView` and returns the extent of the queryset.

Assignment mode's group/FLW selection happens client-side rather than through tile filters, so those two endpoints return `bounds` alongside their work areas instead.

Note:

- **Race handling.** All auto-zoom fits share one request counter (`boundsRequestId`), including the ones whose bounds ride along on an assignment-selection response, so a slow reply can't yank the viewport away from a filter the user has since changed.

Also folded in: `currentTileQuery()` is now the single source of truth for the tile query in both modes (previously the filter form was re-serialized in three places and assignment mode composed its own), and a date-range filter switches the visits layer on, since that filter is otherwise invisible.

### Safety story

Read-only — one new GET endpoint, no writes, no schema change. Behind the existing `MICROPLANNING` flag. Tested locally in both progress and assignment mode: each filter type, search box, group and FLW selection, filter clearing, no-match filters, and an opportunity with no work areas.

### Automated test coverage

`TestWorkAreaBoundsView` covers the new endpoint.

The client-side zoom logic isn't covered by tests.

### QA Plan

Filter the map in both modes and confirm the camera follows the filtered set and widens back on clear. Worth checking a filter matching a single work area (should stop at zoom 14, not street level) and one spanning widely separated areas (layer must stay visible).

Also confirm that a filter matching nothing doesn't move the viewport.

### Labels & Review

[CCCT-2749]: https://dimagi.atlassian.net/browse/CCCT-2749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+            },
+
+            invalidateBoundsRequest() {
+                return ++this.boundsRequestId;
+            },
+
+            // Bounding box of the work areas matching `query`, or null when nothing matches. Work
+            // areas arrive as vector tiles, so their geometry only exists server-side. Superseded
+            // responses resolve to undefined, so a slow reply can't yank the viewport away from a
+            // filter the user has since changed.
```
snopoke:
```suggestion
            // Bounding box of the work areas matching `query`, or null when nothing matches.
```

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+            fetchWorkAreaBounds(query) {
+                const requestId = this.invalidateBoundsRequest();
+                const baseUrl = '{{ bounds_url|escapejs }}';
+                return fetch(query ? `${baseUrl}?${query}` : baseUrl)
+                    .then((res) => {
+                        if (!res.ok) throw new Error('Failed to load work area bounds');
+                        return res.json();
+                    })
+                    .then((data) => (requestId === this.boundsRequestId ? data.bounds : undefined))
```
snopoke, line 259:
```suggestion
                    // Ignore responses if the request has been superseded.
                    .then((data) => (requestId === this.boundsRequestId ? data.bounds : undefined))
```

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+                const options = { padding: 40, maxZoom: {{ max_autozoom_zoom }} };
+                // Omit `duration` entirely to animate. Mapbox merges the options over its defaults
+                // with Object.assign, so an explicit `duration: undefined` overwrites the default
+                // instead of falling back to it — leaving the camera with a NaN zoom.
+                if (!animate) options.duration = 0;
+                // Work area tiles are not served below the layer's minzoom, so a fit wide enough
+                // to drop under it lands on a viewport where nothing renders. fitBounds has no
+                // floor of its own, so take the camera it would use and, only when that camera is
+                // too far out, override its zoom — the framing is sacrificed to keep the layer
+                // visible. A fit that already clears the floor is left exactly as it was.
```
snopoke:
```suggestion
                // Work area tiles are not served below the layer's minzoom, so check the camera zoom and reset if we're below the minimum.
```

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+                    options.center = camera.center;
+                    options.zoom = {{ workarea_min_zoom }};
+                }
+                this.map.fitBounds(bounds, options);
+            },
+
+            // Fits the map to `query`'s work areas. A null bounds means the filters match nothing:
+            // hold the viewport and say so, rather than leaving an unexplained no-op. With no
+            // filters set there is nothing to blame it on — an opportunity with no work areas yet
+            // is the expected state before an import, so that case stays silent.
```
snopoke:
```suggestion
```

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+                        return;
+                    }
+                    this.zoomToBounds(bounds, options);
+                });
+            },
+
+            // Whether `query` actually narrows anything. A serialized filter form is never the
+            // empty string — the hidden work_area field, the payment unit select and both date
+            // inputs each contribute a `name=` pair while blank — so the question has to be asked
+            // of the values, not of the querystring's length.
```
snopoke:
```suggestion
            // Whether `query` actually narrows anything. A serialized filter form is never the
            // empty string — the hidden work_area field, the payment unit select and both date
            // inputs each contribute a `name=` pair while blank.
```

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -518,21 +646,33 @@
...
+                // after — and undo — the fit from a search or filter the user has since used.
+                const requestId = this.invalidateBoundsRequest();
+
                 // Show the summary spinner while this filter's areas load. Until they
                 // land, flwFilterWorkAreaIds is empty, so the live delta would transiently
                 // count the FLW's own existing areas as "new" and inflate the total.
                 this.flwSummaryLoading = true;
                 try {
-                    const workAreas = await fetchWorkAreas();
+                    const { workAreas, bounds } = await fetchSelection();
```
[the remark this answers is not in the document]
snopoke: Leaving this out of this PR but let me know if I should create a followup PR for it.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,114 @@
...
+                if (this.assignmentMode) {
+                    // No filter sidebar here — the query is composed from the unassigned toggle and
+                    // the search box, which apply together.
+                    const params = new URLSearchParams();
+                    if (this.showOnlyUnassigned) params.set('unassigned_only', 'true');
+                    const option = this.searchSelection();
+                    if (option) params.set(option.filter_name, option.object_id);
+                    return params.toString();
+                }
+                return new URLSearchParams(new FormData(this.$refs.filterForm)).toString();
```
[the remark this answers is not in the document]
snopoke: Passing on this feedback. I did consider this as a pre-factor the complexity warrants a new ticket.

commcare_connect/templates/microplanning/map_handler.html
```diff
@@ -213,12 +215,103 @@
...
+                // with Object.assign, so an explicit `duration: undefined` overwrites the default
+                // instead of falling back to it — leaving the camera with a NaN zoom.
+                if (!animate) options.duration = 0;
+                // Work area tiles are not served below the layer's minzoom, so check the camera zoom and reset if we're below the minimum.
+                const camera = this.map.cameraForBounds(bounds, options);
+                if (camera && camera.zoom < {{ workarea_min_zoom }}) {
+                    options.center = camera.center;
+                    options.zoom = {{ workarea_min_zoom }};
+                }
+                this.map.fitBounds(bounds, options);
```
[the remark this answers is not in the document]
snopoke: According to the mapbox docs:

> Options supports all properties from [AnimationOptions](https://docs.mapbox.com/mapbox-gl-js/api/properties/#animationoptions) and [CameraOptions](https://docs.mapbox.com/mapbox-gl-js/api/properties/#cameraoptions) in addition to the fields below.

`CameraOptions` includes `center` and `zoom`.

ajeety4 (approved):
Looks good to me. Appreciate the edge cases handling.

Just confirming that this is going through QA ?

PR #1492, opened 2026-09-01
Add ability to fetch buildings info on map

Charl1996:
## Product Description

A **Show Building Data** toggle on the progress and assignment maps draws building footprints over the work areas, from zoom 15 up. Footprints load for wherever the user moves next, and everything already loaded stays drawn until the toggle goes off. Below zoom 15 the control and overlay both leave the map — the data is meaningless at that scale and too large to fetch. On a screen too large to serve in one read, the map asks the user to zoom in rather than reporting a failure.

[Screencast from 01-09-2026 10:37:18.webm](https://github.com/user-attachments/assets/612bbbf0-63f7-47b6-ab19-a9fb9708e868)

Area too large error message:
image

## Technical Summary
[CCCT-2750](https://dimagi.atlassian.net/browse/CCCT-2750)

We don't store building data in our database so we read this data live from OvertureMaps using their [Python client library](https://docs.overturemaps.org/getting-data/overturemaps-py/). The results are cached with key built from `BUILDINGS_CACHE_KEY`. 

It's useful to understand how map rendering works to understand how the data is fetched and cached.

### How the map is rendered using tiles
Generally GIS solutions breaks the map up into tiles/cells depending on your zoom level. The further you're zoomed in the more tiles will be used to define the map.

Eg. When on zoom level 1 the map is define by 4 tiles. Zooming in to level 2 breaks each tile into 4 additional tiles, so now the map is defined by 4 x 4 = 16 tiles. Each zoom level defines the earth in a set number of tiles where the deeper you're zoomed in the smaller the tiles.

This tile-defined strategy makes it easy for us to say "give me the buildings for the tiles covered by the user's viewport". This is exactly what this PR does, but to be more pragmatic we define the tile size to be sufficiently large (the size you'd get at zoom level 14 - see `GRID_ZOOM`) to cover a reasonable amount of ground so we don't have to trigger another read request as soon as the user pans the map slightly (e.g. at zoom level 15, the minimum zoom level the user can view buildings data at). These results are then cached based on `BUILDINGS_CACHE_KEY`.

### Calculating tiles
In order to calculate which grid tiles certain coordinates falls in (answering the question "which tiles' data should I fetch for the user's viewport") we need to use [web mercator projection](https://gisgeography.com/web-mercator-projection/). Luckily there's a package, `mercantile`, that does this for us — already installed as a transitive dependency of `django-vectortiles`, and now declared directly since we use it ourselves.

### Good followup work to investigate
1. Store the Overture data in the database
2. Have a task check and update the Overture release on a weekly basis

### Safety story

Additive and flag-gated: one read-only endpoint plus an opt-in layer, no migrations, no writes, no existing data touched. Worst case for a user is a failed fetch, which toasts and leaves the map as it was; for the server, a slow Overture read on a view holding no DB transaction.

### Dependencies

Two direct additions to `pyproject.toml`: `pyarrow` and `mercantile`
(web mercator tile maths, previously installed only as a transitive dependency of
`django-vectortiles`).

`pyarrow` is large - roughly ~152MB installed and by far the largest thing in this PR, but we need it since Overture's data is GeoParquet on S3; pyarrow is what reads it, and its bbox predicate pushdown is what keeps us from downloading whole country-sized files per map pan.

### Automated test coverage

`test_buildings.py` covers bbox parsing, the grid (snapping, counting, sub-epsilon and
cell-boundary cases) and the cache path — only missing cells fetched, empty cells cached, a
building on a seam served once. `test_views.py` covers the endpoint: collection shape, bad bbox
(400), over-cap area (422), upstream failure (503), and org-admin gating.

### QA Plan
QA planned.

### Deployment

### Labels & Review

[CCCT-2750]: https://dimagi.atlassian.net/browse/CCCT-2750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mkangia:
Hey @calellowitz 
Can you review this please? specifically interested in your inputs about the libraries included.

commcare_connect/microplanning/tests/conftest.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+from django.core.cache import cache
+
+
+@pytest.fixture
+def local_cache(settings):
+    """Keep cache reads off the shared Redis, and out of each other's way."""
+    settings.CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"}}
```
snopoke: why not make this a global setting for tests?

hemant10yadav (commented):
Did we consider a client-side approach instead of server-side fetching/caching?

Overture publishes pre-built PMTiles archives per release ([docs](https://docs.overturemaps.org/examples/overture-tiles/)), and the bucket appears to support CORS and range requests. Mapbox GL JS has native `.pmtiles` vector-source support since v3.21, and we're currently on v3.24 — see [Add a PMTiles vector source](https://docs.mapbox.com/mapbox-gl-js/example/pmtiles-vector-source/).

If that works for our use case, the browser could potentially read the Overture archive directly via a standard Mapbox vector source, which would remove the need for server-side GeoParquet reads, Redis tile caching, and the proxy view. It would also simplify the client-side logic since Mapbox GL would handle tile loading and caching.

I'd also expect this to reduce server-side overhead compared with computing and caching responses per request.

Curious whether this was something you looked at and moved away from for a reason, or if it just wasn't on the radar when this was built

Charl1996:
> Did we consider a client-side approach instead of server-side fetching/caching?

I did not. Thanks for bringing this up! Looking into it

Charl1996:
@hemant10yadav Your suggestion to load it client-side seems very promising. Much faster and way less complicated! I'll get it going and close this PR once the other one is ready for review.

Charl1996:
Closing in favour of https://github.com/dimagi/commcare-connect/pull/1522

calellowitz:
> Hey @calellowitz Can you review this please? specifically interested in your inputs about the libraries included.

Do you have specific concerns about the libraries? They both seem reasonable if we need to handle this data. Making the old one explicit is mostly no change, and the new one appears to be the standard python library to deal with arrow files, which I gather from this is the format overture data comes in. If we didn't need the new library because we could use other formats, that would be nice, but if we need it, the library seems like the correct one.

commcare_connect/microplanning/overture.py
```diff
@@ -0,0 +1,115 @@
+"""
```
calellowitz: this file is full of very long claude comments and docstrings that aren't providing much value. We should be able to shorten most of them.

commcare_connect/microplanning/const.py
```diff
@@ -39,3 +39,21 @@
     "wag": "work_area_group",
     "ia": "implementation_area",
 }
+
+BUILDING_MIN_ZOOM = 15
+
+# Each zoom level divides the map into fixed grid tiles. We can use these grid tiles to cache
```
calellowitz: are these long comments necessary?

commcare_connect/microplanning/buildings.py
```diff
@@ -0,0 +1,214 @@
...
+def fetch_buildings_for_grid_tiles(grid_tiles):
+    """
+    Fetch the given grid tiles from Overture, returning ``{(x, y): [feature, ...]}``.
+
+    Every grid tile asked for is present in the result, empty ones included, so the caller can cache the
+    fact that an area has no buildings. One query covers the rectangle enclosing all of them rather
+    than one query per grid tile: each Overture read pays seconds of fixed overhead opening the remote
+    dataset, and neighbouring grid tiles largely share the parquet row groups that then get scanned.
+    """
+    # Imported lazily: reading Overture means pyarrow, which costs a process 60MB of resident
```
calellowitz: This lazy import feels like premature optimization that makes the code a bit more confusing to read.

mkangia:
>  Do you have specific concerns about the libraries?

@calellowitz Nothing specific. Just wanted to hear your understanding of including new libraries here.

PR #1493, opened 2026-09-02
Redesign invoice PDF with itemized service summary and approvals

hemant10yadav:
## Product Description

The invoice PDF now shows the organization's details, breaks payments down by month and payment unit instead of one blended total, and includes payment terms, a "Bill to" section, and a note showing who approved it and when.

| Custom Invoice | Service Delivery |
|---|---|
| Custom Invoice | Service Delivery |

## Technical Summary

[CCCT-2736](https://dimagi.atlassian.net/browse/CCCT-2736)

- Service delivery invoices now list one line per billed month and payment unit, plus one combined network management fee line, instead of a single blended total across the whole invoice.
- Added `approval_event` and `approved_by` to `PaymentInvoice` so the PDF can show who approved it.

### Safety story

Rendered the PDF locally against real invoice data, including multi-unit, multi-month, and custom invoices, and checked the output by eye. This only changes what the PDF shows, not how invoices are created or billed.

### Automated test coverage

Added and updated tests for the new service summary grouping and the approval properties; the full opportunity test suite passes.

### QA Plan

Qa(WIP), tested manually on local and on staging.

### Labels & Review

[CCCT-2736]: https://dimagi.atlassian.net/browse/CCCT-2736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -778,6 +778,28 @@ class PaymentInvoice(models.Model):
     def get_status_display(self):
         return InvoiceStatus.get_label(self.status)
 
+    @cached_property
+    def approval_event(self):
+        return (
+            self.status_events.filter(status=InvoiceStatus.READY_TO_PAY)
+            .select_related("pgh_context")
+            .order_by("-pgh_created_at", "-pgh_id")
+            .first()
+        )
+
+    @cached_property
+    def approved_by(self):
+        event = self.approval_event
+        if not event:
+            return None
+
+        email = event.pgh_context.metadata.get("user_email")
+        if not email:
+            return event.pgh_context.metadata.get("username", "")
+
+        user = User.objects.filter(email=email).first()
+        return user.name if user else email
+
 
 class Payment(models.Model):
     payment_id = models.UUIDField(editable=False, default=uuid4, unique=True)
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -779,26 +779,34 @@ class PaymentInvoice(models.Model):
         return InvoiceStatus.get_label(self.status)
 
     @cached_property
-    def approval_event(self):
-        return (
-            self.status_events.filter(status=InvoiceStatus.READY_TO_PAY)
+    def nm_certification(self):
+        """The Network Manager's certification: who submitted the invoice for PM review, and when."""
+        return self._certification_for(InvoiceStatus.PENDING_PM_REVIEW)
+
+    @cached_property
+    def pm_certification(self):
+        """The Program Manager's certification: who approved the invoice for payment, and when."""
+        return self._certification_for(InvoiceStatus.READY_TO_PAY)
+
+    def _certification_for(self, status):
+        event = (
+            self.status_events.filter(status=status)
             .select_related("pgh_context")
             .order_by("-pgh_created_at", "-pgh_id")
             .first()
         )
-
-    @cached_property
-    def approved_by(self):
-        event = self.approval_event
         if not event:
             return None
+        return {"name": self._certifier_name(event), "certified_at": event.pgh_created_at}
 
+    @staticmethod
+    def _certifier_name(event):
         email = event.pgh_context.metadata.get("user_email")
         if not email:
             return event.pgh_context.metadata.get("username", "")
 
         user = User.objects.filter(email=email).first()
-        return user.name if user else email
+        return f"{user.name} ({email})" if user and user.name else email
 
 
 class Payment(models.Model):
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -795,7 +795,7 @@ class PaymentInvoice(models.Model):
             .order_by("-pgh_created_at", "-pgh_id")
             .first()
         )
-        if not event:
+        if not event or not event.pgh_context:
             return None
         return {"name": self._certifier_name(event), "certified_at": event.pgh_created_at}
 
--- a/commcare_connect/opportunity/models.py
+++ b/commcare_connect/opportunity/models.py
@@ -802,9 +802,6 @@ class PaymentInvoice(models.Model):
     @staticmethod
     def _certifier_name(event):
         email = event.pgh_context.metadata.get("user_email")
-        if not email:
-            return event.pgh_context.metadata.get("username", "")
-
         user = User.objects.filter(email=email).first()
         return f"{user.name} ({email})" if user and user.name else email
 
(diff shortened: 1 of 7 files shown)


commcare_connect/opportunity/models.py
```diff
@@ -778,6 +778,28 @@ def is_paid(self):
...
+            .first()
+        )
+
+    @cached_property
+    def approved_by(self):
+        event = self.approval_event
+        if not event:
+            return None
+
+        email = event.pgh_context.metadata.get("user_email")
```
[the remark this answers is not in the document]
hemant10yadav: I think this should be a hard failure. We’d want to know if this case ever occurs, and Sentry will catch it so we can investigate.

commcare_connect/templates/opportunity/invoice_download.html
```diff
@@ -1,87 +1,170 @@
-{% load static i18n %}
+{% load static i18n humanize %}
```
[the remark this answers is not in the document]
hemant10yadav: N/A: rendered via WeasyPrint, not a browser, so no quirks-mode effect

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -215,6 +216,80 @@ def get_invoice_line_items(invoice):
...
+
+    lines = [
+        ServiceSummaryLine(
+            label=gettext(
+                "Service delivery payments for %(payment_unit)s, %(period)s "
+                "(%(units)s verified visits x %(currency)s %(rate)s)"
+            )
+            % {
+                "payment_unit": item.payment_unit_name,
+                "period": f"{item.month:%B %Y}",
```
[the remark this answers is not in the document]
hemant10yadav: Nit: %B is the existing convention in this codebase.

commcare_connect/opportunity/models.py
```diff
@@ -778,6 +778,28 @@ def is_paid(self):
...
+        if not email:
+            return event.pgh_context.metadata.get("username", "")
```
ajeety4: Should we be returning `Name of user`  here similar to the case when email is present ?

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -215,6 +216,80 @@ def get_invoice_line_items(invoice):
...
+                "units": item.number_approved,
+                "currency": currency,
+                "rate": (
```
ajeety4: Can we not simply use the `rate` already available in the line item instead of calculating ?

commcare_connect/opportunity/utils/invoice_line_items.py
```diff
@@ -215,6 +216,80 @@ def get_invoice_line_items(invoice):
...
+def _invoice_period_label(invoice):
+    if (
+        invoice.start_date
+        and invoice.end_date
+        and invoice.start_date.strftime("%Y-%m") != invoice.end_date.strftime("%Y-%m")
+    ):
+        return f"{invoice.start_date:%B %Y} - {invoice.end_date:%B %Y}"
+    if invoice.start_date:
+        return f"{invoice.start_date:%B %Y}"
+    return f"{invoice.date:%B %Y}" if invoice.date else ""
```
ajeety4: I don't expect this to happen in practice. However, it would be safer to return something like `Unknown` rather than return a value that might be incorrect.

ajeety4 (approved):
Looks great to me.
A few non-blocking feedback

commcare_connect/opportunity/models.py
mkangia, line 804: I don't understand the decision making here.
If there is no user email we return the username from context else we find the user by email and then return the user name?
Can you share the context/reasoning behind this?

commcare_connect/templates/opportunity/invoice_download.html
```diff
@@ -1,87 +1,170 @@
...
+        {% if invoice.is_paid %}
+          <p>{% trans "Payment date" %}: {{ invoice.payment.date_paid }}</p>
+        {% endif %}
+      </div>
+      <div class="section">
+        <p class="section-heading">{% trans "Bill to" %}</p>
+        <p>{% trans "Dimagi, Inc." %}</p>
+        <p>{% trans "245 Main Street, 2nd Floor" %}</p>
+        <p>{% trans "Cambridge, MA 02142, USA" %}</p>
+        <p>+1 617.649.2214</p>
```
mkangia: Can we have this in a constant somewhere for easier management?

commcare_connect/templates/opportunity/invoice_download.html
```diff
@@ -1,87 +1,170 @@
-{% load static i18n %}
+{% load static i18n humanize %}
 <html lang="en">
   <head>
     <meta charset="utf-8" />
     <style>
       body {
         font-family: sans-serif;
+        color: #1a1a1a;
```
mkangia, line 8: nit: what makes it okay to have custom css here and not in tailwand?

commcare_connect/opportunity/models.py
hemant10yadav: I added this as a safe check, but ideally it should not happen because it is mandatory that every web user should have an email, so I have removed this check now. Fixed in e27da5785.

commcare_connect/opportunity/models.py
hemant10yadav: I added this as a safe check, but ideally it should not happen because it is mandatory that every web user should have an email, so I have removed this check now. Fixed in e27da5785.

commcare_connect/opportunity/utils/invoice_line_items.py
hemant10yadav: There's no stored per-visit rate to reuse here — this rate is basically the per-visit pay amount (flw_pay.local / number_approved), not the same thing as exchange_rate. I have just renamed it to avoid this confusion. Fixed in e27da5785.

commcare_connect/templates/opportunity/invoice_download.html
hemant10yadav: Extracted the Dimagi bill-to block into a DIMAGI_ADDRESS constant. Fixed in e27da5785.

commcare_connect/templates/opportunity/invoice_download.html
hemant10yadav: This is a WeasyPrint view that generates a standalone PDF rather than serving an HTML page. It doesn't extend base.html, so it doesn't have access to the compiled Tailwind bundle.

commcare_connect/opportunity/utils/invoice_line_items.py
hemant10yadav: Good point — returning "Unknown" instead of a blank string when there's no date, so it's visible rather than silent. Fixed in b955f9dab.

commcare_connect/opportunity/utils/invoice_line_items.py
ajeety4: Thanks for the clarification and the rename

ajeety4 (approved):

PR #1494, opened 2026-09-02
Update PR template and run check

Charl1996:
## Technical Summary
1. The PR template now asks the user to explicitly state whether the PR is ready to be deployed

2. A check runs on defined pull request events which evaluates whether the PR can be merged (depending on the status of the checkbox)

### Safety story
Does not touch code - only PR template.

### Automated test coverage
N.A

### QA Plan
N.A

### Deployment

### Labels & Review

### DIFF
--- a/.github/workflows/pr-checklist.yml
+++ b/.github/workflows/pr-checklist.yml
@@ -1,0 +1,37 @@
+name: PR Checklist
+
+# Blocks merging until the author confirms the PR is safe to deploy.
+# Runs on `edited` too, so re-running is just a matter of ticking the box.
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize]
+    branches: ['main']
+
+permissions:
+  contents: read
+
+jobs:
+  deploy-confirmation:
+    name: Deploy confirmation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check the deployment checkbox is ticked
+        # Dependabot opens PRs without the template, so it can't tick the box.
+        if: github.actor != 'dependabot[bot]'
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          if printf '%s' "$PR_BODY" | tr -d '\r' | grep -qiE '^[[:space:]]*-[[:space:]]*\[x\][[:space:]]*This PR can be deployed'; then
+            echo "Ready for deployment."
+            exit 0
+          fi
+
+          echo "::error::The \"This PR can be deployed\" checkbox in the PR description is not ticked."
+          echo ""
+          echo "Before merging to main, confirm this change can be deployed as-is."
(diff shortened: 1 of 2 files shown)


ajeety4 (approved):
Nice :+1:

.github/workflows/pr-checklist.yml
```diff
@@ -0,0 +1,37 @@
+name: PR Checklist
+
+# Blocks merging until the author confirms the PR is safe to deploy.
+# Runs on `edited` too, so re-running is just a matter of ticking the box.
+on:
+  pull_request:
+    types: [opened, edited, reopened, synchronize]
+    branches: ['main']
```
mkangia: will this short circuit if the base branch is other than main?

Possibly an edge case, but just so you can consider,
Branch B is to be merged in Branch A,
Branch A has set this flag to "yes", then branch B is merged which had it set as "no" but was not blocked from merging.

.github/workflows/pr-checklist.yml
```diff
@@ -0,0 +1,37 @@
+name: PR Checklist
```
mkangia, line 1: I am surprised we didn't have setup already for other checks on the PR? how come?

mkangia (commented):
non blocking, mostly gaps in my understanding.

.github/workflows/pr-checklist.yml
Charl1996: I don't think it was particularly necessary before. This is the first item I think we want to particularly block if the description does not meet a particular set of criteria.

.github/workflows/pr-checklist.yml
Charl1996: Yes, it will only run if the base branch is `main`.

mkangia: okay, any thoughts on the edge case I mentioned?

Charl1996: I think given the intention of this checkbox it makes more sense to check this on every branch. I'll update the file.

Charl1996: https://github.com/dimagi/commcare-connect/pull/1494/commits/373e30fbddbcf1c0e83ab78a074a1414ee3830ed

.github/workflows/pr-checklist.yml
```diff
@@ -0,0 +1,36 @@
...
+permissions:
+  contents: read
+
+jobs:
+  deploy-confirmation:
+    name: Deploy confirmation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check the deployment checkbox is ticked
+        # Dependabot opens PRs without the template, so it can't tick the box.
```
mkangia, line 18: 👍 
The comment answered the query I was thinking about.

mkangia (approved):

PR #1499, opened 2026-09-04
Prelogin: render per-route head tags server-side

AndreaMKing5:
## Product Description

Every marketing URL on connect.dimagi.com renders the same `home.html`, which carries one hardcoded `<head>`. `app.js` corrects it in the browser, but link unfurlers (LinkedIn, Slack, X, iMessage) never run JavaScript. Sharing a page today gives every link the home page's title, description, and image:

```
title      Connect by Dimagi
og:image   .../images/field-photos/join-hero.jpg   ← 404s, missing /static/
canonical  https://connect.dimagi.com/
```

Head tags are now per-route and rendered server-side. The broken og:image path is fixed and the five blog posts carry their own covers.

Three posts also link back to their original dimagi.com post; since dimagi.com now 301s those URLs to Connect, those links just send the reader back to the page they're already on. Those lines are removed.

## Technical Summary

`prelogin/route_meta.py` is the single source of truth for title, description, canonical, og:type and image. `HomeView` resolves the request path against it; `home.html` renders the tags and serializes the same table into a `json_script` block that `app.js` parses for in-page navigation, replacing its own copy (about 100 lines). An unknown path canonicalizes to `/`, matching what the router renders.

Cache-buster bumped to `v=8`. `route_meta.py` added to ruff's `per-file-ignores` for `E501`, since its title/description strings have to stay on one line to stay greppable against the rendered `<head>`.

This promotes [dimagi-internal/connect-labs#1424](https://github.com/dimagi-internal/connect-labs/pull/1424), already verified and deployed on labs, into this repo per `docs/prelogin-marketing-site.md` (labs is staging; this repo is the source of truth for connect.dimagi.com).

## Safety Assurance

New tests cover the server-rendered head, the unknown-path fallback, and that every referenced image is a real static file (`TestRouteMeta` in `commcare_connect/prelogin/tests/test_views.py`). `ruff check` and `ruff format --check` pass on the changed Python files, and `node --check` passes on `app.js`. I could not run the full `pytest` suite locally in this environment (no Docker/PostGIS available) — please confirm CI is green before merging.

### DIFF
--- a/commcare_connect/prelogin/route_meta.py
+++ b/commcare_connect/prelogin/route_meta.py
@@ -1,0 +1,185 @@
+"""Per-route title, description, and social image for the marketing SPA.
+
+Every marketing URL renders the SAME home.html, so without this the whole site
+would ship one <head>: a reader sharing a blog post got the home page's title,
+description, and image. app.js fixed that in the browser, but link unfurlers
+(LinkedIn, Slack, X, iMessage) and non-rendering crawlers never run JavaScript,
+so they only ever saw the home page.
+
+This module is the single source of truth. views.HomeView reads it to render the
+head server-side, and home.html serializes it into a <script type="application/
+json"> block that app.js parses for client-side route changes, so the two can no
+longer drift.
+
+Adding a page means adding an entry here, a data-page section in home.html, a
+route in urls.py, and a sitemap.xml entry.
+"""
+
+SITE_ORIGIN = "https://connect.dimagi.com"
+
+# Static path, resolved through {% static %} / django.templatetags.static so it
+# survives the hashed-static storage used in production. The old hardcoded
+# "/images/field-photos/join-hero.jpg" in home.html was missing the /static/
+# prefix and 404d, so every share of this site had a broken preview image.
+DEFAULT_IMAGE = "prelogin/images/field-photos/join-hero.jpg"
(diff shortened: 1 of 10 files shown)


calellowitz (changes requested):
this includes a lot of changes to dependencies that I suspect are unnecessary (given that it doesn't use most of the changed libraries). can you try to limit the changes to just the pages, and not any of the dependency files (the lockfiles and pyproject). The tests are also unnecessary, but no real fight if you or claude want to leave them

AndreaMKing5:
@calellowitz me (and Claude) did the following:
  - Reverted the three unrelated dependency-bump commits' file changes: uv.lock and package-lock.json are back to main's state,
    and the djangorestframework<3.18 version cap in pyproject.toml is removed.
  - Kept the [tool.ruff.lint.per-file-ignores] addition for route_meta.py in pyproject.toml, since that's actually needed for
    the feature (long title/description strings).
  - Left the tests in test_views.py.
  - Verified ruff check, ruff format --check, and eslint all pass on the remaining changes; couldn't run the full pytest suite
    locally (no Docker/PostGIS here, same limitation noted in the original PR description).
  - Pushed the commit (cacf9557) to the fork remote — the PR diff now shows only the 8 page-related files, no lockfiles.

Can you review again?

calellowitz (approved):
thanks

PR #1505, opened 2026-09-07
Make AGENTS.md the instructions file and correct its drift

snopoke:
Review by commit.

`CLAUDE.md` becomes a one-line `@AGENTS.md` include so both Claude Code and other agent tooling read the same file, and the content that had drifted from the code is corrected. Everything asserted in AGENTS.md was checked against the source — the directory map was missing 8 of 22 apps (including `audit/`, which the text already referenced), and the conftest fixture list omitted four program-manager fixtures while telling readers to prefer fixtures over factories.

Two things worth a look because they're claims about behaviour, not prose:

- The multidb note: `DATABASE_ROUTERS` is only installed when `SECONDARY_DATABASE_URL` is set (`config/settings/base.py:52-56`), so `migrate_multi` is single-DB locally. Please sanity-check that against how you actually run the secondary.
- `pyproject.toml` gains two `filterwarnings` ignores (WhiteNoise's missing-static-dir, allauth importing its own deprecated module). Naive-datetime and `RemovedInDjango60Warning` are left visible on purpose.

`.claude/settings.json` allowlists pytest, prek and `docker compose ps` so agent-driven verification doesn't stop on permission prompts. `uv run inv *`, `./manage.py *` and `psql *` are deliberately excluded — they contain `inv deploy`, `migrate`/`shell -c`, and arbitrary SQL respectively.

## Safety Assurance

No application code changed. The pytest config change is the only thing that can affect CI — I ran subsets locally (~150 tests across `utils`, `opportunity/test_models`, `opportunity/test_forms`) rather than the full suite, so CI is the real check here.

- [x] This PR can be deployed: no configuration, commands, or manual steps required.


AGENTS.md
```diff
@@ -1 +1,120 @@
-@CLAUDE.md
+# CommCare Connect
```
Charl1996, line 1: Why the change?

I get that it's semantically probably more correct to have claude.md include the agents.md file as it's the "generic" file across agents, but I'm curious if there's other functional reasons as well?

AGENTS.md
```diff
@@ -94,6 +94,7 @@ config/
...
 - **Never render broken UI**: don't emit a button, link or map that 500s or dead-ends when the config is absent
 - **Set an explicit timeout on outbound calls**: all HTTP goes through **httpx**, which defaults to 5s — fine for quick calls, too short for bulk work, so pass a timeout sized to the operation (see `_make_request` in `commcare_connect/connect_id_client/main.py`: `timeout=10` default, 15–30s for bulk sends; `commcare_connect/ocs_provider/views.py`)
+- **Dispatch Celery tasks after commit**: `ATOMIC_REQUESTS = True` means a bare `task.delay()` in a request can run before the transaction commits, so the worker may not see the rows it was told about. Wrap the dispatch: `transaction.on_commit(partial(task.delay, obj.pk))` (see `commcare_connect/form_receiver/processor.py`). Most views still call `.delay()` directly — don't copy that.
```
Charl1996, line 119: +1

AGENTS.md
```diff
@@ -115,7 +117,7 @@ config/
...
 - **`.env` leaks into test settings**: `base.py` calls `env.read_env(BASE_DIR / ".env")`, so local `.env` values also apply under `config.settings.test`. An _empty_ value overrides a code default with `""` rather than falling back — `CONNECTID_URL=` breaks the suite. CI has no `.env`, so these failures are local-only
 - **`--reuse-db` + Currency/Country data**: These models get flushed between tests. The `ensure_currency_country_data` autouse fixture handles this — don't remove it
-- **API UUID transition**: The `API_UUID` waffle switch controls whether API endpoints accept integer PKs or UUIDs. Use `get_object_or_list_by_uuid_or_int()` from `utils/db.py` for API lookups
+- **API UUID transition**: The `API_UUID` waffle switch controls whether API endpoints accept integer PKs or UUIDs. Use `get_object_or_list_by_uuid_or_int()` from `utils/db.py` for API lookups. The migration is unfinished — `program/api/views.py`, `opportunity/api/views/automation.py` and `.../task_completion.py` still look up by `pk` directly, so don't take a neighbouring view as the pattern
```
Charl1996: I don't think `commcare_connect/opportunity/api/views/task_completion.py` still looks up by `pk`? It query's by `assigned_task_id` which is a UUID.

AGENTS.md
snopoke: Claude supports `@imports` but none of the others do so either we should delete AGENTS.md or make it the authority. I chose the latter.

AGENTS.md
snopoke:
```suggestion
- **API UUID transition**: The `API_UUID` waffle switch controls whether API endpoints accept integer PKs or UUIDs. Use `get_object_or_list_by_uuid_or_int()` from `utils/db.py` for API lookups.
```

Charl1996 (approved):

AGENTS.md
```diff
@@ -1 +1,150 @@
...
+- **Secondary DB is opt-in**: `DATABASE_ROUTERS` is only installed when `SECONDARY_DATABASE_URL` is set (`config/settings/base.py:52-56`). Without it `SECONDARY_DB_ALIAS` is `None`, routing is off and `migrate_multi` just migrates `default` — so multidb routing behaviour can't be reproduced locally by default
+- **API UUID transition**: The `API_UUID` waffle switch controls whether API endpoints accept integer PKs or UUIDs. Use `get_object_or_list_by_uuid_or_int()` from `utils/db.py` for API lookups.
+- **CSRF via sessions**: `CSRF_USE_SESSIONS = True`. Templates use `hx-headers='{"X-CSRFToken": "{{ csrf_token }}"}'` on `<body>` for htmx
+- **Webpack output**: Bundles are built to `commcare_connect/static/bundles/` and referenced with plain `{% static 'bundles/...' %}`, served via `STATICFILES_DIRS`. `webpack-stats.json` is written but unused — django-webpack-loader is not installed
+- **CI uses**: `postgis/postgis:15-3.5` image, Python 3.11, requires `gdal-bin libproj-dev` apt packages
+
+## Further reading
+
+- `pr_guidelines.md` — PR size, description and review conventions
```
mkangia, line 148: is `PULL_REQUEST_TEMPLATE` mentioned anywhere else?

.claude/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+  "permissions": {
+    "allow": [
```
mkangia, line 3: these seems safe though wondering if someone didn't want these allowed on their machines.

AGENTS.md
snopoke: I don't think so

.claude/settings.json
snopoke: This is a minimal safe list. None of these are destructive.

AGENTS.md
mkangia: Okay, would be good to mention if this is the right PR to do so.

snopoke: The PR template doesn't have any useful info in to so I'm not going to include it.

PR #1506, opened 2026-09-08
Skip audit reports for ended, archived, test, or empty opportunities

hemant10yadav:
## Product Description

Weekly audit reports (and their email notifications) are no longer generated for opportunities that have ended, are archived, marked as test, or have no accepted workers.

## Technical Summary

Ticket: [CI-881](https://dimagi.atlassian.net/browse/CI-881)

- `generate_audit_report_for_opportunity` now returns `None` and skips creating a report when an opportunity has no active accesses, instead of creating an empty report with no entries.
- The task filters on `end_date__gte=period_start` rather than an "is active" flag, so an opportunity that ended partway through the just-completed week still gets its final report.

### Safety story

Ran the updated test suite locally and confirmed each exclusion case (ended, archived, test, no deliveries) is skipped while a normal opportunity and one ending mid-period still generate a report.

### Automated test coverage

Added tests for each exclusion case and the mid-period edge case, and updated the existing service/task tests for the new `None` return; all pass.

### QA Plan

 tested manually on local. Will request QA

### Deployment

### Labels & Review

[CI-881]: https://dimagi.atlassian.net/browse/CI-881?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/audit/services.py
+++ b/commcare_connect/audit/services.py
@@ -100,13 +100,7 @@ def generate_audit_report_for_opportunity(
     opportunity: Opportunity,
     period_start: datetime.date,
     period_end: datetime.date,
-) -> AuditReport:
-    report = AuditReport.objects.create(
-        opportunity=opportunity,
-        period_start=period_start,
-        period_end=period_end,
-    )
-
+) -> AuditReport | None:
     active_accesses = (
         OpportunityAccess.objects.filter(
             opportunity=opportunity,
@@ -115,6 +109,15 @@ def generate_audit_report_for_opportunity(
         )
         .select_related("user")
         .order_by("user__name")
+    )
+
+    if not active_accesses.exists():
+        return None
+
+    report = AuditReport.objects.create(
+        opportunity=opportunity,
+        period_start=period_start,
+        period_end=period_end,
     )
 
     calcs = calculations.get_registered_calculations()
--- a/commcare_connect/audit/tasks.py
+++ b/commcare_connect/audit/tasks.py
@@ -35,6 +35,9 @@ def generate_audit_reports() -> None:
         Opportunity.objects.filter(
             Q(pk__in=flag.opportunities.values_list("pk", flat=True)) | Q(program__in=flag.programs.all()),
             active=True,
+            archived=False,
+            is_test=False,
+            end_date__gte=period_start,
         )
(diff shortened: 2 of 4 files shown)


commcare_connect/audit/tasks.py
```diff
@@ -35,6 +35,9 @@ def generate_audit_reports() -> None:
         Opportunity.objects.filter(
             Q(pk__in=flag.opportunities.values_list("pk", flat=True)) | Q(program__in=flag.programs.all()),
             active=True,
+            archived=False,
+            is_test=False,
+            end_date__gte=period_start,
```
mkangia, line 40: just confirming that this does not let to any weird situations where the end date for an opportunity is between the end of a period and the start of the next period?

commcare_connect/audit/services.py
```diff
@@ -117,6 +111,15 @@ def generate_audit_report_for_opportunity(
         .order_by("user__name")
     )
 
+    if not active_accesses.exists():
```
mkangia, line 114: I assume this is additional to what the ticket was originally about?
So this now skips generating the report entirely, though the user does not really know why it was skipped?

commcare_connect/audit/tasks.py
hemant10yadav: Yeah, I have considered that case as well, and it will produce a report for that too.

commcare_connect/audit/services.py
hemant10yadav: If there are no users in the opportunity, there’s no point in generating a report or notifying the admins, right?

mkangia: Agreed.

I am just thinking from user perspective.

Are we expecting that report generation won't happen initially but then will always continue to happen since users once assigned will continue to exist and we won't have a scenario again where they won't be any accepted access until the opportunity finishes?

hemant10yadav: Yes

mkangia: okay :+1:

mkangia (approved):

PR #1508, opened 2026-09-08
Fix invoice USD total rounding drift

hemant10yadav:
## Product Description

The invoice review page was showing a Total Pay (USD) that was slightly too low for payment types with many deliveries. This fix corrects the displayed total, and a follow-up command corrects the stored total on existing unpaid invoices.

## Technical Summary

[CI-927](https://dimagi.atlassian.net/browse/CI-927)

- Each delivery's pay is rounded to USD individually when it's frozen, and summing hundreds of those already-rounded cents drifts below the true total. The fix re-derives every total (per work, per group, and the invoice as a whole) from the exact local amount divided once by the group's rate, instead of adding up rows that were each rounded on their own.
- Applies at three places that compute a total: the unfrozen preview line items, the frozen invoice line items shown on the review page, and the invoice's own `amount_usd` field.
- Added `backfill_invoice_amount_usd` to correct `amount_usd` on invoices that were already frozen with the old, drifted total. It only touches unpaid, non-cancelled/rejected service-delivery invoices, since a paid invoice's `Payment` record already reflects the old total and rewriting it afterward would just make the two disagree. Supports `--dry-run`.

### Safety story

The frozen per-work data is untouched, so there's no data migration or risk to existing invoice records. The code change only affects how totals are computed for display and for the invoice's stored `amount_usd`. The backfill command is scoped to unpaid invoices only and re-checks under a row lock that an invoice is still unpaid before writing, so a concurrent payment or cancellation can't race it. I verified both the new totals and the backfill's output against the exact math from the ticket's example, and dry-ran the command locally before writing this.

### Automated test coverage

Added and updated tests in `test_invoice_line_items.py` covering the group, frozen-invoice, and invoice-total cases with a currency conversion example matching this ticket's drift; all pass locally. The backfill command itself is a one-time script and isn't covered by automated tests.

### QA Plan

tested manually on local. Will request QA

### Deployment

- This PR can be deployed on its own, but `amount_usd` on invoices that were already frozen before this deploy will still show the old drifted total until `./manage.py backfill_invoice_amount_usd` is run. Run it (optionally with `--dry-run` first) against production once this is deployed.

### Labels & Review

[CI-927]: https://dimagi.atlassian.net/browse/CI-927?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/opportunity/management/commands/backfill_invoice_amount_usd.py
+++ b/commcare_connect/opportunity/management/commands/backfill_invoice_amount_usd.py
@@ -1,0 +1,79 @@
+from django.core.management import BaseCommand
+from django.db import transaction
+
+from commcare_connect.opportunity.models import InvoiceStatus, PaymentInvoice
+from commcare_connect.opportunity.utils.invoice_line_items import (
+    Money,
+    get_invoice_delivery_rows_for_export,
+    group_line_items,
+)
+
+CHUNK_SIZE = 500
+
+
+# TODO One time run command (CI-927). Remove this once it has run.
+class Command(BaseCommand):
+    help = (
+        "Recompute PaymentInvoice.amount_usd for unpaid service-delivery invoices (CI-927). These "
+        "were frozen by summing each delivery's already-rounded USD instead of dividing the exact "
+        "local total once, so the stored total drifts below the correct amount. Local amounts "
+        "(invoice.amount) were not affected and are left alone. Paid invoices are skipped: their "
+        "Payment record already reflects the old total, so this would only make it inconsistent."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument("--dry-run", action="store_true", help="Show what would change without saving.")
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
+        invoices = (
+            PaymentInvoice.objects.filter(service_delivery=True, payment__isnull=True)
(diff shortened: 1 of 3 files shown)


hemant10yadav:
@ajeety4  Since you worked on this recently, I’d like to get your review on this and hear your thoughts on it.

ajeety4:
Hi @hemant10yadav ,

I am still yet to review this PR.

However, before that I think we should confirm with product/Cal with the approach used (**if not already done**). I say this as this has been raised before, and it was decided to continue using the individual sums. We added a tooltip on the USD column for better understanding. 
This is the [ticket ](https://dimagi.atlassian.net/browse/CI-528) for reference.

hemant10yadav:
> Hi @hemant10yadav ,
> 
> I am still yet to review this PR.
> 
> However, before that I think we should confirm with product/Cal with the approach used (**if not already done**). I say this as this has been raised before, and it was decided to continue using the individual sums. We added a tooltip on the USD column for better understanding. This is the [ticket ](https://dimagi.atlassian.net/browse/CI-528) for reference.

Thanks for pointing that out. I’ll discuss this on the ticket and confirm the approach before proceeding.

PR #1515, opened 2026-09-09
Fix query fan-out in active_flags_for_user

hemant10yadav:
## Product Description

No user-facing change.

## Technical Summary

[CI-847](https://dimagi.atlassian.net/browse/CI-847)

- `Flag.active_flags_for_user` joined organizations, opportunities (via opportunityaccess) and programs into one query with `.distinct()`. Joining independent relations like that multiplies their row counts against each other before Postgres can filter and dedupe them, so the cost scales with the product of all three relation sizes instead of the sum.

  | | Old query | New query |
  |---|---|---|
  | Execution time | 147-208 ms | 18-19 ms |
  | Rows evaluated before filtering down to the result | ~501,300 | ~6 |
  | Disk reads | up to 6 | 0 |

  "Rows evaluated" is the actual row count Postgres had to check against the `WHERE` clause before landing on the same handful of matching flags in both cases, not a planner guess. The old query has to join every flag's organizations, opportunities and programs together first, so it ends up checking half a million rows to find 3 matches. The new query checks each relation independently via `EXISTS`, so it only ever looks at a handful of rows.

  **Net result: ~8-11x faster execution, roughly 83,000x fewer rows evaluated, and zero disk reads instead of up to 6.** Both queries return the same flags in every case tested.
  
  Prod shell screenshot
Screenshot 2026-09-09 at 11 43 30 AM

### Safety story
Reproduced the fan-out locally with a synthetic dataset first, then confirmed the improvement against real prod data with a read-only `EXPLAIN ANALYZE` (no writes) using one of the busiest real users I could find. Both queries returned identical flag sets in every case tested.

### Automated test coverage
Existing tests for `active_flags_for_user` (direct user flags, org/opportunity/program relation paths, role flags) pass unchanged, along with the full flags and toggle-view test suites (56 tests).

### QA Plan
No QA ticket needed, tested manually on local (full test suite) and verified against production data with a read-only EXPLAIN ANALYZE comparison.

### Deployment

### Labels & Review

[CI-847]: https://dimagi.atlassian.net/browse/CI-847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/flags/models.py
+++ b/commcare_connect/flags/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.db.models import Exists, OuterRef
 from django.db.models.signals import m2m_changed
 from django.utils.translation import gettext_lazy
 from waffle.models import CACHE_EMPTY, AbstractUserFlag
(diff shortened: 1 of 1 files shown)


zandre-eng (approved):
Nice improvement!

Charl1996 (approved):

PR #1519, opened 2026-09-10
Draw building footprints client-side from Overture PMTiles

Charl1996:
## Product Description

A **Show Building Data** toggle on the progress and assignment maps draws Overture building footprints over the work areas from zoom 15 up, with a loading indicator under the toggle while tiles are in flight. Below zoom 15 the control and overlay both leave the map.

[Screencast from 10-09-2026 09:13:42.webm](https://github.com/user-attachments/assets/2e7dbb70-4a95-441e-85eb-233b0e440010)

## Technical Summary

[CCCT-2750](https://dimagi.atlassian.net/browse/CCCT-2750)

Mapbox GL has supported PMTiles vector sources since 3.21 and we ship 3.24, so the whole server side of the feature becomes unnecessary: no proxy view, no Redis tile cache, no bbox grid snapping, no area cap, no `pyarrow` (~152MB installed, ~60MB resident per worker), no `mercantile`. Mapbox owns tile fetching, caching and eviction, so the client loses its accumulated-area bookkeeping too, and the "area too large" state stops existing. 291 insertions and zero dependency changes here, against 1,348 insertions and two new dependencies in #1492.

Measured against the server-side approach on one z14 tile over Kibera:

**Where to focus review:** the archive stops at zoom 14 and footprints draw from 15, so every rendered zoom is overzoomed from z14 tiles. I verified this loses no features in the dense areas that matter — Kibera's z14 tile carries 11,568 buildings, Kano's 34,067, with no thinning — and geometry is quantized to roughly 0.6m by the tile grid. Whether that *looks* acceptable at z19–20 is a judgement call I couldn't make without a browser, and it's the one thing that could sink this approach. I have a local A/B harness (this overlay vs. full-precision GeoJSON from #1492's own code path, in Kibera and Kano) if it's useful to anyone.

### Note
Overture's buckets drop releases after 60 days and keep only the latest two. We need a followup PR to fetch and update the latest version automatically.

### Safety story

Flag-gated behind the existing `MICROPLANNING` flag: one opt-in map layer, no new endpoint, no migrations, no writes, no existing data touched, no new dependencies. The server does not fetch, parse or store building data at any point — it only computes a URL. Worst case for a user is an empty or failed tile layer on a map that otherwise works; there is no server-side failure mode.

### Automated test coverage

`test_buildings.py` covers the entire Python surface: the tile URL for a configured release, the OSM + Overture attribution, and the unconfigured path (`None`/empty/whitespace release) that hides the overlay instead of emitting a bad URL. 

### QA Plan

QA planned. Worth exercising specifically: footprint alignment and legibility at z19–20 in a dense area, the zoom-15 threshold in both directions, and clicking the toggle with the toast recently shown.

### Deployment

### Labels & Review

https://claude.ai/code/session_01MgpMVxf9EuqYBYdB7W1ou3

[CCCT-2750]: https://dimagi.atlassian.net/browse/CCCT-2750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/commcare_connect/microplanning/buildings.py
+++ b/commcare_connect/microplanning/buildings.py
@@ -1,0 +1,69 @@
+"""
+Building footprints for the microplanning map.
+
+Overture publishes its buildings as a PMTiles archive and Mapbox GL reads PMTiles natively, so the
+browser fetches footprints straight from Overture. This module only works out what to point it at:
+there is no proxy view, no cache and no table behind the overlay, and no building data passes
+through this process.
+
+The one thing needing care is the release. Overture's buckets drop everything older than 60 days,
+so a release that is fine today is a dead URL in two months -- see ``OVERTURE_RELEASE``.
+"""
+
+# The Overture release footprints are read from. Overture keeps only the two most recent releases
+# -- its buckets carry a 60 day retention rule -- so this has to move forward every month or so,
+# which currently means a deploy. Pinned here rather than in settings because it is not per
+# environment: every environment wants the same, current release.
+OVERTURE_RELEASE = "2026-08-19.0"
+
+# https://docs.overturemaps.org/examples/overture-tiles/
+OVERTURE_TILES_URL = (
+    "https://overturemaps-extras-us-west-2.s3.us-west-2.amazonaws.com/tiles/{release}/buildings.pmtiles"
+)
+
+OVERTURE_BUILDINGS_LAYER = "building"
+
+# A fact about the archive, handed to the Mapbox *source* as `maxzoom`: Overture builds tiles down
+# to zoom 14 and no deeper. Declaring it is what makes Mapbox overzoom those z14 tiles for closer
+# views; without it Mapbox requests z15+ tiles that do not exist and the overlay comes up empty.
+OVERTURE_ARCHIVE_MAX_ZOOM = 14
+
+# Our display policy, handed to the Mapbox *layers* as `minzoom`: the zoom at which footprints
+# start being drawn at all.
+#
+# Set to the archive's max zoom, so footprints appear at Overture's own deepest tiles and are drawn
+# at native resolution there; only the zooms past it are overzoomed. These two are not a range --
+# one is a property of the data, this one is a choice.
+#
+# Do not drop it below the archive max. z14 is the last complete level: Overture thins the lower
+# zooms hard (a z13 tile over Kibera carries about a fifth of the buildings its z14 tiles do), so
+# drawing there would show a partial set of buildings while looking complete.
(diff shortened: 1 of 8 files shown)


Charl1996:
### Code review

Found 2 issues:

1. `escapejs` on an `x-tooltip.raw` value will render literal `\uXXXX` escapes once the string is translated. The `.raw` modifier passes the attribute value straight to `setupTooltip(expression)` without evaluating it as JS (`node_modules/@ryangjchandler/alpine-tooltip/dist/module.cjs.js`), so the content is used literally. `escapejs` turns `'`, `;`, `=` and `<`/`>` into `\uXXXX` sequences — `escapejs("Don't show; see notes=none.")` returns `Don\u0027t show\u003B see notes\u003Dnone.` It is a no-op for the current ASCII English copy, so nothing shows today, but any translation or copy edit containing an apostrophe will display the escape sequence in the tooltip. Every other `x-tooltip.raw` usage in the codebase passes the translated string unfiltered and relies on normal attribute autoescaping (e.g. [`home.html:253`](https://github.com/dimagi/commcare-connect/blob/de11bbb056b86b5e99ce01b8b68198160773fad3/commcare_connect/templates/microplanning/home.html#L252-L254)).

https://github.com/dimagi/commcare-connect/blob/de11bbb056b86b5e99ce01b8b68198160773fad3/commcare_connect/templates/microplanning/map_layer_toggles.html#L8-L10

2. The new control panel visually occludes map toasts. `.map-control-panel` is `absolute right-4 top-4 z-[60]` with an opaque `bg-white`, and `.map-toast` is `absolute right-4 top-4 z-50` — the same corner, with the opaque panel stacking above it. `buildingsAvailable` gates the panel on zoom alone, not on whether the layer is toggled on, so the panel is present whenever the user is zoomed past the threshold, which is exactly when the `showToast(...)` calls in `map_handler.html` fire ("Work area updated successfully!", exclude/assign results, filter updates). The panel's comment reasons about this shared corner from the pointer-events angle only — `978f01976` fixed the invisible toast eating clicks — but the same overlap now means the panel hides the message. Before this PR the toast was the sole occupant of `top-4 right-4`.

https://github.com/dimagi/commcare-connect/blob/de11bbb056b86b5e99ce01b8b68198160773fad3/tailwind/tailwind.css#L901-L906

The toast rule it stacks over: [`tailwind.css:917-923`](https://github.com/dimagi/commcare-connect/blob/de11bbb056b86b5e99ce01b8b68198160773fad3/tailwind/tailwind.css#L916-L924)

PR #1523, opened 2026-09-10
Fetch the Overture release dynamically

Charl1996:
https://dimagi.atlassian.net/browse/CCCT-2828

## Product Description

Building footprints came from a hardcoded Overture release. Overture ships a new one roughly monthly and [drops anything older than 60 days](https://docs.overturemaps.org/release-calendar/#data-retention-policy), so the map fell behind until someone bumped the constant. It now tracks the latest release itself, and the tooltip names the one being drawn.

image

## Technical Summary

The release value is read from a single-row `OvertureRelease` table, refreshed daily by a Celery beat task.

#### Why in the database? 
If we only store the release value in Redis there's a big risk of the feature breaking if the cache expires and we cannot read the latest value for some reason. We thus need a persistent value. 

### Safety story

A bad write would persist, so a release is recorded only if the catalog answered with a JSON object and it matches `\d{4}-\d{2}-\d{2}\.\d+`. An unreachable catalog retries, then leaves the row for tomorrow.

`buildings_overlay_config()` now does one primary-key lookup per microplanning home render, deliberately uncached.

### Automated test coverage

19 tests. The ones that matter: an unreachable catalog leaves the previous release standing, malformed and non-object payloads are refused, and the seed stops being written past Overture's retention.

### QA Plan

No QA ticket.

### Deployment

### Labels & Review


commcare_connect/microplanning/models.py
```diff
@@ -192,3 +193,31 @@ class Meta:
 
     def __str__(self):
         return f"WorkAreaInaccessibilityRequest {self.xform_id} - {self.work_area}"
+
+
+class OvertureRelease(BaseModel):
+    """
+    The Overture Maps release building footprints are drawn from.
```
ajeety4: super nit: `The Overture Maps release that building footprints are drawn from.`

commcare_connect/microplanning/tasks.py
```diff
@@ -523,3 +533,48 @@ def cluster_work_areas_task(opp_id, max_buildings=DEFAULT_BUILDING_COUNT):
...
+        catalog = None
+
+    if not isinstance(catalog, dict):
+        logger.error("Overture's catalog at %s did not answer with a JSON object", OVERTURE_CATALOG_URL)
+        return None
+
+    release = catalog.get("latest")
+
+    if not isinstance(release, str) or not OVERTURE_RELEASE_RE.fullmatch(release):
+        logger.error("Overture's catalog named %r as its latest release, which is not one", release)
```
ajeety4: nit: `Overture's catalog reported %r as its latest release, but it does not match the expected release format`

ajeety4 (approved):
LGTM

commcare_connect/microplanning/migrations/0019_create_update_overture_release_periodic_task.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def create_update_overture_release_periodic_task(apps, schema_editor):
+    schedule, _ = CrontabSchedule.objects.get_or_create(
+        minute="0",
+        hour="0",
+        day_of_week="*",
+        day_of_month="*",
+        month_of_year="*",
+    )
```
hemant10yadav, line 12: Nit: Is daily necessary here? I was wondering if weekly would be sufficient. Overture releases roughly monthly and retires releases after 60 days, so a weekly check should still give us 3+ weeks of margin before a stale release could 404. Just wondering if daily is deliberate.

commcare_connect/microplanning/tasks.py
```diff
@@ -523,3 +533,48 @@ def cluster_work_areas_task(opp_id, max_buildings=DEFAULT_BUILDING_COUNT):
...
+    except ValueError:
+        catalog = None
+
+    if not isinstance(catalog, dict):
+        logger.error("Overture's catalog at %s did not answer with a JSON object", OVERTURE_CATALOG_URL)
+        return None
+
+    release = catalog.get("latest")
+
+    if not isinstance(release, str) or not OVERTURE_RELEASE_RE.fullmatch(release):
```
hemant10yadav, line 576: Should we add RegexValidator(OVERTURE_RELEASE_RE) to the model field itself?

hemant10yadav (dismissed):

commcare_connect/microplanning/migrations/0019_create_update_overture_release_periodic_task.py
Charl1996: I went with daily to make sure we get the latest as soon as possible instead of having to wait for a week in case a new release is publised..

commcare_connect/microplanning/tasks.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1523/commits/cb9bc7f510a2fd513dbd60bb0735297132cdccf8

commcare_connect/microplanning/models.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1523/commits/cb9bc7f510a2fd513dbd60bb0735297132cdccf8

ajeety4 (dismissed):
Thanks for addressing the nits

ajeety4 (approved):

hemant10yadav (commented):
Happy to approve once this open question is addressed: https://github.com/dimagi/commcare-connect/pull/1523#discussion_r4012061023

commcare_connect/microplanning/models.py
```diff
@@ -192,3 +193,31 @@ class Meta:
...
+
+    @classmethod
+    def current(cls):
+        """The release to draw footprints from, or None if we have never recorded one."""
+        row = cls.objects.filter(pk=cls.SINGLETON_PK).first()
+        return row.release if row else None
+
+    @classmethod
+    def set_current(cls, release):
+        cls.objects.update_or_create(pk=cls.SINGLETON_PK, defaults={"release": release})
```
pxwxnvermx: What was the reason for going with a Singleton model for managing these releases?

An alternative approach could have been to store this release in the django settings file and fetch it  whenever the server is started by using the `fetch_latest_overture_release` function and storing it in an env variable. That would remove this need for having a singleton model. With our current release cadence we can ensure that we have a fairly new overture release version always set as an environment variable.

commcare_connect/microplanning/tasks.py
Charl1996: https://github.com/dimagi/commcare-connect/pull/1523/commits/64e185e62e03c85df13504646424f800e0cff262

Charl1996: Good suggestion, thanks!

commcare_connect/microplanning/models.py
Charl1996: > An alternative approach could have been to store this release in the django settings file and fetch it whenever the server is started

The downside here is that the startup is dependant on an external service and there is a question of what we should do / fall back to if the service isn't reachable during a given deploy. Since the releases are only kept alive for 60 days the fallback value is tricky to determine and we cannot simply rely on a default value that was potentially set far back in the past. If Overture kept access to release data somewhat indefinitely your suggested approach would have been the ideal one.  

The biggest reason I went with the singleton approach was that we have a durable release value for cases where the service isn't reachable and we know this release value is, if not the latest, the second latest value that still has data that is accessible.

hemant10yadav (approved):

commcare_connect/microplanning/models.py
pxwxnvermx: Thanks for the information. Looking at the overture releases it feels like we might need two fold work to update the release version automatically and also keep the default value up-to-date, so going with the singleton approach does make sense.

pxwxnvermx (approved):

commcare_connect/microplanning/models.py
Charl1996: > we ... need two fold work to update the release version automatically and also keep the default value up-to-date

Exactly

PR #1526, opened 2026-09-14
Trigger QA web test suite after deploy

Charl1996:
## Technical Summary
As part of QA's effort to push automated QA tests they want each production deploy to trigger a post-deploy test suite. This PR adds the necessary step to the deploy workflow to accomplish this.

### How it works
After a deploy a new job invokes the _CommCare-Connect Playwright Web Suite_ (see [example](https://github.com/dimagi/dimagi-qa-ccc/actions/runs/34855143118) run) in the [dimagi-qa-ccc repo](https://github.com/dimagi/dimagi-qa-ccc). After successful invocation the step is marked as "complete" regardless of the test suites outcome (it's fire-and-forget). The test suite then runs happily in the background while the new code runs live.

### What happens when the QA test invocation fails?
The deploy job is still marked as successful and the new code is deployed successfully (slack will still say "success" since we report to slack _before_ invoking the QA suite, which I think is the right way). The workflow will also be marked green due to the `continue-on-error` setting on the QA job, but the QA job itself will be marked red. The failed QA job will also put a comment on slack for any failures.

### What happens when the QA suite fails?
The maintainers of the `dimagi-qa-ccc` repo is notified - **_the Connect deploy is NOT blocked or rolled back or anything_**, for 2 reasons
1. The test suite takes a while to run - almost 30min - and blocking the deploy on this seems counter-productive. 
2. The deploy's outcome should not be dependant on the QA test suite's outcome as it's not playing the same role as in-code unit tests. The test suite (maintained by the QA team) can still fail due to non-related issues such as out of date tests, so we don't want to block Connect deploys for this reason.

See an [example here](https://github.com/dimagi/commcare-connect/actions/runs/34839428406) where I tested this new workflow with staging.

### Safety story
I tested the workflow against staging by changing the env condition to `staging` and checking for `autostaging` in `github.ref_name`.

### Automated test coverage
N.A

### QA Plan
N.A

### Deployment

The `QA_SUITE_DISPATCH_TOKEN` token is already added and have been tested by doing a [staging deploy](https://github.com/dimagi/commcare-connect/actions/runs/34839428406)

### Labels & Review


mkangia:
> but the job will be marked as having "failed" and the dev having triggered the deploy will be notified.

I am not sure about this. Are we saying if I were to see the list of actions, it would show failed but the deploy would have actually happened? If yes, this can get really confusing.

.github/workflows/deploy.yml
```diff
@@ -176,3 +176,27 @@ jobs:
...
+
+  running-qa-test:
+    name: Trigger QA Playwright suite
+    needs: deploy
+    if: ${{ success() && inputs.environment == 'production' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for the deploy to be serving
+        run: |
+          url=${{ github.ref_name == 'main' && 'https://connect.dimagi.com' }}
```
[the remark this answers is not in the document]
hemant10yadav: @Charl1996 this is a real issue — deploying prod from a hotfix branch would set url=false.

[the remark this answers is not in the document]
Charl1996: https://github.com/dimagi/commcare-connect/pull/1526/commits/c1a933589330e387dc2b5f3b14bd2535c8aa9f63

Charl1996:
@mkangia See [this commit](https://github.com/dimagi/commcare-connect/pull/1526/commits/f0a178626187d5800584b7416e9ab9739a1119cf) where we
1. `continue-on-error`, so only the failed job will be marked as failed but the workflow itself will be marked as having succeeded
2. notify on slack when there's a failure to invoke the QA test suite.

.github/workflows/deploy.yml
```diff
@@ -176,3 +178,41 @@ jobs:
...
+    continue-on-error: true
+    steps:
+      - name: Wait for the deploy to be serving
+        run: |
+          url=https://connect.dimagi.com
+          for i in $(seq 1 30); do
+            curl -fsS -o /dev/null "$url" && exit 0
+            sleep 10
+          done
+          echo "site not responding after 5m" >&2; exit 1
```
snopoke, line 196: won't the site always be serving? I mean that it doesn't go down during deploy right?

snopoke (approved):

.github/workflows/deploy.yml
Charl1996: That is correct, but I thought it's still probably better to have it than not

.github/workflows/deploy.yml
```diff
@@ -35,6 +35,8 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    outputs:
+      slack_ts: ${{ steps.slack.outputs.ts }}
```
mkangia, line 39: just curious what this change does?

mkangia (approved):

.github/workflows/deploy.yml
snopoke: It makes the value available for later steps

PR #1532, opened 2026-09-16
Fix delivery notification sent to unclaimed opportunities

hemant10yadav:
## Product Description

Fixed a bug where users who finished learning but hadn't claimed a job could get a push notification that raised 500 error on server.

## Technical Summary

[CI-795](https://dimagi.atlassian.net/browse/CI-795)

- Inactive users with 100% learn progress but no claim were being sent the delivery-progress notification, which routes mobile to a screen it can't load without a claim (the 500 in the Sentry issue). They now get the learn-progress notification instead, the same one used for anyone still mid-learning.

### Safety story

Ran the existing test suite locally; the change only affects which notification unclaimed, inactive users receive, and doesn't touch the claimed-user delivery flow at all.

### Automated test coverage

Updated the existing unclaimed-user inactivity test to match the new behavior; full suite passes.

### QA Plan

QA pending

### Deployment

- [x] This PR can be deployed: no config, migration, or manual step needed.

### Labels & Review

[CI-795]: https://dimagi.atlassian.net/browse/CI-795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ajeety4 (commented):
I’m trying to understand this a bit better, so please correct me if I’m wrong.

For the case where the user has completed all the modules, I agree that we should not send the delivery message. However, it seems misleading to send the learn message here, since the user has already completed all the learning modules and the next step for them is to claim the opportunity.

I’m wondering if we can send an appropriate message for this case while still keeping the action as `ccc_learn_progress`.

Any thoughts?

PR #1536, opened 2026-09-17
Fix slow query powering the Work Area Assignments tab

hemant10yadav:
## Product Description

The Work Area Assignments tab was timing out for opportunities with a lot of workers and work areas. It now loads normally.

## Technical Summary

[CI-953](https://dimagi.atlassian.net/browse/CI-953)

- The worker aggregate query joined each worker to their work areas before grouping, which multiplied every worker row by their work area count and forced a per-worker subquery to re-run once per multiplied row instead of once per worker. Swapped the join for per-worker correlated subqueries so there's no multiplication for the query planner to get wrong.
- Dropped a redundant opportunity filter on the visits-done subquery, since filtering by opportunity_access already scopes to a single opportunity, and the extra filter was forcing Postgres into a slower two-index scan.

### Safety story

Verified against real prod data

**~420x faster execution and ~4,700x fewer rows scanned.**

| | Rows scanned | Execution time |
|---|---|---|
| Old (current prod) | 380,755,722 | 36,782 ms |
| New (this PR) | 80,566 | 88 ms |

image

The results are identical across every worker row with 0 mismatches.

### Automated test coverage

Existing tests for this helper and the related view (217 total) pass unchanged; no behavior change, just how the same result is computed.

### QA Plan

No QA ticket needed, tested manually on local and verified query performance against real prod data (read-only).

### Deployment

### Labels & Review

[CI-953]: https://dimagi.atlassian.net/browse/CI-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Charl1996 (approved):
Nice!

PR #1539, opened 2026-09-18
feat(data_export): opportunities carry their budget, not only their visit count

jjackson:
## Product Description

The opportunity export reports how much work an opportunity has **produced** and
nothing about how much was **funded** for it. A consumer can see 1.4 million
delivered services and has no way to ask the obvious next question — how much
committed work is still outstanding.

This adds the budget to the rows that already carry the visit count. Nothing
changes for existing consumers; the fields are additive.

## Technical Summary

`GET /export/opp_org_program_list/` gains:

| on | fields |
|---|---|
| `opportunities` | `total_budget`, `claimed_budget`, `remaining_budget`, `currency`, `start_date` |
| `programs` | `budget`, `start_date`, `end_date` |

All are plain columns already on `Opportunity` and `Program`. `currency` reads
the FK id rather than joining, exactly as `ProgramDataExportSerializer` already
does — `Currency`'s primary key is the code itself.

### claimed / remaining, without the N+1

`claimed_budget` and `remaining_budget` are the fields that actually answer
"how much funded work is still outstanding", so they are here. Two real
obstacles had to be got round rather than accepted:

- The **model properties** walk `OpportunityAccess` → `OpportunityClaim` →
  `OpportunityClaimLimit` per row. On a list endpoint covering every
  opportunity a member can see, that is hundreds of queries.
- A **second multi-valued aggregate** cannot be annotated beside the existing
  `Count("uservisit")` — both fan out through the join and inflate.

A `Subquery` has neither problem: it evaluates independently of the outer
aggregate, returns one scalar per row, and adds no per-row query.

Two details worth knowing if you touch this:

- The annotation is `claimed_budget_total`, **not** `claimed_budget`. The latter
  is a property with no setter, so Django cannot attach the value and the query
  dies with `can't set attribute`. The serializer maps the public name onto the
  annotation.
- `remaining_budget` is `null` when no budget is set, not `0`. Subtracting from
  an unknown total gives an unknown remainder, and `0` reads as "fully
  committed".

### Who wants it

Connect Labs' marketplace surface (`labs.connect.dimagi.com/labs/marketplace/`)
presents the partner network by programme. It can currently show what each
programme has delivered but not what is funded and unfinished, because
`PulseOpportunity.total_budget` is a column the ingest has never been able to
fill — the export does not carry it.

### Safety story

Additive fields on a read-only export. No model change, no migration, no write
path, and no change to any existing field's name, type or meaning — a consumer
that ignores the new keys is unaffected.

The one risk on a list endpoint is cost, which is why the expensive properties
were left out and a query-count test pins that decision.

`total_budget` is nullable in the model and is exported as `null` rather than
coerced, because "no budget set" and "a budget of zero" are different facts and
a consumer that cannot tell them apart will report the second as the first.

### Automated test coverage

`TestOpportunityBudgetIsExported`, eight tests, all passing locally along with
the full `commcare_connect/data_export/` suite (**130 passed**). Ruff clean.

- An opportunity carries its budget, currency and start date.
- Claimed and remaining budget are exported, against hand-computed figures.
- **The annotation agrees with the model property.** The property is the
  definition; the annotation exists only because the property is too expensive
  to call per row, so if they disagree the cheap one is wrong.
- **An unfunded opportunity reports `null`, not `0`** — for both `total_budget`
  and `remaining_budget`.
- Claims do not inflate `visit_count`.
- A programme carries its budget and dates.
- **Listing opportunities does not query per row.**

Mutation-checked, and one of these is worth flagging: swapping the subquery for
a fan-out aggregate **initially passed**, because `visit_count` is counted with
`distinct=True` and the test opportunity had a single visit — nothing to
multiply. The agreement test now creates three visits, and the mutation fails
as it should. The first version of that test proved nothing.

### QA Plan

`GET /export/opp_org_program_list/` as a user with org membership and confirm
each opportunity row carries `total_budget`, `currency` and `start_date`, and
each programme row `budget`, `start_date` and `end_date`. Compare one
opportunity's `total_budget` against the same value in the admin. Confirm the
response time is unchanged for an org with many opportunities.

### Deployment

Merging is enough. No config, no flag, no migration. Labs picks the fields up on
its next poll once this is deployed.


jjackson:
Closing — opened without checking whether this was wanted in prod, and the answer is that changes here are not mine to make. The labs-side need (committed vs delivered funding on the marketplace) will be met from what the existing APIs already expose. Leaving the branch on my fork in case the export question is ever worth revisiting.

